PR #27, opened 2013-11-16
Update travis.yml for scipy

ankurankan:


ankurankan:
Will try to find a better solution for this using shell scripts.

Issue #79, opened 2014-03-06
setup.py does not specify include_dir for numpy headers


johnymontana:
This results in an error when building pgmpy: `'numpy/arrayobject.h' file not found`. setup.py should use `include_dir = [np.get_include()]` to specify include directory.

PR #80, opened 2014-03-06
Add include_dirs for distutils to find numpy header files

johnymontana:
An include directory was not specified for numpy header files,
resulting in an error when building. Specify include_dirs for
distutil to include numpy header directory. See issue #79


ankurankan:
Thanks for the patch.

PR #82, opened 2014-03-10
Add Cython compiled(Machine generated) C code to gitignore

ankit-maverick:


abinashpanda:
@ankit-maverick: The cython compiled c codes are deliberately added into the library. We don't want to have cython as a additional dependency for the library. So if cython is not present, the user could still install the library.

ankit-maverick:
@abinashpanda : I just noticed that Cython is an optional dependency. Thanks. Closing this.

PR #84, opened 2014-03-15
Implementing MarkovModel class

ankit-maverick:
Although most of it has been replicated from the `BayesianModel` class, please review this. Implementation of more public methods in progress. Also, please suggest a better example that can be used here like the student example for `BayesianModel`.


ankurankan:
@ankit-maverick You will also need to write tests for the code.
But before writing tests, I would suggest you write a draft for the application so that we could work on improving it.

PR #85, opened 2014-03-15
Resolved bug#83 && new version of is_active_trail && is_iequivalent implementation

navinchandak:
Hello,
1. I think this was a minor bug because of which is_active_trail was not working. Basically when observed was None, then it was being added to the list of observed variables.
2. I have implemented a more efficient version of is_active_trail when the end node is given. Earlier all active nodes were returned by the function active_trail_nodes and out is_active_trail was just checking if the end-node was in it. Now I have changed this to break the function on finding the end_node on the search-tree starting from the start node.
I had one doubt about this however. I don't like the fact that function of active_trail_nodes had to be almost copied to is_active_trail. So one thing which I could have done is changed active_trail_nodes to return a list when the end-node was not specified and return a True/False when the end-node was specified. However such a structure looks really bad to me, having programmed mostly in C++. However this feature is perhaps what lends power to python. Which implementation do you recommend?
3. I also felt the comment was inappropriate and not in line with the algorithm. So I changed it.
4. With reference to 2, I have merged the two functions into 1, and the function is_active_trail now returns the active_nodes in a list if the end-node is None
5. I implemented the is_iequivalent method.  The current algorithm is linear in the graph size currently. I have also documented the method
6. I had a few doubts regarding this. So I had all the test_cases in a separate python file. However for documentation, I needed the examples and hence I had to copy-paste the code and add ">>>" everywhere. Is there an automated way for that?
Also, should I also include those examples in a test_is_iequivalent method as a test_routine?

### DIFF
--- a/pgmpy/BayesianModel/BayesianModel.py
+++ b/pgmpy/BayesianModel/BayesianModel.py
@@ -1080,21 +1080,21 @@ class BayesianModel(nx.DiGraph):
         #TODO: refer to IMap class for explanation why this is not implemented.
         pass
  
-    def _findImm(self, nodes):
-      """
-      Returns the immoralities centred at a node X when all the predecessors of X
-      are given in nodes. So just check for every pair in nodes if they have an edge 
-      in between. If not, then it is an immorality centred at X
-      """
-      immoralities1 = set()
-      for i in range(len(nodes)):
-         for j in range(i+1,len(nodes)):
-            if not self.has_edge(nodes[i],nodes[j]):
-               if nodes[i]<nodes[j]:
-                  immoralities1.add((nodes[i],nodes[j]))
-               else:
-                  immoralities1.add((nodes[j],nodes[i]))
-      return immoralities1
+    def _find_immoralities(self, nodes):
+        """
+        Returns the immoralities centred at a node X when all the predecessors of X
+        are given in nodes. So just check for every pair in nodes if they have an edge 
+        in between. If not, then it is an immorality centred at X
+        """
+        immoralities = set()
+        for i in range(len(nodes)):
+            for j in range(i+1,len(nodes)):
+                if not self.has_edge(nodes[i],nodes[j]):
+                    if nodes[i] < nodes[j]:
+                        immoralities.add((nodes[i],nodes[j]))
+                    else:
+                        immoralities.add((nodes[j],nodes[i]))
+        return immoralities
 
     def is_iequivalent(self, model):
         """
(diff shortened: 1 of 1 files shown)


ankurankan:
@navinchandak : You are correct about the first point. I myself had written that piece of code and god knows how I did that mistake. Thanks for pointing that.

You do have a great point for `is_active_trail`. For huge networks we would be doing much unnecessary computation.
I think the best way to improve this would be to move the whole code to `is_active_trail` and inside `active_trail_nodes` we should call `is_active_trail` with `end=None`. You will then have to change the `active_trail_nodes` code a little bit. Check the inline comments.

@abinashpanda : Comments??

navinchandak:
Hi.. I have made the changes as suggested. !

ankurankan:
There are many other PEP8 issues (4 spaces tab, spaces around operators etc) in the code. Please fix them.  About variable names I meant to replace `1`'s and `2`'s in their name to something more suggestive. You should also replace the function name `_findImm` with `_find_immoralities`. Python recommends using UpperCamelCase for class names and lowercase_separated_by_underscores for other names.

ankurankan:
You could replace this block:

``` python
           predecessors1 = self.predecessors(node)
           sucessors1 = self.successors(node)
           predecessors2 = model.predecessors(node)
           sucessors2 = model.successors(node)
           " Need to check if the neighbour set is same "
           " So take all neighbours, add them to a set and check if the two are same "
           " O(size of neighbour set) operation because of using sets "
           nbr1 = set(predecessors1) | set(sucessors1)
           nbr2 = set(predecessors2) | set(sucessors2)
           if not nbr1 == nbr2:
```

with 

``` python
if not set(self.predecessors(node) + self.successors(node)) == set(model.predecessors(node) + model.successors(node)):
```

ankurankan:
Also this code block:

``` python
           immoralities1 = self._findImm(predecessors1)
           immoralities2 = self._findImm(predecessors2)
           if not (immoralities1==immoralities2):
```

with:

``` python
if not self._findImm(self.predecessors()) == self._findImm(model.predecessors()):
```

ankurankan:
Please remove all the comments in between the code. I think they are all obvious. Also remove all the print statements which are in comments.

ankurankan:
Line 1140:

``` python
        node1=self.nodes()[0]
        if node1 not in model.nodes():
           return False
```

replace with:

``` python
if self.nodes()[0] not in model.nodes():
    return False
```

ankurankan:
@navinchandak : I am sorry, I guess I have given you some wrong comments about your code. 
Just write the tests for now and we will improve the code later.

navinchandak:
@ankurankan You had said that you had given some wrong comments about the code and that we will change the code later. It will be good to get done with that issue.

Also, I have made a change in get_independencies. As far as I could see, you were trying to iterate over all possible observed variables in get_independencies, which was not happening. I have changed that piece.

ankurankan:
@navinchandak I am getting merge conflicts on pulling your changes. Update your local repo by pull the main repo and then make the changes to it and then send the PR.

navinchandak:
@ankurankan Hi. I have pulled the changes from the main repo and fixed the merge conflicts. However, I have kept my version and I am not sure (not so comfortable with git), but , I guess you will again have a conflict. Because it is my version which is there in my repo. So just keep my changes.

And I am not sure why but I get a message saying "Error — The Travis CI build could not complete due to an error " and the details of that are not helping.

ankurankan:
@navinchandak: Please send a new PR. It is very difficult for me now to see your changes with all these extra commits. I will also suggest you to start with a fresh clone of the main pgmpy repo (since during merge conflict you committed your changes, either revert those changes back or start fresh).

Before sending PR's always check that you have updated your local repo so that it doesn't result in  a merge conflicts. And I am still seeing some PEP8 issues in your code. Before sending PR run pylint on your file to check for some common errors.

Also in your commit messages try to refer to issues if any https://github.com/blog/831-issues-2-0-the-next-generation . 

Travis CI is a Continuous Integration tools which automatically runs test every time we commit something or when someone sends a PR. It was working fine 2 days ago and have no idea why it's failing right now. I will have to look into it.

PR #98, opened 2014-04-03
Initial pass over UAI reader class

gunner272:
I am trying to implement reader class for UAI file format.
PS:To test the code, one would have to make changes to **init**.py  file, but as they should not form part of PR I have left them out.
Please provide feedback about parsing the CPD


gunner272:
@ankurankan, @gargshashank5 : please have a look

PR #228, opened 2014-11-29
parallel for loops in factor operations using joblib

ankurankan:


ankurankan:
Throwing some error will have to check

PR #231, opened 2014-12-01
pgmpy refactor

ankurankan:


pgmpy/base/DirectedGraph.py
```diff
@@ -1,238 +1,290 @@
...
+    True
+    >>> len(G)  # number of nodes in graph
+    3
     """
 
-    def get_node_name_with_suffix(self, node):
+    def __init__(self, ebunch=None):
+        super(DirectedGraph, self).__init__(ebunch)
+        self.cpds = []
+        self.cardinalities = defaultdict(int)
```
ankurankan: Why are you associating cpds and cardinalities with the base DirectedGraph class? Shouldn't it be a properties of model classes? I think DirectedGraph should just have the basic structure details of the network.

And things like structures for storing the states of the variables etc should be in the DirectedGraph

abinashpanda: Instead of having `self.cpds` in `DirectedGraph`, we can have something like `self.factors` as we don't know whether the factors associated with the _directed graphical model_ would be a CPD or not. Furthermore as the nodes in the `DirectedGraph` class represents some random variable, factors or a cluster of factors, we can associate cpds and cardinalities to `DirectedGraph` class. Just a suggestion.

abinashpanda:
- Both `UndirectedGraph` and `DirectedGraph` serves as an intermediate class between pure `networkx` graph classes and `pgmpy` models. All the methods that can be applied to all the _undirected graphical models_ should reside in `UndirectedGraph` and all the methods that can be applied to all the _directed graphical models_ should reside in `DirectedGraph` class.
- For example, `triangulate()` method should reside in `UndirectedGraph` class as any _undirected graphical model_ can be triangulated.

PR #234, opened 2014-12-06
Fix incorrect reference in docs/conf.py to logo.jpeg

jaidevd:
`make latex` fails without this. The logo is a png file, not a jpeg.


abinashpanda:
@jaidevd Thanks for the patch :)

jaidevd:
@abinashpanda Looking forward to many more.

PR #251, opened 2014-12-24
Predict method for Bayesian Model

ankurankan:


abinashpanda:
@ankurankan Don't you think there should be a provision to select method for performing inference. Currently the `predict` method uses `VariableElimination` implicitly.

ankurankan:
@abinashpanda Yep. Will add..

PR #277, opened 2015-02-02
Fix ve

ankurankan:


pgmpy/inference/ExactInference.py
```diff
@@ -30,9 +31,21 @@ def _variable_elimination(self, variables, operation, evidence=None, elimination
...
+        working_factors = {node: {factor for factor in self.factors[node]}
                            for node in self.factors}
 
+        # Dealing with evidence. Reducing factors over it before VE is run.
+        if evidence:
+            for evidence_var in evidence:
+                for factor in working_factors[evidence_var]:
+                    factor_reduced = factor.reduce('{evidence_var}_{state}'.format(evidence_var=evidence_var,
+                                                                                   state=evidence[evidence_var]),
+                                                   inplace=False).normalize(inplace=False)
```
abinash-panda: We should only reduce the factor. Don't normalize the factors. By normalizing the factors we are again doing the same thing that was present earlier.

Issue #267, opened 2015-01-16
Faster implementation of marginalize


ankurankan:
Currently marginalize method uses `_marginalize_single_variable` and in the case of marginalizing over multiple variables, it iterates over the variables marginalizing each of them. Improve this implementation so that marginalization over all the variables is done in a single iteration (implementation of maximize method is done is this way).

ankurankan:
Fixed in https://github.com/pgmpy/pgmpy/pull/282

PR #282, opened 2015-02-16
Faster implementation of marginalize . bug #267

palashahuja:


ankurankan:
@palashahuja Tests are failing for this PR also.

palashahuja:
Just have a look at the gist .  I have passed a lot of test cases now . As per to my knowledge, there is no syntax error. Here is the [link](https://gist.github.com/palashahuja/488050fac7a03b4acfa8)

ankurankan:
@palashahuja Whatever changes you have made, please do it on the same branch (`patch-2`) and github will automatically update this PR.

abinash-panda:
@palashahuja Yeah there is no syntax error now, but look at the failing test cases.

pgmpy/factors/Factor.py
```diff
@@ -211,18 +211,18 @@ def _marginalize_single_variable(self, variable):
...
-                                       i*cum_cardinality[index+1]]
-        return marg_factor
+        marginalize_index = np.where(np.in1d(self.scope(),variable))
+        assign = np.array(self.cardinality)
+        assign[marginalize_index] = -1
+        marg_factor = []
+        for i in itertools.product(*[range(index) for index in self.cardinality[np.where(assign!=-1)[0]]]):
+            temp_sum = 0
+            sum_variable = np.array(i)
+            sum_variable = np.insert(sum_variable,np.int(marginalize_index[0]),-1)
```
abinash-panda: There should be a space after **,**. It should be 

``` python
np.insert(sum_variable, np.int(marginalize_index[0]), -1)
```

Use `pep8` checker to check for all PEP8 errors. Please make sure that your code doesn't have any PEP8 violations (except for max line length which we had to 120 characters) before committing it. Use `pep8` checker as

```
pep8 file_name.py --max-line-length=120
```

pgmpy/factors/Factor.py
```diff
@@ -211,18 +211,18 @@ def _marginalize_single_variable(self, variable):
...
-        num_elements = cum_cardinality[0]
-        sum_index = [j for i in range(0, num_elements,
-                                      cum_cardinality[index])
-                     for j in range(i, i+cum_cardinality[index+1])]
-        marg_factor = np.zeros(num_elements/self.cardinality[index])
-        for i in range(self.cardinality[index]):
-            marg_factor += self.values[np.array(sum_index) +
-                                       i*cum_cardinality[index+1]]
-        return marg_factor
+        marginalize_index = np.where(np.in1d(self.scope(),variable))
```
abinash-panda: PEP8 error.

palashahuja:
The code is still not optimized yet. I was planning to optimize over all the marginalizing variables using a similar subroutine and then removing the "marginalize_single_variable" entirely. Is the above method really optimal or there is some other method that can do something faster ?

ankurankan:
@palashahuja I am not getting the point of improving `marginalize_single_variable`. Am I missing something?  
We need to completely remove it and use `_index_for_assignment` in a way similar to what we have done in `maximize`.
The problem with the current implementation of `marginalize` is that if multiple variables are given it iterates over each of them and marginalizes over one variable in one iteration. But if you look at `maximize`, by using `_index_for_assignment` we are able to do it in a single iteration.

palashahuja:
@ankurankan That's exactly what I am trying to say . The reason I did this just because there was a bug that said to remove these redundancies.What I will do now is create a single marginalize function such that there is no need to call marginalize_single_variable

ankurankan:
@palashahuja: Cool.. Waiting to see your final PR.

palashahuja:
@ankurankan  I have still not optimized the deletion part yet. Also, I have to ask, are there are any dependencies for marginalize_single_variable?If not , then I think it will be really safe to erase  that  code.

ankurankan:
@palashahuja `_marginalize_single_variable` is used just by `marginalize` method. So once you get this code inside `marginalize`, it should be safe to delete it.

palashahuja:
Fine @ankurankan  I will make a final commit, with the _marginalize_single_variable code removed.

ankurankan:
@palashahuja And please use more specific commit messages that give a gist of changes that you have made in the commit.

palashahuja:
@ankurankan  So is the above commit mergeable?

pgmpy/factors/Factor.py
```diff
@@ -191,39 +191,26 @@ def marginalize(self, variables, inplace=True):
...
         else:
             factor = Factor(self.scope(), self.cardinality, self.values)
-
+        marginalize_index = np.array(np.where(np.in1d(factor.scope(),variables)))
+        assign = np.array(factor.cardinality)
+        assign[marginalize_index] = -1
+        sum_variable = np.zeros(len(factor.cardinality))
+        sum_variable[marginalize_index] = -1
+        marg_factor = []
+        for i in itertools.product(*[range(index) for index in factor.cardinality[np.where(assign!=-1)[0]]]):
```
ankurankan: PEP 8. Missing whitespace around operator

pgmpy/factors/Factor.py
```diff
@@ -191,39 +191,26 @@ def marginalize(self, variables, inplace=True):
...
+            temp_sum = 0
+            sum_variable[sum_variable!=-1] = i
```
ankurankan: PEP 8. Missing whitespace around operator

pgmpy/factors/Factor.py
```diff
@@ -191,39 +191,26 @@ def marginalize(self, variables, inplace=True):
             factor = self
         else:
             factor = Factor(self.scope(), self.cardinality, self.values)
-
+        marginalize_index = np.array(np.where(np.in1d(factor.scope(),variables)))
```
ankurankan: PEP 8. missing whitespace after ','

ankurankan:
@palashahuja Please always run pep8 check [1] on your changes before sending a PR.

[1] https://pypi.python.org/pypi/pep8/1.6.2

pgmpy/factors/Factor.py
```diff
@@ -18,7 +18,7 @@ class Factor:
     product(*Factor)
     reduce([variable_values_list])
     """
-
+
```
abinashpanda: PEP8 Trailing spaces.

pgmpy/factors/Factor.py
ankurankan: What not simply use `assign[assign != -1]` instead of `factor.cardinality[np.where(assign != -1)[0]]` ?

palashahuja: good point though. I am sure that the same thing can be done in the maximize routine also.

pgmpy/factors/Factor.py
```diff
@@ -191,39 +191,26 @@ def marginalize(self, variables, inplace=True):
...
+        assign = np.array(factor.cardinality)
+        assign[marginalize_index] = -1
+        sum_variable = np.zeros(len(factor.cardinality))
+        sum_variable[marginalize_index] = -1
+        marg_factor = []
+        for i in itertools.product(*[range(index) for index in factor.cardinality[np.where(assign!=-1)[0]]]):
+            temp_sum = 0
+            sum_variable[sum_variable!=-1] = i
+            for index in factor._index_for_assignment(sum_variable):
+                temp_sum += factor.values[index]
```
ankurankan: @palashahuja Can't we replace this loop with something like `temp_sum = np.sum(factor.values[factor._index_for_assignment(sum_variable)])` ??

palashahuja: @ankurankan Yes , we can . I will make the changes now.

pgmpy/factors/Factor.py
```diff
@@ -191,39 +191,26 @@ def marginalize(self, variables, inplace=True):
             factor = self
         else:
             factor = Factor(self.scope(), self.cardinality, self.values)
-
+        marginalize_index = np.array(np.where(np.in1d(factor.scope(),variables)))
+        assign = np.array(factor.cardinality)
+        assign[marginalize_index] = -1
+        sum_variable = np.zeros(len(factor.cardinality))
+        sum_variable[marginalize_index] = -1
```
ankurankan: And also why are you creating two variables `assign` and `sum_variable`? I think you should be able to work with only one of them.

palashahuja: Can we really do that ? Won't that change the value of assign inside the loop ?

ankurankan: Since you are using `assign[assing != -1]` in the for loop, it returns a
new object. So changing the value of `assign` inside the loop shouldn't
affect it.

pgmpy/factors/Factor.py
```diff
@@ -191,39 +190,22 @@ def marginalize(self, variables, inplace=True):
             factor = self
         else:
             factor = Factor(self.scope(), self.cardinality, self.values)
-
+        marginalize_index = np.array(np.where(np.in1d(factor.scope(), variables)))
+        assign = np.array(factor.cardinality)
+        assign[marginalize_index] = -1
+        marg_factor = []
```
ankurankan: @palashahuja I think we can have a better name for `marg_factor`. Maybe `new_values`? Any better name in your mind?

pgmpy/factors/Factor.py
```diff
@@ -191,39 +190,22 @@ def marginalize(self, variables, inplace=True):
...
+        for i in itertools.product(*[range(index) for index in assign[assign != -1]]):
+            assign[assign != -1] = i
+            temp_sum = np.sum(factor.values[factor._index_for_assignment(assign)])
+            marg_factor.append(temp_sum)
```
ankurankan: Merge lines 199 and 200 in one. No need to assign to `temp_sum`and then append it.

ankurankan:
@palashahuja Also please check if our `marginalize` tests are missing some edge cases. If missing please also write some tests covering that.

pgmpy/factors/Factor.py
palashahuja: @ankurankan I was using `marg_factor` maybe because `maximize` was also using the same. I have 
`marginalized-values` , `modified-values` in mind. I will replace with whichever you want to go with.

pgmpy/factors/Factor.py
palashahuja: @ankurankan please tell me the new-name for the `marg-factor` list so that I can make the appropriate changes.

pgmpy/factors/Factor.py
ankurankan: @palashahuja I don't think `maximize` is using `marg_factor` as any variable name. I think `marginalized_values` makes more sense.

palashahuja: @ankurankan Fine, I will make the changes then.

ankurankan:
@palashahuja Just check if you need to add more tests. Other than that I think PR is good to go.

palashahuja:
For the case 
`phi = Factor(['x1'],[2],np.arange(2))
 phi.marginalize('x1')`
Should it produce an empty list  or `0` for the value? If `0` then I have to make some changes.

ankurankan:
@palashahuja I think the current behavior of giving the total sum is correct.

palashahuja:
@ankurankan  So is it good to merge for now ?

ankurankan:
@palashahuja For future, you can use `[refs #issue]` and `[fixes #issue]` in your commit messages to reference commits in your issues or close issues.

Issue #276, opened 2015-01-30
Use prettytable for __repr__ methods


ankurankan:
prettytable: https://pypi.python.org/pypi/PrettyTable/0.7.2

kislayabhi:
Hey @ankurankan , I am on it. 

Any nuances that you might want to add...

abinash-panda:
@kislayabhi Just before you start working on it, lets discuss whether we want to add dependency for such a trivial task. Your views @ankurankan ?

ankurankan:
I really think this dependency would be worth it. Right now we have hard coded spaces between the variables while printing the `Factors`. So in many cases the representation doesn't look good at all. Also we are using the same method for `TabularCPD`  which I think should have a separate method which could print it properly. And I don't think that there is a simple way to do that.

So, we could either write something similar to prettytable for our needs or use prettytable because we will be needing this at other places as well.

abinash-panda:
My point against adding such trivial library as a dependency is that we don't want users to install another library just for printing `Factors` or `TabularCPD`. I have a good suggestion to use this library without asking the users to install it. The library is very small (21Kb), so we can add the code to our library. My proposal
- Create another submodule `extern` where we would add all the libraries that we would be using but we don't want users to install it such as `OpenGM` (if we are going to use it).
- Add the library over here.
- Import `prettytable` as `pgmpy.extern.prettytable`.

@ankurankan @kislayabhi Any suggestions?

kislayabhi:
Yeah. That appears cool as in a way we are using the same code but without bothering the users. It satisfies @ankurankan points too, I think.

ankurankan:
@abinash-panda @kislayabhi Only problem would be that we will have to manually update prettytables in our codebase but since it's nothing of much importance so I think we should be fine with that.

kislayabhi:
Looks like the last update of that library is quite old. It dates back to 2013. https://groups.google.com/forum/#!forum/prettytable-announce

ankurankan:
@kislayabhi Cool.. I think we can go with it then.

ankurankan:
#55 can also be handled along with this.

kislayabhi:
I am trying to get the pgmpy's way of handling TabularCPD. For creating something like this:

For a distribution of P(grade|intel, diff)

```
+-----------+--------------------+------------------+
|diff      |        easy         |    hard         |
+------------+-----+------+-------+------+----+------+
[...]

PR #289, opened 2015-02-20
Tabulate fix for the issue #276

kislayabhi:
Here I have tried to get the `tabulate` module working with that of ours. Some serious hacking done. Help me in cleaning this a bit. #276

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -5,6 +5,8 @@ import numpy as np
 import networkx as nx
 from pgmpy.factors import Factor
 from pgmpy import exceptions
+from itertools import repeat
+from pgmpy.extern import tabulate
 
 
 class TabularCPD(Factor):
@@ -63,6 +65,7 @@ class TabularCPD(Factor):
     normalize()
     reduce([values_list])
     """
+
     def __init__(self, variable, variable_card, values,
                  evidence=None, evidence_card=None):
 
@@ -116,14 +119,32 @@ class TabularCPD(Factor):
 
         super(TabularCPD, self).__init__(variables, cardinality, values.flatten('C'))
 
-    def get_cpd(self):
+    def get_cpd(self, pretty=True):
         """
         Returns the cpd
         """
         if self.variable in self.variables:
+            if pretty:
+                table_list = []
+                for i in range(len(self.cardinality))[1:]:
+                    var_rep = np.prod(self.evidence_card[i:])
+                    row = [self.scope()[-i]]
+                    row.extend(self.variables[row[0]] * var_rep)
+                    if i > 1:
+                        row[1:] = [x for item in row[1:] for x in
+                                   repeat(item, int(np.prod(self.evidence_card[0:i - 1])))]
+                    table_list.insert(0, row)
+
+                value_list = self.values.reshape(self.cardinality[0], np.prod(self.cardinality[1:])).tolist()
+                for item, grade in zip(value_list, self.variables[self.variable]):
+                    item.insert(0, grade)
+                table_list = table_list + value_list
+
+                print(tabulate(table_list, tablefmt="grid"))
             return self.values.reshape(self.cardinality[0], np.prod(self.cardinality[1:]))
         else:
             return self.values.reshape(1, np.prod(self.cardinality))
+
 
     def _repr_html_(self):
         """
@@ -303,6 +324,7 @@ class TreeCPD(nx.DiGraph):
     """
     Base Class for Tree CPD.
     """
+
     def __init__(self, data=None):
         """
         Base Class for Tree CPD.
@@ -535,6 +557,7 @@ class RuleCPD:
         assignment.
         """
         from itertools import combinations
+
         for rule, another_rule in combinations(self.rules, 2):
             rule, another_rule = (rule, another_rule) if len(rule) < len(another_rule) else (another_rule, rule)
             if not set(rule) - set(another_rule):
@@ -564,7 +587,7 @@ class RuleCPD:
         verify = self._verify()
         if not verify[0]:
             for rule in rules:
-                del(self.rules[rule])
+                del (self.rules[rule])
             raise ValueError(str(verify[1]) + " and " + str(verify[2]) + " point to the same assignment")
 
     def scope(self):
@@ -611,6 +634,7 @@ class RuleCPD:
         """
         from itertools import chain
         from collections import Counter
+
         assignments = list(set(chain(*self.rules)))
         cardinality = dict(Counter([element.split('_')[0] for element in assignments]))
         if variable:
@@ -649,8 +673,8 @@ class RuleCPD:
             start, end = 0, np.product(list(cardinality_dict.values()))
             for var in sorted(rule):
                 if var.split('_')[0] != self.variable:
-                    start, end = (start + (end-start)/cardinality_dict[var] * int(var.split('_')[1]),
-                                  start + (end-start)/cardinality_dict[var] * (int(var.split('_')[1]) + 1))
+                    start, end = (start + (end - start) / cardinality_dict[var] * int(var.split('_')[1]),
+                                  start + (end - start) / cardinality_dict[var] * (int(var.split('_')[1]) + 1))
                 else:
                     var_assignment = int(var.split('_')[1])
             for index in range(start, end):
@@ -710,13 +734,14 @@ class RuleCPD:
         <CPD.TreeCPD object at 0x7f6b6f952fd0>
         """
         from collections import OrderedDict
+
         tree_cpd = TreeCPD()
         merged_rules = OrderedDict(sorted(self._merge().items(), key=lambda t: len(t[0])))
 
         for assignments, value in merged_rules.items():
             for assignment_index in range(len(assignments) - 1):
                 tree_cpd.add_edge(assignments[assignment_index].split('_')[0],
-                                  assignments[assignment_index+1].split('_')[0],
+                                  assignments[assignment_index + 1].split('_')[0],
                                   assignments[assignment_index].split('_')[1])
             tree_cpd.add_edge(assignments[-1].split('_')[0],
                               Factor([self.variable], [len(value)], value),
@@ -725,6 +750,7 @@ class RuleCPD:
 
     def __str__(self):
         from collections import OrderedDict
+
         string = ""
         for index, key in enumerate(OrderedDict(sorted(self.rules.items(), key=lambda t: len(t[0])))):
             key_string = ', '.join(key)
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -5,7 +5,7 @@ import numpy as np
 import networkx as nx
 from pgmpy.factors import Factor
 from pgmpy import exceptions
-from itertools import repeat
+from itertools import repeat, product
 from pgmpy.extern import tabulate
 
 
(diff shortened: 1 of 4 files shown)


kislayabhi:
RIght now the results are a bit like this:

``` python
final_cpd = TabularCPD('grade', 4, 
                                 [[0.1,0.1,0.1,0.1,0.1,0.1,0.1,0.1,0.8,0.8,0.8,0.8],
                                  [0.1,0.1,0.1,0.1,0.1,0.1,0.1,0.1,0.8,0.8,0.8,0.8],
                                  [0.8,0.8,0.8,0.8,0.8,0.8,0.1,0.1,0.8,0.8,0.8,0.8],
                                  [0.1,0.1,0.1,0.1,0.1,0.1,0.1,0.1,0.8,0.8,0.8,0.8]],
                     evidence=['intel', 'diff', 'iq'],
                     evidence_card=[3, 2, 2])

final_cpd.get_cpd(pretty=True)
```

git

abinash-panda:
@kislayabhi I have few concerns
- Firstly, this code should not be in `get_cpd`. The whole motive of `get_cpd` is to get the tabular cpd array. This should be present in ``__str__` method. This method is  called when you use `print` function.
- Secondly does tabulate also help in producing HTML tables.

ankurankan:
@kislayabhi I think a better place to keep this code would be in `__str__` method of `TabularCPD` because right now `get_cpd` is printing the table as well as returning the array of values.

pgmpy/factors/CPD.py
```diff
@@ -116,15 +119,33 @@ def __init__(self, variable, variable_card, values,
...
-    def get_cpd(self):
+    def get_cpd(self, pretty=True):
         """
         Returns the cpd
         """
         if self.variable in self.variables:
+            if pretty:
+                table_list = []
+                for i in range(len(self.cardinality))[1:]:
+                    var_rep = np.prod(self.evidence_card[i:])
```
abinash-panda: Don't compute the product in each loop. It would be better to use `np.cumprod` method and save these values apriori to the loop.

pgmpy/factors/CPD.py
```diff
@@ -116,15 +119,33 @@ def __init__(self, variable, variable_card, values,
...
+                    row = [self.scope()[-i]]
+                    row.extend(self.variables[row[0]] * var_rep)
+                    if i > 1:
+                        row[1:] = [x for item in row[1:] for x in
+                                   repeat(item, int(np.prod(self.evidence_card[0:i - 1])))]
```
abinash-panda: Same as above. Try to reduce the call to `prod` method as much as possible.

pgmpy/factors/CPD.py
```diff
@@ -116,15 +119,33 @@ def __init__(self, variable, variable_card, values,
...
+                    table_list.insert(0, row)
+
+                value_list = self.values.reshape(self.cardinality[0], np.prod(self.cardinality[1:])).tolist()
+                for item, grade in zip(value_list, self.variables[self.variable]):
+                    item.insert(0, grade)
+                table_list = table_list + value_list
```
abinash-panda: Please don't extend table like this. `+` operation always create a new object which is very slow. Instead use `table_list.extend(value_list)`.

kislayabhi:
Yeah. I agree. I will get the code in `__str__` method.

@abinash-panda There is a mention of `html` in tabulate library under the table formatting options. So I think yes.

pgmpy/factors/CPD.py
```diff
@@ -116,15 +119,33 @@ def __init__(self, variable, variable_card, values,
...
+                    row = [self.scope()[-i]]
+                    row.extend(self.variables[row[0]] * var_rep)
+                    if i > 1:
+                        row[1:] = [x for item in row[1:] for x in
+                                   repeat(item, int(np.prod(self.evidence_card[0:i - 1])))]
+                    table_list.insert(0, row)
+
+                value_list = self.values.reshape(self.cardinality[0], np.prod(self.cardinality[1:])).tolist()
+                for item, grade in zip(value_list, self.variables[self.variable]):
+                    item.insert(0, grade)
```
abinash-panda: I have a hunch that this thing can be implement better. This doesn't feel very robust. Let me know if you need any help in this.

kislayabhi: Yes. I feel that too. Let me go through your above mentioned comments first. I think that may clean this a bit.

abinash-panda:
@kislayabhi So here is my suggestion. 
- Create a generic method such as `_str(html=False)`. This method can be used both for `__str__` (required from printing) as well as `_repr_html_` method (required for creating HTML tables in IPython Notebook).
- In the `__str__` method call `_str(html=False)` whereas in `_repr_html_` method call `_str(html=True)`. This would help in removing code redundancy as well as use the functionality of `tabulate` to create tables instead of writing our own.
- Once this exercise is completed we can also create the same thing for `Factor` class as well.

abinash-panda:
@kislayabhi One more thing before going to change the `_repr_html_` method just check whether we are able to produce the same kind of HTML table that we are able to produce now using `tabulate`. If not lets discuss before working on it. Let me know if you have any doubt.

pgmpy/factors/CPD.py
```diff
@@ -116,15 +119,33 @@ def __init__(self, variable, variable_card, values,
...
 
-    def get_cpd(self):
+    def get_cpd(self, pretty=True):
         """
         Returns the cpd
         """
         if self.variable in self.variables:
+            if pretty:
+                table_list = []
+                for i in range(len(self.cardinality))[1:]:
```
ankurankan: Would be better to simply use `range(1, len(self.cardinality))`. It would be much faster and also is more readable.

pgmpy/factors/CPD.py
ankurankan: You are adding values in `row` in Line 132 and then replacing them in line 134 except for one case when `i = 1`. There should be a generalized way to do this and if not you can atleast put an if-else.

kislayabhi: Here I need to convert an arrays like `["diff","diff_0","diff_1"]` into `["diff","diff_0","diff_0","diff_1","diff_1"]`.

This was what I found after some pondering. 
Any more pythonic way to achieve something like this in general.?

pgmpy/factors/CPD.py
ankurankan: @kislayabhi: Check `numpy.column_stack`. You should be able to replace this loop with a single statement.

pgmpy/factors/CPD.py
ankurankan: @kislayabhi: Maybe try something in these lines:

``` python
indexes = np.array(list(product(*[range(i) for i in self.cardinality[1:]])))
```

and then have a loop on columns of `indexes` to replace them:

``` python
indexes[:, 0] = np.array(self.variables['iq'])[np.atleast_2d(indexes[:, 0])]
```

And then finally get the transpose and then could use `np.column_stack` to append the variable names.

I feel that this method should be a cleaner. What do you think?

kislayabhi: @ankurankan: what is `product` doing here?

abinashpanda: @kislayabhi `product` is `itertools.product`

kislayabhi:
I have taken care of most of the comments. However, `_repr_html_` is left. I am working on it.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +127,19 @@ def get_cpd(self):
...
 
+    def __str__(self):
+        return self._str(html=False)
+
+    def _str(self, html=False):
+        table_list = []
+        indexes = np.array(list(product(*[range(i) for i in self.cardinality[1:]])))
+        for i in range(1, len(self.cardinality)):
+            row = self.scope()[i]
+            row = [row]+np.array(self.variables[row])[indexes[:, i-1]].tolist()
```
abinashpanda: Please don't add list to create a new list. It would always result in creation of new object. Instead use `row.extend(np.array(self.variables[row])[indexes[:, i-1]].tolist())`.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +127,19 @@ def get_cpd(self):
...
             return self.values.reshape(1, np.prod(self.cardinality))
 
+    def __str__(self):
+        return self._str(html=False)
+
+    def _str(self, html=False):
+        table_list = []
+        indexes = np.array(list(product(*[range(i) for i in self.cardinality[1:]])))
+        for i in range(1, len(self.cardinality)):
+            row = self.scope()[i]
```
abinashpanda: In each iteration you are calling the method `scope()` which is an overhead. Instead store `scope = self.scope()` prior to the loop.

abinashpanda:
@kislayabhi Please add the LICENSE of this module as well in the `extern` module as well as `LICENCE_TABULATE`.

abinashpanda:
@kislayabhi Once you have also completed the `_repr_html_()` method, we can merge it.

kislayabhi:
@abinash-panda @ankurankan : There is a difference in the output of "html" tables for tabulate and ours.
Below is that of tabulate:

```
-------  -------  -------  -------  -------  -------  -------  -------  -------  -------  -------  -------  -------
iq       iq_0     iq_0     iq_0     iq_0     iq_0     iq_0     iq_1     iq_1     iq_1     iq_1     iq_1     iq_1
diff     diff_0   diff_0   diff_0   diff_1   diff_1   diff_1   diff_0   diff_0   diff_0   diff_1   diff_1   diff_1
intel    intel_0  intel_1  intel_2  intel_0  intel_1  intel_2  intel_0  intel_1  intel_2  intel_0  intel_1  intel_2
grade_0  0.1      0.1      0.1      0.1      0.1      0.1      0.1      0.1      0.8      0.8      0.8      0.8
grade_1  0.1      0.1      0.1      0.1      0.1      0.1      0.1      0.1      0.8      0.8      0.8      0.8
grade_2  0.8      0.8      0.8      0.8      0.8      0.8      0.1      0.1      0.8      0.8      0.8      0.8
grade_3  0.1      0.1      0.1      0.1      0.1      0.1      0.1      0.1      0.8      0.8      0.8      0.8
-------  -------  -------  -------  -------  -------  -------  -------  -------  -------  -------  -------  ------- 
```

The current implementation of html tables in pgmpy is much awesomer than this!!

abinashpanda:
@kislayabhi Yeah. The current implementation is much better than this. So no need to change it. What are your views @ankurankan ?

abinashpanda:
@kislayabhi Add the LICENSE as well. The PR is good to merge.

kislayabhi:
@abinashpanda: Here I have just added the tabulate license file that came along with its tarball in the extern module. Is there more to add?

abinashpanda:
@kislayabhi Yeah its enough.

abinashpanda:
@kislayabhi If you want to make the similar changes to `Factor` class as well.

kislayabhi:
@abinashpanda  You mean the pretty printing of tables in Factor class?

abinashpanda:
@kislayabhi Yeah. pretty printing for `Factor` class. It would be straight forward.

kislayabhi:
:+1:

ankurankan:
@kislayabhi Could you look into some styling options for tabulate? There are a few on the https://pypi.python.org/pypi/tabulate like using double lines to separate column names from values with double lines. I think that would make our tables much more cleaner. @abinashpanda Your opinion?

abinashpanda:
@kislayabhi @ankurankan Yeah. I am fine with this.

PR #294, opened 2015-02-27
Fixes Factor __eq__ method

ankurankan:


pgmpy/factors/Factor.py
```diff
@@ -496,11 +496,24 @@ def __eq__(self, other):
...
-        if self.variables == other.variables and all(self.cardinality == other.cardinality) \
-                and all(self.values == other.values):
-            return True
-        else:
+        if set(self.scope()) != set(other.scope()):
             return False
+        else:
+            order_dict = {}
+            for var_index in range(len(self.scope())):
+                order_dict[var_index] = other.scope().index(self.scope()[var_index])
```
abinashpanda: Can you make it more readable. Currently its very confusing and scary at the first look.

pgmpy/factors/Factor.py
```diff
@@ -496,11 +496,24 @@ def __eq__(self, other):
...
+            if any(self.cardinality[[order_dict[i] for i in range(len(self.cardinality))]] != other.cardinality):
```
abinashpanda: Phew!! The same thing here also.

pgmpy/factors/Factor.py
```diff
@@ -496,11 +496,24 @@ def __eq__(self, other):
...
+                return False
+            indexes = np.array(list(itertools.product(*[range(i) for i in self.cardinality])))
```
abinashpanda: I am not very supportive of converting iterator to `list` and then converting to `np.array`. Have a look at `np.fromiter`. Converting to `list` would not be very scalable (I think for large factors it would suck the whole memory). Your thoughts?

pgmpy/factors/Factor.py
```diff
@@ -496,11 +496,24 @@ def __eq__(self, other):
...
+            transformed_assign = np.zeros(indexes.shape, dtype=np.int)
+            for k, v in order_dict.items():
+                transformed_assign[:, v] = indexes[:, k]
+            transformed_indexes = np.array(list(map(
+                    other._index_for_assignment, transformed_assign))).ravel()
```
abinashpanda: The same thing as suggested above. Your thoughts?

pgmpy/factors/Factor.py
ankurankan: @abinashpanda I agree.. Fixing this.

ankurankan: @abinashpanda The problem with `np.fromiter` is that we can't create 2D arrays using it. So, the solution that I could find is to either `chain` our 2D array, use `fromiter` and then reshape it or use structured arrays. With structured arrays we lose indexing functionality because it creates an array of tuples and therefore numpy considers it as a 1D array.

Comparison current method and chain:

``` python
%timeit np.array(list(itertools.product(*[range(i) for i in range(2, 10)])))
1 loops, best of 3: 201 ms per loop

%timeit np.fromiter(itertools.chain(*itertools.product(*[range(i) for i in range(2, 10)])), np.int).reshape(np.prod(range(2, 10)), len(range(2, 10)))
10 loops, best of 3: 121 ms per loop

Line #    Mem usage    Increment   Line Contents
================================================
    10   26.410 MiB    0.000 MiB   @profile
    11                             def my_fun2():
    12   26.410 MiB    0.000 MiB       a = itertools.product(*[range(i) for i in range(2, 10)])
    13   65.660 MiB   39.250 MiB       b = list(a)
    14   65.660 MiB    0.000 MiB       c = np.array(a)


Filename: testing.py

Line #    Mem usage    Increment   Line Contents
================================================
     4   23.352 MiB    0.000 MiB   @profile
     5                             def my_fun():
     6   23.352 MiB    0.000 MiB       a = itertools.product(*[range(i) for i in range(2, 10)])
     7   66.180 MiB   42.828 MiB       b = itertools.chain(*a)
     8   26.406 MiB  -39.773 MiB       np.fromiter(b, np.int).reshape(np.prod(range(2, 10)), len(range(2, 10)))
```

So creating a list is slower and using `chain` is using almost the same amount of memory as creating a list but we are losing readability. So, which method do you think we should go with?

abinashpanda: Okay. Here are my thoughts: 
- See the whole purpose of using `numpy.array` is not just indexing it is also about memory efficiency by having homogeneous data structure with fixed memory size. Creating a list would diminish the later advantage.
- Furthermore converting a generator into a list also looses the advantage of generator over the list.
- We can create this array with very basic implementation (without harming the code readability). We would be knowing the size of array in advance. So just create an array of zeros of that size (`np.zeros(...)`) and fill the elements as they are generated on the fly.
- I think (not pretty sure) that it would be efficient both in terms of memory and time. 

What are your thoughts on this?

pgmpy/factors/Factor.py
ankurankan: @abinashpanda Suggest me something to make it more simpler. Assigning `self.scope()[var_index]` to a new variable and then using it doesn't make much sense to me.

abinashpanda: @ankurankan In each iteration we are making a call to `scope()` method. Instead we should save the scope of the factor apriori and use it in each iteration and the same thing applies for `other.scope()`.

abinashpanda:
@ankurankan Apart from that every thing looks good. Make the above stated changes and then we can merge.

pgmpy/factors/Factor.py
ankurankan: @abinashpanda Oops.. Fixing this.

Issue #255, opened 2015-01-07
Make changes to marginalize and reduce


ankurankan:
Marginalize and reduce methods have some redundant code that can be replaced with `_index_for_assignment`

palashahuja:
I have tried to replace marginalize over single variable with _index_for_assignment method in the [following gist-](https://gist.github.com/palashahuja/488050fac7a03b4acfa8). Please have a look.
@ankurankan.,@abinashpanda

abinash-panda:
@palashahuja Send a PR with the changes and make sure the tests pass. It is easier to review the code.

palashahuja:
Sure I will send it.

palashahuja:
@abinash-panda  What is the error ? Can you please explain ?

ankurankan:
Fixed in #316

Issue #222, opened 2014-11-04
Bug in Factor's reduce method


ankurankan:
``` python3
f = Factor(['x1'], [2], [0.4, 0.6])
f.reduce('x1_0')
f.values
Out[32]: array([ 0.4])

f.variables
Out[33]: OrderedDict()

f.cardinality
Out[35]: array([], dtype=int64)
```

Expected Behavior:

``` python3
f = Factor(['x1'], [2], [0.4, 0.6])
f.reduce('x1_0')

f.values
Out: array([0.4])

f.variables
Out: OrderedDict([('x1', ['x1_0'])])

f.cardinality
Out: array([1])
```

palashahuja:
@ankurankan expected behaviour for 

``` python3
f = Factor (['x1','x2','x3'], [2,3,2],np.arange (12))
f.reduce (['x1_0','x2_0','x3_0'])
```

ankurankan:
@palashahuja 

``` python3
f.variables
Out: OrderedDict([('x1', ['x1_0']), ('x2', ['x2_0']), ('x3', ['x3_0'])])

f.cardinality
Out: array([1, 1, 1])

f.values
Out: array([0])
```

palashahuja:
@ankurankan 
Could you please explain why do all the terms should not reduce to null value when they are reduced simultaneously. Instead,what we can do  is make some changes in the `_repr_` function  so that when the table doesn't contain any value, there should be a default error statement. Also, should I add the above two cases in the test cases ?

abinashpanda:
@palashahuja Reduce means reducing the whole distribution to some context. For example when we reduce a factor f (as you have defined above) to the context of `x1_0`, we should be only having those entries where `x1` had assignment `0`. So if you reduce the whole table to the context of `[x1_0, x2_0, x3_0]`, the table should contain value corresponding to the assignment `x1 = 0`, `x2 = 0` and `x3 = 0`. It should not reduce to null. 

I hope this explanation helps. Let me know if you still have any doubt.

ankurankan:
@palashahuja From the docstring of `reduce`: `Reduces the factor to the context of given variable values.`. So in the above example when we are trying to reduce the factor to `[x1_0, x2_0, x3_0]` we want to remove all the values from the factor that are not consistent with these states which leaves us with a single value in the factor. And this also reduces the cardinalities to `[1, 1, 1]` because for each variable we only have one state.

palashahuja:
[...]

PR #297, opened 2015-02-28
reduced the redundant code in reduce [refs #255]

palashahuja:


palashahuja:
@ankurankan Please let me know any changes in the name of variables that have to be made. Also, the above commit removes bug #222 .
Also,
- There are no pep8 errors for now.
- I couldn't think of a better way to delete elements from a dictionary other than using a `for` loop.
- If there is one element that is to be reduced, then what should `factor.values` be replaced by? Although, I have assigned it an empty `np.array([],dtype = np.int32)`

palashahuja:
Also in the 
`if not all((itertools.starmap(lt,zip(value_indices,phi.cardinality[reduced_indices]))))` statement takes around `7.56us` while `if not all(value_indices < phi.cardinality[reduced_indices])` takes `10.4us`, although the former is bit longer and is quite difficult to read. What should I use ? .

ankurankan:
@palashahuja: About #222, I have updated the issue with the expected output. Please fix your code accordingly.

pgmpy/factors/Factor.py
```diff
@@ -261,28 +262,26 @@ def reduce(self, values, inplace=True):
...
+            
+        if not all('_' in value for value in values):
+            raise TypeError("Values should be in the form of variablename_index")
+        value_row = list(zip(*[value.split('_') for value in values]))
+        reduced_variables = list(value_row[0])
+        value_indices = list(map(int, value_row[1]))
+        if not set(reduced_variables).issubset(set(factor.scope())):
+            raise Exceptions.ScopeError("%s not in scope" % list(set(reduced_variables)-set(factor.scope())))
+        reduced_indices = np.where(np.in1d(factor.scope(), reduced_variables))
+        if not all(itertools.starmap(lt, zip(value_indices, (factor.cardinality)[reduced_indices]))):
```
ankurankan: Redundant parenthesis in `(factor.cardinality)`

ankurankan:
@palashahuja Also add a few blank lines in between to separate parts of code. Presently it looks too jumbled up.

palashahuja:
@ankurankan Please let me know about all the changes. So that I can stash all the changes in one single commit.

ankurankan:
@palashahuja And I think you can go with using `itertools.starmap`, it's not that complex and is save a good amount of time. 

Make these changes and I think the PR should be good to go.

palashahuja:
I have already fixed the code though but I am facing the following conundrum :

```
In [6]:phi = Factor(['x1','x2','x3'],[2,3,2],np.arange(12))
In [7]:phi.reduce(['x1_0','x2_1','x3_1'])
In [8]:print(phi)
x1      x2      x3      phi(x1, x2, x3)
------------------------------------------------------
x1_0        x2_0        x3_0        3.0000
```

although

```
In [9]:phi.variables
OrderedDict([('x1', ['x1_0']), ('x2', ['x2_1']), ('x3', ['x3_1'])])
In [10]:phi.cardinality
array([1, 1, 1])
```

Am I doing something wrong over here or is that what you are exactly expecting ?

abinashpanda:
@palashahuja This is the correct behaviour.

ankurankan:
@palashahuja The behavior of reduce is correct but there's a bug in `__str__` function #302

palashahuja:
@ankurankan @abinashpanda  any further changes , please have a look.

ankurankan:
@palashahuja Could you please also add a few more tests for `reduce` method?

palashahuja:
@ankurankan Sure no problem...

pgmpy/factors/Factor.py
```diff
@@ -261,27 +262,31 @@ def reduce(self, values, inplace=True):
...
+        if not set(reduced_variables).issubset(set(factor.scope())):
+            raise Exceptions.ScopeError('%s not in scope' % list(set(reduced_variables)-set(factor.scope())))
+
+        reduced_indices = np.where(np.in1d(factor.scope(), reduced_variables))
+        if not all(itertools.starmap(lt, zip(value_indices, factor.cardinality[reduced_indices]))):
+            raise Exceptions.SizeError('Value is greater than max possible value')
+
+        reduce_assign = np.full(len(factor.cardinality), -1)
+        reduce_assign[reduced_indices] = value_indices
+        factor.values = factor.values[list(map(int,factor._index_for_assignment(reduce_assign)))]
```
abinashpanda, line 281: PEP8 space after `,`. @palashahuja please use pep8 checker before submitting a PR.

PR #298, opened 2015-02-28
reduced the redundant code in reduce [refs #255]

palashahuja:


abinashpanda:
@palashahuja Please don't create and close multiple PRs. It becomes really tough to track. It would be better if you could keep 1 PR and commit all your changes over there. This way we could keep track all the comments and modifications in a single place.

PR #307, opened 2015-03-03
adding test case in reduce [refs #297]

palashahuja:
Test when reducing all the variables simultaneously


ankurankan:
@palashahuja Don't worry about failing travis. I had to revert back your previous PR of reduce because some tests were failing because of that. I will look into that.

palashahuja:
@ankurankan  But I have tried the same reduce code on my local machine and it doesn't seem to give any error as such. AFAIK , there is some problem with `exceptions` which I was facing initially and then resolved out magically.

ankurankan:
@palashahuja Something weird is happening. I haven't been able to yet figure it out. Give me some time to investigate.

ankurankan:
@palashahuja Please add a check for the variables of the reduced factor also.

Issue #248, opened 2014-12-24
Parallel for loops in predict method


ankurankan:

gokul-uf:
Hi @ankurankan, could you give me more info about this?

ankurankan:
@gokul-uf: So there is still a pending PR for predict method for Bayesian Network (and I don't remember it's state whether it was working or not). And I think it was iterating over data to calculate predictions. So we want to either replace this with vectorization or at least make the iteration parallel by using `joblib`

ankurankan:
Creating new issue for vectorizing predict. Closing this.

PR #312, opened 2015-03-06
Parallel for implemenation [issue #248]

gokul-uf:
I have added support for parallel for loops [issue #248]
This created a dependency on the joblib library.

I have some questions
1.    What should be the optimal value for the n_jobs parameter? A very small value won't result in significant performance gains. On the other hand, a large value would cause high thread switching overheads. Do we have a default hard-coded value or do we decide based on the size of the model?
2.    Do we follow this approach to parallelize all the for loops or are there iterations which cannot be parallelized or doesn't need to be parallelized?

NOTE: The docs for joblib warn us about use cases in Windows (https://pythonhosted.org/joblib/parallel.html#id1). Need to be careful if we are going ahead with using jobib

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import numpy as np
+from joblib import Parallel,delayed
 from pgmpy.inference import Inference
 from pgmpy.factors.Factor import factor_product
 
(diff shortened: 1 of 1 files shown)


ankurankan:
@gokul-uf: `n_jobs` parameter should have a default value of 1. No we don't need to make all our for loops parallel as it is not possible in all the cases and also we don't need it everywhere. In this the user may want to predict for a lot of data and this can surely be done in parallel so makes sense in this case.

palashahuja:
But what should be the value of `n_jobs`
@ankurankan

ankurankan:
Depends on how much you want to run the operation in parallel. So I have a 8-core machine so I at times use all the cores to make things parallel. Also joblib accepts `-1` as the value to use all the cores.

ankurankan:
@gokul-uf Please always run pep8 on your code before sending PRs and also check landscape-bot's comments.

PR #316, opened 2015-03-07
Improved reduce code (@palashahuja's changes)

ankurankan:


pgmpy/factors/Factor.py
```diff
@@ -261,27 +262,31 @@ def reduce(self, values, inplace=True):
...
+        if not set(reduced_variables).issubset(set(factor.scope())):
+            raise Exceptions.ScopeError('%s not in scope' % list(set(reduced_variables)-set(factor.scope())))
+
+        reduced_indices = np.where(np.in1d(factor.scope(), reduced_variables))
+        if not all(itertools.starmap(lt, zip(value_indices, factor.cardinality[reduced_indices]))):
+            raise Exceptions.SizeError('Value is greater than max possible value')
+
+        reduce_assign = np.full(len(factor.cardinality), -1)
+        reduce_assign[reduced_indices] = value_indices
+        factor.values = factor.values[list(map(int, factor._index_for_assignment(reduce_assign)))]
```
ankurankan: @palashahuja Why have you used map here? I think `_index_for_assignment` returns an array of dtype=int only.

pgmpy/factors/Factor.py
```diff
@@ -261,27 +262,31 @@ def reduce(self, values, inplace=True):
...
+        reduced_variables = list(value_row[0])
+        value_indices = list(map(int, value_row[1]))
+        if not set(reduced_variables).issubset(set(factor.scope())):
+            raise Exceptions.ScopeError('%s not in scope' % list(set(reduced_variables)-set(factor.scope())))
+
+        reduced_indices = np.where(np.in1d(factor.scope(), reduced_variables))
+        if not all(itertools.starmap(lt, zip(value_indices, factor.cardinality[reduced_indices]))):
+            raise Exceptions.SizeError('Value is greater than max possible value')
+
+        reduce_assign = np.full(len(factor.cardinality), -1)
```
ankurankan: @palashahuja Please use `dtype=int` argument to have the elements as integers. No need for having them as float.

ankurankan:
@palashahuja I have fixed the errors. So basically we had changed the behavior of reduce to have `variables` attribute even when all the variables were reduced and because of this behavior `map_query` was throwing errors. So I have fixed these. Have a look at my last commit of this PR to understand the changes better.

palashahuja:
@ankurankan  So is that how reduce is supposed to behave from now on ? But then that will create errors in **str** function . It is good for now though but there should be something to indicate the null values while printing it...

ankurankan:
@palashahuja Yeah even I think that there should be something that should represent this. Although `__str__` would still work but won't show any variables. Any ideas for how should we represent this situation?

palashahuja:
@ankurankan  Maybe raise a valueerror showing that the Factor doesn't contain any values any more ... I think that will be good for now ...

palashahuja:
@ankurankan  Closes a couple of other bugs too ... bug #255 #222

ankurankan:
@palashahuja I don't think raising an error is a good idea. Let leave this for now. We will get a better perspective of what we should do once we run into more issues regarding this.

palashahuja:
@ankurankan  okay :+1:

PR #320, opened 2015-03-08
Added the tests for ProbModelXML. Also removed some KeyError.

vivek425ster:


ankurankan:
@vivek425ster Could you please fix the errors in tests?

vivek425ster:
@ankurankan When i try to run tests in my local system they are running successfully. 
What is the problem here ?
I am running tests using nosetests3

gokul-uf:
@vivek425ster 
Here's what Travis reports

```
ERROR: test_comment (pgmpy.tests.test_readwrite.test_ProbModelXML.TestProbModelXMLReaderString)

----------------------------------------------------------------------

File "/home/travis/build/pgmpy/pgmpy/pgmpy/tests/test_readwrite/test_ProbModelXML.py", line 94, in setUp

self.reader_string = ProbModelXMLReader(string=string)

File "/home/travis/build/pgmpy/pgmpy/pgmpy/readwrite/ProbModelXML.py", line 439, in __init__

self.create_probnet()

File "/home/travis/build/pgmpy/pgmpy/pgmpy/readwrite/ProbModelXML.py", line 459, in create_probnet

for variable in probnet_elem.find('Variables').iterchildren():

AttributeError: 'xml.etree.ElementTree.Element' object has no attribute 'iterchildren'
```

This should hopefully help
 http://stackoverflow.com/questions/7684333/converting-xml-to-dictionary-using-elementtree

ankurankan:
I think this might be because of dependencies. Check the  implementation of
probmodel we have used several fallback modules. So you might be having
some module in your local system that is being used for parsing but Travis
doesn't have that module, it falls back to some other (looks like falls
back to xml from the Travis error) and that module doesn't have that
attribute that you are using because of which it's throwing error.

So avoid using that attribute so that it also works with all the fall back
modules.

vivek425ster:
@ankurankan Why are we using xpath instead of find ?
If we use find instead of xpath then also it will work fine

ankurankan:
Why do you think that should solve the problem? AFAIR the error was because
of etree not having iterchildren.

I am not really sure but I think that xpath should be  computationally much
less expensive than find and in cases of huge files this will make a huge
difference.

vivek425ster:
@ankurankan But the iterchildren method is giving error in the travis build.
I am not able to understand the error as i am unable to reproduce in my system.
Is there any way to reproduce it on my system ?

ankurankan:
Try uninstalling lxml from your system and you should be able to see the
error.

vivek425ster:
@ankurankan Since lxml has no iterchildern method. 
Should we use iterchildren or not ?

ankurankan:
lxml does support iterchildren and you have lxml installed on your system
because of which tests are passing on your system but since Travis doesn't
have lxml installed so it tries to use the xml module which doesn't support
iterchildren and that's why tests fail.

So replace iterchildren with something that is supported by both lxml and
xml.

vivek425ster:
Also there is Xpath method in lxml but not in xml module. 
http://lxml.de/xpathxslt.html
In xml documentation they have mentioned xpath queries used with findall method.
http://docs.activestate.com/activepython/3.3/python/library/xml.etree.elementtree.html#elementtree-xpath
Should i replace it xpath method with these findall with xpath queries ?

ankurankan:
@vivek425ster I think it should work. Both lxml and xml support `find` and `findall` methods which can take xpath strings as arguments, so that should work.

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +486,33 @@ def add_node(self, variable):
         variable_name = variable.attrib['name']
         self.probnet['Variables'][variable_name] = {}
         self.probnet['Variables'][variable_name]['type'] = variable.attrib['type']
-        self.probnet['Variables'][variable_name]['roles'] = variable.attrib['roles']
-        if variable.find('Comment'):
+        self.probnet['Variables'][variable_name]['role'] = variable.attrib['role']
+        if variable.find('Comment') is not None:
```
abinashpanda, line 503: @vivek425ster 

``` python
if variable.find('Comment') is not None
```

is not **Pythonthic**. Instead just keep it as

``` python
if variable.find('Comment')
```

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +486,33 @@ def add_node(self, variable):
...
             self.probnet['Variables'][variable_name]['Comment'] = variable.find('Comment').text
-        if variable.find('Coordinates'):
+        if variable.find('Coordinates') is not None:
```
abinashpanda, line 505: Same as above.

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +486,33 @@ def add_node(self, variable):
...
             self.probnet['Variables'][variable_name]['Coordinates'] = variable.find('Coordinates').attrib
-        if variable.xpath('AdditionalProperties'):
-            for prop in variable.find('AdditionalProperties').iterchildren():
+        if variable.find('AdditionalProperties/Property') is not None:
```
abinashpanda, line 507: Same as above.

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +486,33 @@ def add_node(self, variable):
...
+            for prop in variable.findall('AdditionalProperties/Property'):
                 self.probnet['Variables'][variable_name]['AdditionalProperties'][prop.attrib['name']] = \
                     prop.attrib['value']
-        if not variable.xpath('States'):
+        if variable.find('States/State') is None:
```
abinashpanda, line 511: Same as above.

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +486,33 @@ def add_node(self, variable):
...
+            self.probnet['Variables'][variable_name]['States'] = {state.attrib['name']: {prop.attrib['name']: prop.attrib['value'] for prop in state.findall('AdditionalProperties/Property')} for state in variable.findall('States/State')}
 
     def add_edge(self, edge):
         var1 = edge.attrib['var1']
         var2 = edge.attrib['var2']
+        self.probnet['edges'] = {}
         self.probnet['edges'][(var1, var2)] = {}
         self.probnet['edges'][(var1, var2)]['directed'] = edge.attrib['directed']
-        if edge.xpath('Comment'):
+        if edge.find('Comment') is not None:
```
abinashpanda, line 522: Same as above.

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +486,33 @@ def add_node(self, variable):
...
         #TODO: check for the case of undirected graphs if we need to add to both elements of the dic for a single edge.
-            self.probnet['edges'][(var1, var2)]['Comment'] = edge.xpath('Comment').text
-        if edge.xpath('Label'):
-            self.probnet['edges'][(var1, var2)]['Label'] = edge.xpath('Label').text
-        if edge.xpath('AdditionalProperties'):
-            for prop in edge.xpath('AdditioanlProperties')[0].iterchildren():
+            self.probnet['edges'][(var1, var2)]['Comment'] = edge.find('Comment').text
+        if edge.find('Label') is not None:
```
abinashpanda, line 524: Same as above.

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +486,33 @@ def add_node(self, variable):
...
+            self.probnet['edges'][(var1, var2)]['Label'] = edge.find('Label').text
+        if edge.find('AdditionalProperties/Property') is not None:
```
abinashpanda, line 526: @vivek425ster Correct all the places where explicitly `is not None` is mentioned.

pgmpy/tests/test_readwrite/test_ProbModelXML.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python
+import unittest
+from io import StringIO
+import networkx as nx
+from pgmpy.readwrite import ProbModelXMLWriter, ProbModelXMLReader, parse_probmodelxml, generate_probmodelxml
+try:
+    from lxml import etree
+except ImportError:
+    try:
+        # Python 2.5
```
abinashpanda: @vivek425ster Why Python 2.5? `pgmpy` is only compatible with Python 3.

pgmpy/tests/test_readwrite/test_ProbModelXML.py
```diff
@@ -0,0 +1,101 @@
...
+                import cElementTree as etree
+                print("running with cElementTree")
+            except ImportError:
+                try:
+                    # normal ElementTree install
+                    import elementtree.ElementTree as etree
+                    print("running with ElementTree")
+                except ImportError:
+                    print("Failed to import ElementTree from any known place")
+
```
ankurankan, line 17: @vivek425ster And don't you think that using `warnings` is better than `print` statements?

ankurankan:
@vivek425ster And also please add more tests for this. Like comparing variables, edges etc.

pgmpy/readwrite/ProbModelXML.py
vivek425ster, line 503: If i use if variable.find('Comment') then while running tests it gives me following warning:

FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if variable.find('Comment'):

pgmpy/tests/test_readwrite/test_ProbModelXML.py
vivek425ster: In test_PomdpX also it is mentioned liked this. Therefore i have used it.
https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/tests/test_readwrite/test_PomdpX.py#L14
Should i change it in that test file also ?

ankurankan:
@vivek425ster Tests are failing. Could you please look into it?

vivek425ster:
@ankurankan  Whenever i try to print variables everytime the order changes. 
Is there any specific reason for this ?

ankurankan:
@vivek425ster Are you trying to print dictionaries?

vivek425ster:
Yes

ankurankan:
Dictionaries don't store the keys in any specific order. Read the docs for python `dict`. 
And you can use `assertDictEqual` to compare the values of two dicts.

palashahuja:
use an ordered dict

ankurankan:
@palashahuja: We shouldn't be using OrderedDict here. No need for that.

pgmpy/readwrite/ProbModelXML.py
```diff
@@ -483,32 +487,32 @@ def add_node(self, variable):
...
-        if edge.xpath('Comment'):
+        if edge.find('Comment'):
         #TODO: check for the case of undirected graphs if we need to add to both elements of the dic for a single edge.
-            self.probnet['edges'][(var1, var2)]['Comment'] = edge.xpath('Comment').text
-        if edge.xpath('Label'):
-            self.probnet['edges'][(var1, var2)]['Label'] = edge.xpath('Label').text
-        if edge.xpath('AdditionalProperties'):
-            for prop in edge.xpath('AdditioanlProperties')[0].iterchildren():
+            self.probnet['edges'][(var1, var2)]['Comment'] = edge.find('Comment').text
+        if edge.find('Label'):
```
vivek425ster: @abinashpanda @ankurankan 
If  i use  
    if edge.find('Label'): 
then it is not getting inside if statement
but if i use 
    if edge.find('Label') is not None: 
then it is getting inside if statement. 
and same is the case with all other if also 
What is the problem ?
http://stackoverflow.com/questions/3901144/difference-between-if-x-and-if-x-is-not-none
I think it is better to use is not None also.

ankurankan:
@vivek425ster What does `find` return if it doesn't match anything ?

ankurankan:
@vivek425ster And also I am seeing a lot of pep8 errors. Please fix them.

ankurankan:
@vivek425ster I just checked that `findall` returns an empty list if it can't find anything matching. So why are you even checking for `not None`. You can simply iterate over it and if there's nothing in the list it will exit from the loop without executing. What do you think?

vivek425ster:
@ankurankan find method returns None.
https://docs.python.org/2/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.find

yes for findall we don't need to check for None.
But replacing find with findall will be problematic since it may return all of the grandchildren element also .
Or may be we can have a xpath query to search only in the immediate children.

Also where can i see the pep8 errrors?

palashahuja:
use a pep8 checker...

palashahuja:
https://pypi.python.org/pypi/pep8

vivek425ster:
@ankurankan Also consider this code. https://gist.github.com/vivek425ster/b896ccea61773b9b77e5
In this code only in the first case it is going inside the if condition whereas it is not going inside if condition in the second case

saketkc:
> In this code only in the first case it is going inside the if condition whereas it is not going inside if condition in the second case

The reason for that is the `ElementTree`'s implementation:
http://stackoverflow.com/questions/20129996/why-does-boolxml-etree-elementtree-element-evaluate-to-false

ankurankan:
@vivek425ster I am getting a couple of `FutureWarning`:

``` bash
ankur@ankur:~/pgmpy (pr/320 $%)$ nosetests3 
.....................................................................................SSS........................................................................................................................................./home/ankur/pgmpy/pgmpy/readwrite/PomdpX.py:359: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if parameter.find('SubDAGTemplate'):
............/home/ankur/pgmpy/pgmpy/readwrite/ProbModelXML.py:511: FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is not None' test instead.
  if not variable.find('States/State'):
.......................
----------------------------------------------------------------------
Ran 260 tests in 4.657s
```

Please fix these. Otherwise PR is good to go.

PR #324, opened 2015-03-13
Minor readme change

### LINKED ISSUE #322: nosetest fails after fresh install
I am runing `Arch Linux`

``` sh
$ uname -a
Linux tardis 3.18.6-1-ARCH #1 SMP PREEMPT Sat Feb 7 08:44:05 CET 2015 x86_64 GNU/Linux
```

I followed these [installation instructions](http://pgmpy.org/install.html). Specifically, my actions were:

``` sh
$ git clone https://github.com/pgmpy/pgmpy
$ cd pgmpy
$ virtualenv --system-site-packages vipgmpy
Using base prefix '/usr'
New python executable in vipgmpy/bin/python3
Also creating executable in vipgmpy/bin/python
Installing setuptools, pip...done.
$ source ./vipgmpy/bin/activate
$ python setup.py install
```

At this point, the `setup.py` finds all the required packages except one: `numpy`. That is because in `Arch Linux` the version of this package compiled with `openblas` support is called `numpy-openblas`. So, `setup.py` installs the vanilla `numpy` into the `vipgmpy` virtual environment. To be precise, here is the [verbose list of my dependencies](https://gist.github.com/all3fox/ff8a27901cae376519a0).

At this point everything is fine. I fire up `ipython`, issue `import pgmp`. It succeeds. So I decide to run the tests. From the source directory, I run `nosetests` and get a huge sheet of [failed tests](https://gist.github.com/all3fox/582fb271cb885fa72c53).

What am I doing wrong?

alisianoi:
version of NetworkX library should be `1.8.1`

close #322

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Using `pip`                       | Using `conda`
 ----------------------------------|-------------
 `pip install -r requirements.txt` | `conda install --file requirements.txt`
 
-Please note that you should use NetworkX 1.8.1, consult #322 for details.
+Please note that you should use NetworkX 1.8.1, consult issue #322 for details.
 
 Installation:
 =============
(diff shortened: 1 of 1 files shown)


ankurankan:
@all3fox I don't think that these changes are needed. 
1. Removing the white spaces has messed up the formatting and everything is on a single line.
2. We don't need to explicitly specify the networkx version as it is already in the `requirements.txt`.
3. And I don't think that making a table does really make any difference.

So, I am closing this PR.

Issue #200, opened 2014-08-28
Complete to_tabular_cpd method of TreeCPD


ankurankan:

palashahuja:
@ankurankan @abinashpanda I have finished writing the code but I just wanted to confirm whether this is what is exactly expected

```
tree = TreeCPD([('B', Factor(['A'], [2], [0.8, 0.2]), '0'),('B', 'C', '1'),('C', Factor(['A'], [2], [0.1, 0.9]), '0'),('C', 'D', '1'),('D', Factor(['A'], [2], [0.9, 0.1]), '0'),('D', Factor(['A'], [2], [0.4, 0.6]), '1')])
print(tree.to_tabular_cpd())
A       B       C       D       phi(A, B, C, D)
------------------------------------------------------
A_0     B_0     C_0     D_0     0.8000
A_0     B_0     C_0     D_1     0.8000
A_0     B_0     C_1     D_0     0.8000
A_0     B_0     C_1     D_1     0.8000
A_0     B_1     C_0     D_0     0.9000
A_0     B_1     C_0     D_1     0.4000
A_0     B_1     C_1     D_0     0.1000
A_0     B_1     C_1     D_1     0.1000
A_1     B_0     C_0     D_0     0.2000
A_1     B_0     C_0     D_1     0.2000
A_1     B_0     C_1     D_0     0.2000
A_1     B_0     C_1     D_1     0.2000
A_1     B_1     C_0     D_0     0.1000
A_1     B_1     C_0     D_1     0.6000
A_1     B_1     C_1     D_0     0.9000
A_1     B_1     C_1     D_1     0.9000


```

ankurankan:
@palashahuja I think you have made a minor mistake in constructing the tree. The tree should look like this:

``` python3

                          B
                0  /             \  1
                  /               \
            a_0     0.8            C
            a_1     0.2       /            \
                           0 /              \ 1
                            /                \
                      a_0   0.1               D
                      a_1   0.9          /            \
                                     0  /              \  1
                                       /                \ 
                               a_0        0.9              a_0        0.4
                               a_1        0.1              a_1        0.6
```

And the TabularCPD should be:

``` python3
A       B       C       D       phi(A, B, C, D)
------------------------------------------------------
A_0     B_0     C_0     D_0     0.8000
A_0     B_0     C_0     D_1     0.8000
A_0     B_0     C_1     D_0     0.8000
A_0     B_0     C_1     D_1     0.8000
[...]

PR #325, opened 2015-03-14
to_tabular_cpd method of TreeCPD [refs #200]

palashahuja:


ankurankan:
@palashahuja Please add tests for these.

PR #328, opened 2015-03-15
Added method to compute induced width

pratyakshs:


abinashpanda:
@pratyakshs Could you please explain what does the method `eliminate` do? We have an implementation of `reduce` method which helps in reducing the variables in factor. Correct me if I am wrong eliminate is trying to do the same?

pratyakshs:
@abinashpanda Yes, eliminate does the same. I'll remove the redundant code. 

Another thing, on running the README.md code, I am getting an error in the line "if list(nx.simple_cycles(self)):" in BayesianModel.py. Are there any known issues?

abinashpanda:
@pratyakshs Could you tell the version of `python` as well as the version of `networkx` you are using? If you are using the latest version of `networkx` i.e. `1.9.1`, I would suggest you to downgrade it to `1.8.1`. There are some issues with the latest one and we have not yet figured that one.

pratyakshs:
I was using version `1.9.1` of `networkx` on `python 3.4.2`. Downgraded it to `1.8.1` and it works fine. Thanks.

alisianoi:
@ankurankan here is a similar situation to issue #322 in terms that the README should be more verbose about the fact that `networkx` should not be the latest version but rather the `1.8.1`

Could you please make the README more precise?

saketkc:
But the `requirements.txt` does explicitly mention `1.8.1`, and the recommended way is inside a `virtualenv`.

alisianoi:
Many people do not follow the "put everything into a separate `virtualenv`" installation strategy. Here is a possible way of their thinking process:
1. There already is the whole SciPy stack installed on my system, I do not want to wait for it to compile and occupy extra space. I'd rather add `--system-site-packages`. 
2. So, I will not be using the `pip install -r requirements.txt` approach but I'd rather construct the `virtualenv` myself. 

At this point there is a series of `pip install somelib`, where `pip` fetches the latest versions of those libs and you really could not be bothered to compare those versions with the `requirements.txt`. Or, even if you are, it is common to think that if the version of the required lib is older than the available one, it is probably because the list of dependencies is a bit outdated and not because there actually is a problem with the latest version of that library.

PR #335, opened 2015-03-16
Fixed issues with networkx 1.9.1

pratyakshs:
Output of `nosetests3`:
`$ nosetests3`
`.....................................................................................SSS....................................................................................``...................................................../home/pratyaksh/workspace/gsoc/pgmpy/pgmpy/readwrite/PomdpX.py:359:``FutureWarning: The behavior of this method will change in future versions. Use specific 'len(elem)' or 'elem is``not None' test instead.`
 `if parameter.find('SubDAGTemplate'):`
`...............................`
`----------------------------------------------------------------------`
`Ran 256 tests in 2.660s`

`OK (SKIP=3)`

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -383,10 +383,11 @@ class TreeCPD(nx.DiGraph):
         >>> tree.add_edge('C', Factor(['A'], [2], [0.1, 0.9]), label=0)
         """
         if u != v:
-            super(TreeCPD, self).add_edge(u, v, label=label)
-            if list(nx.simple_cycles(nx.DiGraph(self.edges()))):
-                super(TreeCPD, self).remove_edge(u, v)
+            if u in self.nodes() and v in self.nodes() and nx.has_path(self, v, u): 
+                # check if adding edge (u, v) forms a cycle
                 raise ValueError("Self Loops and Cycles are not allowed")
+            else:
+                super(TreeCPD, self).add_edge(u, v, label=label)
         else:
             raise ValueError("Self Loops and Cycles are not allowed")
 
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -384,7 +384,7 @@ class TreeCPD(nx.DiGraph):
         """
         if u != v:
             super(TreeCPD, self).add_edge(u, v, label=label)
-            if list(nx.simple_cycles(self)):
+            if list(nx.simple_cycles(nx.DiGraph(self.edges()))):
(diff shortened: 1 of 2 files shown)


abinashpanda:
@pratyakshs Thanks for looking into this issue. I have only one concern with your hack, every time you add an edge using `add_edge` method it would create a `DiGraph` object which for larger graphs would not only be time-intensive, it would be also memory-intensive. If you could find some cleaner solution that would be better.

pratyakshs:
@abinashpanda I agree with your point. But well, isn't cycle detection running in O(V+E) for each add_edge call?

abinashpanda:
@pratyakshs I am not worried about the time taken for cycle detection, my concern is regarding creating new objects in each iteration. I am against object creation.

pratyakshs:
@abinashpanda There's another way to do almost the same thing. Instead of passing `nx.DiGraph(self)` as argument, we can instead pass `type(nx.DiGraph())(self)` which is basically a typecast. Does it fit the bill?

abinashpanda:
@pratyakshs This would also cause the same issue of creation of `DiGraph` objects. Have a look at the code sample

``` python
In [1]: import networkx as nx

In [2]: a = nx.DiGraph([('a', 'b'), ('c', 'd')])

In [3]: a
Out[3]: <networkx.classes.digraph.DiGraph at 0x7fbb343c5a90>

In [4]: b = type(nx.DiGraph())(a)

In [5]: b
Out[5]: <networkx.classes.digraph.DiGraph at 0x7fbb343cff28>
```

See the objects `a` and `b` are totally different objects stored at different memory location. There is no such typecasting over here. So if you still use `type(nx.DiGraph())(self)` it would create still create new `DiGraph` objects.

pratyakshs:
@abinashpanda Thanks for pointing that out. I wasn't myself convinced when I wrote that!

pratyakshs:
@abinashpanda If the issue is only with the function call `nx.simple_cycles()`, I think it would be better to write a simple code for reachability of node `u` from node `v` when an edge `(u, v)` is added. The implementation of `nx.symple_cycles` uses Tarjan's SCC algorithm (after making a copy of the original graph), then tries to find all simple cycles iteratively using these components. This looks a bit overkill (and memory intensive too) for every `add_edge` operation. Would you be interested if I add the code for reachability(BFS/DFS)?

abinashpanda:
@pratyakshs my suggestion is keep the code as it is (as it working). If you still think that you can create a robust method for finding edges then surely go for it. Send a PR and lets review :)

PR #336, opened 2015-03-16
Using nx.has_path() instead of nx.simple_cycles()

pratyakshs:


pgmpy/factors/CPD.py
```diff
@@ -383,10 +383,11 @@ def add_edge(self, u, v, label):
         >>> tree.add_edge('C', Factor(['A'], [2], [0.1, 0.9]), label=0)
         """
         if u != v:
-            super(TreeCPD, self).add_edge(u, v, label=label)
-            if list(nx.simple_cycles(self)):
-                super(TreeCPD, self).remove_edge(u, v)
+            if u in self.nodes() and v in self.nodes() and nx.has_path(self, v, u): 
+                # check if adding edge (u, v) forms a cycle
                 raise ValueError("Self Loops and Cycles are not allowed")
```
abinashpanda, line 389: Could you add a better error message here (similar to what it is present in `BayesianModel`. Apart from that its ready to merge.

abinashpanda:
@pratyakshs Whats the status of this bug? Have a look at the comments and fix them as soon as possible. It would be ready to merge.

PR #338, opened 2015-03-19
Replaced recursive call with `while` loop.

ashwch:
- Replaces the recursive version of [fun](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/factors/Factor.py#L445) with an iterative one, much easier to read in my opinion. This should be slightly fast as well because recursive calls are expensive in Python(results in new stack frame each time), plus Python has a default limit of 1000 for recursive calls(though this is highly unlikely to occur in our case).
- I am not sure why the library is using Python 2 based `super()` calls considering the fact that dependency includes Python 3.3. In Python 3 thanks to the cell variable `__class__` we can simply use `super().method_name(...)`([PEP 3135 -- New Super](http://legacy.python.org/dev/peps/pep-3135/)).

### DIFF
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -444,9 +444,10 @@ class Factor:
         # by 1.
         def fun(b, index=len(self.cardinality)-1):
             b[index] += 1
-            if b[index] == self.cardinality[index]:
+            while b[index] == self.cardinality[index]:
                 b[index] = 0
-                fun(b, index-1)
+                index -= 1
+                b[index] += 1
             return b
 
         def gen():
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -444,10 +444,9 @@ class Factor:
         # by 1.
         def fun(b, index=len(self.cardinality)-1):
             b[index] += 1
-            while b[index] == self.cardinality[index]:
+            if b[index] == self.cardinality[index]:
                 b[index] = 0
-                index -= 1
-                b[index] += 1
+                fun(b, index-1)
             return b
 
         def gen():
--- a/pgmpy/base/DirectedGraph.py
+++ b/pgmpy/base/DirectedGraph.py
@@ -67,7 +67,7 @@ class DirectedGraph(nx.DiGraph):
     """
 
     def __init__(self, ebunch=None):
-        super(DirectedGraph, self).__init__(ebunch)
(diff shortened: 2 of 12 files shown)


ashwch:
I guess Landscape has [no support](https://github.com/landscapeio/landscape-issues/issues/32) for Python 3. :-(

pgmpy/factors/Factor.py
```diff
@@ -443,8 +443,11 @@ def _str(self, html=False):
         # to increment it
         # by 1.
         def fun(b, index=len(self.cardinality)-1):
+            #print ('---->', self.cardinality, index, b)
```
abinashpanda: @ashwch When submitting a PR make sure that you remove the debug statements completely instead of commenting them.

pgmpy/factors/Factor.py
```diff
@@ -443,8 +443,11 @@ def _str(self, html=False):
...
             b[index] += 1
+            #print(b[index], self.cardinality[index], b[index] == self.cardinality[index])
```
abinashpanda: Same as above.

pgmpy/factors/Factor.py
```diff
@@ -453,6 +456,7 @@ def gen():
             b = [0] * len(self.variables)
             yield b
             for i in range(np.prod(self.cardinality)-1):
+                #print ('', fun(b))
                 yield fun(b)
```
abinashpanda, line 457: Clear all the comments.

ashwch:
@abinashpanda Fixed.

ashwch:
**[b73ca6a](https://github.com/ashwch/pgmpy/commit/b73ca6a75dd9e217de4fe6a8d77320b713824edb)** fixes `NoisyOrModel` and its tests. Changes are:
- We should store `inhibitor_probability` as a list because it would be tougher(and possibly expensive) to check its size and shape each time `add_variables` and `del_variables` are called and then we pad or slice off zeroes from it accordingly. If we really want a NumPy array then better convert it to one when we actually want to do some vectorized operations on it.
- We can store `variables` as a NumPy array though. As it is always 1-D we don't have to worry about its  shape.
- As `inhibitor_probability` is not always a nested list a special check is required to convert it to one when number of `variables` equal 1.

abinashpanda:
@ashwch I think this PR has multiple changes. From the next time could you please send separate PR for each major change.

PR #345, opened 2015-03-24
Fix __repr__ of TabularCPD

siteshjaiswal:
@abinashpanda 
I am getting this

```
In [1]: cpd_a = TabularCPD('A', 2, [[0.1, 0.1, 0.1, 0.1, 0.1, 0.1],
                                    [0.9, 0.9, 0.9, 0.9, 0.9, 0.9]],
                           ['B', 'C'], [2, 3])

# __repr__ method should return something like this
In [2]: cpd_a
<TabularCPD representing P(A:2 |B:2, C:3)>
```

Instead of 

```
In [1]: cpd_a = TabularCPD('A', 2, [[0.1, 0.1, 0.1, 0.1, 0.1, 0.1],
                                    [0.9, 0.9, 0.9, 0.9, 0.9, 0.9]],
                           ['B', 'C'], [2, 3])

# __repr__ method should return something like this
In [2]: cpd_a
<TabularCPD representing P(A:2 |B:2, C:3) at 0x7f34ac8c5160>
```

Please look into it.

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -121,13 +121,14 @@ class TabularCPD(Factor):
     def __repr__(self):
         if not self.evidence:
             str_1 = '<TabularCPD representing P({variable}:{cardinality}) at {address}>'\
-            .format(variable=self.variable,cardinality=self.variable_card,address=str(repr(hex(id(self)))))
+                .format(variable=self.variable, cardinality=self.variable_card, address=str(repr(hex(id(self)))))
             return str_1
         else:
             str_2 = ', '.join(['{variable}:{cardinality}'.format(variable=variable, cardinality=cardinality)
-                for variable, cardinality in zip(self.evidence, self.evidence_card)])
+                               for variable, cardinality in zip(self.evidence, self.evidence_card)])
             str_1 = '<TabularCPD representing P({variable}:{cardinality} |{string2}) at {address}>'\
-                .format(variable=self.variable,cardinality=self.variable_card,string2=str_2,address=str(repr(hex(id(self)))))
+                .format(variable=self.variable, cardinality=self.variable_card, string2=str_2,
+                        address=str(repr(hex(id(self)))))
             return str_1
 
     def get_cpd(self):


pgmpy/factors/CPD.py
```diff
@@ -117,6 +117,21 @@ def __init__(self, variable, variable_card, values,
             raise TypeError("Values must be a 2D list/array")
 
         super(TabularCPD, self).__init__(variables, cardinality, values.flatten('C'))
+        
+    def __repr__(self):
+        if not self.evidence:
+            str_1 = ""
+            str_1 += '<TabularCPD representing P(' + str(self.variable) + ':' + str(self.variable_card) + ')>'
```
abinashpanda: Why are you creating an empty string `str_1` here if you are going to add something to it?

siteshjaiswal: Ya i agree........i will remove it in final commit

abinashpanda: Please don't use `+` operator with string objects. Its a costly operation as it create new objects. Instead you can use something like this

``` python
str_1 = '<TabularCPD representing P({variable} : {cardinality})>'.format(variable=self.variable, cardinality=self.variable_card)
```

pgmpy/factors/CPD.py
```diff
@@ -117,6 +117,21 @@ def __init__(self, variable, variable_card, values,
...
+            return str_1
+
+        else:
+            str_1 = ""
+            str_1 += '<TabularCPD representing P(' + str(self.variable) + ':' + str(self.variable_card) + ' |'
+            str_2 = ""
+            for i in range(len(self.evidence)):
```
abinashpanda: Don't add to `str_2` in each iteration. Adding to any immutable object creates a new immutable object. Instead try to use string formatting as much as possible. Furthermore, the `for` loop can also be replaced by list comprehension as

``` python
str_2 = ' '.join(['{variable}:{cardinality}'.format(variable=variable, cardinality=cardinality)
                  for variable, cardinality in zip(self.evidence, self.evidence_card)])
```

pgmpy/factors/CPD.py
```diff
@@ -117,6 +117,21 @@ def __init__(self, variable, variable_card, values,
...
+                str_2 += str(self.evidence[i]) + ':' + str(self.evidence_card[i]) + ','
+            str_2 = str_2[:-1]
+            return str_1 + str_2 + ')>'
```
abinashpanda: Again the same. Please use string formatting here.

ankurankan:
@siteshjaiswal For getting the address you can either make a call to the actual `__repr__` method and it will return a string which you can parse to get the object address. Or instead of using the address we can use the `id`of the object as it is also unique. @abinashpanda What do you think?

abinashpanda:
@ankurankan @siteshjaiswal I think internally `__repr__` method uses `hex(id())` to compute the addresss. So you can use either `hex(id())` or `id()`. I am fine with both of them.

ankurankan:
@siteshjaiswal @abinashpanda I think we should go with `hex(id(obj))` as it would be consistent with the default `__repr__` behaviour.

siteshjaiswal:
@ankurankan @abinashpanda 
Ya i agree hex(id(obj)) is more consistent.

ankurankan:
@siteshjaiswal Can you please fix this quickly? Needs to go out ASAP

siteshjaiswal:
@ankurankan @abinashpanda 
Please look into the modified code.

pgmpy/factors/CPD.py
```diff
@@ -117,6 +117,18 @@ def __init__(self, variable, variable_card, values,
             raise TypeError("Values must be a 2D list/array")
 
         super(TabularCPD, self).__init__(variables, cardinality, values.flatten('C'))
+        
+    def __repr__(self):
+        if not self.evidence:
+            str_1 = '<TabularCPD representing P({variable}:{cardinality}) at {address}>'\
+            .format(variable=self.variable,cardinality=self.variable_card,address=str(repr(hex(id(self)))))
```
abinashpanda: Always put a **space** after `,`. Follow PEP8.

abinashpanda:
Fix all the PEP8 issues. Run PEP8 checker as `pep8 --max-line-length=120` and make sure that there are no errors.

ankurankan:
@siteshjaiswal Please add tests for this.

pramttl:
@siteshjaiswal There is superset tool of `pep8` called `flake8` which also helps detect import errors and remove redundant imports other than just pep8 style check. You can use that as well.

```
 pip install flake8
 flake8 <python_filename>
```

PR #347, opened 2015-03-25
Implemented AdditionalConstraints in Reader class of ProbModelXML file.

vivek425ster:
Link from where it is implemented: http://www.cisiad.uned.es/techreports/ProbModelXML.pdf


ankurankan:
@vivek425ster Please add a reference of the above paper at the top of the `ProbModelXML.py` file.

vivek425ster:
@ankurankan  Done.

PR #348, opened 2015-03-25
induced_graph

pratyakshs:


pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +228,47 @@ def induced_graph(self, elimination_order):
...
+        if len(elimination_order) < len(self.variables):
+            raise ValueError("Elimination order incomplete")
+
+        # Removing all the factors containing the variables which are
+        # eliminated (as all the factors should be considered only once)
+        for var in elimination_order:
+            factors = [factor for factor in working_factors[var]
+                       if not set(factor.variables).intersection(eliminated_variables)]
+            phi = factor_product(*factors)
+            phi.reduce(var + '_0')
```
ankurankan: @pratyakshs Never use `+` to join strings. Better use `.format`

pratyakshs: @ankurankan Why is that? 
So, should I fix it and do another pull request?

ankurankan: @pratyakshs: `+` creates a new object. So you will be first creating a string `_0` then add this to var which will again create a new object whereas format simply replaces in the string. Also format looks cleaner.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +228,47 @@ def induced_graph(self, elimination_order):
...
+
+        edges = []
+        # add edges corresponding to each clique
+        for clique in cliques:
+            if len(clique) > 1:
+                for i in range(len(clique)):
+                    for j in range(i + 1, len(clique)):
+                        edges.append((clique[i], clique[j]))
+
+        import networkx as nx
```
ankurankan: @pratyakshs Please put imports at the top of the file.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +228,47 @@ def induced_graph(self, elimination_order):
...
+                           for node in self.factors}
+
+        # The set of cliques should be in the induced graph
+        cliques = set()
+        for factors in working_factors.values():
+            for factor in factors:
+                cliques.add(tuple(factor.scope()))
+
+        # If the elimination_order list is incomplete, raise an error
+        if len(elimination_order) < len(self.variables):
```
ankurankan: @pratyakshs: Please put all such check at the top in the method. You should also check if all the variables listed in `elimination_order` are in the model as in such cases we will get an error somewhere deep in code and will be very tough to trace back. So, it's better to have a check for that and raise a informative error.

pgmpy/inference/ExactInference.py
pratyakshs: @ankurankan Cool! I didn't know that. :)

pgmpy/inference/ExactInference.py
pratyakshs: @ankurankan Sure, I'll do that. I'll check if `set(elimination_order) == set(self.variables)` apart from doing the `len` check.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +228,46 @@ def induced_graph(self, elimination_order):
...
+            if len(clique) > 1:
+                for i in range(len(clique)):
+                    for j in range(i + 1, len(clique)):
+                        edges.append((clique[i], clique[j]))
+
+        # Final induced graph
+        graph = nx.Graph()
+        graph.add_nodes_from(elimination_order)
+        graph.add_edges_from(edges)
+        return graph
```
ankurankan, line 231: @pratyakshs Add an extra new line here. According to PEP8 class definitions should be separated by 2 blank lines.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +228,46 @@ def induced_graph(self, elimination_order):
...
+        # add edges corresponding to each clique
+        for clique in cliques:
+            if len(clique) > 1:
+                for i in range(len(clique)):
+                    for j in range(i + 1, len(clique)):
+                        edges.append((clique[i], clique[j]))
+
+        # Final induced graph
+        graph = nx.Graph()
+        graph.add_nodes_from(elimination_order)
```
ankurankan: @pratyakshs Check if you can remove this line as networkx automatically adds variables if you add an edge between variables that don't exist in the graph.

pratyakshs: @ankurankan Yes, this can be removed.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +228,46 @@ def induced_graph(self, elimination_order):
...
+            for variable in phi.variables:
+                working_factors[variable].add(phi)
+            eliminated_variables.add(var)
+
+        edges = []
+        # add edges corresponding to each clique
+        for clique in cliques:
+            if len(clique) > 1:
+                for i in range(len(clique)):
+                    for j in range(i + 1, len(clique)):
```
ankurankan: @pratyakshs I think you should be able to replace this nested for loop using `itertools.combinations`.

pgmpy/inference/ExactInference.py
```diff
@@ -3,7 +3,7 @@
 import numpy as np
 from pgmpy.inference import Inference
 from pgmpy.factors.Factor import factor_product
-
+import networkx as nx
```
ankurankan: @pratyakshs Have a look at PEP8 (https://www.python.org/dev/peps/pep-0008/#imports) regarding the order of imports in a file.

ankurankan:
@pratyakshs Also check why the tests are failing.

So just checked, there is some weird character at the end of line 55 that github is not showing. Remove that.

pratyakshs:
@ankurankan Let me do another pull request then.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +229,48 @@ def induced_graph(self, elimination_order):
         >>> inference.induced_graph(['C', 'D', 'A', 'B', 'E'])
         <networkx.classes.graph.Graph at 0x7f34ac8c5160>
         """
+        # If the elimination_order list is incomplete, raise an error
+        if len(elimination_order) < len(self.variables):
```
ankurankan: @pratyakshs Why do you need this check when you are already checking for equality in the next line? This check is redundant.

PR #349, opened 2015-03-25
Added method to compute induced graph

pratyakshs:


pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +230,45 @@ def induced_graph(self, elimination_order):
...
+                           for node in self.factors}
+
+        # The set of cliques should be in the induced graph
+        cliques = set()
+        for factors in working_factors.values():
+            for factor in factors:
+                cliques.add(tuple(factor.scope()))
+
+        # If the elimination_order list is incomplete, raise an error
+        if len(elimination_order) < len(self.variables):
```
ankurankan: @pratyakshs Please always work on a single PR (helps us keep a track of everything). And you did the same mistake again.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +230,45 @@ def induced_graph(self, elimination_order):
...
+            raise ValueError("Elimination order incomplete")
+
+        # Removing all the factors containing the variables which are
+        # eliminated (as all the factors should be considered only once)
+        for var in elimination_order:
+            factors = [factor for factor in working_factors[var]
+                       if not set(factor.variables).intersection(eliminated_variables)]
+            phi = factor_product(*factors)
```
ankurankan: @pratyakshs Do we really need to actually multiply the factors? We will not be needing the actual values anywhere and just be needing the scopes. So, I think this should be avoidable.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +230,45 @@ def induced_graph(self, elimination_order):
         >>> inference.induced_graph(['C', 'D', 'A', 'B', 'E'])
         <networkx.classes.graph.Graph at 0x7f34ac8c5160>
         """
+        eliminated_variables = set()
+        working_factors = {node: {factor for factor in self.factors[node]}
```
ankurankan: @pratyakshs: Should not even be needing to create this huge structure. Should be okay with just the scopes of the factors.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +230,48 @@ def induced_graph(self, elimination_order):
...
+            cliques.add(tuple(phi))
+            del working_factors[var]
+            for variable in phi:
+                working_factors[variable].append(list(phi))
+            eliminated_variables.add(var)
+
+        edges = []
+        # add edges corresponding to each clique
+        for clique in cliques:
+            if len(clique) > 1:
```
abinashpanda: You can remove the `if` statement using `filter` such as

``` python
for clique in filter(lambda x: len(x) > 1, cliques):
```

pratyakshs: Done

pgmpy/tests/test_inference/test_ExactInference.py
```diff
@@ -86,6 +86,14 @@ def test_map_query_with_evidence(self):
                                                       {'J': 0, 'Q': 1, 'G': 0})
         self.assertDictEqual(map_query, {'A': 1, 'R': 0, 'L': 0})
 
+    def test_induced_graph(self):
+        induced_graph = self.bayesian_inference.induced_graph(['G', 'Q', 'A', 'J', 'L', 'R'])
+        result_edges = sorted([sorted(x) for x in induced_graph.edges()])
```
abinashpanda, line 91: For sorting list of lists don't use `sorted` twice. You can use something like this

``` python
result_edges = sorted(induced_graph.edges(), key=lambda x: (x[0], x[1]))
```

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +230,47 @@ def induced_graph(self, elimination_order):
...
+
+        edges = []
+        # add edges corresponding to each clique
+        for clique in filter(lambda x: len(x) > 1, cliques):
+            for i, j in itertools.combinations(clique, 2):
+                edges.append((i, j))
+
+        # Final induced graph
+        graph = nx.Graph()
+        graph.add_edges_from(edges)
```
ankurankan: @pratyakshs No need to first create a graph object and then adding edges. You could simply do `return nx.Graph(edges)`

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +230,47 @@ def induced_graph(self, elimination_order):
         >>> inference.induced_graph(['C', 'D', 'A', 'B', 'E'])
         <networkx.classes.graph.Graph at 0x7f34ac8c5160>
         """
+        # If the elimination_order list is incomplete, raise an error
+        if len(elimination_order) < len(self.variables):
```
ankurankan: @pratyakshs One of the conditions covers the other.

ankurankan:
@pratyakshs Please do remove some obvious comments. Fix the code as mentioned in the comments and the PR should be fine to merge.

pgmpy/inference/ExactInference.py
```diff
@@ -228,6 +230,47 @@ def induced_graph(self, elimination_order):
...
+            del working_factors[var]
+            for variable in phi:
+                working_factors[variable].append(list(phi))
+            eliminated_variables.add(var)
+
+        edges = []
+        # add edges corresponding to each clique
+        for clique in filter(lambda x: len(x) > 1, cliques):
+            for i, j in itertools.combinations(clique, 2):
+                edges.append((i, j))
```
ankurankan: @pratyakshs One possible improvement that can be done here is that rather than creating a list of edges just create a list of `combinations` objects and then while creating the graph use `itertools.chain` to get the argument for `nx.Graph`. This would be really beneficial in the case when we have thousands of nodes, we will then be having multiple thousand edges (also repeating edges in different cliques) and hence we will be ending up with a huge `edges` list. @abinashpanda What do you think?

abinashpanda: @ankurankan @pratyakshs Yep. I also agree. It is more scalable.

pgmpy/tests/test_inference/test_ExactInference.py
pratyakshs, line 91: This isn't exactly what I want to do. The induced graph is undirected, so I sort each edge in order to get (u, v) with u <= v.

pgmpy/inference/ExactInference.py
pratyakshs: Will do that.

pgmpy/inference/ExactInference.py
pratyakshs: Yes it does. But checking this one is more efficient than the other one. Anyway, I'll remove it.

ankurankan: @pratyakshs It's more probable that the arguments would be correct and in that case you will be doing more computation.

pgmpy/inference/ExactInference.py
pratyakshs: Looks good now?

ankurankan:
@pratyakshs And also add some tests for the exceptions and check for any edge cases that you can think.

PR #355, opened 2015-03-29
Improving Variable Elimination (VE)

jhonatanoliveira:
Here, we finish the implementation of VE with few missing steps, namely **computing good elimination orderings** (with 4 heuristics - _min neighbors_, _min fill_, _min weight_, and _weighted min fill_) and **safely removing irrelevant variables from the model** (_barren_ and _independent by evidence_ nodes).
In order to make these improvements, few new methods were necessary and modifying other members too. In our preliminary experimental results, queries that were taking up to 30 minutes using VE now take less than 2 minutes. Please, help us test this new code for a robust implementation and also looking for better ways of coding the algorithms. Thanks.


palashahuja:
@jhonatanoliveira  I am not really sure if this is of any use but please have a look. This method is quite different than the usual one. Here is the [link](http://www.aaai.org/ocs/index.php/AAAI/AAAI14/paper/download/8596/8512) . Please correct me if I am wrong.

jhonatanoliveira:
Thanks for the comment, @palashahuja. I didn't know this paper and it's a very interesting approach, but for relational models. I assumed that our class _BayesianModel_ may mostly be suitable for representing a Bayesian network (please, correct me if I'm wrong). Given that, as highlighted in this paper you posted, these heuristics I've implemented "are known to produce good orderings". The way I've implemented _EliminationOrdering_ is as proposed on page 314 in "Probabilistic Graphical Models - Principles and Techniques". But anyway, I'll be happy to modify something here in order to improve VE :)

pgmpy/base/UndirectedGraph.py
```diff
@@ -174,3 +174,29 @@ def is_triangulated(self):
...
+        >>> print(ug.fill_in_edges("bronc"))
+        [('dysp', 'smoke'), ('either', 'smoke')]
+        """
+        if not isinstance(variables, (list, set)):
+            variables = [variables]
+        edges = []
+        for variable in variables:
+            neighbors = self.neighbors(variable)
+            edges = [e for e in itertools.combinations(neighbors, 2) if not self.has_edge(*e)]
+        return edges
```
abinashpanda, line 191: @jhonatanoliveira Correct me if I am wrong. The method `fill_in_edges` takes input a `list` of nodes to be removed. But by looking into the code it seems that the method would return only the edges to be removed only for the last variable as `edges` variable is updated in each iteration without storing the previous results.

Furthermore could you please write the tests for each of the method that you wrote with all possible test cases. That would make these methods more robust.

pgmpy/inference/EliminationOrdering.py
```diff
@@ -0,0 +1,101 @@
...
+    Min-Fill and Weighted-Min-Fill tend to work better on more problems".
+
+    Parameters
+    ----------
+    bayesian_model: input bayesian model
+        A bayesian model to be moralized (an undirected graph where the parents
+        of a v-structures are connected).
+    """
+
+    def __init__(self, bayesian_model):
```
abinashpanda: This class becomes very specific to `BayesianModel`. Although its rare to see variable elimination applied to `MarkovModel` but we want our classes and methods to be as generic as possible. So instead of `bayesian_model` use `model` and put explicit checks that if it is a `BayesianModel` then moralize it.

jhonatanoliveira:
Hi, @abinashpanda . Sorry the delay. Indeed, that's a mistake in `fill_in_edges`, thanks for spotting that. I'll fix it right away, and I'll make your other proposes (test cases and changing from `bayesian_model` to `model`) as soon as possible.

jhonatanoliveira:
@abinashpanda I've finally finished the improvements you've proposed. Please, let me know if there is any further improvement on this PR. We'll be happy to help. Regards.

ankurankan:
@jhonatanoliveira I will try to review and merge this PR by this weekend.

jhonatanoliveira:
@ankurankan is dealing with this in another PR.

PR #358, opened 2015-03-31
added method to swap variables in TabularCPD

kislayabhi:
Please see the architecture. The code is working fine, but again I need help to make it more pythonic.

```
final_cpd=TabularCPD('grade', 5,  [[0.1,0.6,0.2,0.9,0.5,0.4,0.5,0.5,0.4,0.9,0.8,0.8],
                                   [0.2,0.7,0.3,0.7,0.4,0.3,0.6,0.6,0.3,0.8,0.8,0.8],
                                   [0.3,0.8,0.4,0.8,0.8,0.8,0.7,0.7,0.2,0.8,0.8,0.4],
                                   [0.4,0.9,0.5,0.7,0.3,0.2,0.8,0.8,0.1,0.6,0.8,0.8],
                                   [0.5,0.1,0.6,0.6,0.2,0.1,0.9,0.9,0.8,0.5,0.8,0.8]],
                      evidence=['intel', 'diff', 'iq'],
                      evidence_card=[2, 2, 3])
```

```

>>> print(final_cpd)
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| iq      | iq_0    | iq_0    | iq_0    | iq_0    | iq_1    | iq_1    | iq_1    | iq_1    | iq_2    | iq_2    | iq_2    | iq_2    |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| diff    | diff_0  | diff_0  | diff_1  | diff_1  | diff_0  | diff_0  | diff_1  | diff_1  | diff_0  | diff_0  | diff_1  | diff_1  |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| intel   | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_0 | 0.1     | 0.6     | 0.2     | 0.9     | 0.5     | 0.4     | 0.5     | 0.5     | 0.4     | 0.9     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_1 | 0.2     | 0.7     | 0.3     | 0.7     | 0.4     | 0.3     | 0.6     | 0.6     | 0.3     | 0.8     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_2 | 0.3     | 0.8     | 0.4     | 0.8     | 0.8     | 0.8     | 0.7     | 0.7     | 0.2     | 0.8     | 0.8     | 0.4     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_3 | 0.4     | 0.9     | 0.5     | 0.7     | 0.3     | 0.2     | 0.8     | 0.8     | 0.1     | 0.6     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_4 | 0.5     | 0.1     | 0.6     | 0.6     | 0.2     | 0.1     | 0.9     | 0.9     | 0.8     | 0.5     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+


>>> final_cpd.replace_variables('iq', 'diff')
>>> print(final_cpd)
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| diff    | diff_0  | diff_0  | diff_0  | diff_0  | diff_0  | diff_0  | diff_1  | diff_1  | diff_1  | diff_1  | diff_1  | diff_1  |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| iq      | iq_0    | iq_0    | iq_1    | iq_1    | iq_2    | iq_2    | iq_0    | iq_0    | iq_1    | iq_1    | iq_2    | iq_2    |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| intel   | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 | intel_0 | intel_1 |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_0 | 0.1     | 0.6     | 0.5     | 0.4     | 0.4     | 0.9     | 0.2     | 0.9     | 0.5     | 0.5     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_1 | 0.2     | 0.7     | 0.4     | 0.3     | 0.3     | 0.8     | 0.3     | 0.7     | 0.6     | 0.6     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_2 | 0.3     | 0.8     | 0.8     | 0.8     | 0.2     | 0.8     | 0.4     | 0.8     | 0.7     | 0.7     | 0.8     | 0.4     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_3 | 0.4     | 0.9     | 0.3     | 0.2     | 0.1     | 0.6     | 0.5     | 0.7     | 0.8     | 0.8     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_4 | 0.5     | 0.1     | 0.2     | 0.1     | 0.8     | 0.5     | 0.6     | 0.6     | 0.9     | 0.9     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+

>>> final_cpd.replace_variables('intel', 'diff')
>>> print(final_cpd)
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| intel   | intel_0 | intel_0 | intel_0 | intel_0 | intel_0 | intel_0 | intel_1 | intel_1 | intel_1 | intel_1 | intel_1 | intel_1 |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| iq      | iq_0    | iq_0    | iq_1    | iq_1    | iq_2    | iq_2    | iq_0    | iq_0    | iq_1    | iq_1    | iq_2    | iq_2    |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| diff    | diff_0  | diff_1  | diff_0  | diff_1  | diff_0  | diff_1  | diff_0  | diff_1  | diff_0  | diff_1  | diff_0  | diff_1  |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_0 | 0.1     | 0.2     | 0.5     | 0.5     | 0.4     | 0.8     | 0.6     | 0.9     | 0.4     | 0.5     | 0.9     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_1 | 0.2     | 0.3     | 0.4     | 0.6     | 0.3     | 0.8     | 0.7     | 0.7     | 0.3     | 0.6     | 0.8     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_2 | 0.3     | 0.4     | 0.8     | 0.7     | 0.2     | 0.8     | 0.8     | 0.8     | 0.8     | 0.7     | 0.8     | 0.4     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_3 | 0.4     | 0.5     | 0.3     | 0.8     | 0.1     | 0.8     | 0.9     | 0.7     | 0.2     | 0.8     | 0.6     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
| grade_4 | 0.5     | 0.6     | 0.2     | 0.9     | 0.8     | 0.8     | 0.1     | 0.6     | 0.1     | 0.9     | 0.5     | 0.8     |
+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+---------+
```

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -125,26 +125,78 @@ class TabularCPD(Factor):
 
         return var_str + evidence_str + ') at {address}>'.format(address=hex(id(self)))
 
-    def replace_variables(self, var1, var2):
-        new_scope = self.scope()
-        new_scope.reverse()
-        new_cardinality = self.cardinality[::-1].copy()
-        index_1 = new_scope.index(var1)
-        index_2 = new_scope.index(var2)
-        index_11 = self.scope().index(var1)
-        index_22 = self.scope().index(var2)
-        new_scope[index_1], new_scope[index_2] = new_scope[index_2], new_scope[index_1]
-        new_cardinality[index_1], new_cardinality[index_2] = new_cardinality[index_2], new_cardinality[index_1]
-        values = self.values.reshape([self.cardinality[0], self.cardinality[1:].prod()])
-        old_heading = np.array(list(product(*[range(i) for i in self.cardinality[1:]])))
-        new_heading = np.array(list(product(*[range(i) for i in new_cardinality[-2::-1]])))
-        new_heading[:, [index_11-1, index_22-1]] = new_heading[:, [index_22-1, index_11-1]]
+    def set_parents_order(self, new_order):
+        """
+        Change the order of the parents
+
+        Parameters
+        ----------
+        new_order: list
+            reordered list of evidences.
+        The returned tabular cpd has values reordered according to the new order provided by the user.
+
+        Example
+        --------
+        >>> from pgmpy.factors import TabularCPD
+        >>> final_cpd = TabularCPD('grade', 5, [
+        ... [0.12, 0.6, 0.2, 0.9, 0.5, 0.4, 0.5, 0.5, 0.4, 0.9, 0.8, 0.8],
+        ... [0.12, 0.6, 0.2, 0.9, 0.5, 0.4, 0.5, 0.5, 0.4, 0.9, 0.8, 0.8],
+        ... [0.3, 0.8, 0.4, 0.8, 0.8, 0.8, 0.7, 0.7, 0.2, 0.8, 0.8, 0.4],
+        ... [0.4, 0.9, 0.5, 0.7, 0.3, 0.2, 0.8, 0.8, 0.1, 0.6, 0.8, 0.8],
+        ... [0.5, 0.1, 0.6, 0.6, 0.2, 0.1, 0.9, 0.9, 0.8, 0.5, 0.8, 0.8]],
+        ...                                   evidence=['intel', 'diff', 'iq'], evidence_card=[2, 3, 2])
+        >>> print(final_cpd)
+╒═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
+│ iq      │ iq_0    │ iq_0    │ iq_0    │ iq_0    │ iq_0    │ iq_0    │ iq_1    │ iq_1    │ iq_1    │ iq_1    │ iq_1    │ iq_1    │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ diff    │ diff_0  │ diff_0  │ diff_1  │ diff_1  │ diff_2  │ diff_2  │ diff_0  │ diff_0  │ diff_1  │ diff_1  │ diff_2  │ diff_2  │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ intel   │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_0 │ 0.12    │ 0.6     │ 0.2     │ 0.9     │ 0.5     │ 0.4     │ 0.5     │ 0.5     │ 0.4     │ 0.9     │ 0.8     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_1 │ 0.12    │ 0.6     │ 0.2     │ 0.9     │ 0.5     │ 0.4     │ 0.5     │ 0.5     │ 0.4     │ 0.9     │ 0.8     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_2 │ 0.3     │ 0.8     │ 0.4     │ 0.8     │ 0.8     │ 0.8     │ 0.7     │ 0.7     │ 0.2     │ 0.8     │ 0.8     │ 0.4     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_3 │ 0.4     │ 0.9     │ 0.5     │ 0.7     │ 0.3     │ 0.2     │ 0.8     │ 0.8     │ 0.1     │ 0.6     │ 0.8     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_4 │ 0.5     │ 0.1     │ 0.6     │ 0.6     │ 0.2     │ 0.1     │ 0.9     │ 0.9     │ 0.8     │ 0.5     │ 0.8     │ 0.8     │
+╘═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
+        >>> final_cpd.set_parents_order(['iq', 'intel', 'diff'])
+        >>> print(final_cpd)
+╒═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
+│ diff    │ diff_0  │ diff_0  │ diff_0  │ diff_0  │ diff_1  │ diff_1  │ diff_1  │ diff_1  │ diff_2  │ diff_2  │ diff_2  │ diff_2  │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ intel   │ intel_0 │ intel_0 │ intel_1 │ intel_1 │ intel_0 │ intel_0 │ intel_1 │ intel_1 │ intel_0 │ intel_0 │ intel_1 │ intel_1 │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ iq      │ iq_0    │ iq_1    │ iq_0    │ iq_1    │ iq_0    │ iq_1    │ iq_0    │ iq_1    │ iq_0    │ iq_1    │ iq_0    │ iq_1    │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_0 │ 0.12    │ 0.5     │ 0.6     │ 0.5     │ 0.2     │ 0.4     │ 0.9     │ 0.9     │ 0.5     │ 0.8     │ 0.4     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_1 │ 0.12    │ 0.5     │ 0.6     │ 0.5     │ 0.2     │ 0.4     │ 0.9     │ 0.9     │ 0.5     │ 0.8     │ 0.4     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_2 │ 0.3     │ 0.7     │ 0.8     │ 0.7     │ 0.4     │ 0.2     │ 0.8     │ 0.8     │ 0.8     │ 0.8     │ 0.8     │ 0.4     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_3 │ 0.4     │ 0.8     │ 0.9     │ 0.8     │ 0.5     │ 0.1     │ 0.7     │ 0.6     │ 0.3     │ 0.8     │ 0.2     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_4 │ 0.5     │ 0.9     │ 0.1     │ 0.9     │ 0.6     │ 0.8     │ 0.6     │ 0.5     │ 0.2     │ 0.8     │ 0.1     │ 0.8     │
+╘═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
+        """
+        old_scope = self.scope()[::-1].copy()
+        new_cardinality = np.asarray(self.evidence_card)
+        new_order.append(self.variable)
+        list_a = [i for i in range(len(self.evidence_card))]
+        # list_b is to be used for cardinality mapping and column mapping.
+        list_b = [new_order[:].index(i) for i in old_scope]
+        new_cardinality[[x for x in list_b[:-1]]] = new_cardinality[[x for x in list_a]]
+        values = self.values.reshape([self.cardinality[0], new_cardinality.prod()])
+        new_heading = np.array(list(product(*[range(i) for i in new_cardinality[::-1]])))
+        new_heading[:, [x for x in list_b[:-1]]] = new_heading[:, [x for x in list_a]]
         new_value_list = []
-        for i in new_heading:
-            temp = (i == old_heading)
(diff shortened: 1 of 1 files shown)


pgmpy/factors/CPD.py
```diff
@@ -112,6 +112,27 @@ def __init__(self, variable, variable_card, values,
 
         super().__init__(variables, cardinality, values.flatten('C'))
 
+    def replace_variables(self, var1, var2):
```
abinashpanda: @kislayabhi Please add docstrings as well as doctests.

pgmpy/factors/CPD.py
```diff
@@ -112,6 +112,27 @@ def __init__(self, variable, variable_card, values,
...
+        new_scope = self.scope()
+        new_scope.reverse()
```
abinashpanda: Use

``` python```
new_scope = self.scope()[::-1]
```

pgmpy/factors/CPD.py
abinashpanda: Have an additional input argument `inplace`; when set as `inplace=True` it should modify the existing CPD else should return a new modified CPD.

ankurankan:
@kislayabhi: Rather than taking the two variable names in the argument, take the new parent order as the argument. This will be useful when we want to reorder more than two variables. And therefore rename the method name to something like `set_parents_order`.

ankurankan:
@kislayabhi I haven't gone through your code completely to see how you are replacing the order but I can think of a very simple way to do this. If you are doing something in these lines then let me know but if not please do consider this approach:

Take a simple case of a CPD `P(A | B, C)` and we want to reorder parents to `P(A | C, B)`. So start by creating a dict of index changes (mapping from final representation to the original) like for this case:

``` python3
{0: 0, 1: 2, 2: 1}
```

Now create a numpy array of assignments for the final representation like:

```
A          C            B
0          0             0
0          0             1
0          1             0
0          1             1
....
```

Now replace the columns in this array according to the above dict (0th column will be at 0th place, 1st column will go to 2nd columns place and 2nd will be at 1st columns place) so that we get something like this:

```
A         B            C
0           0            0
0           1            0
0           0            1
0           1            1
...
```

And using `index_for_assignment` using these assignments, get the indexes from the original CPD and use that indexes to get the values for the new CPD.

You can also use the above dict to get the new cardinality.

kislayabhi:
@ankurankan : Bingo! Same here. :+1: 

Just that I haven't used `index_for_assignment`. I used manual loops to find them.

pgmpy/factors/CPD.py
```diff
@@ -112,6 +112,27 @@ def __init__(self, variable, variable_card, values,
...
+        new_cardinality = self.cardinality[::-1].copy()
```
kislayabhi: @ankurankan @abinashpanda : I am using `.copy()` here. If I don't use that, my whole code goes berserk. The problem is that in line number 124, I am trying to swap two values of the numpy array `new_cardinality`. If I don't use `copy()` here, the values are also swapped in `self.cardinality()`

Is this consistent?

pgmpy/factors/CPD.py
kislayabhi: @ankurankan  @abinashpanda  yes, the above one is cool but I have some inconsistent behavior when `[::-1]` is used to reverse the list. I have to swap some values in `new_scope` but this inturn changes `self.scope()` too if used with `[::-1]`. Something very similar happened with `new_cardinality` too. Please refer to the comment I did below.

pgmpy/factors/CPD.py
abinashpanda: @kislayabhi When you are just using `new_cardinality = self.cardinality[::-1]` you are basically doing a shallow copy. You can find more about about shallow copy and deep copy at https://docs.python.org/2/library/copy.html. So when you change the elements of `new_cardinality` then it would change the values of `self.cardinality` as `new_cardinality` contains the references of the objects present in `self.cardinality`.

So I would suggest you to do something like this

``` python
new_cardinality = np.array([var_card for var_card in self.cardinality[::-1]])
```

This way you are not creating a shallow copy rather you are creating a new array object with recursively adding the copy of the objects in `self.cardinality` so changing the elements of `new_cardinality` won't have any effect on `self.cardinality`. 

Hope this makes sense :)

kislayabhi: @abinashpanda : Yup. I get it now. But is it any way better than `.copy()` as it happens to be more readable.

abinashpanda: @kislayabhi I did some tests and found that `.copy()` is much faster as compared to the one that I had suggested.

``` python
In [1]: import numpy as np
In [2]: a = np.arange(100000)
In [3]: %timeit new_cardinality = np.array([var_card for var_card in a[::-1]])
100 loops, best of 3: 13 ms per loop
In [4]: %timeit new_cardinality = a.copy()
The slowest run took 6.74 times longer than the fastest. This could mean that an intermediate result is being cached 
10000 loops, best of 3: 50.9 µs per loop
```

So I would suggest you to go with `.copy()`.

kislayabhi: :+1:

kislayabhi:
@ankurankan, I have taken care of most of the comments. 

@abinashpanda, If the `inplace=False` , how to return a new modified CPD. (Should I choose any name myself?)

ankurankan:
@kislayabhi Just create a new CPD object and return it.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +125,79 @@ def __repr__(self):
...
+        >>> from pgmpy.factors import TabularCPD
+        >>> final_cpd = TabularCPD('grade', 5, [
+        ... [0.12, 0.6, 0.2, 0.9, 0.5, 0.4, 0.5, 0.5, 0.4, 0.9, 0.8, 0.8],
+        ... [0.12, 0.6, 0.2, 0.9, 0.5, 0.4, 0.5, 0.5, 0.4, 0.9, 0.8, 0.8],
+        ... [0.3, 0.8, 0.4, 0.8, 0.8, 0.8, 0.7, 0.7, 0.2, 0.8, 0.8, 0.4],
+        ... [0.4, 0.9, 0.5, 0.7, 0.3, 0.2, 0.8, 0.8, 0.1, 0.6, 0.8, 0.8],
+        ... [0.5, 0.1, 0.6, 0.6, 0.2, 0.1, 0.9, 0.9, 0.8, 0.5, 0.8, 0.8]],
+        ...                                   evidence=['intel', 'diff', 'iq'], evidence_card=[2, 3, 2])
+        >>> print(final_cpd)
+╒═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
```
abinashpanda, line 149: @kislayabhi This make the doc string looks ugly. So instead of 

``` python
>>> print(final_cpd)
```

use 

``` python
>>> final_cpd
```

It would call the `__repr__` method which also shows the ordering of the variables.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +125,79 @@ def __repr__(self):
...
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_2 │ 0.3     │ 0.8     │ 0.4     │ 0.8     │ 0.8     │ 0.8     │ 0.7     │ 0.7     │ 0.2     │ 0.8     │ 0.8     │ 0.4     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_3 │ 0.4     │ 0.9     │ 0.5     │ 0.7     │ 0.3     │ 0.2     │ 0.8     │ 0.8     │ 0.1     │ 0.6     │ 0.8     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_4 │ 0.5     │ 0.1     │ 0.6     │ 0.6     │ 0.2     │ 0.1     │ 0.9     │ 0.9     │ 0.8     │ 0.5     │ 0.8     │ 0.8     │
+╘═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
+        >>> final_cpd.set_parents_order(['iq', 'intel', 'diff'])
+        >>> print(final_cpd)
+╒═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
```
abinashpanda, line 168: @kislayabhi Same as above.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +125,79 @@ def __repr__(self):
...
+│ grade_1 │ 0.12    │ 0.5     │ 0.6     │ 0.5     │ 0.2     │ 0.4     │ 0.9     │ 0.9     │ 0.5     │ 0.8     │ 0.4     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_2 │ 0.3     │ 0.7     │ 0.8     │ 0.7     │ 0.4     │ 0.2     │ 0.8     │ 0.8     │ 0.8     │ 0.8     │ 0.8     │ 0.4     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_3 │ 0.4     │ 0.8     │ 0.9     │ 0.8     │ 0.5     │ 0.1     │ 0.7     │ 0.6     │ 0.3     │ 0.8     │ 0.2     │ 0.8     │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_4 │ 0.5     │ 0.9     │ 0.1     │ 0.9     │ 0.6     │ 0.8     │ 0.6     │ 0.5     │ 0.2     │ 0.8     │ 0.1     │ 0.8     │
+╘═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
+        """
+        old_scope = self.scope()[::-1].copy()
```
ankurankan, line 186: @kislayabhi Add checks to raise an exception if `new_order` has some variable that is not in the actual factor.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +125,79 @@ def __repr__(self):
...
+        new_order.append(self.variable)
+        list_a = [i for i in range(len(self.evidence_card))]
+        # list_b is to be used for cardinality mapping and column mapping.
+        list_b = [new_order[:].index(i) for i in old_scope]
+        new_cardinality[[x for x in list_b[:-1]]] = new_cardinality[[x for x in list_a]]
+        values = self.values.reshape([self.cardinality[0], new_cardinality.prod()])
+        new_heading = np.array(list(product(*[range(i) for i in new_cardinality[::-1]])))
+        new_heading[:, [x for x in list_b[:-1]]] = new_heading[:, [x for x in list_a]]
+        new_value_list = []
+        [new_value_list.append(self._index_for_assignment([0] + list(i))[0]) for i in new_heading]
```
ankurankan, line 197: @kislayabhi Never do this. Better use a loop. More better if you can use `map`.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +125,79 @@ def __repr__(self):
...
+        ...                                   evidence=['intel', 'diff', 'iq'], evidence_card=[2, 3, 2])
+        >>> print(final_cpd)
+╒═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
+│ iq      │ iq_0    │ iq_0    │ iq_0    │ iq_0    │ iq_0    │ iq_0    │ iq_1    │ iq_1    │ iq_1    │ iq_1    │ iq_1    │ iq_1    │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ diff    │ diff_0  │ diff_0  │ diff_1  │ diff_1  │ diff_2  │ diff_2  │ diff_0  │ diff_0  │ diff_1  │ diff_1  │ diff_2  │ diff_2  │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ intel   │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │ intel_0 │ intel_1 │
+├─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
+│ grade_0 │ 0.12    │ 0.6     │ 0.2     │ 0.9     │ 0.5     │ 0.4     │ 0.5     │ 0.5     │ 0.4     │ 0.9     │ 0.8     │ 0.8     │
```
ankurankan, line 156: @kislayabhi And please also do fix these values. Since its a CPD each column should sum to 1.

pgmpy/factors/CPD.py
```diff
@@ -125,6 +125,79 @@ def __repr__(self):
...
+│ grade_4 │ 0.5     │ 0.9     │ 0.1     │ 0.9     │ 0.6     │ 0.8     │ 0.6     │ 0.5     │ 0.2     │ 0.8     │ 0.1     │ 0.8     │
+╘═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
+        """
+        old_scope = self.scope()[::-1].copy()
+        new_cardinality = np.asarray(self.evidence_card)
+        new_order.append(self.variable)
+        list_a = [i for i in range(len(self.evidence_card))]
+        # list_b is to be used for cardinality mapping and column mapping.
+        list_b = [new_order[:].index(i) for i in old_scope]
+        new_cardinality[[x for x in list_b[:-1]]] = new_cardinality[[x for x in list_a]]
```
ankurankan, line 192: @kislayabhi Simply doing `list_b[:-1]` would give the same result as `[x for x in list_b[:-1]]`.

ankurankan:
@kislayabhi I still feel that the code can be made much more cleaner.

PR #364, opened 2015-04-07
Upgraded the networkx version.

abinashpanda:


abinashpanda:
The new version of `networkx` is required for some functions that were not present in the earlier versions such as `is_tree`. I checked locally and the new version of `networkx` is not creating any issue. If all test passes then it is go to merge.

PR #361, opened 2015-04-03
Added FactorSet class for approximate inference.

abinashpanda:

PR #366, opened 2015-04-07
Cluster graph

abinashpanda:


abinashpanda:
@ankurankan  Please merge the PR #361 before merging this PR as it would have duplicate commits.

pgmpy/factors/FactorSet.py
```diff
@@ -0,0 +1,191 @@
+#!/usr/bin/env python3
+from pgmpy.factors import Factor
+from functools import reduce
```
ankurankan: @abinashpanda No need to import `reduce` from `functools`. It's builtin https://docs.python.org/2/library/functools.html#functools.reduce

pgmpy/factors/FactorSet.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class FactorSet:
```
ankurankan, line 6: @abinashpanda I think we should also have `add` and `remove` methods to add/remove factors from `FactorSet`? What do you think?

pgmpy/factors/FactorSet.py
abinashpanda: @ankurankan ```reduce``` is built-in in ```Python2.7``` not in ```Python3```. https://docs.python.org/3.0/whatsnew/3.0.html#builtins

pgmpy/factors/FactorSet.py
abinashpanda, line 6: @ankurankan Yeah. I agree on this. I would add these methods as well.

pgmpy/factors/FactorSet.py
```diff
@@ -0,0 +1,191 @@
...
+        >>> from pgmpy.factors import FactorSet
+        >>> from pgmpy.factors import Factor
+        >>> phi1 = Factor(['x1', 'x2', 'x3'], [2, 3, 2], range(12))
+        >>> phi2 = Factor(['x3', 'x4', 'x1'], [2, 2, 2], range(8))
+        >>> factor_set = FactorSet(phi1, phi2)
+        """
+        from copy import deepcopy
+        if not all(isinstance(phi, Factor) for phi in factors_list):
+            raise TypeError("Input parameters must be all factors")
+        self.factors_set = set(deepcopy(factors_list))
```
ankurankan: @abinashpanda Don't you think just `self.factors` would be a more appropriate name? Like when we want to access the factors of a FactorSet we would do `fset.factors`. Right now we have to do `fset.factors_set`. What do you think?

abinashpanda: @ankurankan Okay. I am fine with `factors`. I would make the changes.

pgmpy/factors/FactorSet.py
```diff
@@ -0,0 +1,191 @@
...
+    >>> phi2 = Factor(['x3', 'x4', 'x1'], [2, 2, 2], range(8))
+    >>> factor_set1 = FactorSet(phi1, phi2)
+    >>> phi3 = Factor(['x5', 'x6', 'x7'], [2, 2, 2], range(8))
+    >>> phi4 = Factor(['x5', 'x7', 'x8'], [2, 2, 2], range(8))
+    >>> factor_set2 = FactorSet(phi3, phi4)
+    >>> factor_set3 = factorset_product(factor_set1, factor_set2)
+    """
+    if not all(isinstance(factorset, FactorSet) for factorset in factorsets_list):
+        raise TypeError("Input parameters must be factor sets")
+    return reduce(lambda x, y: FactorSet(*(x.factors_set.union(y.factors_set))), factorsets_list)
```
ankurankan: @abinashpanda This doesn't look very good. We are creating FactorSet objects in each iteration. What do you think of something like this:

``` python3
FactorSet(*set.union(*map(lambda t: t.factors_set, factorsets_list)))
```

Also this is faster. For `factorsets_list` having 200 elements:

``` python3
%timeit FactorSet(*set.union(*map(lambda t: t.factors_set, factorsets_list)))
1000 loops, best of 3: 305 µs per loop

%timeit reduce(lambda x, y: FactorSet(*(x.factors_set.union(y.factors_set))), factorsets_list)
1 loops, best of 3: 178 ms per loop
```

pgmpy/factors/FactorSet.py
```diff
@@ -0,0 +1,191 @@
...
+    >>> phi1 = Factor(['x1', 'x2', 'x3'], [2, 3, 2], range(12))
+    >>> phi2 = Factor(['x3', 'x4', 'x1'], [2, 2, 2], range(8))
+    >>> factor_set1 = FactorSet(phi1, phi2)
+    >>> phi3 = Factor(['x5', 'x6', 'x7'], [2, 2, 2], range(8))
+    >>> phi4 = Factor(['x5', 'x7', 'x8'], [2, 2, 2], range(8))
+    >>> factor_set2 = FactorSet(phi3, phi4)
+    >>> factor_set3 = factorset_divide(factor_set1, factor_set2)
+    """
+    if not isinstance(factorset1, FactorSet) or not isinstance(factorset2, FactorSet):
+        raise TypeError("factorset1 and factorset2 must be factor sets")
```
ankurankan: @abinashpanda `must be FactorSet instances`?? Makes the exception a bit more clearer.

pgmpy/models/ClusterGraph.py
```diff
@@ -0,0 +1,276 @@
+#!/usr/bin/env python3
+
+from pgmpy.base import UndirectedGraph
```
ankurankan, line 3: @abinashpanda Order of imports

pgmpy/factors/FactorSet.py
ankurankan, line 2: @abinashpanda Order of imports PEP8

pgmpy/factors/FactorSet.py
```diff
@@ -0,0 +1,191 @@
...
+        >>> phi1 = Factor(['x1', 'x2', 'x3'], [2, 3, 2], range(12))
+        >>> phi2 = Factor(['x3', 'x4', 'x1'], [2, 2, 2], range(8))
+        >>> factor_set1 = FactorSet(phi1, phi2)
+        >>> factor_set1.marginalize('x1')
+        """
+        if not isinstance(variables, (list, set, tuple)):
+            variables = [variables]
+
+        factors_to_be_marginalized = set(filter(lambda x: len(set(x.scope()).intersection(variables)) != 0,
+                                                self.factors_set))
```
ankurankan: @abinashpanda: We could simply do:

``` python3
factors_to_be_marginalized = set(filter(lambda x: set(x.scope()).intersection(variables), self.factors_set))
```

PR #368, opened 2015-04-07
Induced width

alekdimi:
I think I have implemented what needed to be done, though I am unsure if the method is in the right class.


ankurankan:
@alekdimi We want the `induced_width` method to take in an elimination order as its argument and give the width as the result. So, please make the necessary changes.

alekdimi:
My apologies, I must have overlooked that part, will get right on it.

ankurankan:
@alekdimi: Can you please also add a few tests for this method?

PR #378, opened 2015-04-16
Efficient factor product

jhonatanoliveira:
A factor product following "Probabilistic Graphical Models" (Koller 09) on page 359, Algorithm 10.A.1 - Efficient implementation of a factor product operation.
Koller's algorithm was modified to fit the configuration used in pgmpy. For example, in pgmpy the configurations of Factor are supposed to be like (0,0,0) (0,0,1) (0,1,0) (1,0,0) and so on, instead of (0,0,0) (1,0,0) (0,1,0) (0,0,1) as expected for Koller's algorithm.

Koller's implementation is around 98% faster than the current one in pgmpy. This benchmark was done by using a simple python script as follows:

```
from pgmpy.factors import Factor
from pgmpy.factors import factor_product
from time import time

phi = Factor(['x1', 'x2'], [2, 2], range(4))
phi1 = Factor(['x3', 'x4'], [2, 2], range(4))
t0 = time()
prod = factor_product(phi, phi1)
t1 = time()
print(t1-t0)
```

After running 6 time each implementation, here is the results:

Comparison

Unfortunately, we don't know how to use _JobLib_. But we leave this **TODO** with the hope that using parallel computation can improve  this implementation even further.

### DIFF
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -612,34 +612,35 @@ def _bivar_factor_operation(phi1, phi2, operation, n_jobs=1):
                        for var in phi2.variables
                        if var not in phi1.variables)
     quantity_values = np.prod(np.array(cardinality))
-    values = [0] * quantity_values
-    # Algorithm 10.A.1
+    # Constants for jumping in j and i
+    j_star = {}
+    k_star = {}
+    for idx, variable in zip(reversed(range(len(variables))),
+                             reversed(variables)):
+        j_star[variable] = (cardinality[idx] - 1) * phi1.stride(variable)
+        k_star[variable] = (cardinality[idx] - 1) * phi2.stride(variable)
+    # Counter for the assignments within the domain cardinality of variables
+    assignment = {var: 0 for var in variables}
+    # Loop building the product table
+    values = []
     j = 0
     k = 0
-    assignment = {var: 0 for var in variables}
     for i in range(quantity_values):
-        # We re-organize the values list to match the configuration
-        # setting used in pgmpy. Ex: (0,0,0), (0,0,1), (0,1,0),
-        # (1,0,0), and so on instead of (0,0,0) (1,0,0),
-        # (0,1,0), (0,0,1)
-        index = 0
-        for v in variables:
-            index = index + assignment[v] * _stride(
-                                            v, variables, cardinality)
         if operation == "M":
-            values[index] = phi1.values[j] * phi2.values[k]
+            values.append(phi1.values[j] * phi2.values[k])
         elif operation == "D":
             # Zero division should return zero
             if phi2.values[k] == 0:
-                values[index] = 0
+                values.append(0)
             else:
-                values[index] = phi1.values[j] / phi2.values[k]
-        for idx, variable in enumerate(variables):
+                values.append(phi1.values[j] / phi2.values[k])
+        for idx, variable in zip(reversed(range(len(variables))),
+                                 reversed(variables)):
             assignment[variable] = assignment[variable] + 1
             if assignment[variable] == cardinality[idx]:
                 assignment[variable] = 0
-                j = j - (cardinality[idx] - 1) * phi1.stride(variable)
-                k = k - (cardinality[idx] - 1) * phi2.stride(variable)
+                j = j - j_star[variable]
+                k = k - k_star[variable]
             else:
                 j = j + phi1.stride(variable)
                 k = k + phi2.stride(variable)
@@ -647,6 +648,20 @@ def _bivar_factor_operation(phi1, phi2, operation, n_jobs=1):
 
     # Construct the product Factor
     phi = Factor(variables, cardinality, values)
+
+    # Re-organize the structure for resultant factor
+    lhs = []
+    if operation == 'M':
+        lhs = list(set(phi1.left_hand_side).union(set(phi2.left_hand_side)))
+    elif operation == 'D':
+        lhs = [v for v in set(phi1.left_hand_side).union(
+               set(phi2.left_hand_side))
+               if v not in phi2.left_hand_side]
+    rhs = [v for v in set(phi1.right_hand_side).union(
+            set(phi2.right_hand_side))
+           if v not in lhs]
+    phi.left_hand_side = lhs
+    phi.right_hand_side = rhs
 
     return phi
 
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -532,46 +532,62 @@ class Factor:
         return hash(' '.join(self.variables) + ' '.join(map(str, self.cardinality)) +
                     ' '.join(list(map(str, self.values))))
 
-    def stride(self, variable):
-        if variable not in self.strides:
-            self.strides[variable] = _stride(variable,
-                                             self.variables, self.cardinality)
-        return self.strides[variable]
-
-
-def _stride(variable, variables, cardinalities):
+def _index_to_assignment(indices, cardinalities):
     """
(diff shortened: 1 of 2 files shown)


ankurankan:
@jhonatanoliveira I getting an error when trying to compute factor product.

``` python3
phi1 = Factor(['x_{i}'.format(i=i) for i in range(10)], [2] * 10, np.arange(2**10))

phi2 = Factor(['x_{i}'.format(i=i) for i in range(10, 20)], [2] * 10, np.arange(2**10))

phi1 * phi2
AttributeError                            Traceback (most recent call last)
<ipython-input-17-e0a56e1ddb88> in <module>()
----> 1 phi1 * phi2

/home/ankur/pgmpy/pgmpy/factors/Factor.py in __mul__(self, other)
    488 
    489     def __mul__(self, other):
--> 490         return self.product(other)
    491 
    492     def __truediv__(self, other):

/home/ankur/pgmpy/pgmpy/factors/Factor.py in product(self, n_jobs, *factors)
    310                 ('x3', ['x3_0', 'x3_1']), ('x4', ['x4_0', 'x4_1'])])
    311         """
--> 312         return factor_product(self, *factors, n_jobs=n_jobs)
    313 
    314     def divide(self, factor, n_jobs=1):

/home/ankur/pgmpy/pgmpy/factors/Factor.py in factor_product(n_jobs, *args)
    717         raise TypeError("Input parameters must be factors")
    718     return functools.reduce(lambda phi1, phi2: _bivar_factor_operation(phi1, phi2, operation='M',
--> 719                                                                        n_jobs=n_jobs), args)
    720 
    721 

/home/ankur/pgmpy/pgmpy/factors/Factor.py in <lambda>(phi1, phi2)
    717         raise TypeError("Input parameters must be factors")
    718     return functools.reduce(lambda phi1, phi2: _bivar_factor_operation(phi1, phi2, operation='M',
--> 719                                                                        n_jobs=n_jobs), args)
    720 
    721 

/home/ankur/pgmpy/pgmpy/factors/Factor.py in _bivar_factor_operation(phi1, phi2, operation, n_jobs)
    654     lhs = []
    655     if operation == 'M':
--> 656         lhs = list(set(phi1.left_hand_side).union(set(phi2.left_hand_side)))
    657     elif operation == 'D':
    658         lhs = [v for v in set(phi1.left_hand_side).union(

AttributeError: 'Factor' object has no attribute 'left_hand_side'

```

ankurankan:
@jhonatanoliveira And are you sure that you have installed the latest version of pgmpy? For the example you have used my machine took 0.0000531s to compute.

ankurankan:
@jhonatanoliveira And also please merge the latest develop branch in your current branch. And resolve the merge conflicts.

jhonatanoliveira:
Hi, @ankurankan . I've merged with the latest dev branch (which had only one commit inside `_bivar_factor_operation` ignored by me, since we've re-written this function). I've also fixed the error you've experienced: a portion of code that shouldn't be there (my fault). If there is anything further to improve in this PR, please let me know. Regards.

ankurankan:
@jhonatanoliveira: This implementation is taking much longer on my machine than the current implementation. Results:
This implementation:

``` python3
phi1 = Factor(['x_{i}'.format(i=i) for i in range(10)], [2] * 10, np.arange(2**10))

phi2 = Factor(['x_{i}'.format(i=i) for i in range(10, 20)], [2] * 10, np.arange(2**10))

%timeit phi1 * phi2
1 loops, best of 3: 2.3 s per loop
```

For the current implementation:

``` python3
%timeit phi1 * phi2
100 loops, best of 3: 5.43 ms per loop
```

And I feel this is mainly because you are computing each final value using a loop.

jhonatanoliveira:
Thanks for noticing this, @ankurankan. Indeed, for bigger tables the current implementation is faster. And you're correct, probably the `for` loops are slowing it down. Please, hold this PR for while and let us try a matrix-wise implementation of this function, using `numpy`. Regards.

jhonatanoliveira:
Hi, @ankurankan.

We just finished implementing two factor operations (multiplication and marginalization) using Koller's algorithm, but now with matrices (from numpy). Fortunately, we got good results (look at the tables below). This implementation is following Koller's book and tips from her classes in Coursera. But notice that hers assignments (configurations) pattern is different from the one used in PgmPy.
For example, a factor with two binary variables would have the following assignments: `[[ 0.  0.] [ 1.  0.] [ 0.  1.] [ 1.  1.]]` for her, while for PgmPy it would be `[[ 0.  0.] [ 1.  0.] [ 0.  1.] [ 1.  1.]]`. Our implementation modifies her algorithm to produce the assignments in our order.

Our benchmark set up is very simple and used the following code:

```
import numpy as np
from math import ceil
from pgmpy.factors import Factor
from pgmpy.factors import factor_product
from time import time

size = 15  # 2 variables or 15.
# Quantity of variables to be marginalized
vars_to_marg = 1  # One variable
# vars_to_marg = ceil(size/2)  # Half variables
# vars_to_marg = size  # All variables

# Define factors
phi1 = Factor(['x_{i}'.format(i=i)
               for i in range(size)], [2] * size, np.random.rand(2**size))
phi2 = Factor(['x_{i}'.format(i=i)
               for i in range(size)], [2] * size, np.random.rand(2**size))

# Timing factor product
t0_prod = time()
prod = factor_product(phi1, phi2)
t1_prod = time()
print("Product: %s" % str(t1_prod - t0_prod))

# Timing factor marginalization
t0_marg = time()
marg = prod.marginalize(['x_{i}'.format(i=i)
                         for i in range(vars_to_marg)],
                        inplace=False)
t1_marg = time()
print("Marginalization: %s" % str(t1_marg - t0_marg))
```

Basically, we run 6 times each configuration. For the marginalization we also changed the number of variables being marginalized, in order to see the effect of this on timing the operation.
The following tables gives the results, where the _Improvements_ rows shows how many time faster is the new implementation compared with the current one in PgmPy.

benchmarks

ankurankan:
@jhonatanoliveira The current implementation of factor product is still faster on my machine. For this PR's implementation I am getting:

``` python3
phi1 = Factor(['x_{i}'.format(i=i) for i in range(10)], [2] * 10, np.arange(2**10))

phi2 = Factor(['x_{i}'.format(i=i) for i in range(10, 20)], [2] * 10, np.arange(2**10))

%timeit phi1 * phi2
1 loops, best of 3: 1.17 s per loop
```

as compared to 5.34 ms for the current implementation. But this PR's marginalization is around 4x faster than the current implementation. 

So, I will try to merge the `marginalization` changes.

@abinashpanda Can you also check if you are getting similar results on your machine ?

jhonatanoliveira:
Thanks for noticing this @ankurankan! Indeed, we were missing another setting of benchmark: the number of common variables. Notice that for a factor product with all variables in common, the current implementation is poor, but for no variables in common our implementation is the poor one. Now, I am wondering if this is due to the use of joblib in current implementation. Anyway, we'll figure it out later. I think that the marginalization is better in all cases we could test. If you agree, go on merging it, please. Thanks again.

jhonatanoliveira:
I think it's better to close this Pull Request by now. The reason is that I couldn't find a better solution for optimizing the factor manipulations and @ankurankan is working on this by now.
Trying to catch all benchmarks settings, I could get the following results:

screen shot 2015-04-22 at 11 14 30 am

But the problem with the vectorized version of Koller (though it's the fast one) is memory consumption. So, I concluded that we have to choose between both worlds: either a fast alg with high memory consumption or a slow one but we reasonable memory consumption.

We can discuss this later, but by now it's better to do not change PgmPy current implementation.

Best regards.

PR #382, opened 2015-04-30
Improved performance of _index_for_assignment and issue fix

### LINKED ISSUE #381: Factor reduce scope and variable ordering
**Problem 1: factor reduction makes an assumption on the order in which variables are passed in:**

Original factor:
╒═════╤═════╤════════════╕
│ 3   │ 2   │   phi(3,2) │
╞═════╪═════╪════════════╡
│ 3_0 │ 2_0 │     0.3900 │
├─────┼─────┼────────────┤
│ 3_0 │ 2_1 │     0.0600 │
├─────┼─────┼────────────┤
│ 3_1 │ 2_0 │     0.6100 │
├─────┼─────┼────────────┤
│ 3_1 │ 2_1 │     0.9400 │
╘═════╧═════╧════════════╛
Observed (2_0, 3_1):
╒═════════╕
│   phi() │
╞═════════╡
│  0.0600 │
╘═════════╛

Expected output: 0.61. The problem occurs because the evidence was passed in as (2_0, 3_1) instead of (3_1, 2_0).

**Problem 2: when a variable is passed in that is not in scope, an error occurs.**

Original factor:
╒═════╤═════╤════════════╕
│ 2   │ 1   │   phi(2,1) │
╞═════╪═════╪════════════╡
│ 2_0 │ 1_0 │     0.5900 │
├─────┼─────┼────────────┤
│ 2_0 │ 1_1 │     0.2200 │
├─────┼─────┼────────────┤
│ 2_1 │ 1_0 │     0.4100 │
├─────┼─────┼────────────┤
│ 2_1 │ 1_1 │     0.7800 │
╘═════╧═════╧════════════╛

Observed (2_0, 3_1):

ERROR: test_observe_evidence

  File "../pgmpy/factors/Factor.py", line 276, in reduce
    raise Exceptions.ScopeError('%s not in scope' % list(set(reduced_variables)-set(factor.scope())))
pgmpy.exceptions.Exceptions.ScopeError: "['3'] not in scope"

The error occurs because '3' is not in scope of the factor.

Problem 2 doesn't happen when the factor is used through one of the inference methods (variable elimination or belief propagation). I'm not sure about problem 1. However, in either case the implementation now couples the Factor class to the VE / BP implementations. If that's the intended use, i.e. if Factor shouldn't be used on its own, it should be a private class.

I propose to fix the above in the Factor class. It's not a difficult fix and shouldn't have performance impact.

sytham:
- fixed problem with reduce() if variables are passed in in different
  order from which they are in factor scope. Fixes pgmpy/pgmpy#381
  (partial fix; chose not to fix problem 2)
- improved performance of _index_for_assignment: 4 - 50x faster. See below timeit results:

setup: from pgmpy.factors import Factor, factor_product
F = Factor(['1','2','3','4','5','6','7','8'], [4,2,2,3,5,2,2,3], range(2880))

timeit.timeit("F._index_for_assignment([1,1,1,1,1,1,1,1])", setup=setup, number=10000)
0.18317501399906178
timeit.timeit("F._index_for_assignment_old([1,1,1,1,1,1,1,1])", setup=setup, number=10000)
0.44581467299940414

timeit.timeit("F._index_for_assignment([1,1,-1,1,-1,1,1,1])", setup=setup, number=10000)
0.9840771199997107
timeit.timeit("F._index_for_assignment_old([1,1,-1,1,-1,1,1,1])", setup=setup, number=10000)
2.7516035469998315

timeit.timeit("F._index_for_assignment([1,1,-1,1,-1,1,-1,1])", setup=setup, number=10000)
1.382639981999091
timeit.timeit("F._index_for_assignment_old([1,1,-1,1,-1,1,-1,1])", setup=setup, number=10000)
4.924537786999281

timeit.timeit("F._index_for_assignment([1,1,-1,1,-1,1,-1,-1])", setup=setup, number=10000)
1.7535868900013156
timeit.timeit("F._index_for_assignment_old([1,1,-1,1,-1,1,-1,-1])", setup=setup, number=10000)
13.60733557200001

timeit.timeit("F._index_for_assignment([1,-1,-1,-1,-1,-1,-1,-1])", setup=setup, number=10000)
3.316042758000549
timeit.timeit("F._index_for_assignment_old([1,-1,-1,-1,-1,-1,-1,-1])", setup=setup, number=10000)
161.2918238050006

### DIFF
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -4,6 +4,7 @@ from itertools import product, starmap
 from collections import OrderedDict
 
 import numpy as np
+from sklearn.utils.extmath import cartesian
 
 from pgmpy.exceptions import Exceptions
 from pgmpy.extern import tabulate
@@ -67,6 +68,7 @@ class Factor:
         >>> from pgmpy.factors import Factor
         >>> phi = Factor(['x1', 'x2', 'x3'], [2, 2, 2], np.ones(8))
         """
+
         self.variables = OrderedDict()
         if len(variables) != len(cardinality):
             raise ValueError("The size of variables and cardinality should be same")
@@ -77,6 +79,8 @@ class Factor:
         self.values = np.array(value, dtype=np.double)
         if not self.values.shape[0] == np.prod(self.cardinality):
             raise Exceptions.SizeError("Incompetant value array")
+        
+        self._stride = np.cumprod(np.r_[self.cardinality[1:],1][::-1])[::-1]
 
     def scope(self):
         """
@@ -200,12 +204,11 @@ class Factor:
         marginalized_values = []
         for i in product(*[range(index) for index in assign[assign != -1]]):
             assign[assign != -1] = i
-            marginalized_values.append(np.sum(factor.values[factor._index_for_assignment(assign)]))
+            idxa = factor._index_for_assignment(assign)
+            marginalized_values.append(np.sum(factor.values[idxa]))
         factor.values = np.array(marginalized_values)
-        for variable in variables:
-            index = list(factor.variables.keys()).index(variable)
-            del(factor.variables[variable])
-            factor.cardinality = np.delete(factor.cardinality, index)
+        factor._remove_variables(variables)
+        
         if not inplace:
             return factor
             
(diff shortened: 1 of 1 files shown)


pgmpy/factors/Factor.py
```diff
@@ -4,6 +4,7 @@
 from collections import OrderedDict
 
 import numpy as np
+from sklearn.utils.extmath import cartesian
```
abinashpanda, line 7: Please don't add any other dependencies unless we can't deal with it. `cartesian` method is not a very complicated method and can be implemented very easily in `pgmpy` as well. So adding `sklearn` as a dependency just for this method doesn't make any sense.

abinashpanda:
@sytham As per your comments in the PR, you said that you had made some changes in the `_index_for_assignment` method. But I don't see them in this PR. Could you please check that.

sytham:
If you check the diff on Factor.py, lines 410 - 422 in red are what I removed from `_index_for_assignment` and lines 414 - 430 are what I replaced it with. I essentially vectorized the code.

Sure, I can add the cartesian function in pgmpy itself. Where do you think it's best to add it? Create a utils package?

ankurankan:
Not relevant anymore since `_index_for_assignment` has been removed.

PR #401, opened 2015-05-26
Added the XMLBIFWriter class. Also removed some pep8 errors.

vivek425ster:


pgmpy/readwrite/XMLBIF.py
```diff
@@ -159,9 +160,10 @@ def get_property(self):
          'hear-bark': ['position = (154, 241)'],
          'light-on': ['position = (73, 165)']}
         """
-        variable_property = {variable.find('NAME').text: [property.text for property in variable.findall('PROPERTY')]
-                             for variable in self.network.findall('VARIABLE')}
-        return variable_property
+        self.variable_property = {variable.find('NAME').text:
+                                  [property.text for property in variable.findall('PROPERTY')]
+                                  for variable in self.network.findall('VARIABLE')}
+        return self.variable_property
```
ankurankan: @vivek425ster Why this change?

vivek425ster: @ankurankan I messed this while resolving conflicts. Just changed it.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,83 @@ def get_model(self):
...
+                elem.tail = i
+
+    def get_variables(self):
+        """
+        Add variables to XMLBIF
+
+        Return
+        ---------------
+        xml containing variables tag
+        """
```
jaidevd, line 271: Write an example in the docstring here.

pgmpy/tests/test_readwrite/test_XMLBIF.py
```diff
@@ -2,7 +2,17 @@
...
+try:
+    from lxml import etree
+except ImportError:
+    try:
+        import xml.etree.cElementTree as etree
+    except ImportError:
+        try:
+            import xml.etree.ElementTree as etree
+        except ImportError:
+            warnings.warn("Failed to import ElementTree from any known place")
```
jaidevd, line 17: There has to be a better way of handling this than having so many nested try / except catches.

vivek425ster, line 17: @jaidevd i can't find any better way. Stackoverflow shows the use of nested try except only.http://stackoverflow.com/questions/342437/best-way-to-import-version-specific-python-modules

ankurankan:
@vivek425ster The `__init__` method of the writer should be accepting an instance of some model (Bayesian or Markov) not model data. Also the write class should have a method that accepts a filename as an argument to write the xml into.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+    Base class for writing XMLBIF network file format.
+    """
+    def __init__(self, model, encoding='utf-8', prettyprint=True):
+        """
+        Initialise a XMLBIFWriter object.
+
+        Parameters
+        ----------
+        model: Model to write
+        encoding: String(optional)
```
ankurankan: @vivek425ster `encoding: str (optional)`

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+                elem.tail = i
+            for elem in elem:
+                self.indent(elem, level+1)
+            if not elem.tail or not elem.tail.strip():
+                elem.tail = i
+        else:
+            if level and (not elem.tail or not elem.tail.strip()):
+                elem.tail = i
+
+    def add_variables(self):
```
ankurankan: @vivek425ster Better name would be `get_variables`.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+        Add variables to XMLBIF
+
+        Return
+        ---------------
+        xml containing variables tag
+
+        Examples
+        -------
+        >>> writer = XMLBIFWriter(model)
+        >>> writer.get_variables()
```
ankurankan, line 265: @vivek425ster Add the output also.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+    """
+    Base class for writing XMLBIF network file format.
+    """
+    def __init__(self, model, encoding='utf-8', prettyprint=True):
+        """
+        Initialise a XMLBIFWriter object.
+
+        Parameters
+        ----------
+        model: Model to write
```
ankurankan: @vivek425ster It should be:

```
model: BayesianModel instance
        Model to write
```

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+        >>> writer.get_variables()
+        """
+        variables = self.model.nodes()
+        variable_tag = {}
+        for var in sorted(variables):
+            variable_tag[var] = etree.SubElement(self.network, "VARIABLE", attrib={'TYPE': 'nature'})
+            etree.SubElement(variable_tag[var], "NAME").text = var
+        return variable_tag
+
+    def add_states(self):
```
ankurankan: @vivek425ster Same as above.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+        Add outcome to variables of XMLBIF
+
+        Return
+        ---------------
+        xml containing outcome tag
+
+        Examples
+        -------
+        >>> writer = XMLBIFWriter(model)
+        >>> writer.get_states()
```
ankurankan, line 290: @vivek425ster Same as above

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+            etree.SubElement(variable_tag[var], "NAME").text = var
+        return variable_tag
+
+    def add_states(self):
+        """
+        Add outcome to variables of XMLBIF
+
+        Return
+        ---------------
+        xml containing outcome tag
```
ankurankan: @vivek425ster It returns a dict. So, change this to something like: `dict: dict of type {node: xml}`

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+        self.model = model
+
+        self.encoding = encoding
+        self.prettyprint = prettyprint
+
+        self.xml = etree.Element("BIF", attrib={'version': '0.3'})
+        self.network = etree.SubElement(self.xml, 'NETWORK')
+        try:
+            etree.SubElement(self.network, 'NAME').text = self.model['network_name']
+        except KeyError:
```
jaidevd: Instead of this, why not 

``` python
etree.SubElement(self.network, 'NAME').text = self.model.get('network_name')
```

Eliminates the need for the try/except.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -176,3 +177,202 @@ def get_model(self):
...
+        """
+        cpds = self.model.get_cpds()
+        definition_tag = self.definition
+        table_tag = {}
+        for cpd in cpds:
+            table_tag[cpd.variable] = etree.SubElement(definition_tag[cpd.variable], "TABLE")
+            table_tag[cpd.variable].text = ''
+            for val in cpd.values:
+                table_tag[cpd.variable].text += str(val)
+                table_tag[cpd.variable].text += ' '
```
jaidevd: These two lines can be combined into one.

pgmpy/tests/test_readwrite/test_XMLBIF.py
jaidevd, line 17: Okay. But do create an issue before the PR is merged, so we know that this needs to be simplified.

pgmpy/readwrite/XMLBIF.py
vivek425ster: @jaidevd  @ankurankan How can we get the name of Bayesian model. The above code will not work i think.
Should i remove it ?

jaidevd: The one that is there should work. The one I suggested may not work because an `nx.DiGraph` object apparently doesn't have a `get` method.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -182,3 +184,225 @@ def get_model(self):
...
+         'bowel-problem': <Element DEFINITION at 0x7f1d48977348>,
+         'light-on': <Element DEFINITION at 0x7f1d48977448>}
+        """
+        cpds = self.model.get_cpds()
+        cpds.sort(key=lambda x: x.variable)
+        definition_tag = {}
+        for cpd in cpds:
+            definition_tag[cpd.variable] = etree.SubElement(self.network, "DEFINITION")
+            etree.SubElement(definition_tag[cpd.variable], "FOR").text = cpd.variable
+            for child in sorted(cpd.evidence):
```
ankurankan: @vivek425ster This will throw an error when there's no evidence variable in the cpd. You can do something like `sorted([] if cpd.evidence is None else cpd.evidence)` to avoid it

ankurankan:
@vivek425ster Just make the changes as in the comments and this PR should be good to be merged.

vivek425ster:
@ankurankan done.

PR #402, opened 2015-05-29
added sample_discrete function

pratyakshs:


pgmpy/utils/mathext.py
```diff
@@ -48,3 +49,30 @@ def cartesian(arrays, out=None):
...
+    >>> import numpy as np
+    >>> from pgmpy.utils.mathext import sample_discrete
+    >>> values = np.array(['v_0', 'v_1', 'v_2'])
+    >>> probabilities = np.array([0.2, 0.5, 0.3])
+    >>> sample_discrete(values, probabilities, 10)
+    array(['v_1', 'v_1', 'v_0', 'v_1', 'v_2', 'v_0', 'v_1', 'v_1', 'v_1', 'v_2'],
+      dtype='<U3')
+    """
+    bins = np.add.accumulate(weights)
+    return values[np.digitize(random_sample(size), bins)]
```
abinashpanda: PEP8 correction. New line at the end of file.

pgmpy/utils/mathext.py
```diff
@@ -48,3 +49,30 @@ def cartesian(arrays, out=None):
...
 
     return out
+
+def sample_discrete(values, weights, size=1):
+    """
+    Generate a sample of given size, given a probability mass function.
+
+    Parameters
+    ----------
+    values: a numpy.array of possible values that the random variable can take.
```
abinashpanda: It should be something like this.

```

Parameters
----------
values: numpy array
    Array of all possible values that the random variable can take.
```

pratyakshs:
@abinashpanda ready to merge?

pgmpy/inference/approximate_inference/Sampling.py
```diff
@@ -0,0 +1,79 @@
...
+        >>> from pgmpy.models.BayesianModel import BayesianModel
+        >>> from pgmpy.factors.CPD import TabularCPD
+        >>> student = BayesianModel([('diff', 'grade'), ('intel', 'grade')])
+        >>> cpd_d = TabularCPD('diff', 2, [[0.6], [0.4]])
+        >>> cpd_i = TabularCPD('intel', 2, [[0.7], [0.3]])
+        >>> cpd_g = TabularCPD('grade', 3, [[0.3, 0.05, 0.9, 0.5], [0.4, 0.25,
+        ...                0.08, 0.3], [0.3, 0.7, 0.02, 0.2]],
+        ...                ['intel', 'diff'], [2, 2])
+        >>> student.add_cpds(cpd_d, cpd_i, cpd_g)
+        >>> from pgmpy.inference.approximate_inference.Sampling import *
```
ankurankan: @pratyakshs Put this on the top (along with other imports). And don't use import *.

pgmpy/utils/mathext.py
```diff
@@ -48,3 +49,34 @@ def cartesian(arrays, out=None):
...
+    values: numpy.array
+        Array of all possible values that the random variable can take.
+    weights: numpy.array
+        Array representing the PMF of the random variable.
+    size: int
+        Size of the sample to be generated.
+
+    Returns
+    -------
+    a numpy.array of values of the random variable sampled from the given PMF.
```
ankurankan: @pratyakshs Should be like `type: description`. So, 'numpy.array: description`

pgmpy/inference/approximate_inference/Sampling.py
```diff
@@ -0,0 +1,79 @@
...
+        model on which inference queries will be computed
+
+
+    Public Methods
+    --------------
+    forward_sample(size)
+    """
+    def __init__(self, model):
+        if not isinstance(model, BayesianModel):
+            raise TypeError("Argument must be of type BayesianModel")
```
ankurankan: @pratyakshs I feel a better message can be `model must be an instance of BayesianModel`. What do you think?

pgmpy/inference/approximate_inference/Sampling.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
```
ankurankan: @pratyakshs Put this inside inference folder only.

pgmpy/inference/approximate_inference/Sampling.py
```diff
@@ -0,0 +1,79 @@
...
+        >>> cpd_g = TabularCPD('grade', 3, [[0.3, 0.05, 0.9, 0.5], [0.4, 0.25,
+        ...                0.08, 0.3], [0.3, 0.7, 0.02, 0.2]],
+        ...                ['intel', 'diff'], [2, 2])
+        >>> student.add_cpds(cpd_d, cpd_i, cpd_g)
+        >>> from pgmpy.inference.approximate_inference.Sampling import *
+        >>> inference = BayesianModelSampling(student)
+        >>> inference.forward_sample(2)
+        [{'intel': 'intel_0', 'grade': 'grade_1', 'diff': 'diff_1'},
+        {'intel':  'intel_1', 'grade': 'grade_0', 'diff': 'diff_0'}]
+        """
```
ankurankan: @pratyakshs I don't think having a list of dicts for storing the samples is a good idea. It will unnecessarily be consuming too much memory. Maybe return a pandas dataframe or return a tuple of order of variables and the values of samples in the same order. @abinashpanda What do you think?

pgmpy/inference/approximate_inference/Sampling.py
```diff
@@ -0,0 +1,79 @@
...
+    def __init__(self, model):
+        if not isinstance(model, BayesianModel):
+            raise TypeError("Argument must be of type BayesianModel")
+        super(BayesianModelSampling, self).__init__(model)
+        self.topological_order = nx.topological_sort(model)
+        self.cpds = {}
+        for node in model.nodes():
+            self.cpds[node] = model.get_cpds(node)
+
+    def forward_sample(self, size=1):
```
ankurankan: @pratyakshs What about the cases when we have some observed variables?

pgmpy/inference/approximate_inference/Sampling.py
pratyakshs: I agree

pgmpy/inference/approximate_inference/Sampling.py
pratyakshs: For that, we use rejection sampling. Will send a PR soon.

pgmpy/inference/approximate_inference/Sampling.py
pratyakshs: @ankurankan @abinashpanda I've replaced the dict to DataFrame in the final list I return, but I'm keeping a single dict to represent one single sample. This, although a bit slower than using just dicts, should save some memory. Please check the latest commit.

ankurankan:
@pratyakshs `cartesian` is used in one of the PRs and will be merged soon, so please revert this.

pratyakshs:
Oh okay, reverted.

pratyakshs:
@ankurankan @abinashpanda Any would like to review this?

ankurankan:
@pratyakshs I will review it by tomorrow..

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,166 @@
...
+                    weights = cpd.reduce(evid, inplace=False).values
+                else:
+                    weights = cpd.values
+                particle[node] = sample_discrete(cpd.variables[cpd.variable],
+                                                 weights)
+            sampled.loc[i] = [particle[node] for node in self.topological_order]
+        return sampled
+
+    @staticmethod
+    def check_if_consistent(dict1, dict2):
```
abinashpanda: This function although being a `staticmethod` doesn't seem fit in this class. See `BayesianModelSampling.check_if_consistent` doesn't make any sense as it has nothing to do with sampling method. You can go with two options:
- As this function is generic one, we could save it in some `util` module from where we can call as and when we want to compare two dictionaries. (Although I am not very much convinced about it).
- The other option is make is a nested function in the function where it is used. I am seeing that it is only used in `rejection_sample` method, so you can create a nested function `_check_if_consistent` within that method and  use it. (This is a much more cleaner way).

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,166 @@
...
+            for node in self.topological_order:
+                cpd = self.cpds[node]
+                if cpd.evidence:
+                    evid = []
+                    for var in cpd.evidence:
+                        evid.append(particle[var])
+                    weights = cpd.reduce(evid, inplace=False).values
+                else:
+                    weights = cpd.values
+                particle[node] = sample_discrete(cpd.variables[cpd.variable],
```
abinashpanda: I am not convinced what happens at this step. `cpd.variables` is a list of variables and `cpd.variable` is a string representing the variable of the CPD. How are you able to index a list using a string. Could you please add test cases and make sure that all the test cases pass?

pgmpy/inference/Sampling.py
pratyakshs: Fine, I'll make it a nested function.

pgmpy/inference/Sampling.py
pratyakshs: `cpd.variables` is a dict, in which keys are the name of the variable (say `'diff'`), and the values are the list of corresponding values (say `['diff_0', 'diff_1']`) that can be taken by the variable. 
If there's any other way to get `['diff_0', 'diff_1']` from `cpd`, it would be better. 
I'll add the tests for this and `rejection_sampling`.

pgmpy/tests/test_inference/test_Sampling.py
```diff
@@ -0,0 +1,68 @@
...
+        sample = self.sampling_inference.forward_sample(25)
+        self.assertEquals(len(sample), 25)
+        self.assertEquals(len(sample.columns), 6)
+        self.assertIn('A', sample.columns)
+        self.assertIn('J', sample.columns)
+        self.assertIn('R', sample.columns)
+        self.assertIn('Q', sample.columns)
+        self.assertIn('G', sample.columns)
+        self.assertIn('L', sample.columns)
+        self.assertEquals(set(sample.A), {'A_0', 'A_1', None})
```
pratyakshs: This line should read `self.assertTrue(set(sample.A).issubset({'A_0', 'A_1'}))`. With that change (and a similar one for `test_rejection_sample_basic`, all tests pass on my machine. 
However, the Travis-CI tests fail. From lines 573-74 of the Travis-CI log (Build #863), it seems that the contents of `sample.A` are `State(var='A', state=0)` and `State(var='A', state=1)`, instead of (the expected) `'A_0', 'A_1'`. Why could this be the case?

pgmpy/tests/test_inference/test_Sampling.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_rejection_sample_basic(self):
+        sample = self.sampling_inference.rejection_sample({'A': 'A_0',
+                                                           'J': 'J_1',
+                                                           'R': 'R_0'}, 25)
+        self.assertEquals(len(sample), 25)
+        self.assertEquals(len(sample.columns), 3)
+        self.assertIn('Q', sample.columns)
+        self.assertIn('G', sample.columns)
+        self.assertIn('L', sample.columns)
+        self.assertEquals(set(sample.Q), {'Q_0', 'Q_1', None})
```
pratyakshs: Same as above.

ankurankan:
@pratyakshs: Please see why the tests are failing and also make appropriate changes according to the new representation of states.

pratyakshs:
@ankurankan looks good?

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python3
+
+import networkx as nx
+from pandas import DataFrame
+from pgmpy.inference import Inference
```
ankurankan, line 5: @pratyakshs Add a new line above this.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+    Public Methods
+    --------------
+    forward_sample(size)
+    """
+    def __init__(self, model):
+        if not isinstance(model, BayesianModel):
+            raise TypeError("model must an instance of BayesianModel")
+        super(BayesianModelSampling, self).__init__(model)
```
ankurankan: @pratyakshs Use newer style `super`.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,229 @@
+#!/usr/bin/env python3
+
+import networkx as nx
+from pandas import DataFrame
+from pgmpy.inference import Inference
+from pgmpy.models import BayesianModel
+from pgmpy.utils.mathext import sample_discrete
+
+
+class BayesianModelSampling(Inference):
```
ankurankan, line 10: @pratyakshs Why are you inheriting `Inference` if you are not using it?

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,229 @@
...
+    forward_sample(size)
+    """
+    def __init__(self, model):
+        if not isinstance(model, BayesianModel):
+            raise TypeError("model must an instance of BayesianModel")
+        super(BayesianModelSampling, self).__init__(model)
+        self.topological_order = nx.topological_sort(model)
+        self.cpds = {}
+        for node in model.nodes():
+            self.cpds[node] = model.get_cpds(node)
```
ankurankan: @pratyakshs You can simply do `self.cpds = {node: model.get_cpds(node) for node in model.nodes()}`

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def forward_sample(self, size=1):
```
ankurankan, line 31: @pratyakshs One another way to computes samples would be by looping over the variables rather than looping over the number of samples we want. Taking the example in the doctest and let's say we want 1000 samples, we will start with `diff` and compute all the 1000 samples for it by a single `sample_discrete` call, then can compute all the 1000 samples for `intel` and since now we have the states of all the 1000 samples of `intel` and `diff` we can compute the values for `grade`. 

This method should be much faster and will also give us some space to vectorize computations. For starters for a call on 5 vars for 10000 samples we will need to call `sample_discrete` only 5 times instead of 50000 times.

Also, this method will give us space for memoizing `reduce` calls.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,229 @@
...
+                        evid.append(particle[var])
+                    weights = cpd.reduce(evid, inplace=False).values
+                else:
+                    weights = cpd.values
+                particle[node] = sample_discrete(cpd.variables[cpd.variable],
+                                                 weights)
+            sampled.loc[i] = [particle[node] for node in self.topological_order]
+        return sampled
+
+    def rejection_sample(self, evidence=None, size=1):
```
ankurankan, line 78: I think you should be able to make similar changes as per my comment for `forward_sampling` in this method also.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,229 @@
...
+                    weights = cpd.values
+                particle[node] = sample_discrete(cpd.variables[cpd.variable],
+                                                 weights)
+            if check_if_consistent(evidence, particle):
+                # reject if the sample contradicts the evidence
+                sampled.loc[i] = [particle[node] for node in query_vars]
+                i += 1
+        return sampled
+
+    def likelihood_weighted_sample(self, evidence=None, size=1):
```
ankurankan, line 133: @pratyakshs I think you should be able to make similar changes as per my comment for `forward_sampling` in this method also.

pgmpy/inference/Sampling.py
pratyakshs, line 10: @ankurankan It'll be used later on, should I remove it for now?

pgmpy/inference/Sampling.py
pratyakshs: Right :)

pgmpy/inference/Sampling.py
ankurankan, line 10: @pratyakshs If you think you will need it later then leave it.

pratyakshs:
@ankurankan Have a look at the `forward_sample` method and let me know if that's ok. I'll change the other methods too similarly.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+    Public Methods
+    --------------
+    forward_sample(size)
+    """
+    def __init__(self, model):
+        if not isinstance(model, BayesianModel):
+            raise TypeError("model must an instance of BayesianModel")
+        Inference.__init__(self, model)
```
ankurankan: @pratyakshs Use `super().__init__(model)`

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,235 @@
...
+        self.topological_order = nx.topological_sort(model)
+        self.cpds = {}
```
ankurankan: @pratyakshs No need for this line.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,235 @@
...
+                # memoizing calls to cpd.reduce
+                weights = {}
+                card_iter = [range(card) for card in cpd.evidence_card]
+                for tup in product(*card_iter):
+                    evid = [State(v, s) for v, s in zip(cpd.evidence, tup)]
+                    weights[tup] = cpd.reduce(evid, inplace=False).values
+                # can now fill values, row by row
+                for i in range(size):
+                    tup = tuple(sampled[var][i].state for var in cpd.evidence)
+                    sampled[node][i] = \
```
ankurankan: @pratyakshs This doesn't look good. Try avoiding the back slashes. Also we use 120 characters line in pgmpy. You could do something like this:

``` python
sampled[node][i] = sample_discrete(cpd.variables[cpd.variable],
                                                          weights[tup])[0]
```

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,235 @@
...
+                        sample_discrete(cpd.variables[cpd.variable],
+                                        weights[tup])[0]
+            else:
+                # can generate the column at once
+                weights = cpd.values
+                sampled[node] = \
```
ankurankan: @pratyakshs Same here. Try avoiding the back slash

pgmpy/inference/Sampling.py
pratyakshs: I didn't know that. I was assuming 80 till now. With 120, it fits in one line.

pgmpy/inference/Sampling.py
pratyakshs: @ankurankan apart from these changes, is the method `forward_sample` alright?

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,235 @@
...
+        >>> inference.forward_sample(2)
+                diff       intel       grade
+        0  (diff, 1)  (intel, 0)  (grade, 1)
+        1  (diff, 1)  (intel, 0)  (grade, 2)
+        """
+        sampled = DataFrame(index=range(size), columns=self.topological_order)
+        for node in self.topological_order:
+            cpd = self.cpds[node]
+            if cpd.evidence:
+                # memoizing calls to cpd.reduce
```
ankurankan: @pratyakshs A much better way to memoize is by using decorators. But leave that to me. I will implement it over the `reduce` method as it can be useful in other places also.

pratyakshs: So, I leave this as it is? What about other methods (viz. `rejection_sample` and `likelihood_weighted_sample`), should I memoize calls to `reduce`?

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,235 @@
...
+                weights = {}
+                card_iter = [range(card) for card in cpd.evidence_card]
+                for tup in product(*card_iter):
+                    evid = [State(v, s) for v, s in zip(cpd.evidence, tup)]
+                    weights[tup] = cpd.reduce(evid, inplace=False).values
+                # can now fill values, row by row
+                for i in range(size):
```
ankurankan, line 68: @pratyakshs for loops are costly. So, what you can do here is that you can create a 2-D array of states and a 2-D array for weights as:
For states:
[tuple/array of states of evidence vars for 1st sample,
 tuple/array of states of evidence vars for 2nd sample,
 ... ]

and similarly for weights:
[weight arr for 1st sample,
 weight arr for 2nd sample,
 ...]

and modify `sample_discrete` accordingly so that it computes all the values in a single vector operation.

pgmpy/inference/Sampling.py
ankurankan: @pratyakshs Yes leave it for now.. I will implement it in a way that all reduce calls will be memoized.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,196 @@
...
+        >>> student.add_cpds(cpd_d, cpd_i, cpd_g)
+        >>> inference = BayesianModelSampling(student)
+        >>> evidence = {'diff': State(var='diff',state=0)}
+        >>> inference.likelihood_weighted_sample(evidence, 2)
+                intel       diff       grade  _weight
+        0  (intel, 0)  (diff, 0)  (grade, 1)      0.6
+        1  (intel, 1)  (diff, 0)  (grade, 1)      0.6
+        """
+        sampled = DataFrame(index=range(size), columns=self.topological_order)
+        sampled['_weight'] = [1] * size
```
abinashpanda, line 174: Don't use list over here. Instead you could create a `numpy.array` of ones using `np.ones((1, size))`. It is much more memory efficient.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,196 @@
...
+                if node in evidence:
+                    sampled[node] = [evidence[node]] * size
+                    for i in range(size):
+                        sampled.loc[i, '_weight'] *= weights[i][evidence[node].state]
+                else:
+                    sampled[node] = sample_discrete(cpd.variables[cpd.variable], weights)
+            else:
+                weights = cpd.values
+                if node in evidence:
+                    sampled[node] = [evidence[node]] * size
```
abinashpanda, line 191: Again please don't repeat elements of a list by multiplying with an integer. It is not memory efficient. Rule of thumb **try to avoid lists as much as possible** because when we would be dealing with huge data it won't be efficient at all. Instead you could use `np.tile`.

pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,196 @@
...
+        sampled = DataFrame(columns=self.topological_order)
+        prob = 1
+        while len(sampled) < size:
+            _size = int((size - len(sampled)) / prob)
+            _sampled = self.forward_sample(_size)
+            for i in range(_size):
+                for var in evidence:
+                    if evidence[var] != _sampled[var][i]:
+                        _sampled.drop(i, inplace=True)
+                        break
```
abinashpanda, line 127: Using `break` statement is not a very clean solution. So instead of using a `for` loop with an conditional `if` and `break` could you please change it to `while` loop. It would be much cleaner and easier to read as well.

@ankurankan Your suggestions?

abinashpanda:
@pratyakshs Please make the changes. Rest I think is really good and ready to merge. @ankurankan could you also have a look and suggest if anything needs to be changed.

ankurankan:
@pratyakshs Don't spend time on this.. I have started working on this PR and will make all the necessary changes. I will let you know when I am done. Just please make sure that you don't make similar mistakes in future. @abinashpanda

ankurankan:
Updated PR #424

PR #404, opened 2015-06-02
small changes for named tuples 

palashahuja:


pgmpy/factors/Factor.py
```diff
@@ -194,7 +194,7 @@ def marginalize(self, variables, inplace=True):
             factor = self
         else:
             factor = Factor(self.scope(), self.cardinality, self.values)
-        marginalize_index = np.array(np.where(np.in1d(factor.scope(), variables)))
+        marginalize_index = np.array(np.where(np.array([i in variables for i in factor.scope()])))
```
jaidevd: `marginalize_index` is only being used for indexing, right? So in the interest of better performance, simply doing

``` python
marginalize_index = np.array([i in variables for i in factor.scope()])
```

will create a boolean array which can itself be used for indexing. No need for the extra `np.where` and `np.array` calls.

pgmpy/factors/Factor.py
```diff
@@ -194,7 +194,7 @@ def marginalize(self, variables, inplace=True):
...
         assign = np.array(factor.cardinality)
```
jaidevd: Ignore the comment above if 
```
assign.shape[0] != marginalize_index.shape[0]
```

pgmpy/factors/Factor.py
```diff
@@ -194,7 +194,7 @@ def marginalize(self, variables, inplace=True):
...
         assign[marginalize_index] = -1
```
jaidevd, line 199: Ignore the comment above if 

``` python
assign.shape[0] != marginalize_index.shape[0]
```

and feel free to delete both comments.

palashahuja:
@jaidevd The shapes of `assign` and `marginalize_index` must be same.

jaidevd:
Let's see if something breaks.

jaidevd:
LGTM.

ankurankan:
@palashahuja Are there more changes to the state thing (that we had discussed) or this is it?

palashahuja:
@ankurankan There are some more changes. Please don't merge this branch right now. I will keep updating the pull request.

ankurankan:
@palashahuja Then please don't work on the state representation thing because I have already started fixing it.

palashahuja:
@ankurankan It's almost done.... I have to still look for the errors in bayesian model(file)...

abinashpanda:
@palashahuja Could you please have a look at the test case that is failing. Apart from that we can merge it.

PR #410, opened 2015-06-04
Dynamic Bayesian Network

palashahuja:


pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
+
+from pgmpy.models import BayesianModel
+from pgmpy.factors import TabularCPD, TreeCPD, RuleCPD
+from collections import OrderedDict, namedtuple
+import numpy as np
```
ankurankan, line 3: @palashahuja pep8 order of imports

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+
+class DynamicBayesianNetwork(BayesianModel):
+    def __init__(self, ebunch=None):
+        """
+        Base class for Dynamic Bayesian Network
+
+        This model is a time variant of the static Bayesian model, where each timeslice has
+        some static nodes and is then replicated over a certain time slice.
+
+        The nodes that will be added to the network will have the string form.
```
ankurankan: @palashahuja Don't need the node names to be strings.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+        most of the methods will be imported from Bayesian Model
+        However there will be a few changes
+        >>> dbn.list_of_nodes
+        ['D','G','I','S','L']
+        >>> dbn.edges()
+        [(('D',0),('G',0)),(('I',0),('G',0)),(('G',0),('L',0))]
+        If some edges connect nodes not yet in the model, the nodes
+        are added automatically.  There are no errors when adding
+        nodes or edges that already exist.
+        Methods:
```
ankurankan, line 57: Add a blank line above this

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+            if not isinstance(cpd, (TabularCPD, TreeCPD, RuleCPD)):
+                raise ValueError('Only TabularCPD, TreeCPD or RuleCPD can be'
+                                 ' added.')
+
+            if (cpd.variable, timeslice) not in self.dbn_cpds:
+                self.dbn_cpds[(cpd.variable, timeslice)] = cpd
+
+    def get_dbn_cpds(self, node=None, timeslice=None):
+        """
+
```
ankurankan, line 169: Remove blank line

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+        Returns the cpds that have been added till now to the graph
+
+
```
ankurankan, line 173: Remove blank line

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+        Parameter
+        ---------
+        node: any hashable python object (optional)
+            The node whose CPD we want. If node not specified returns all the
+            CPDs added to the model.
+        timeslice: integer value. should be ranging from 0
+                   to the number of the slices present in
+                   the bayesian network.
+        EXAMPLE
```
ankurankan, line 174: Add blank line

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+        """
+        adding nodes in the static Bayesian Network.
+        Examples
+        --------
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        >>> dbn.add_nodes_from(['D','G','I','S','L'])
+        """
+        for node in nodes:
+            if not isinstance(node, str):
```
ankurankan: @palashahuja Can be any hashable object

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+    def add_edge(self, start, end, **kwargs):
+        """
+        Add an edge between two nodes.
+
+        The nodes will be automatically added if they are
+        not already in the initial static Bayesian Network.
+
+        Parameters
+        ----------
+        start: the starting node of the edge
```
ankurankan: @palashahuja Have a look at other methods to see the proper format of parameters

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+        the index of the time_slice that the node belongs to.
+        EXAMPLE
+        -------
+
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        >>> dbn.add_nodes_from(['d', 'i'])
+        >>> dbn.add_edge((('d',0), ('i',0)))
+        """
+        edge = (start,end)
```
ankurankan: pep8

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+        if not any(map(lambda x: isinstance(x, tuple), iter(edge))):
```
ankurankan: @palashahuja Can also be array.

palashahuja:
@ankurankan I will reformat the file for pep8 errors and blank lines all together.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+            raise AttributeError('The edge is not of a standard form')
```
ankurankan: Rather than a general statement throw some specific error message which will help in debugging.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,480 @@
...
+
+        if not any(map(lambda x: x[0] in self.list_of_nodes, iter(edge))):
+            new_nodes = list(filter(lambda x: x[0] not in self.list_of_nodes, iter(edge)))
+            self.add_nodes_from(set(map(lambda x: x[0], iter(new_nodes))))
+
+        if edge[0][1] == edge[1][1]:
+            if edge[0][1] in (0, 1):
+                new_edge = map(lambda x: (x[0], 1 - edge[0][1]), edge)
+                super(DynamicBayesianNetwork, self).add_edge(*edge_type(*new_edge))
```
ankurankan: @palashahuja Use new style `super`

palashahuja: @ankurankan  Is the new style super --> `super().add_edge(...)`

ankurankan:
@palashahuja Please also check landscape bot's comment and fix those.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        self.inter_edges = []
+        self.intra_edges = []
+        self.dbn_cpds = OrderedDict()
+        self.list_of_nodes = []
+        if ebunch:
+            self.add_nodes_from(ebunch)
+        self.timestate = 1
+        self.computed = False
+
+    def add_nodes_from(self, nodes):
```
abinashpanda: There is a major issue with `add_nodes_from` function. We have not defined a `add_node` method in this class, so when someone tries to use call `add_node()` method it would be using the `add_node` method from `BayesianModel` class. For example 

``` python
G = DynamicBayesianModel()
G.add_node('A')
```

it won't be adding `A` in `list_of_nodes`, as there is no such variable in `BayesianModel`. It would lead to inconsistency. So my suggestion is don't define `add_node` and `add_nodes_from` until you want to have some extra constraints on the nodes. And whenever you want to access the nodes instead of using `list_of_nodes` just call `self.nodes()`.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        >>> dbn.add_nodes_from(['d', 'i'])
+        >>> dbn.add_edge((('d',0), ('i',0)))
+        """
+        edge = (start, end)
+        if not any(map(lambda x: isinstance(x, tuple), iter(edge))):
+            raise AttributeError('The edge is not of a standard form')
+
+        if not any(map(lambda x: isinstance(x[1], int), iter(edge))):
```
abinashpanda: Just like internal state representation, instead of using `tuple` you can use `namedtuple`. 

``` python
Edge = namedtuple('Edge', ['var', 'timeslice'])
```

So instead of using `x[1]` over here you can use `x.timeslice` or something like this. 

@ankurankan What are your suggestions?

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+            raise AttributeError('The edge does not contain the timeslice in integer form')
+
+        if not any(map(lambda x: x[0] in self.list_of_nodes, iter(edge))):
+            new_nodes = list(filter(lambda x: x[0] not in self.list_of_nodes, iter(edge)))
+            self.add_nodes_from(set(map(lambda x: x[0], iter(new_nodes))))
+
+        if edge[0][1] == edge[1][1]:
```
abinashpanda: Similarly here also we are using indexes to represent timeslice. Using `namedtuple` instead of `tuple` would make it more readable and convenient as well.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+                super().add_edge(*edge_type(*new_edge))
+                self.intra_edges.extend([edge_type(*edge), edge_type(*map(lambda x: (x[0], 1 - edge[0][1]), edge))])
+            else:
+                self.intra_edges.extend([edge_type(*edge)])
+        else:
+            self.inter_edges.append(edge_type(*edge))
+
+        super().add_edge(*edge_type(*edge))
+
+    def add_edges_from(self, ebunch, **kwargs):
```
abinashpanda: You don't need to define `add_edges_from` method again. It would directly inherit this method from the parent class, where it does the same thing. So in the child class only define the `add_edge` method.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        -------
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        >>> dbn.add_nodes_from(['D','G','I','S','L'])
+        >>> dbn.add_edges_from([(('D',0),('G',0)),(('I',0),('G',0)),(('G',0),('L',0))])
+        """
+        for edge in ebunch:
+            self.add_edge(edge[0], edge[1])
+
+    def intra_slice(self, timeslice):
```
abinashpanda: As this function returns `intra_slice`, so rename this function as `get_intra_slice`. Using `get` and `set` in method names makes it more clearer.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        """
+        if not isinstance(timeslice, int):
+            raise ValueError('the timeslice is not an integer value')
+        if timeslice < 0:
+            raise ValueError('negative values for a timeslice are not permissible')
+        if timeslice > self.timestate:
+            raise ValueError('the bayesian network does not contain the intra slice edges of this timeslice')
+        return list(filter(lambda x: x.start[1] == x.end[1] == timeslice, self.intra_edges))
+
+    def inter_slice(self, timeslice):
```
abinashpanda: Same for this as well. `get_inter_slice`.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        add_cpds
+        get_cpds
+        compute_initial_state
+        unroll
+        remove_timeslice
+        """
+        super().__init__()
+        self.inter_edges = []
+        self.intra_edges = []
+        self.dbn_cpds = OrderedDict()
```
abinashpanda: Please try to maintain consistency as much as possible. All the directed graphical models are assumed to have `cpds` variable to store the CPDs and `factors` to store the factors.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        +------+------+------+---------+------+------+-------+
+        |gradeC| 0.8  | 0.8  |   0.8   |  0.8 |  0.8 |   0.8 |
+        +------+------+------+---------+------+------+-------+
+        """
+        for cpd in cpds:
+            if not isinstance(cpd, (TabularCPD, TreeCPD, RuleCPD)):
+                raise ValueError('Only TabularCPD, TreeCPD or RuleCPD can be'
+                                 ' added.')
+
+            if (cpd.variable, timeslice) not in self.dbn_cpds:
```
abinashpanda: Don't try to store the CPDs in this fashion. If you will look at the implementation of all the models, you will find that whether it is CPDs or factors, we don't store them in form of a dictionary with key being the variable name. For example in `BayesianModel`, `self.cpds` is a `list` and similarly in `MarkovModel`, `self.factors` is a `list` as well. Don't assign the CPDs to the variable in this stage. It should be only assigned to a particular variable only when you need so, (for example while doing inference). So I would suggest that you make `self.cpds` as `{}` with key only being the time-slice and value being list of all CPDs associated with the model in the particular time-slice.

@ankurankan Your suggestions?

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        """
+        if not isinstance(timeslice, int):
+            raise ValueError('the timeslice is not an integer value')
+        if timeslice < 0:
+            raise ValueError('negative values for a timeslice are not permissible')
+        if timeslice > self.timestate:
+            raise ValueError('the bayesian network does not contain the inter slice edges of this timeslice')
+        return list(filter(lambda x: x.end[1] == timeslice, self.inter_edges))
+
+    def add_dbn_cpds(self, timeslice, *cpds):
```
abinashpanda: Please ensure consistency in naming methods, it should be `add_cpds` for directed graphical models and `add_factors` for undirected graphical models.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        """
+        for cpd in cpds:
+            if not isinstance(cpd, (TabularCPD, TreeCPD, RuleCPD)):
+                raise ValueError('Only TabularCPD, TreeCPD or RuleCPD can be'
+                                 ' added.')
+
+            if (cpd.variable, timeslice) not in self.dbn_cpds:
+                self.dbn_cpds[(cpd.variable, timeslice)] = cpd
+
+    def get_dbn_cpds(self, node=None, timeslice=None):
```
abinashpanda: Rename it to `get_cpds`.

abinashpanda: Also please have a look at the `get_cpds` method in `BayesianModel` to have an idea what it should return and how the API should be structured.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+                    return self.dbn_cpds[(node, timeslice)]
+            else:
+                list_of_cpds = []
+                for time_state in range(self.timestate + 1):
+                    list_of_cpds.append(self.dbn_cpds[(node, time_state)])
+                return list_of_cpds
+        else:
+            return self.dbn_cpds
+
+    def compute_initial_state(self):
```
abinashpanda: The method name is also a misnomer. `compute_initial_state` gives us a sense that it returns something, whereas it  doesn't. Rename it to something like `initialize_initial_state` or something like this.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,500 @@
...
+        >>> i_cpd = TabularCPD('I', 2, [[0.7, 0.3]])
+        >>> student.add_dbn_cpds(0, grade_cpd, d_cpd, i_cpd)
+        >>> student.add_dbn_cpds(1, diff_cpd, intel_cpd)
+        >>> student.compute_initial_state()
+        """
+
+        if not self.computed:
+            for node in self.dbn_cpds:
+                list_of_parents = super().get_parents(node)
+                if all(x[1] == list_of_parents[0][1] for x in list_of_parents) and list_of_parents != [] and (
```
abinashpanda: Could you please add a line of comment above this check stating what are these checks. This makes the code more readable and easy to understand for a newcomer.

abinashpanda:
@palashahuja Please add test cases.

pgmpy/models/DynamicBayesianNetwork.py
palashahuja: @abinashpanda I wanted to store it as `self.cpds` but then it causes conflict with the `self.cpds`  defined inside the `BayesianModel`. Either I could continue to use this, or remove the `super()`entirely.

pgmpy/models/DynamicBayesianNetwork.py
palashahuja: @abinashpanda same here, conflict with the super() definition.

pgmpy/models/DynamicBayesianNetwork.py
palashahuja: @abinashpanda Then could you please explain how the user can enter the static nodes defined in the initial Bayesian Network ? Also, if you call `self.nodes()` , it gives the nodes along with the  time slice too.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
+
+import numpy as np
+from pgmpy.models import BayesianModel
+from pgmpy.factors import TabularCPD, TreeCPD, RuleCPD
+from collections import OrderedDict, namedtuple
```
ankurankan: @palashahuja Pep8 import order: https://www.python.org/dev/peps/pep-0008/#imports
So this should be like this:

``` python
from collections import OrderedDict, namedtuple

import numpy as np

from pgmpy.factors import TabularCPD, TreeCPD, RuleCPD
```

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        data: Data to initialize graph.  If data=None (default) an empty
+              graph is created.  The data can be an edge list, or any
+              NetworkX graph object
+
+        Examples:
+        --------
+        Create an empty Dynamic Bayesian Network with no nodes and no edges
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        adding nodes and edges inside the dynamic bayesian network. A single
```
ankurankan, line 40: @palashahuja Make the docs more clear.. Add a space above this line and use capital a in adding.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        node can be added using the method below.
+        >>> dbn.add_nodes_from(['D','G','I','S','L'])
+        >>> dbn.add_edges_from([(('D',0),('G',0)),(('I',0),('G',0)),(('G',0),('L',0))])
+
+        most of the methods will be imported from Bayesian Model
```
ankurankan: @palashahuja Capital m.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        However there will be a few changes
+        >>> dbn.list_of_nodes
+        ['D','G','I','S','L']
+        >>> dbn.edges()
+        [(('D',0),('G',0)),(('I',0),('G',0)),(('G',0),('L',0))]
+        If some edges connect nodes not yet in the model, the nodes
```
ankurankan, line 53: Add new line above this.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+                    while the time_slice is an integer value, which
+                    denotes the index of the time_slice that the node
+                    belongs to.
+
+        EXAMPLE
+        -------
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        >>> dbn.add_nodes_from(['d', 'i'])
+        >>> dbn.add_edge((('d',0), ('i',0)))
```
ankurankan: @palashahuja Doesn't look like this would work.. Your function definition is `add_edge(self, start, end, **kwargs)` but you are passing a single argument here.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        """
+        edge = (start, end)
+        if not any(map(lambda x: isinstance(x, tuple), iter(edge))):
```
ankurankan: @palashahuja `edge` has only 2 elements. So, you can simply do `if not (isinstance(start, tuple) and isinstance(end, tuple))`. It would be much easier to read.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+            raise AttributeError('The edge is not of a standard form')
+
+        if not any(map(lambda x: isinstance(x[1], int), iter(edge))):
+            raise AttributeError('The edge does not contain the timeslice in integer form')
+
+        if not any(map(lambda x: x[0] in self.list_of_nodes, iter(edge))):
+            new_nodes = list(filter(lambda x: x[0] not in self.list_of_nodes, iter(edge)))
+            self.add_nodes_from(set(map(lambda x: x[0], iter(new_nodes))))
```
ankurankan: @palashahuja No need to manually add nodes which are not present. Networkx does this and since you are calling `super` it will be automatically handled by networkx.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        EXAMPLE
+        -------
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        >>> dbn.add_nodes_from(['d', 'i'])
+        >>> dbn.add_edge((('d',0), ('i',0)))
+        """
+        edge = (start, end)
+        if not any(map(lambda x: isinstance(x, tuple), iter(edge))):
+            raise AttributeError('The edge is not of a standard form')
```
ankurankan: @palashahuja Rather than just showing `not of a standard form`, print something more detailed.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+                    belongs to.
+
+        EXAMPLE
+        -------
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> dbn = DBN()
+        >>> dbn.add_nodes_from(['d', 'i'])
+        >>> dbn.add_edge((('d',0), ('i',0)))
+        """
+        edge = (start, end)
```
ankurankan: @palashahuja Why are you creating a tuple for representing edge? I don't think you are using this tuple anywhere.

ankurankan:
@palashahuja Can you please go through the implementation doc that I had created once again? This PR doesn't look anything like that. You are creating too many extra variables that we won't be needing. Also, it's making things complex unnecessarily. @abinashpanda Can you also please go through the doc [1] once and let us know if you have any other suggestions?

[1]https://docs.google.com/document/d/1JshSdWVEWjYb_T01bULyjzt3ZEVLugcecyWvBIQAMz4/edit

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+                super().add_edge(*edge_type(*new_edge))
+                self.intra_edges.extend([edge_type(*edge), edge_type(*map(lambda x: (x[0], 1 - edge[0][1]), edge))])
+            else:
+                self.intra_edges.extend([edge_type(*edge)])
+        else:
+            self.inter_edges.append(edge_type(*edge))
+
+        super().add_edge(*edge_type(*edge))
+
+    def get_intra_slice(self, timeslice):
```
ankurankan: Not at all clear from the name that it will return edge list.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
+
+import numpy as np
+from pgmpy.models import BayesianModel
+from pgmpy.factors import TabularCPD, TreeCPD, RuleCPD
+from collections import OrderedDict, namedtuple
+
+
+edge_type = namedtuple('edge_type', ('start', 'end'))
```
ankurankan: Why have you used the name `edge_type`?

palashahuja:
@ankurankan Could you please let me know which part you didn't understand? Maybe I will try to reduce some variables

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        get_cpds
+        compute_initial_state
+        unroll
+        remove_timeslice
+        """
+        super().__init__()
+        self.inter_edges = []
+        self.intra_edges = []
+        self.dbn_cpds = OrderedDict()
+        self.list_of_nodes = []
```
ankurankan: No need for this. Use networkx's node dict.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        inter_slice
+        add_cpds
+        get_cpds
+        compute_initial_state
+        unroll
+        remove_timeslice
+        """
+        super().__init__()
+        self.inter_edges = []
+        self.intra_edges = []
```
ankurankan: Use networkx's edge list.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+        """
+        if not isinstance(timeslice, int):
+            raise ValueError('the timeslice is not an integer value')
+        if timeslice < 0:
+            raise ValueError('negative values for a timeslice are not permissible')
+        if timeslice > self.timestate:
+            raise ValueError('the bayesian network does not contain the intra slice edges of this timeslice')
+        return list(filter(lambda x: x.start[1] == x.end[1] == timeslice, self.intra_edges))
+
+    def get_inter_slice(self, timeslice):
```
ankurankan: @palashahuja Why do you need `timeslice`? The edges would be same in all the timeslices except for the first.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,481 @@
...
+import numpy as np
+from pgmpy.models import BayesianModel
+from pgmpy.factors import TabularCPD, TreeCPD, RuleCPD
+from collections import OrderedDict, namedtuple
+
+
+edge_type = namedtuple('edge_type', ('start', 'end'))
+
+
+class DynamicBayesianNetwork(BayesianModel):
```
ankurankan, line 9: @palashahuja  This class will deal only with the representation of DBNs. So, don't add anything that we don't need in representing the network. Also we want to keep things modular. So, we don't need timeslice in most of the places where you have used. For representing a complete DBN we will only need the 0th and 1st timeslice and the cpds associated with it.

pgmpy/models/DynamicBayesianNetwork.py
ankurankan: @palashahuja No need to use timeslice at all.

palashahuja:
@ankurankan The edges are different in every timeslice ... I am using that later ...

ankurankan:
@palashahuja Can you please explain to me how the edges are going to be different in each time slice? And If you are going for such a representation with different edges, different cpds etc, how much memory do you think you will be consuming when we have 10000 or more time slices?

palashahuja:
@ankurankan ok, So how should be the representation be like?

ankurankan:
@palashahuja I wrote a doc for that https://docs.google.com/document/d/1JshSdWVEWjYb_T01bULyjzt3ZEVLugcecyWvBIQAMz4/edit. And I don't know why do you need to have different edges in different time slices. If you are implementing from some specific paper please share it with me.

palashahuja:
@ankurankan So basically I should add the edges for 0th and 1st time slice and the concerned cpds. After that, I should check that bayesian model. That's it?

ankurankan:
@palashahuja Yes.. This class is just for the representation of the DBN. And for the complete representation you won't need anything else than the 0th and 1st slice and their cpds.

palashahuja:
@ankurankan And suppose the user wants to inquire about the cpd of a node in a different time slice, Should the tabular cpd be changed accordingly?. similar method for obtaining edges?

ankurankan:
@palashahuja That's inference part and there would be different class for that for computing all the changes. You can't store everything.

palashahuja:
@ankurankan Should I even use the `edge_type` namedtuple? I don't think it is necessary....

ankurankan:
@palashahuja Simply use `namedtuple('Edge', ['var', 'time'])`.

palashahuja:
@ankurankan That will be for the node right?

ankurankan:
@palashahuja Sorry my bad.. You won't be needing the namedtuple. Please go through the doc thoroughly, I have mentioned all of this there.

pgmpy/models/DynamicBayesianNetwork.py
palashahuja: @ankurankan But I am adding the nodes without the time slice in the `list_of_nodes`.super() won't take care for that

ankurankan: @palashahuja You don't need the time slices now. So, no need for this.

palashahuja: @ankurankan The `list_of_nodes` will contain `['D','G','I','L'],` while self.nodes() contain the nodes along with the time slice. This leads to a lot of confusion. I was thinking to remove the list_of_nodes entirely. What do you think?

ankurankan: @palashahuja Yes remove the `list_of_nodes` variable. Store the node names in `self.nodes` and add a method that overrides the netowrkx's `nodes()` method and it will return just the name of the variables `['D', 'G', 'I', 'L']`.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,233 @@
...
+        if not (start[1] in (0, 1) and end[1] in (0, 1)):
+            raise ValueError('the timeslices inside the node must belong to 0 or 1')
+
+        if start[1] == end[1]:
+            super().add_edge((start[0], 1 - start[1]), (end[0], 1 - end[1]))
+
+        super().add_edge(start, end)
+        self.correct_nodes()
+
+    def correct_nodes(self):
```
abinashpanda, line 108: If this method is only used in `add_edge()` method, then I would suggest to not expose to the user. Rename it to `_correct_nodes()` method.

pgmpy/models/BayesianModel.py
```diff
@@ -84,6 +84,7 @@ def __init__(self, ebunch=None):
             self.add_edges_from(ebunch)
```
ankurankan, line 84: @palashahuja Don't change things in BayesianModel. Better simply inherit from DirectedGraph.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,233 @@
...
+        add_edge
+        add_edges_from
+        intra_slice
+        inter_slice
+        add_cpds
+        initialize_initial_state
+        """
+        super().__init__()
+        if ebunch:
+            self.add_nodes_from(ebunch)
```
ankurankan, line 69: @palashahuja How are you adding the edges in the ebunch??

ankurankan:
@palashahuja Can you please write tests also?

pgmpy/models/DynamicBayesianNetwork.py
palashahuja, line 69: @ankurankan I thought `ebunch` was for the nodes. Sorry about that

palashahuja:
@ankurankan  sure, I will do that too.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -0,0 +1,233 @@
...
+        >>> dbn.add_nodes_from(['D','G','I','S','L'])
+        >>> dbn.add_edges_from([(('D',0),('G',0)),(('I',0),('G',0)),(('D',0),('D',1)),(('I',0),('I',1)))])
+        >>> dbn.get_inter_edges()
+        [(('D', 0), ('D', 1)), (('I', 0), ('I', 1))]
+        """
+        return [edge for edge in self.edges() if edge[0][1] != edge[1][1]]
+
+    def add_cpds(self, *cpds):
+        """
+        This method adds the cpds inside the dynamic bayesian network.
```
ankurankan, line 165: Better would be `Add CPDs to the Dynamic Bayesian Network`

pgmpy/models/DynamicBayesianNetwork.py
ankurankan, line 69: @palashahuja No, it's a list of edges. See the examples in BayesianNetwork or networkx

ankurankan:
@palashahuja Make the changes according to the comments and let me know. Should be good to merge then.

palashahuja:
@ankurankan for the `add_cpd` method, should I alter the `add_cpd` method inside the dbn if the changes aren't to be made in the bayesian network?

ankurankan:
@palashahuja Yes you can.. Just don't change the methods of BayesianNetwork to make them work for DynamicBayesianNetwork.

palashahuja:
@ankurankan Should I replace `ebunch` with something appropriate? Because in the docs, you had mentioned in the api where the user can add the nodes while initialization ...

ankurankan:
@palashahuja ebuch should be a list of edges. We can define the whole network by a list of edges because nodes are automatically added if not present.

PR #411, opened 2015-06-06
Uai reader

kislayabhi:
@ankurankan  Just a minor change worked.

### DIFF
--- a/pgmpy/readwrite/UAI.py
+++ b/pgmpy/readwrite/UAI.py
@@ -39,7 +39,7 @@ class UAIReader:
 
         for i in range(self.no_of_functions):
             function_data = self.functions_data[i]
-            vars_index = list(map(int, function_data.strip().split(' ')))[1:]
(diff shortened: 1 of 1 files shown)


abinashpanda:
Hey @kislayabhi and @ankurankan I have a major question in this commit.

The reader class should not create a model. Its primary job should be to read only from the corresponding file and model creation should be a method in the corresponding model. For example, `UAIReader` class should be having the methods
- `get_factor_scope()`
- `get_factors_table()`
- `get_preamble()`, etc. 

while there should be a method in `BayesianModel` such as `read_from_file(file_object)` that should read from the `file_object` it may be `UAI` file or anything to create the model.

ankurankan:
@abinashpanda @kislayabhi This code was never meant to be included in our repo. This was just a dirty implementation so that @kislayabhi would be able to read from files until we have a proper implementation.

abinashpanda:
If it is so. @kislayabhi you don't require to a send a PR. I am closing it. You can use locally wherever you require.

PR #422, opened 2015-06-16
Markov Chain Monte Carlo

pratyakshs:
~~I'm still working on this, isn't ready to be merged~~


pgmpy/inference/Sampling.py
```diff
@@ -0,0 +1,274 @@
...
+                if node in evidence:
+                    sampled[node] = [evidence[node]] * size
+                    for i in range(size):
+                        sampled.loc[i, '_weight'] *= weights[evidence[node].state]
+                else:
+                    sampled[node] = sample_discrete(cpd.variables[cpd.variable], weights, size)
+        return sampled
+
+
+class MarkovChainMonteCarlo(nx.DiGraph):
```
abinashpanda: Hey don't put any model in `inference` module. Move this class to `models` module.

abinashpanda: Plus don't inherit directly from `nx.DiGraph`. Have a look at `pgmpy.base.DirectedGraph`.

pratyakshs: @abinashpanda  `pgmpy.base.DirectedGraph` inherits directly from `nx.DiGraph`, isn't it? Should I inherit from `DirectedGraph` instead?

ankurankan: @pratyakshs Yes.. Why?

pratyakshs: @abinashpanda One issue that arises when I inherit from `DirectedGraph` is: 
on calling `self.add_weighted_edges_from`, it uses the definition in `nx.Graph.add_weighted_edges_from`, which is simply `self.add_edges_from(((u,v,{weight:d}) for u,v,d in ebunch),**attr)`. But `add_edges_from` has been overridden by the definition `DirectedGraph.add_edges_from`, which in turn calls `DirectedGraph.add_edge` - but this does not support passing an argument for the weight, and thus results in an error. I could modify the definition of `DirectedGraph.add_edge` a bit, would you suggest that?

pratyakshs:
@abinashpanda @ankurankan this is ready for a review.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,226 @@
+#!/usr/bin/env python3
+import numpy as np
+import networkx as nx
+
+from pgmpy.utils import sample_discrete
+
+
+class MarkovChainMonteCarlo(nx.DiGraph):
```
ankurankan, line 8: @pratyakshs Since this class represents a Markov chain. The name should be Markov Chain.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,226 @@
...
+        ---------
+        >>> from pgmpy.models import MarkovChainMonteCarlo as MCMC
+        >>> model = MCMC([(0, 1, 0.25), (0, -1, 0.25), (0, 0, 0.5), (1, 1, 1), (-1, -1, 1)])
+        >>> model.set_start_state(0)
+        >>> model.sample(5)
+        array([ 0.,  0.,  0.,  0.,  1.])
+        """
+        sampled = np.empty(size)
+        sampled[0] = sample_discrete(list(self._weights.keys()), list(self._weights.values()))[0]
+        for i in range(size - 1):
```
ankurankan, line 222: @pratyakshs Please try to think of a better way to do this rather than looping.

ankurankan:
@pratyakshs I think this implementation is just for the case when you have a single kernel. What about the case when we would be having multiple kernels ? Since we will need it for sampling from models.

ankurankan:
@pratyakshs This is not relevant anymore. Should I close this ?

PR #424, opened 2015-06-17
Sampling

ankurankan:


ankurankan:
@pratyakshs I have made changes in `forward_sample`, `rejection_sample` and `sample_discrete`. Please make changes in `likelihood_weighted_sample` accordingly. Also please pay attention to changes in `forward_sample` and try to similarly make likelihood sampling more efficient. Also please fix the tests.

@abinashpanda Please have a look at the changes and see if it can be improved (other than caching). Still taking too much time. forward_sampling over 10000 samples is taking around 2 sec and rejection is taking around 3 sec.

ankurankan:
@pratyakshs Also I am closing your earlier PR as all your commits are included here.

pratyakshs:
@ankurankan `np.random.choice` expects its first argument to be 1-dimensional. In `forward_sample` etc, we call it with a list of `State` namedtuples (for example `[State('a', 1), State('b', 2)]`). When `np.random.choice` treats this as `np.array`, it converts it to `array([['a', '1'], ['b', '2'], dtype='<U1')`, which is 2-dimensional. The problem in general is in the treatment of `np.array` of namedtuples as a 2-dimensional array. Any workaround for this? @abinashpanda

ankurankan:
@pratyakshs See how I have implemented it for `forward_sample`. In short rather than passing the State tuple, just pass the state names which will be a 1D array.

pratyakshs:
Ok, this looks fine

ankurankan:
Updated in #425

PR #435, opened 2015-06-23
dbn inference

palashahuja:


palashahuja:
Currently, the inference does not support the case when the factor having one variable is reduced. I am still currently working on the `map` queries.

ankurankan:
@palashahuja 
1. Fix PEP 8.
2. Use proper variable names (not temp)
3. Add blank lines separating code blocks doing different things.
4. Comments should not telling what you are doing rather should describe what the code does. So, don't start with "I am ... ".

Can you please fix these so that it would be easier for me to understand the code?

palashahuja:
@ankurankan , @abinashpanda Please review this, I need your approval so as to proceed with `map` queries

pgmpy/inference/base.py
```diff
@@ -80,3 +81,25 @@ def __init__(self, model):
...
+            # first timeslice.
+            # 2) one_and_half_model:- This is the bayesian model which consists of the
+            # interface nodes along with the first time slice nodes.
+            # Both of the above parameters will be required by the junction tree algorithm
+            # too.
+            start_model = BayesianModel(model.edges())
+            start_model.add_cpds(*model.cpds)
+            self.start_bayesian_model = start_model
+            self.interface_nodes = [(edge[0][0], 0) for edge in model.get_inter_edges()]
+            interface_nodes_1 = [(edge[0][0], 1) for edge in model.get_inter_edges()]
```
ankurankan: @palashahuja Use descriptive variable names.

pgmpy/inference/base.py
```diff
@@ -80,3 +81,25 @@ def __init__(self, model):
...
+            intra_edges_1 = [edge for edge in model.get_intra_edges() if edge[1][1] == 1]
```
ankurankan: Same

pgmpy/inference/base.py
```diff
@@ -80,3 +81,25 @@ def __init__(self, model):
...
+            self.one_and_half_model = BayesianModel(model.get_inter_edges() + intra_edges_1)
+            nodes_1 = [(node, 1) for node in model.nodes()]
```
ankurankan: Same

pgmpy/inference/base.py
```diff
@@ -80,3 +81,25 @@ def __init__(self, model):
...
+            cpd_1 = [cpd for cpd in model.cpds if set(list(cpd.variables)).issubset(nodes_1)]
```
ankurankan: Same

pgmpy/inference/dbn_inference.py
```diff
@@ -0,0 +1,98 @@
+from collections import defaultdict
+from pgmpy.inference import VariableElimination
+from pgmpy.factors import TabularCPD
+from pgmpy.inference import Inference
+
+class DBNInference(Inference):
+
+    def query(self, variables, evidence= None, elimination_order= None):
```
ankurankan: PEP8

ankurankan:
@palashahuja There are a few other PEP8 issues also.. Please fix those.

pgmpy/inference/base.py
```diff
@@ -80,3 +81,25 @@ def __init__(self, model):
...
+            # This method constructs a temporary Bayesian Model so as to create the
+            # initial potential for inference.
+            # This will also declare the following two parameters
+            # 1) interface nodes:- These are the nodes whose children are in the 
+            # first timeslice.
+            # 2) one_and_half_model:- This is the bayesian model which consists of the
+            # interface nodes along with the first time slice nodes.
+            # Both of the above parameters will be required by the junction tree algorithm
+            # too.
+            start_model = BayesianModel(model.edges())
```
ankurankan: @palashahuja Also please add empty lines between code blocks doing different things.

ankurankan: @palashahuja Also please add empty lines between code blocks doing different things.

pgmpy/inference/dbn_inference.py
```diff
@@ -0,0 +1,98 @@
+from collections import defaultdict
+from pgmpy.inference import VariableElimination
+from pgmpy.factors import TabularCPD
+from pgmpy.inference import Inference
+
+
+class DBNInference(Inference):
+
+    def query(self, variables, evidence=None, elimination_order=None):
```
ankurankan, line 12: @palashahuja How would someone give an elimination order ?

pgmpy/inference/base.py
```diff
@@ -80,3 +81,28 @@ def __init__(self, model):
...
+            start_model.add_cpds(*model.cpds)
+            self.start_bayesian_model = start_model
+            # interface nodes extraction for inference
+            self.interface_nodes = [(edge[0][0], 0) for edge in model.get_inter_edges()]
+            slice_1_interface_nodes = [(edge[0][0], 1) for edge in model.get_inter_edges()]
```
ankurankan: This can be done in a single line with a single loop: `self.interface_nodes, slice_1_interface_nodes = zip(*[((edge[0][0],  0), (edge[0][0], 1)) for edge in model.get_inter_edges()])`

pgmpy/inference/base.py
ankurankan: Couldn't get sense of lines 95 - 97. Replacing them with `self.start_bayesian_model = model` would do the same thing ??

ankurankan:
@palashahuja Add docstrings for all the methods.

ankurankan:
@palashahuja Also please add tests.

palashahuja:
@ankurankan  I will be using `self.start_bayesian_model` for belief propagation and variable elimination, and also bp requires the model to be converted to junction tree for which the dbn does not have the method. This is inconvenient though to do so. So maybe I was wondering whether to add a separate method inside dbn for converting it to junction tree.

palashahuja:
@ankurankan I actually wanted to use `elimination order` for `inference.query`. The elimination order will be given in the sense of considering the skeleton consisting the 0th and 1st timeslice and that will be considered till the end.

pgmpy/inference/base.py
```diff
@@ -80,3 +81,28 @@ def __init__(self, model):
...
+            # first timeslice.
+            # 2) one_and_half_model:- This is the bayesian model which consists of the
+            # interface nodes along with the first time slice nodes.
+            # Both of the above parameters will be required by the junction tree algorithm
+            # too.
+            start_model = BayesianModel(model.edges())
+            start_model.add_cpds(*model.cpds)
+            self.start_bayesian_model = start_model
+            # interface nodes extraction for inference
+            self.interface_nodes = [(edge[0][0], 0) for edge in model.get_inter_edges()]
```
ankurankan: @palashahuja Maybe simply add a method in DBN model to get the interface nodes. You can even accept an argument `slice` and default it to 0.

pgmpy/inference/base.py
```diff
@@ -80,3 +81,28 @@ def __init__(self, model):
...
+            slice_1_interface_nodes = [(edge[0][0], 1) for edge in model.get_inter_edges()]
+
+            #constructing one and half timeslice for inference
+            slice_1_intra_edges = [edge for edge in model.get_intra_edges() if edge[1][1] == 1]
```
ankurankan: @palashahuja Add an argument to `get_intra_edges` to accept the slice and default it to 0.

pgmpy/inference/base.py
```diff
@@ -80,3 +81,28 @@ def __init__(self, model):
...
+            self.one_and_half_model = BayesianModel(model.get_inter_edges() + intra_edges_1)
+            slice_1_nodes = [(node, 1) for node in model.nodes()]
+            slice_1_cpd = [cpd for cpd in model.cpds if set(list(cpd.variables)).issubset(nodes_1)]
+            cpd_inter = [model.get_cpds(node) for node in interface_nodes_1]
```
ankurankan: @palashahuja Add these methods for lines 105-107 also in the model class.

pgmpy/inference/base.py
```diff
@@ -80,3 +81,28 @@ def __init__(self, model):
...
+            self.one_and_half_model.add_cpds(*(cpd_1 + cpd_inter))
```
ankurankan: @palashahuja Can't see where `cpd_1` is defined..

pgmpy/inference/dbn_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        variable_dict = defaultdict(list)
+        for var in variables:
+            variable_dict[var[1]].append(var[0])
+
+        # This method extracts the max value in time slices
+        timerange = max(variable_dict)
```
ankurankan, line 18: A simple one liner can be `timerange = max([slice for slice in itertools.chain(variable_dict, evidence)])

pgmpy/inference/dbn_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        if evidence:
+            for node, timeslice in list(evidence):
+                timerange = max(timerange, timeslice)
+
+        int_nodes = self.interface_nodes
```
ankurankan, line 23: Why assigning it to `int_nodes`? You can simply use `self.interface_nodes`

pgmpy/inference/dbn_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        inference = VariableElimination(self.start_bayesian_model)
+
+        def vect(variables, time):
```
ankurankan, line 26: @palashahuja If any of these methods are going to be used in `map_query` also, then please put this outside of query method and add an `_` before the name.

ankurankan, line 26: And even if it's not being used anywhere else, put it outside the query method until and unless it's very specific to something you are doing in this method.

pgmpy/inference/dbn_inference.py
```diff
@@ -0,0 +1,98 @@
...
+        self.one_and_half_model.add_cpds(*new_cpds)
+
+        for time in range(1, timerange + 1):
+            inference = VariableElimination(self.one_and_half_model)
+            int_time_nodes = [(var, 1) for var, time in int_nodes]
+            if dict_factor(time, 1):
+                new_values = inference.query(int_time_nodes, dict_factor)
+            else:
+                new_values = inference.query(int_time_nodes)
+            new_cpds = [factor_to_cpd(new_values[node]) for node in new_nodes]
```
ankurankan, line 91: Do we really need to convert to cpd everytime and then convert back? I think a better solution would be to use some intermediate structure.

palashahuja:
@ankurankan Should I also add a method inside the model class of dbn for converting it to a bayesian model ? . Something similar to `to_markov_model()` in the bayesian model class because if that is the case then we can further reduce the code inside the base class ...

ankurankan:
@palashahuja I don't think that makes much sense as the size of the network would be too large and don't think anybody would be able to make something sensible out of it.

ankurankan:
@palashahuja Can you please add docstrings to all the methods?

PR #442, opened 2015-07-02
Markov Chain

pratyakshs:


pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,226 @@
...
+
+        sampled = DataFrame(index=range(size), columns=self.variables)
+        sampled.loc[0] = [self.state[var] for var in self.variables]
+        for i in range(size - 1):
+            for var in self.variables:
+                val = self.state[var]
+                next_val = sample_discrete(list(self.transition_models[var][val].keys()),
+                                           list(self.transition_models[var][val].values()))[0]
+                self.state[var] = next_val
+            sampled.loc[i + 1] = [self.state[var] for var in self.variables]
```
pratyakshs: `line_profiler` tells me that this line is a major bottleneck. It's because of `.loc[]`, any faster alternatives to this?

I've tried `.iloc[]` and `.ix[]`, both of which gave similar performance as `.loc[]`

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+        probabilities = np.zeros(windows)
+        for i in range(windows):
+            for j in range(window_size):
+                ind = i * 100 + j
+                state_eq = [sample.loc[ind, k] == state[k] for k in state.keys()]
+                if all(state_eq):
+                    probabilities[i] += 1
+        return probabilities / window_size
+
+    def generate_sample(self, start_state=None, size=1):
```
pratyakshs, line 302: This seems to be about 10 times faster than the `sample` method.

```
%timeit gen = x.generate_sample({'intel': 0, 'diff': 1}, size=10000); [x for x in gen]
1 loops, best of 3: 515 ms per loop

%timeit x.sample({'intel': 0, 'diff': 1}, size=10000)
1 loops, best of 3: 5.7 s per loop
```

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
+#!/usr/bin/env python3
+import numpy as np
+from pandas import DataFrame
+
+from pgmpy.utils import sample_discrete
+
+
+class MarkovChain(object):
```
ankurankan: @pratyakshs No need to inherit from `object` in python3. http://stackoverflow.com/a/9448136/1470009

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+        >>> model = MC()
+        >>> model.add_variable('grade', 3)
+        >>> grade_tm = {0: {0: 0.1, 1: 0.5, 2: 0.4}, 1: {0: 0.2, 1: 0.2, 2: 0.6 }, 2: {0: 0.7, 1: 0.15, 2: 0.15}}
+        >>> model.add_transition_model('grade', grade_tm)
+        """
+        # check if the transition model is valid
+        if not isinstance(transition_model, dict):
+            raise ValueError('Transition model must be a dict.')
+        if not set(transition_model.keys()) == set(range(self.cardinalities[variable])):
+            raise ValueError('Transitions must be defined for all states of variable {var}.'.format(var=variable))
```
ankurankan: @pratyakshs You can also add the expected states and and states got in the error message. Will make debugging more easier.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+        card: list
+            a list of cardinalities of the variables.
+        start_state: dict
+            representing the starting states of the variables.
+        """
+        if variables is None:
+            variables = []
+        if card is None:
+            card = []
+        assert isinstance(card, list)
```
ankurankan: @pratyakshs Acceptable types for `card` or `variables` should be np array, set, tuple or any iterable object.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+        1      1     0
+        2      0     1
+        3      1     0
+        4      0     2
+        """
+        # check if the start state is valid
+        if start_state is not None and self._check_state(start_state):
+            self.state = start_state
+        elif start_state is None and self.state is None:
+            raise ValueError('Start state not set.')
```
ankurankan: @pratyakshs Shouldn't we randomly set some start state if it's not given by the user ?

pratyakshs: Yes we should, thanks for pointing out.

pgmpy/models/MarkovChain.py
ankurankan: @pratyakshs This is a bottleneck because it's inside a for loop and is being called too many times.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+    Sample from it
+    >>> model.sample(size=5)
+       intel  diff
+    0      0     2
+    1      1     0
+    2      0     1
+    3      1     0
+    4      0     2
+    """
+    def __init__(self, variables=None, card=None, start_state=None):
```
ankurankan: @pratyakshs I think we should also have optional transition probabilities in the `__init__` method.. Then we will be able to simply create a complete chain in a single line.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+    >>> model = MC(['intel', 'diff'], [2, 3])
+
+    Add transition models
+    >>> intel_tm = {0: {0: 0.25, 1: 0.75}, 1: {0: 0.5, 1: 0.5}}
+    >>> model.add_transition_model('intel', intel_tm)
+    >>> diff_tm = {0: {0: 0.1, 1: 0.5, 2: 0.4}, 1: {0: 0.2, 1: 0.2, 2: 0.6 }, 2: {0: 0.7, 1: 0.15, 2: 0.15}}
+    >>> model.add_transition_model('diff', diff_tm)
+
+    Set a start state
+    >>> model.set_start_state({'intel': 0, 'diff': 2})
```
ankurankan: @pratyakshs According to the new representation of states, is should be a tuple as (var_name, state).

pgmpy/models/MarkovChain.py
ankurankan: Also using assert for throwing errors is not a very good way as it doesn't give any error message.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+        card: int
+            representing the cardinality of the variable to be added.
+
+        Examples:
+        ---------
+        >>> from pgmpy.models import MarkovChain as MC
+        >>> model = MC()
+        >>> model.add_variable('x', 4)
+        """
+        assert variable not in self.variables
```
ankurankan: @pratyakshs I don't think we should be throwing an error even if the variable is already present. Maybe show a warning message showing that the earlier attributes of the variable will be over written.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+        >>> model.sample(size=5)
+           intel  diff
+        0      0     2
+        1      1     0
+        2      0     1
+        3      1     0
+        4      0     2
+        """
+        # check if the start state is valid
+        if start_state is not None and self._check_state(start_state):
```
ankurankan: You have already checked this before storing it. Do we need to do this again?

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+            self.state = start_state
+        elif start_state is None and self.state is None:
+            raise ValueError('Start state not set.')
+
+        sampled = DataFrame(index=range(size), columns=self.variables)
+        sampled.loc[0] = [self.state[var] for var in self.variables]
+        for i in range(size - 1):
+            for var in self.variables:
+                val = self.state[var]
+                next_val = sample_discrete(list(self.transition_models[var][val].keys()),
```
ankurankan: You can put this list conversion out of the for loop to avoid computing it in each iteration.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+            for _, prob in transition.items():
+                if prob < 0 or prob > 1:
+                    raise ValueError('Transitions must represent valid probability weights.')
+                prob_sum += prob
+            if not np.allclose(prob_sum, 1):
+                raise ValueError('Transition probabilities must sum to 1.')
+
+        self.transition_models[variable] = transition_model
+
+    def sample(self, start_state=None, size=1):
```
ankurankan, line 214: @pratyakshs We need a better way to generate samples. Iterating over the size is not a good idea.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+    1      1     0
+    2      0     1
+    3      1     0
+    4      0     2
+    """
+    def __init__(self, variables=None, card=None, start_state=None):
+        """
+        Parameters:
+        -----------
+        variables: list
```
ankurankan: @pratyakshs Add more detailed docstring. For example: `list, array-like`

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+            a list of variables of the model.
+        card: list
```
ankurankan: Empty line after each parameter

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+    2      0     1
+    3      1     0
+    4      0     2
+    """
+    def __init__(self, variables=None, card=None, start_state=None):
+        """
+        Parameters:
+        -----------
+        variables: list
+            a list of variables of the model.
```
ankurankan: Capital A

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,299 @@
...
+        card: list
+            a list of cardinalities of the variables.
+        start_state: dict
+            representing the starting states of the variables.
```
ankurankan: Rephrase this.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+            A list of cardinalities of the variables.
+
+        start_state: array-like iterable object
+            List of tuples representing the starting states of the variables.
+        """
+        if variables is None:
+            variables = []
+        if card is None:
+            card = []
+        if not hasattr(variables, '__iter__'):
```
ankurankan: @pratyakshs You will also have to check that it's not a string since strings are also iterables.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Add transition models
+    >>> intel_tm = {0: {0: 0.25, 1: 0.75}, 1: {0: 0.5, 1: 0.5}}
+    >>> model.add_transition_model('intel', intel_tm)
+    >>> diff_tm = {0: {0: 0.1, 1: 0.5, 2: 0.4}, 1: {0: 0.2, 1: 0.2, 2: 0.6 }, 2: {0: 0.7, 1: 0.15, 2: 0.15}}
+    >>> model.add_transition_model('diff', diff_tm)
+
+    Set a start state
+    >>> from pgmpy.factors import State
+    >>> model.set_start_state([State('intel': 0), State('diff': 2)])
```
ankurankan: This should be `model.set_start_state([State('intel', 0), State('diff', 2)])`

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Sample from it
+    >>> model.sample(size=5)
```
ankurankan, line 47: Tried to run the code in this docstring.. Got an error here:

``` python
In [16]: model.sample(size=5)
TypeError                                 Traceback (most recent call last)
<ipython-input-16-609b6e0bcfc7> in <module>()
----> 1 model.sample(size=5)

/home/ubuntu/pgmpy/pgmpy/models/MarkovChain.py in sample(self, start_state, size)
    247 
    248         sampled = DataFrame(index=range(size), columns=self.variables)
--> 249         sampled.loc[0] = [st for var, st in self.state]
    250         for i in range(size - 1):
    251             for j, (var, st) in enumerate(self.state):

TypeError: 'NoneType' object is not iterable

```

Please check.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+            Dict (or list) of tuples representing the starting states of the variables.
+
+        Examples:
+        ---------
+        >>> from pgmpy.models import MarkovChain as MC
+        >>> from pgmpy.factors import State
+        >>> model = MC(['a', 'b'], [2, 2])
+        >>> model.set_start_state([State('a', 0), State('b', 1)])
+        """
+        if isinstance(start_state, dict):
```
ankurankan: Don't accept multiple types. Accept just the array of `State`. We will be making this change in the whole library.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+            state_dict = {var: st for var, st in start_state}
+            start_state = [State(var, state_dict[var]) for var in self.variables]
+        if start_state is None or self._check_state(start_state):
+            self.state = start_state
+
+    def _check_state(self, state):
+        """
+        Checks if a list representing the state of the variables is valid.
+        """
+        if not hasattr(state, '__iter__'):
```
ankurankan: Check for string too

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+        # check if the transition model is valid
+        if not isinstance(transition_model, dict):
+            raise ValueError('Transition model must be a dict.')
+
+        exp_states = set(range(self.cardinalities[variable]))
+        tm_states = set(transition_model.keys())
+        if not exp_states == tm_states:
+            raise ValueError('Transitions must be defined for all states of variable {v}. '
+                             'Expected states: {es}, Got: {ts}.'.format(v=variable, es=exp_states, ts=tm_states))
+        for _, transition in transition_model.items():
```
ankurankan, line 199: Add an empty line above this.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+            if not isinstance(transition, dict):
+                raise ValueError('Each transition must be a dict.')
+            prob_sum = 0
```
ankurankan, line 202: Add an empty line above this.

pgmpy/models/MarkovChain.py
```diff
@@ -0,0 +1,341 @@
...
+        Examples:
+        ---------
+        >>> from pgmpy.models import MarkovChain as MC
+        >>> model = MC(['intel', 'diff'], [2, 3])
+        >>> model.set_start_state({'intel': 0, 'diff': 2})
+        >>> intel_tm = {0: {0: 0.25, 1: 0.75}, 1: {0: 0.5, 1: 0.5}}
+        >>> model.add_transition_model('intel', intel_tm)
+        >>> diff_tm = {0: {0: 0.1, 1: 0.5, 2: 0.4}, 1: {0: 0.2, 1: 0.2, 2: 0.6 }, 2: {0: 0.7, 1: 0.15, 2: 0.15}}
+        >>> model.add_transition_model('diff', diff_tm)
+        >>> model.sample(size=5)
```
ankurankan, line 239: Getting error for this docstring.. Same as for the init method.

``` python
In [8]: model.sample(size=5)
TypeError                                 Traceback (most recent call last)
<ipython-input-8-609b6e0bcfc7> in <module>()
----> 1 model.sample(size=5)

/home/ubuntu/pgmpy/pgmpy/models/MarkovChain.py in sample(self, start_state, size)
    247 
    248         sampled = DataFrame(index=range(size), columns=self.variables)
--> 249         sampled.loc[0] = [st for var, st in self.state]
    250         for i in range(size - 1):
    251             for j, (var, st) in enumerate(self.state):

TypeError: 'NoneType' object is not iterable
```

pgmpy/models/MarkovChain.py
ankurankan, line 47: @pratyakshs And while fixing this make sure to add a test for this too.

ankurankan:
@pratyakshs Please fix these minor things and also add a few blank lines in between code blocks doing separate things. Once you fix these let me know, I will do some profiling and let's see what can we improve on.

pratyakshs:
@ankurankan ready for final review.

PR #447, opened 2015-07-11
Max Product LP for approximate inference

kislayabhi:
I have:
- added map_query method
- inherited from a base class.(I really thought the need of this)
- Made it UAI independent. (only a markov model is needed)
- Added doc-strings.
- Pep8 

Currently working on:
- shortening the code

### DIFF
--- a/pgmpy/inference/base.py
+++ b/pgmpy/inference/base.py
@@ -80,3 +80,40 @@ class Inference:
             for factor in model.get_factors():
                 for var in factor.variables:
                     self.factors[var].append(factor)
+
+
+class ApproxInference:
+    """
+    Base class for the approximate inference algorithm.
+    Parameters
+    ----------
+    model: pgmpy.models.MarkovModel
+    We need a pairwise MarkovModel for Approx. inference.
+    """
+    def __init__(self, model):
+
+        # List of the variables in the present pairwise Markov Model
+        self.variables = []
+
+        # List of the cardinalities of the above variables in that order
+        self.cardinality = []
+
+        # List of all the factors that are present in the current model
+        self.factors = []
+
+        # List of all the factors values that are present in the current model
+        self.factor_values = []
+
+        # In a pairwise markov model, the order of the single node factors will define the cardinality order in the
+        # list. ( Following the way in which the data is provided in a UAI file )
+        for i in model.factors:
+            self.factor_values.append(i.values)
+            if len(i.cardinality) == 1:
+                self.cardinality.append(i.cardinality[0])
+                self.variables.append(i.scope()[0])
+
+        for i in model.factors:
+            temp = []
+            for factor_variables in i.variables.keys():
+                temp.extend([self.variables.index(variable_name) for variable_name in factor_variables])
+            self.factors.append(temp)
--- a/pgmpy/inference/mplp.py
+++ b/pgmpy/inference/mplp.py
@@ -1,0 +1,298 @@
+from itertools import combinations
+from math import log
+from pgmpy.models import BayesianModel, MarkovModel
+from pgmpy.factors import TabularCPD, Factor
+import itertools
+import numpy as np
+from operator import sub
+from operator import add
+
+class UAIReader:
+    def __init__(self, path, var_names=None):
+        self.file = open(path)
+
+        model = self.file.readline().strip()
+        if model == 'MARKOV':
+            self.model = MarkovModel()
+        elif model == 'BAYES':
+            self.model = BayesianModel()
+
+        self.no_of_nodes = int(self.file.readline().strip())
+        if var_names and len(var_names) == self.no_of_nodes:
+            self.variables = np.array(var_names)
+            self.model.add_nodes_from(var_names)
+        else:
+            self.variables = np.array(list(map(str, range(self.no_of_nodes))))
+            self.model.add_nodes_from(list(map(str, range(self.no_of_nodes))))
+
+        self.cardinality = np.array(list(map(int, self.file.readline().strip().split(' '))))
+
+        self.no_of_functions = int(self.file.readline().strip())
+
+        self.functions_data = []
+        self.functions = []
+
+        for i in range(self.no_of_functions):
+            self.functions_data.append(self.file.readline().strip())
+
+        self.file.readline()
+        self.function_vars=[]
+        self.params=[]
+        for i in range(self.no_of_functions):
+            function_data = self.functions_data[i]
+            vars_index = list(map(int, function_data.strip().split()))[1:]
+            function_vars = self.variables[vars_index]
+            self.function_vars.append(function_vars)
+
+            self.model.add_edges_from(combinations(function_vars, 2))
+            if model == 'MARKOV':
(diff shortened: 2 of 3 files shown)


pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,752 @@
...
+class MPLP(ApproxInference):
+    """
+    Class for performing inference using Max-Product Linear Programming Relaxations.
+    We use the Max-Product Linear Programming (MPLP) algorithm to minimize the dual .
+    """
+
+    def __init__(self, model):
+        super().__init__(model)
+
+        if not isinstance(model, MarkovModel):
```
abinashpanda, line 21: @kislayabhi First check for the model and then call the `__init__`. Object should not be initialized if its not a correct one.

pgmpy/inference/base.py
```diff
@@ -80,3 +80,40 @@ def __init__(self, model):
...
+    model: pgmpy.models.MarkovModel
+    We need a pairwise MarkovModel for Approx. inference.
+    """
+    def __init__(self, model):
+
+        # List of the variables in the present pairwise Markov Model
+        self.variables = []
+
+        # List of the cardinalities of the above variables in that order
+        self.cardinality = []
```
abinashpanda, line 99: You don't need to compute the cardinality of all the variables again. I think all the models have `check_model()` method, which computes the cardinality of the variables. Have a look at `MarkovModel.check_model()` method.

pgmpy/inference/base.py
```diff
@@ -80,3 +80,40 @@ def __init__(self, model):
...
+
+        # List of all the factors that are present in the current model
+        self.factors = []
```
abinashpanda, line 102: Same. Don't initialize variables for storing the factors in inference class. You can get them directly from the `model.get_factors()`. If you want to store factors corresponding to each variable (in form of a dictionary), then you can inherit from the `Inference` class; it computes this during initialization itself.

pgmpy/inference/base.py
```diff
@@ -80,3 +80,40 @@ def __init__(self, model):
...
+
+        # List of all the factors values that are present in the current model
+        self.factor_values = []
+
+        # In a pairwise markov model, the order of the single node factors will define the cardinality order in the
+        # list. ( Following the way in which the data is provided in a UAI file )
```
abinashpanda, line 108: Please don't code things according to any file format. The inference class shouldn't be written how the values are provided in UAI file format. And please don't put UAI file again in the comments. Its misleading and makes the reader thinks that this class is specific to that file format.

pgmpy/inference/base.py
```diff
@@ -80,3 +80,40 @@ def __init__(self, model):
...
+    Base class for the approximate inference algorithm.
+    Parameters
+    ----------
+    model: pgmpy.models.MarkovModel
+    We need a pairwise MarkovModel for Approx. inference.
+    """
+    def __init__(self, model):
+
+        # List of the variables in the present pairwise Markov Model
+        self.variables = []
```
abinashpanda, line 96: Instead of creating another new class, use the `Inference` class. It has all the variables that you require.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,752 @@
...
+            raise ValueError('The model given is not MarkovModel.')
+
+        # For containing the cardinalities of all the variables present in the model
+        self.m_var_sizes = self.cardinality
```
abinashpanda, line 25: Don't use `m_var_name` in `Python`. It is a C++ style used to setting private variables; we don't require over here. So please remove `m_` from the start of all the variables.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,752 @@
...
+        self.m_all_regions = []
+        # Initial messages for more than single node regions are kept here. These don't change.
+        self.m_region_lambdas = []
+        # Maximum integral solution of the single nodes makes the m_best_val
+        self.m_best_val = -1000000
+        # This is to be used for calculating the breaking point of RunMPLP
+        self.last_obj = 1e40
+        self.m_intersect_map = []
+        for i in range(len(self.factors)):
+            self.objective.append([np.log(y) for y in self.factor_values[i]])
```
abinashpanda, line 40: Why store all the factor values in another variable again? Don't create `factor_values` class. You can change the code to 

``` python
for factor in self.factors:
    self.objective.append(np.log(factor.values))
```

I think it would serve the same purpose and there is no need to create another list to store all the values. I strongly suggest you to minimize the use of list as much as possible; they are not memory efficient.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,752 @@
...
+            if int_gap < 1:
+                n_iter_later = 600
+            ncluster_added = self.tighten_cycle()
+            if ncluster_added == 0:
+                print("This is the best result")
+                break
+            self.run_mplp(n_iter_later, objective_threshold, int_gap_threshold)
+
+
+class Region:
```
abinashpanda, line 733: I don't think `Region` class belongs here. Please update the documentation so that we can decide where to move this class.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,752 @@
...
+                curr_region = Region(self.factors[i], self.factors[:i + 1], self.factors[i], self.cardinality, i)
+                self.m_all_regions.append(curr_region)
+            if len(self.factors[i]) == 2:
+                self.m_intersect_map.append((self.factors[i], i))
+
+            # Initialise output vector
+            self.m_decoded_res = [0] * len(self.cardinality)
+            self.m_best_decoded_res = [0] * len(self.cardinality)
+
+    def _get_flat_ind(self, base_ind, base_card):
```
abinashpanda, line 52: As I said earlier, move this method from this class. ASAP

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,752 @@
...
+        for i in range(niter):
+            # For each of the 24 regions
+            for r in self.m_all_regions:
+                self.objective = self._update_msgs(self.objective, r)
+            obj = self._local_decode()
+            obj_del = self.last_obj - obj
+            self.last_obj = obj
+            int_gap = obj - self.m_best_val
+
+            print("m_best_val: {}".format(self.m_best_val))
```
abinashpanda, line 106: Don't put `print` statements in the final code. You can use it while debugging but make sure to remove them while sending a PR.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,752 @@
...
+        for vi in range(len(big_message)):
+            for si in range(no_subsets):
+                ind = self._get_idx(lst[vi], region.rintersects_relative_pos[si], region)
+                subsets[si][ind] = max(big_message[vi], subsets[si][ind])
+        return subsets
+
+    def _get_idx(self, bit_value_big, active_index, region):
+        """
+        Converts the bit value's active indexes into decimal number corresponding to the region's cardinality.
+        For eg:
```
abinashpanda, line 131: Please stick to the documentation format used everywhere. Do you even look at the code written elsewhere?

PR #449, opened 2015-07-17
Mplp reworked

kislayabhi:
@abinashpanda , @ankurankan 
I have written the update_message method here. 
Also did some changes in the factor class to meet my needs. (As it is getting reworked so did some hacks there)


pgmpy/factors/Factor.py
```diff
@@ -314,7 +315,26 @@ def product(self, *factors, n_jobs=1):
...
+        """
+        Returns a new factors instance after addition with factor
+
+        Parameters
+        ----------
+        factor : factors
+            The factor to be added
+        """
+        """
+        :param factor:
```
ankurankan: @kislayabhi Remove this

kislayabhi: @ankurankan : okay.
I needed an add functionality for factors. Any other work around or should i just wait for the reworked Factor?

abinashpanda: @kislayabhi I think by this @ankurankan means to remove this docstring, not this method.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+    >>> factor_c_d = Factor(['C', 'D'], cardinality=[2, 2], value=np.array([0.0043227, 231.34, 231.34, 0.0043227]))
+    >>> factor_d_e = Factor(['E', 'F'], cardinality=[2, 2], value=np.array([31.228, 0.032023, 0.032023, 31.228]))
+    >>> student.add_factors(factor_a, factor_b, factor_c, factor_d, factor_e, factor_f, factor_a_b,
+    ...    factor_b_c, factor_c_d, factor_d_e)
+    >>> mplp = Mplp(student)
+    """
+    def __init__(self, model):
+        super().__init__(model)
+
+        if not isinstance(model, MarkovModel):
```
abinashpanda, line 42: @kislayabhi check the model before calling the `__init___()` method. Don't initialize the object if there is any error.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+            raise ValueError('Only MarkovModel is supported')
+
+        # S = { c ∩ c': c, c' ∈ C, c ∩ c' != Null}
```
abinashpanda: Don't use unicode characters like ∩. Instead of this you can write its latex expression.

``` latex
S = \{ c \cap c^' : c, c^' \in C, c \cap c^' \neq \emptyset \}
```

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+        # It should be ready before initializing Cluster.
+        # Nothing but union of all the edge variables.
+        self.intersection_set_variables = set()
+
+        # The generation of all the Intersections of all
+        # the pairwise edges taken one at a time
+        for edge_pair in it.combinations([sorted(i) for i in model.edges()], 2):
```
abinashpanda: Do we need to sort the edges. I think we can do without this as well.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+            self.intersection_set_variables.add(frozenset(edge_pair[0]) & frozenset(edge_pair[1]))
+
+        # Initialize each cluster for variables appearing in each of the edges.
+        # We also need the potentials of the current edges.
+        self.cluster_set = []
+        scope_list = [i.scope() for i in model.get_factors()]
```
abinashpanda: Please don't use variables like `i`. You can converted this a

``` python
scope_list = [factor.scope() for factor in model.get_factors()]
```

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+
+    class Cluster:
+
+        """
+        Inner class for representing a cluster.
+        A cluster is a subset of variables.
+
+        Parameters
+        ----------
+        set of variables: tuple
```
abinashpanda: Put underscores.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+            self.intersection_sets_for_cluster_c = [i.intersection(self.cluster_variables)
+                                                    for i in intersection_set_variables
+                                                    if i.intersection(self.cluster_variables)]
+
+            # Initialize messages from this cluster to its respective intersection sets
+            # \lambda_{c \rightarrow \s} = 1/|S(c)| * max_{x_{c-s}}{\theta_{ij}}
+            self.message_from_cluster = {}
+            for i in self.intersection_sets_for_cluster_c:
+                other_variables = list(set(set_of_variables_for_cluster_c)-i)
+                phi = copy.deepcopy(cluster_potential)
```
abinashpanda: @kislayabhi I think instead of using `copy.deepcopy`, we can write a `copy` method in `Factor` class. This method would be returning a new factor using the same variables present in the cluster.

@ankurankan Your suggestions on this?

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+
+        Reference
+        ---------
+        Fixing Max-Product: Convergent Message-Passing Algorithms for MAP LP Relaxations
+        by Amir Globerson and Tommi Jaakkola.
+        Section 6, Page: 5; Beyond pairwise potentials: Generalized MPLP
+        """
+        updated_results = []
+        # The new updates will take place for the intersection_sets of this cluster.
+        for i in sending_cluster.intersection_sets_for_cluster_c:
```
abinashpanda: Please use appropriate variables instead of `i`. In future don't use any variable using i, j, k, l; use proper variable names.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+            message_not_from_cluster = []
+
+            # 2. Summation of messages not originating from the current sending_cluster and going into other
+            #    intersects than i of this cluster only.
+            sum_other_intersects = []
+
+            cluster_list_except_current = [j for j in self.cluster_set if j != sending_cluster]
+            other_intersects = [j for j in sending_cluster.intersection_sets_for_cluster_c if j != i]
+
+            for cluster in cluster_list_except_current:
```
abinashpanda: Why are you creating another list just for the purpose of iterations? Lists are very bad in terms of memory management.

Instead you can use something like this:

``` python
for cluster in self.cluster_set:
    if cluster == sending_cluster:
        pass
```

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+                for intersection_set in cluster.intersection_sets_for_cluster_c:
+                    if i == intersection_set:
+                        message_not_from_cluster.append(cluster.message_from_cluster[intersection_set])
+
+                for other_intersect in other_intersects:
```
abinashpanda: Same as above.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+                        for intersection_set in cluster.intersection_sets_for_cluster_c:
+                            if other_intersect == intersection_set:
+                                sum_other_intersects.append(cluster.message_from_cluster[intersection_set])
+
+            # lambda_1 = \lambda_{s}^{-c}
+            lambda_1 = copy.deepcopy(sending_cluster.message_from_cluster[i])
+            lambda_1.values = np.zeros(np.prod(lambda_1.cardinality))
+            # lambda_2 = \sum_{}{\lambda_{s'}^{-c}}
+            lambda_2 = copy.deepcopy(lambda_1)
```
abinashpanda: @kislayabhi Lets discuss on how to create a method in Factor class for copying. I don't using `copy.deepcopy` method so many times.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,172 @@
...
+            # maximize (lambda_2 + sending cluster potential) w.r.t nodes in the present cluster but not in intersect
+            phi = sending_cluster.cluster_potential + lambda_2
+            # phi = max_{x_{c-s}}{lambda_2 + /theta_c}
+            phi.maximize(list(set(sending_cluster.cluster_variables)-i))
+            intersection_length = len(sending_cluster.intersection_sets_for_cluster_c)
+            # \lambda_{c \rightarrow \s} = -(1 - 1 / |S(c)|) lambda_1 + (1 / |S(c)|) phi
+            lambda_c_s = -(1 - 1 / intersection_length) * lambda_1 + (1 / intersection_length) * phi
+            updated_results.append(lambda_c_s)
+
+        for i, j in zip(sending_cluster.intersection_sets_for_cluster_c, range(len(updated_results))):
```
abinashpanda: @kislayabhi I don't think its a good solution to modify the input variable passed to this method. It would lead to so many bugs which are not trivial to debug. Please think of some more elegant solution if this can't be removed, else remove this.

pgmpy/inference/mplp.py
kislayabhi: Yes. It will make using inplace functions like `maximize()` a lot easier.

pgmpy/inference/mplp.py
kislayabhi: Yup. Copying a Factor would be a one line thing mostly.
We just need to return a new factor with variables, cardinalities and values taken from the given factor.

kislayabhi: Actually I can just use `lambda_2 = Factor(lambda_1.variables, lambda_1.cardinality, lambda_1.values)` if it makes it better memory-wise

kislayabhi:
@abinashpanda : This is the same algorithm that I implemented except that Sontag later modified it in his PHD thesis a bit from (the original one in the Fixing-Max Product 2007 paper) that I implemented.

Should I modify for the below one?
Of the set of 3 papers, the Later 2 papers are written by Sontag along with the cpp code so I have an intention to go for that

image

AFAIK, the modifications are very much trivial

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,303 @@
...
+        self.cluster_set = {}
+        for factor in model.get_factors():
+            scope = tuple(factor.scope())
+            self.objective[scope] = factor
+            # For every factor consisting of more that a single node, we initialize a cluster.
+            if len(scope) > 1:
+                self.cluster_set[scope] = self.Cluster(self.intersection_set_variables, factor)
+
+        # L(\delta)
+        self.L = sum([max(self.objective[obj].values) for obj in self.objective])
```
abinashpanda: Don't use variable `L`, rename it to `loss` or `delta`, but please don't use `L`.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,303 @@
...
+        Later Modified by Sontag in "Introduction to Dual decomposition for Inference" Pg: 7 & 17
+        """
+
+        # The new updates will take place for the intersection_sets of this cluster.
+        # The new updates are:
+        # \delta_{f \rightarrow i}(x_i) = - \delta_i^{-f} +
+        # 1/{\| f \|} max_{x_{f-i}}\left[{\theta_f(x_f) + \sum_{i' in f}{\delta_{i'}^{-f}}(x_i')} \right ]
+
+        # Step. 1) Calculate {\theta_f(x_f) + \sum_{i' in f}{\delta_{i'}^{-f}}(x_i')}
+        objective_cluster = self.objective[tuple(sending_cluster.cluster_variables)]
```
abinashpanda: As we have discussed earlier don't use `tuple` instead convert it to a `frozenset`.

pgmpy/tests/test_inference/test_Mplp.py
```diff
@@ -0,0 +1,100 @@
...
+                                      factor_k_o, factor_d_h, factor_h_l, factor_l_p)
+
+        for factor in self.markov_model.factors:
+            factor.values = np.log(factor.values)
+        self.mplp = Mplp(self.markov_model)
+
+    def test_query_single_variable(self):
+        query_result = self.mplp.map_query(1000, 0.0002, 0.0002)
+        # Results from the Sontag code for a mplp run without tightening is:
+        sontag_result = {
```
abinashpanda: Don't use the term `sontag_result` over here. Instead rename it to `expected_result` and put the reference of the paper from where you got these values.

abinashpanda:
@kislayabhi PR looks good. There are few things that you need to change
- Are you sure the max-character length is 120. I don't think so; if its not the case then make the PEP8 corrections.
- As per our last discussion remove the `Cluster` class and make the changes accordingly.

Rest I don't think we need any major change. Once all these changes are done, this is good to merge.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,302 @@
...
+
+            # Initialize messages from this cluster to its respective intersection sets
+            # \lambda_{c \rightarrow \s} = 0
+            self.message_from_cluster = {}
+            for intersection in self.intersection_sets_for_cluster_c:
+                # Present variable. It can be a node or an edge too. (that is ['A'] or ['A', 'C'] too)
+                present_variables = list(intersection)
+
+                # Present variables cardinality
+                present_variables_card = [cluster_potential.cardinality[cluster_potential.scope().index(variable)]
```
abinashpanda: Don't do such complex operation to just get the cardinality of a variable. Instead you can simply use

``` python
present_variables_card = [cluster_potential.get_cardinality(variable) 
                          for variable in present_variables]
```

abinashpanda:
@kislayabhi Everything is good, just few last changes and a another important thing to add is to raise a `NotImplementedException` whenever the user tries to call `query()` method. After that its good to go.

kislayabhi:
I will be adding the triplet cluster tightening. Just few hours. Once that is done, we will have the complete implementation of the 1st two papers.

kislayabhi:
@abinashpanda , While writing the tests, can we use the new UAI reader to build the model and use it. That way, we can show the users that UAI files are easy to deal with when working with approximate inference.

abinashpanda:
@kislayabhi Yeah go for it. It would also be a good integration test.

kislayabhi:
@abinashpanda , As mentioned in the paper, we have to break the ties arbitrarily. This causes the result to change and we are left with 2 possible results of which both are correct.

See the test_Mplp file. I hope this is the correct way of handling these issues.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,502 @@
...
+            self.best_assignment = decoded_result_assignment
+
+    def _is_converged(self, dual_threshold=None, integrality_gap_threshold=None):
+        """
+        This method checks the integrality gap to ensure either:
+            * we have found a near to exact solution or
+            * stuck on a local minima.
+
+        Parameters
+        ---------
```
abinashpanda: Dashed line should be completely covering the previous line.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,502 @@
...
+        dual_threshold: double
+                        This sets the minimum width between the dual objective decrements. If the decrement is lesser
+                        than the threshold, then that means we have stuck on a local minima.
+
+        integrality_gap_threshold: double
+                                   This sets the threshold for the integrality gap below which we say that the solution
+                                   is satisfactory.
+
+        Reference:
```
abinashpanda: It should be rather 

```
References
----------
```

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,502 @@
...
+        >>> mm.add_nodes_from(['x1', 'x2', 'x3', 'x4', 'x5', 'x6', 'x7'])
+        >>> mm.add_edges_from([('x1', 'x3'), ('x1', 'x4'), ('x2', 'x4'),
+        ...                    ('x2', 'x5'), ('x3', 'x6'), ('x4', 'x6'),
+        ...                    ('x4', 'x7'), ('x5', 'x7')])
+        >>> phi = [Factor(edge, [2, 2], np.random.rand(4)) for edge in mm.edges()]
+        >>> mm.add_factors(*phi)
+        >>> mplp = Mplp(mm)
+        >>> mplp.find_triangles()
+        """
+        return [clique for clique in nx.find_cliques(self.model) if len(clique) == 3]
```
abinashpanda: Use `filter` for this

``` python
return list(filter(lambda x: len(x) == 3, nx.find_cliques(self.model)])
```

This would be much faster as compared to list comprehension with conditions.

pgmpy/inference/mplp.py
```diff
@@ -0,0 +1,502 @@
...
+    def _run_mplp(self, no_iterations):
+        """
+        Updates messages until either Mplp converges or if doesn't converges; halts after no_iterations.
+
+        Parameters
+        --------
+        no_iterations:  integer
+                        Number of maximum iterations that we want MPLP to run.
+        """
+        for niter in range(no_iterations):
```
abinashpanda, line 348: Rename `niter` to `n_iter`. Trivial changes :P

abinashpanda:
@kislayabhi There are a few trivial changes to be made. After that its ready to be merged.

Issue #462, opened 2015-08-09
printing factor gives error when variables are in tuple form


palashahuja:
```
>>> phi = Factor([('a', 0),('b', 0),('c', 0)], [2,2,2], list(range(8)))
>>> print(phi)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/palash/pgmpy/pgmpy/factors/Factor.py", line 428, in __str__
    return self._str(phi_or_p='phi', html=False)
  File "/home/palash/pgmpy/pgmpy/factors/Factor.py", line 446, in _str
    variables=','.join(string_header)))
TypeError: sequence item 0: expected str instance, tuple found
```

palashahuja:
fixed ...

Issue #461, opened 2015-08-09
factor reduce bug


palashahuja:
The factor reduce does not seem to work when the variables are in tuple form.

```
>>> phi = Factor([('a', 0),('b', 0),('c', 0)], [2,2,2], list(range(8)))
>>> phi = phi.reduce((('a', 0), 0), inplace=False)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/palash/pgmpy/pgmpy/factors/Factor.py", line 296, in reduce
    return Factor(new_vars, new_card, new_values)
  File "/home/palash/pgmpy/pgmpy/factors/Factor.py", line 77, in __init__
    self.variables[variable] = [State(variable, index) for index in range(card)]
  File "/home/palash/anaconda3/lib/python3.4/collections/__init__.py", line 67, in __setitem__
    if key not in self:
TypeError: unhashable type: 'numpy.ndarray'
```

palashahuja:
fixed ..

PR #463, opened 2015-08-09
bugs in factor.py refs #462 and #461

palashahuja:


ankurankan:
@palashahuja Can you also add some test cases for the failing cases?

PR #464, opened 2015-08-11
Dbn inference

palashahuja:


ankurankan:
@palashahuja Fix PEP8 and check other files to see how docstrings are written.

PR #470, opened 2015-08-22
New factor representation

ankurankan:


ankurankan:
Waiting on #471 & #472.

PR #477, opened 2015-09-03
adding reference to the algorithm

palashahuja:


ankurankan:
@palashahuja: Add the link to the paper also

Issue #247, opened 2014-12-24
Improve ModelError


ankurankan:
Right now `ModelError` just prints a message that something is wrong with the model. Improve this so that it could exactly list out things that are wrong

ankurankan:
`check_model` method is handling this.

PR #478, opened 2015-09-18
issue #247 improved ModelError

yashu-seth:
@ankurankan 
Please look into it.
Please suggest changes if required.
If its fine I will edit the other models as well.

### DIFF
--- a/pgmpy/inference/base.py
+++ b/pgmpy/inference/base.py
@@ -56,7 +56,8 @@ class Inference:
 
     def __init__(self, model):
         if not model.check_model():
-            raise ModelError("Model of type {!r} not valid".format(type(model).__name__))
+            model_error = "Model of type {!r} not valid, ".format(type(model).__name__) + model.get_error()
+            raise ModelError(model_error)
 
         if isinstance(model, JunctionTree):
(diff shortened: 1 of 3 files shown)


ankurankan:
@yashu-seth I think this issue has already been fixed since `model.check_model` will raise the appropriate exceptions. You can check once if something is missing in `check_model`. 

And also please let us know once before you start working on some issue so that you don't spend time on working on something that has already been fixed or not relevant anymore.

yashu-seth:
I think the check_model function should only return True or False and there should be a separate function to get the type of error because there could be cases where the ValueError raised has to be displayed along with some other error message. Like in this case I think ModelError should also be displayed. 
Please tell me if the changes are unnecessary and I will focus on a different issue. 
I will let you know before starting to work on an issue in the future.

ankurankan:
@yashu-seth  I don't think that there is much need of raising model error when check_model is already giving enough information for debugging. Also adding a new method will does almost the same thing will lead to code repetition.

yashu-seth:
Okay. The issue got me thinking over it in the first place. I will definitely notify you before starting on an issue from now on.

PR #479, opened 2015-09-24
Added TypeError for string parameters in Factor

yashu-seth:
Please review it.

### DIFF
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -99,7 +99,7 @@ class Factor:
         Examples
         --------
         >>> from pgmpy.factors import Factor
-        >>> phi = Factor(['x1', 'x2', 'x3'], [2, 3, 2], np.ones(8))
+        >>> phi = Factor(['x1', 'x2', 'x3'], [2, 3, 2], np.ones(12))
         >>> phi.scope()
         ['x1', 'x2', 'x3']
         """
@@ -154,7 +154,7 @@ class Factor:
         >>> from pgmpy.factors import Factor
         >>> phi = Factor(['diff', 'intel'], [2, 2], np.ones(4))
         >>> phi.assignment([1, 2])
-        [['diff_0', 'intel_1'], ['diff_1', 'intel_0']]
+        [[('diff', 0), ('intel', 1)], [('diff', 1), ('intel', 0)]]
         """
         index = np.array(index)
 
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -76,7 +76,11 @@ class Factor:
         >>> phi
         <Factor representing phi(x1:2, x2:2, x3:2) at 0x7f8188fcaa90>
         """
+
         values = np.array(values)
+        
+        if values.dtype != int and values.dtype != float:
+            raise TypeError("Values: Expected type int or type float elements only")
 
         if len(cardinality) != len(variables):
             raise ValueError("Number of elements in cardinality must be equal to number of variables")
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -379,6 +379,10 @@ class Factor:
         if isinstance(values, str):
             raise TypeError("values: Expected type list or array-like, got type str")
 
+        if any(isinstance(value, str) for value in values) or \
+               not all(isinstance(var, str) and isinstance(state, int) for var,state in values):
+            raise TypeError("values: must contain tuples or array-like elements of the form (type str, type int)")
+
         phi = self if inplace else self.copy()
 
         var_index_to_del = []
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -127,6 +127,9 @@ class Factor:
         >>> phi.get_cardinality(['x1', 'x2'])
         {'x1': 2, 'x2': 3}
         """
+        if isinstance(variables, str):
+            raise TypeError("variables: Expected type list or array-like, got type str")
+
         if not all([var in self.variables for var in variables]):
             raise ValueError("Variable not in scope")
 
@@ -227,6 +230,10 @@ class Factor:
         >>> phi.variables
         ['x2']
         """
+
+        if isinstance(variables, str):
+            raise TypeError("variables: Expected type list or array-like, got type str")
+
         phi = self if inplace else self.copy()
 
         for var in variables:
@@ -277,6 +284,9 @@ class Factor:
                [ 0.05,  0.07],
                [ 0.15,  0.21]]
         """
+        if isinstance(variables, str):
+            raise TypeError("variables: Expected type list or array-like, got type str")
+
         phi = self if inplace else self.copy()
 
         for var in variables:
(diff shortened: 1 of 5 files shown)


pgmpy/factors/Factor.py
```diff
@@ -127,6 +127,9 @@ def get_cardinality(self, variables):
         >>> phi.get_cardinality(['x1', 'x2'])
         {'x1': 2, 'x2': 3}
         """
+        if type(variables)==str:
```
ankurankan: @yashu-seth Always use `isinstance` to compare type.

pgmpy/factors/Factor.py
```diff
@@ -229,6 +232,9 @@ def marginalize(self, variables, inplace=True):
         """
         phi = self if inplace else self.copy()
 
+        if type(variables)==str:
```
ankurankan: @yashu-seth Put this above `phi = self if inplace else self.copy()`.

pgmpy/factors/Factor.py
```diff
@@ -127,6 +127,9 @@ def get_cardinality(self, variables):
         >>> phi.get_cardinality(['x1', 'x2'])
         {'x1': 2, 'x2': 3}
         """
+        if type(variables)==str:
+            raise TypeError("List or array-like parameter expected, got string")
```
ankurankan: @yashu-seth: Make the error message a bit more informative. Like you can include the argument name which you are expecting to be a list. Rather than showing just `string` write `type str`.

yashu-seth:
I have made the suggested changes. Please review it.

pgmpy/factors/Factor.py
```diff
@@ -127,6 +127,9 @@ def get_cardinality(self, variables):
         >>> phi.get_cardinality(['x1', 'x2'])
         {'x1': 2, 'x2': 3}
         """
+        if isinstance(variables, str):
+            raise TypeError("List or array-like parameter expected, given parameter {variables} is type str".format(variables=variables))
```
ankurankan: @yashu-seth This seems to be a better more clearer message to me: `variables: Expected type list or array-like, got type str`. And don't put the actual parameter value in the message, will just be confusing. What do you think?

ankurankan:
@yashu-seth And also please write a few tests for these cases.

yashu-seth:
I will write the tests for these cases.

pgmpy/factors/Factor.py
yashu-seth: Yes actual parameters should not be put. Let me change it.

pgmpy/factors/Factor.py
```diff
@@ -76,6 +76,9 @@ def __init__(self, variables, cardinality, values):
         >>> phi
         <Factor representing phi(x1:2, x2:2, x3:2) at 0x7f8188fcaa90>
         """
+        if not all(isinstance(value, int) or isinstance(value, float) for value in values):
+            raise TypeError("Values: can contain type int or type float elements only")
+
```
ankurankan, line 79: @yashu-seth Since we are converting the values in a numpy array in the next line. You can move this check into the next line and check for the dtype of the array. That would be much faster than this approach.

pgmpy/factors/Factor.py
```diff
@@ -365,6 +378,16 @@ def reduce(self, values, inplace=True):
...
         array([0., 1.])
         """
+
+        if isinstance(values, str):
+            raise TypeError("values: Expected type list or array-like, got type str")
+
+        if any(isinstance(value, str) for value in values):
+            raise TypeError("values: must contain tuples or array-like elements of the form (type str, type int)")
+
+        if not all(isinstance(var, str) and isinstance(state, int) for var,state in values):
```
ankurankan: @yashu-seth Since the error message is the same in both this case and the one just above this. You can merge the conditions with an `or` .

ankurankan:
@yashu-seth Please make the changes according to the comments. Everything else looks fine.

yashu-seth:
@ankurankan I have made the suggested changes.

pgmpy/factors/Factor.py
```diff
@@ -76,7 +76,11 @@ def __init__(self, variables, cardinality, values):
         >>> phi
         <Factor representing phi(x1:2, x2:2, x3:2) at 0x7f8188fcaa90>
         """
+
         values = np.array(values)
+        
+        if values.dtype != int and values.dtype != float:
+            raise TypeError("Values: can contain type int or type float elements only")
```
ankurankan: @yashu-seth Stick to a general format of error messages. As we have been using `Expected type ....` , use that here also.

pgmpy/factors/Factor.py
```diff
@@ -76,7 +76,11 @@ def __init__(self, variables, cardinality, values):
         >>> phi
         <Factor representing phi(x1:2, x2:2, x3:2) at 0x7f8188fcaa90>
         """
+
         values = np.array(values)
+        
+        if values.dtype != int and values.dtype != float:
```
ankurankan, line 82: @yashu-seth This wouldn't throw any error if we give some other type than `int` or `float`. Both conditions would be false. I think it should be `not(values.dtype != int or values.dtype != float)`

ankurankan:
@yashu-seth And please also check the tests. A few of them are failing.

pgmpy/factors/Factor.py
yashu-seth, line 82: I think it is okay.
If we give some other type both conditions will be true since it is checking not equal to.
Please correct me if I am making a mistake.

ankurankan, line 82: @yashu-seth Sorry, my bad here. You are correct. I think I assumed not equal as equal to.

yashu-seth, line 82: Okay. I will change the error message and stick to the format in future.

yashu-seth:
@ankurankan 
https://github.com/pgmpy/pgmpy/issues/485
please look into it.

yashu-seth:
Okay. I will look into the failing tests.

ankurankan:
@yashu-seth I already did reply on the issue #485

yashu-seth:
@ankurankan But wouldn't checking the parameters for their type cause the the same optimization problems as they used to earlier?

yashu-seth:
@ankurankan I have corrected most of the failing tests. Some are still failing but I guess they used to fail previously too since they are not consistent with the input parameter format.

ankurankan:
@yashu-seth Can you please merge in the latest develop in this branch ?

yashu-seth:
@ankurankan I have merged the latest develop.

ankurankan:
@yashu-seth Thanks. I will fix the other tests and merge this.

yashu-seth:
@ankurankan Thank you.

pgmpy/factors/Factor.py
```diff
@@ -365,6 +379,14 @@ def reduce(self, values, inplace=True):
         >>> phi.values
         array([0., 1.])
         """
+
+        if isinstance(values, str):
+            raise TypeError("values: Expected type list or array-like, got type str")
+
+        if any(isinstance(value, str) for value in values) or \
+               not all(isinstance(var, str) and isinstance(state, int) for var,state in values):
+            raise TypeError("values: must contain tuples or array-like elements of the form (type str, type int)")
```
ankurankan, line 388: @yashu-seth This is wrong. Variable names can be any hashable python object not just str. Please fix this.

yashu-seth, line 388: @ankurankan I do not understand, the code is dealing with restrictions to the values in the Factor class. Please tell me more.

PR #483, opened 2015-10-04
Fixes in cpd

yashu-seth:


yashu-seth:
I also found that the<code>  **str**  </code>method in CPD is not working. Should I start working on it?

pgmpy/factors/CPD.py
```diff
@@ -110,9 +110,12 @@ def __init__(self, variable, variable_card, values,
                                     " of strings.")
             self.evidence = evidence
             variables.extend(evidence[::-1])
-            if not len(evidence_card) == len(evidence):
+            if len(evidence_card) < len(evidence):
```
ankurankan, line 113: @yashu-seth I don't think we need to check two different cases for this. The previous behavior seems fine to me.

yashu-seth, line 113: @ankurankan  But in the previous behavior, the same error message - 
"Cardinality of all evidences not specified"
 would be given even if there are suppose, 4 evidences and we are giving 5 elements in the cardinality.
It wouldn't make much of a difference but the error message would be logically incorrect.
Please tell me if you feel otherwise.

ankurankan, line 113: @yashu-seth We can change the error message to a more clear one. But adding an extra condition doesn't make much sense to me. Since, it would be very few times that these errors would be thrown and for the normal case we will be adding an extra condition which will be run every time a CPD is defined.

yashu-seth, line 113: Okay. For now should I change it to a more clear error message. Any suggestions for the error message?

yashu-seth:
And please update me on the status of the **str** method.

ankurankan:
@yashu-seth Yeah, you can work on that. Please just create an issue for that and work on a separate branch so that you can send a separate PR for that.

PR #487, opened 2015-10-08
update in check_model method

yashu-seth:


yashu-seth:
``` python
def check_model(self):
        for node in self.nodes():
            cpd = self.get_cpds(node=node)
            if isinstance(cpd, TabularCPD):
                evidence = cpd.evidence
                if set(evidence if evidence else []) != set(parents if parents else []):
                    raise ValueError("CPD associated with %s doesn't have "
                                     "proper parents associated with it." % node)
                if not np.allclose(cpd.marginalize([node], inplace=False).values,
                                   np.ones(np.product(cpd.evidence_card)),
                                   atol=0.01):
                    raise ValueError('Sum of probabilites of states for node %s'
                                     ' is not equal to 1.' % node)
        return True
```

In the second condition, where it checks for the sum of probabilities, it marginalizes the cpd. In cpd marginalization, cpd normalization is also done. Hence, I don't understand when will this case occur, since cpds are already getting normalized.
Please tell me if I am missing out on certain cases.

ankurankan:
@yashu-seth This seems to be a bug to me. Can you fix this ?
Simplest way to fix this would be to convert the cpd to factor before marginalizing. You can use `to_factor` method for that.

yashu-seth:
@ankurankan  Okay. I will look into it.

yashu-seth:
@ankurankan I have fixed the bug and added the tests. Please review them.

yashu-seth:
@ankurankan I will soon look into the check_model for other models too.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -403,6 +403,37 @@ def get_cpds(self, node=None, time_slice=0):
...
+        ...                                      [0.3,0.7,0.02,0.2]], [('I', 0),('D', 0)],[2,2])
+        >>> dbn.add_cpds(grade_cpd)
+        >>> dbn.get_cpds()
+        [<TabularCPD representing P(('G', 0):3 | ('I', 0):2, ('D', 0):2) at 0x3348ab0>]
+        >>> dbn.remove_cpds(grade_cpd)
+        >>> dbn.get_cpds()
+        []
+        """
+        for cpd in cpds:
+            if isinstance(cpd, tuple):
```
ankurankan, line 433: @yashu-seth Not just tuple. Can be a list with 2 elements.

yashu-seth, line 433: @ankurankan I'll change it.

ankurankan, line 433: @yashu-seth I already did and merged this. I was working on something which was getting blocked because of this.

yashu-seth, line 433: Okay.

PR #491, opened 2015-10-14
updates in check_model method

yashu-seth:
@ankurankan I have removed cardinalities from the model attributes but it seems it is being used at other places as well. Should cardinalities be computed every single time it is required? Is there a particular problem in having it as an attribute?

### DIFF
--- a/pgmpy/models/ClusterGraph.py
+++ b/pgmpy/models/ClusterGraph.py
@@ -208,10 +208,17 @@ class ClusterGraph(UndirectedGraph):
         for factor in factors:
             self.factors.remove(factor)
 
-    def get_cardinality(self):
+    def get_cardinality(self, check_cardinality=False):
         """
         Returns a dictionary with the given factors as keys and their respective
         cardinality as values.
+        
+        Parameters
+        ----------
+        check_cardinality: boolean, optional
+            If, check_cardinality=True it checks if cardinality information
+            for all the variables is availble or not. If not it raises an error.
+
         Examples
         --------
         >>> from pgmpy.models import ClusterGraph
@@ -229,6 +236,8 @@ class ClusterGraph(UndirectedGraph):
         for factor in self.factors:
             for variable, cardinality in zip(factor.scope(), factor.cardinality):
                 cardinalities[variable] = cardinality
+        if check_cardinality and len(set((x for clique in self.nodes() for x in clique))) != len(cardinalities):
+            raise ValueError('Factors for all the variables not defined.')
         return cardinalities
 
     def get_partition_function(self):
--- a/pgmpy/models/ClusterGraph.py
+++ b/pgmpy/models/ClusterGraph.py
@@ -265,12 +265,13 @@ class ClusterGraph(UndirectedGraph):
     def check_model(self):
         """
         Check the model for various errors. This method checks for the following
-        errors. In the same time also updates the cardinalities of all the random
-        variables.
+        errors.
 
-        * Checks if clique potentials are defined for all the cliques or not.
+        * Checks if factors are defined for all the cliques or not.
         * Check for running intersection property is not done explicitly over
         here as it done in the add_edges method.
+        * Check if cardinality of random variable remains same across all the
+        factors.
 
         Returns
         -------
@@ -278,15 +279,10 @@ class ClusterGraph(UndirectedGraph):
             True if all the checks are passed
         """
         for clique in self.nodes():
-            if self.get_factors(clique):
-                pass
-            else:
-                raise ValueError('Factors for all the cliques or clusters not'
+            factors = filter(lambda x: set(x.scope()) == set(clique), self.factors)
+            if not any(factors):
+                raise ValueError('Factors for all the cliques or clusters not '
                                  'defined.')
-
-        if len(self.factors) != len(self.nodes()):
-            raise ValueError('One to one mapping of factor to clique or cluster'
-                             'is not there.')
 
         cardinalities = self.get_cardinality()
         for factor in self.factors:
--- a/pgmpy/models/ClusterGraph.py
+++ b/pgmpy/models/ClusterGraph.py
@@ -59,7 +59,6 @@ class ClusterGraph(UndirectedGraph):
         if ebunch:
             self.add_edges_from(ebunch)
         self.factors = []
-        self.cardinalities = defaultdict(int)
 
     def add_node(self, node, **kwargs):
         """
@@ -147,7 +146,7 @@ class ClusterGraph(UndirectedGraph):
         >>> student = ClusterGraph()
         >>> student.add_node(('Alice', 'Bob'))
         >>> factor = Factor(['Alice', 'Bob'], cardinality=[3, 2],
-        ...                 value=np.random.rand(6))
+        ...                 values=np.random.rand(6))
         >>> student.add_factors(factor)
         """
         for factor in factors:
@@ -209,6 +208,29 @@ class ClusterGraph(UndirectedGraph):
         for factor in factors:
             self.factors.remove(factor)
 
+    def get_cardinality(self):
+        """
+        Returns a dictionary with the given factors as keys and their respective
+        cardinality as values.
+        Examples
+        --------
+        >>> from pgmpy.models import ClusterGraph
+        >>> from pgmpy.factors import Factor
+        >>> student = ClusterGraph()
+        >>> factor = Factor(['Alice', 'Bob'], cardinality=[2, 2],
+        ...                 values=np.random.rand(4))
+        >>> student.add_node(('Alice', 'Bob'))
+        >>> student.add_factors(factor)
+        >>> student.get_cardinality()
+        defaultdict(<class 'int'>, {'Bob': 2, 'Alice': 2})
+        
+        """
+        cardinalities = defaultdict(int)
+        for factor in self.factors:
+            for variable, cardinality in zip(factor.scope(), factor.cardinality):
+                cardinalities[variable] = cardinality
+        return cardinalities
+
     def get_partition_function(self):
         r"""
         Returns the partition function for a given undirected graph.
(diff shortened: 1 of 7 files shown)


ankurankan:
@yashu-seth `self.cardinalities` is also used `triangulation` so may write a method for computing the cardinality? And then we can use that in triangulation method also.

yashu-seth:
@ankurankan Okay. I'll write a method to get cardinality.

yashu-seth:
@ankurankan I think get_cardinality should be added to FactorGraph and ClusterGraph as well.

ankurankan:
@yashu-seth Yes.. Please do..

yashu-seth:
FactorGraph - 

``` python
def check_model(self):
        variable_nodes = set([x for factor in self.factors for x in factor.scope()])
        if not bipartite.is_bipartite_node_set(self, variable_nodes):
            raise ValueError('Factors not associated for all the random'
                             'variables.')
```

Shouldn't this check be done for factor nodes also? And shouldn't the error message say something like edges between variable nodes and between factor nodes are not allowed?

ankurankan:
@yashu-seth Yes there should be.

yashu-seth:
@ankurankan I think the check done in add_edges will not always ensure that there are no edges between factor nodes and variable nodes. So why keep it there like that?

ankurankan:
@yashu-seth Oops.. Sorry my bad.. Maybe just add a comment there in `check_model` that `add_edge` handles this.

yashu-seth:
@ankurankan but in add_edge it checks for bipartite property only.  it will not satisfy the bipartite property even if there are some edges between two factor nodes. I think the factor nodes and variable nodes should be checked for bipartite_set property separately and not the entire graph.

ankurankan:
@yashu-seth Yes, you are right. We will need to check for bipartite_set for both variable and factor nodes.

yashu-seth:
@ankurankan okay I'll fix it.

yashu-seth:
@ankurankan I did not know why it was showing merge conflicts so I tried to pull the latest changes, but somehow other past commits also got uploaded. So I rebased those commits but it seems I have really messed up the branch.
Please help me on this.

ankurankan:
@yashu-seth I can't exactly understand what problem you are facing.
This PR is showing merge conflicts because while you were working on this branch some other changes were pushed to the main dev branch and now git is unable to automatically merge this branch to dev.
So, you will have to fetch the dev branch and merge it to your branch. While merging you will get error that there are conflicts and git will ask you to resolve it. Do `git status` to check which files are conflicting and you will have to manually fix those conflicts. Once you have resolved everything and removed all the conflict markers, stage those files on git and the commit.

yashu-seth:
@ankurankan thank you. I was also trying the same. I don't know where the problem was. Ill try it again.

yashu-seth:
@ankurankan I have added the tests for check_model and made the necessary fixes. Tests for <code> check_model </code> for ClusterGraph is still left but i'll add them later. Please look into this and if you think everything is okay, please merge this.

yashu-seth:
@ankurankan I have added tests for check_model in ClusterGraph too. I have also made certain changes in the check_model method that I felt necessary.

pgmpy/models/ClusterGraph.py
```diff
@@ -210,6 +209,29 @@ def remove_factors(self, *factors):
...
+        >>> student = ClusterGraph()
+        >>> factor = Factor(['Alice', 'Bob'], cardinality=[2, 2],
+        ...                 values=np.random.rand(4))
+        >>> student.add_node(('Alice', 'Bob'))
+        >>> student.add_factors(factor)
+        >>> student.get_cardinality()
+        defaultdict(<class 'int'>, {'Bob': 2, 'Alice': 2})
+        
+        """
+        cardinalities = defaultdict(int)
```
ankurankan, line 235: @yashu-seth We should also have a check somewhere which checks if we have cardinality information for all the variables in the network. It is possible that CPD/factor for all the variables have not yet been added to the network.

pgmpy/models/ClusterGraph.py
```diff
@@ -256,24 +279,16 @@ def check_model(self):
...
-            raise ValueError('One to one mapping of factor to clique or cluster'
-                             'is not there.')
-
+        cardinalities = self.get_cardinality()
         for factor in self.factors:
             for variable, cardinality in zip(factor.scope(), factor.cardinality):
-                if ((self.cardinalities[variable]) and
-                        (self.cardinalities[variable] != cardinality)):
+                if (cardinalities[variable] != cardinality):
                     raise CardinalityError(
```
ankurankan: @yashu-seth Also please replace this `CardinalityError` with appropriate inbuilt python Exception. We are trying to remove all custom exceptions from the codebase.

pgmpy/models/FactorGraph.py
```diff
@@ -133,16 +133,35 @@ def remove_factors(self, *factors):
...
+        >>> phi2 = Factor(['b', 'c'], [2, 2], np.random.rand(4))
+        >>> G.add_nodes_from([phi1, phi2])
+        >>> G.add_edges_from([('a', phi1), ('b', phi1),
+        ...                   ('b', phi2), ('c', phi2)])
+        >>> G.add_factors(phi1, phi2)
+        >>> G.get_cardinality()
+            defaultdict(<class 'int'>, {'c': 2, 'b': 2, 'a': 2})
+        
+        """
+        cardinalities = defaultdict(int)
```
ankurankan, line 167: @yashu-seth Same here. Add a check.

pgmpy/models/FactorGraph.py
```diff
@@ -150,29 +169,32 @@ def check_model(self):
...
         if len(factor_nodes) != len(self.factors):
             raise ValueError('Factors not associated with all the factor nodes.')
 
+        cardinalities = self.get_cardinality()
         for factor in self.factors:
             for variable, cardinality in zip(factor.scope(), factor.cardinality):
-                if ((self.cardinalities[variable]) and
-                        (self.cardinalities[variable] != cardinality)):
+                if (cardinalities[variable] != cardinality):
                     raise CardinalityError(
```
ankurankan: @yashu-seth Replace `CardinalityError` here also.

pgmpy/models/FactorGraph.py
```diff
@@ -307,14 +330,14 @@ def get_factors(self, node=None):
...
-        ...                   ('b', 'phi2'), ('c', 'phi2')])
         >>> phi1 = Factor(['a', 'b'], [2, 2], np.random.rand(4))
         >>> phi2 = Factor(['b', 'c'], [2, 2], np.random.rand(4))
         >>> G.add_factors(phi1, phi2)
+        >>> G.add_nodes_from([phi1, phi2])
+        >>> G.add_edges_from([('a', phi1), ('b', phi1),
+        ...                   ('b', phi2), ('c', phi2)])
         >>> G.get_factors()
-        >>> G.get_factors(node='phi1')
+        >>> G.get_factors(node = phi1)
```
ankurankan: @yashu-seth PEP 8. In case of arguments, there shouldn't be space around =.

pgmpy/models/MarkovModel.py
```diff
@@ -171,18 +170,39 @@ def remove_factors(self, *factors):
...
+        >>> from pgmpy.factors import Factor
+        >>> student = MarkovModel([('Alice', 'Bob'), ('Bob', 'Charles')])
+        >>> factor = Factor(['Alice', 'Bob'], cardinality=[2, 2],
+        ...                 values=np.random.rand(4))
+        >>> student.add_factors(factor)
+        >>> student.get_cardinality()
+        defaultdict(<class 'int'>, {'Bob': 2, 'Alice': 2})
+        
+        """
+        cardinalities = defaultdict(int)
```
ankurankan, line 202: @yashu-seth Same here. Check.

pgmpy/models/MarkovModel.py
```diff
@@ -192,18 +212,15 @@ def check_model(self):
         check: boolean
             True if all the checks are passed
         """
+        cardinalities = self.get_cardinality()
         for factor in self.factors:
             for variable, cardinality in zip(factor.scope(), factor.cardinality):
-                if ((self.cardinalities[variable]) and
-                        (self.cardinalities[variable] != cardinality)):
+                if cardinalities[variable] != cardinality:
                     raise CardinalityError(
```
ankurankan: @yashu-seth Same here. Inbuilt Exception

ankurankan:
@yashu-seth Please fix the code according to the comments. Everything else looks fine.

pgmpy/models/ClusterGraph.py
yashu-seth: @ankurankan Okay i'll change it. But why are we removing the custom exceptions? What is the problem?

pgmpy/models/FactorGraph.py
yashu-seth, line 167: @ankurankan Should the check be performed every time the function is called and shoud it throw an error? Or maybe we can add a parameter to the function that on being given will perform this check.

pgmpy/models/ClusterGraph.py
ankurankan: @yashu-seth Because most of the cases can be handled by the in built Exceptions (along with the error message). So, we don't need these custom exceptions.

pgmpy/models/FactorGraph.py
ankurankan, line 167: @yashu-seth Adding an optional parameter looks like the best solution. And I feel like throwing an error would be the best way.

yashu-seth, line 167: @ankurankan I think the get_cardinality method can be added in the parent class of the models. Can it?

ankurankan, line 167: @yashu-seth Yeah, we should be able to. But I don't think this will be a very good idea because cardinality is a property of the model. And the base graph classes are written for representing just the graph structure and properties related to that. 

A better place to put `get_cardinailty` would be in an intermediate class between base graph and model. But doesn't make much sense to create a new class for just one method. So, leave it like this only for now.

yashu-seth, line 167: @ankurankan if we don't want to add factors attribute in the <code>UndirectedGraph</code> we can add another derived class from it that can act as a parent class for MarkovModel, FactorGraph and ClusterGraph. I think it can save a lot of redundancy.

yashu-seth, line 167: @ankurankan Okay. Lets leave it like that for now. I'll make the changes in <code>get_cardinality</code> .

yashu-seth:
@ankurankan Should I make it check_card_info?

yashu-seth:
@ankurankan I have changed the parameter name from <code>check_info</code> to <code>check_cardinality</code>.

Issue #493, opened 2015-10-17
Fix tests for python2


ankurankan:
- [x] MPLP
- [x] readwrite module (mostly unicode problems)
- [x] Gibbs Sampling

khalibartan:
I just have to check the compatibility issues in these modules or there is some catch in it?

ankurankan:
@khalibartan Better check for the tests that we are skipping and according to that will need to fix wherever there are issues.

khalibartan:
Okay I'll give a try. I'm new to community so it might take me some time to send a pr as I have to understand some things. I think I can take up this issue. Do I need some background in probabilistic graphical models for this issue?

ankurankan:
Not a problem, feel free to post to mailing list if you face any issues.

And except for MPLP, I don't think you will need any background in PGM.

khalibartan:
Thanks

khalibartan:
@ankurankan take a look at pr 507. I think issue can be closed now.

Issue #448, opened 2015-07-14
python 2 support


ankurankan:

khalibartan:
Could you elaborate the issue a bit more

ankurankan:
@khalibartan Currently pgmpy only support python3. We need to add support for python 2.7. And it's almost done except for a few functionalities. And currently we are skipping tests for these. You can search for `@unittest.skipIf(six.PY2` in the project for checking which tests we are skipping because of compatibility issues. There is another issue listing the things which are not compatible yet: #493

ankurankan:
Fixed

PR #507, opened 2015-11-03
Removed compatibility issues from PomdpX and test_PomdpX

khalibartan:
In TestPomdpXReaderString io.StringIO is replaced by six.StringIO because in py2 io.StringIO doesn't read string inputs.
Also in PomdpXWriter class and TestPomdpXWriter class etree is being import form lxml or xml.etree.ElementTree because cElementTree is having some problems in taking dictionary attributes.
check issue #493 #448

### DIFF
--- a/pgmpy/inference/Sampling.py
+++ b/pgmpy/inference/Sampling.py
@@ -257,10 +257,10 @@ class GibbsSampling(MarkovChain):
         self.variables = np.array(model.nodes())
         self.cardinalities = {var: model.get_cpds(var).variable_card for var in self.variables}
 
-        for var in self.variables:
-            other_vars = [v for v in self.variables if var != v]
+        for Var in self.variables:
+            other_vars = [v for v in self.variables if Var != v]
             other_cards = [self.cardinalities[v] for v in other_vars]
-            cpds = [cpd for cpd in model.cpds if var in cpd.scope()]
+            cpds = [cpd for cpd in model.cpds if Var in cpd.scope()]
             prod_cpd = factor_product(*cpds)
             kernel = {}
             scope = set(prod_cpd.scope())
@@ -268,7 +268,7 @@ class GibbsSampling(MarkovChain):
                 states = [State(var, s) for var, s in zip(other_vars, tup) if var in scope]
                 prod_cpd_reduced = prod_cpd.reduce(states, inplace=False)
                 kernel[tup] = prod_cpd_reduced.values / sum(prod_cpd_reduced.values)
-            self.transition_models[var] = kernel
+            self.transition_models[Var] = kernel
 
     def _get_kernel_from_markov_model(self, model):
         """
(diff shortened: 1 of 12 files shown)


khalibartan:
@ankurankan for `test_ProbModelXML`, the fix isn't good. I have replace some strings by unicode in class `TestProbModelXMLReaderString`, method `test_potential` ,dictionary `potential_expected` because these strings were not being converted to unicode. 
Suggest some better solution

khalibartan:
Should I replace print by six.print_ because this landscape bot is showing repository health decreased.
I don't think that's an issue?

ankurankan:
@khalibartan I think you will need to specify strings to be unicode in case of python2. A better way to do this rather to append `u` in front of strings is to add `# -*- coding: utf-8 -*-` at the top of the file.
Don't worry too much about landscape.

khalibartan:
`# -*- coding: utf-8 -*-` is there in  test_ProbModelXML.py .But it wasn't working as intended. Not all strings should be converted to unicode only some of the strings in `potential_expected` should be converted.

khalibartan:
@ankurankan  can take a look at the pull request. Anything that needs to be changed ?

khalibartan:
Don't merge these commits. I think we should use cElementTree in testing and ElementTree in normal because only problem with cElementTree is that it is not taking dictionary arguments but it is faster.

khalibartan:
@ankurankan can give views on the commits if something needs to be done?

pgmpy/inference/Sampling.py
```diff
@@ -257,10 +257,10 @@ def _get_kernel_from_bayesian_model(self, model):
         self.variables = np.array(model.nodes())
         self.cardinalities = {var: model.get_cpds(var).variable_card for var in self.variables}
 
-        for var in self.variables:
-            other_vars = [v for v in self.variables if var != v]
+        for Var in self.variables:
```
ankurankan, line 260: @khalibartan Why was this failing earlier?

ankurankan:
@khalibartan I think except for MPLP everything else is fixed ? 
And if I remember correctly there was some issue in the algorithm of MPLP. So if you don't want to go through the algorithm, leave that and I will merge this ?

pgmpy/inference/Sampling.py
khalibartan, line 260: if look down below the person who wrote this was using var in a for loop also. Which caused some values to skipped in self.transition_models.See line 268.

khalibartan:
@ankurankan Yes apart from mplp all compatibility issues have been fixed in this pull request.You can merge this one. 
 Also I want to go for mplp .I have just began working on mplp and I'm just about to find the origin of the bug. I have some good lead on it.

ankurankan:
@khalibartan Cool.. No problem.

khalibartan:
@ankurankan look at the latest commit. mplp is fixed . Now you can merge this pr..

Issue #506, opened 2015-11-03
Add support for BIF file format


ankurankan:
Can get some file to test on from http://www.bnlearn.com/bnrepository/.

khalibartan:
Does this require any background in Bayesian Learning .If it doesn't , then I'm onto it.

ankurankan:
@khalibartan Not much. You will just need to know how to build a bayesian network in pgmpy.

khalibartan:
Okay, I think there are examples in pgmpy for this? I'm will start working on this issue tomorrow .

ankurankan:
@khalibartan Yeah, I think so.

ankurankan:
@khalibartan Also there is a dirty implementation of this here: https://github.com/pgmpy/pgmpy/pull/488/files#diff-e84ca375a827eebb64245fbfa41835caR1. And we would like to have the API as we have for other formats (check how other file formats are implemented). You can also check this notebook (might be helpful): https://github.com/pgmpy/pgmpy_notebook/blob/master/Reading%20and%20Writing%20from%20pgmpy%20file%20formats.ipynb

khalibartan:
Okay thanks.

khalibartan:
@ankurankan Now we can close this issue .

ankurankan:
Fixed in #512

PR #512, opened 2015-11-07
Added BIF.py into readwrite

khalibartan:
All functions are not implemented. only get_variable,get_states and
get_property are implemented.

Creating this pull request for easy review. 
I have tested these methods on [munin2.bif](http://www.bnlearn.com/bnrepository/munin4/munin2.bif.gz) and [dog-problem.bif](http://www.cs.cmu.edu/~javabayes/Examples/DogProblem/dog-problem.bif) and they are working fine.
Implemented these functions in accordance with [BIF v0.15 as given here](http://www.cs.cmu.edu/~javabayes/Home/node7.html#SECTION00730000000000000000) .
Time taken to run if not importing pgmpy and numpy was 0.06s on average with printing the complete variable_states for munin2.bif (PS: 1003 nodes)
Check issue #506 .

### DIFF
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -37,9 +37,6 @@ class BIFReader(object):
         >>> reader = BIFReader("bif_test.bif")
         <pgmpy.readwrite.BIF.BIFReader object at 0x7f2375621cf8>
         """
-        # An pyparsing expression for checking mulitple spaces
-        multi_space = OneOrMore(Literal(' '))
-
         if path:
             with open(path, 'r') as network:
                 self.network = network.read()
@@ -56,9 +53,6 @@ class BIFReader(object):
             # or "true""false" and "true" "false" and true false
             self.network = self.network.replace('"', ' ')
 
-        # replacing mulitple spaces or tabs by one space
-        #if '  ' in self.network :
-        #    self.network = multi_space.setParseAction(lambda t: ' ').transformString(self.network)
         if '/*' in self.network or '//' in self.network:
             self.network = cppStyleComment.suppress().transformString(self.network)  # removing comments from the file
 
@@ -113,13 +107,13 @@ class BIFReader(object):
     def variable_block(self):
         start = re.finditer('variable', self.network)
         for index in start:
-            end = self.network.find('}\n',index.start())
+            end = self.network.find('}\n', index.start())
             yield self.network[index.start():end]
 
     def probability_block(self):
         start = re.finditer('probability', self.network)
         for index in start:
-            end = self.network.find('}\n',index.start())
+            end = self.network.find('}\n', index.start())
             yield self.network[index.start():end]
 
     def get_network_name(self):
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -170,11 +170,11 @@ class BifReader(object):
         -------------
         >>> reader = BIF.BifReader("bif_test.bif")
         >>> reader.get_property()
-        {'bowel-problem': ['position = (190, 69)'],
-         'dog-out': ['position = (155, 165)'],
-         'family-out': ['position = (112, 69)'],
-         'hear-bark': ['position = (154, 241)'],
-         'light-on': ['position = (73, 165)']}
+        {{'bowel-problem': ['position = (335, 99)'],
+          'dog-out': ['position = (300, 195)'],
+          'family-out': ['position = (257, 99)'],
+          'hear-bark': ['position = (296, 268)'],
+          'light-on': ['position = (218, 195)']}
         """
         
         return self.variable_properties
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -1,4 +1,4 @@
-#import numpy as np
+import numpy as np
 #from pgmpy.models import BayesianModel
 #from pgmpy.factors import TabularCPD, State
 #from pgmpy.extern.six.moves import range
@@ -8,7 +8,7 @@ import re
 
 block_comment_regex = re.compile('/\*[^/]*/')   # A regular expression to check for block comments
 line_comment_regex = re.compile(r'//[^"\n"]*')  # A regular expression to check for line comments
-
+remove_multipule_spaces = re.compile(r'[" ""\t""\r""\f"][" ""\t""\r""\f"]*')
 class BifReader(object):
     
     
@@ -35,7 +35,7 @@ class BifReader(object):
         if path:
 
             path = open(path, 'r').read().replace('"', '')  # Opening the file and replacing qoutes by null string
-            
+            path = remove_multipule_spaces.sub(' ', path)
             if '/*' or '//' in FILE:
                 
                 path = block_comment_regex.sub('', path)
@@ -46,11 +46,11 @@ class BifReader(object):
         elif string:
 
             string = string.replace('"', '')
-            
+            string = remove_multipule_spaces.sub(' ', string)
             if '/*' or '//' in string:
                 
                 string = block_comment_regex.sub('', string)
-                string = line_comment_regex.sub('', string) #Striping comments off
+                string = line_comment_regex.sub('', string)     #Striping comments off
             
             self.network = string
 
@@ -58,9 +58,26 @@ class BifReader(object):
             
             raise ValueError("Must specify either path or string")
         
+        self.network_name()
         self.get_variables_info()
+        self.get_cpd()
     
-    
+    def network_name(self):
+        
+        """
+        Retruns the name of the network
+
+        Examples
+        ---------------
+        >>> reader = BIF.BifReader("bif_test.bif")
+        >>> reader.network_name()
+        'Dog-Problem'
+        """
+        start = self.network.find('network')
+        end = self.network.find('{',start)
+        self.network_name = self.network[start+8:end].strip()
+        return self.network_name
+
     def get_variables_info(self):
         
         """
@@ -103,17 +120,6 @@ class BifReader(object):
         self.variable_states = variable_states
         self.variable_properties = variable_properties
 
-    def network_name(self):
-        
-        """
-        Retruns the name of the network
-
-        Examples
-        ---------------
-        >>> reader = BIF.BifReader("bif_test.bif")
-        >>> reader.network_name()
-        'Dog-Problem'
-        """
     
     def get_variables(self):
         
@@ -199,3 +205,34 @@ class BifReader(object):
          'light-on': array([[ 0.6 ,  0.4 ],
                             [ 0.05,  0.95]])}
         """
+        probability_starts = [x.end()+1 for x in re.finditer('probability',self.network)]
+        probability_block = []
+        strip_parenthesis = re.compile('[\(\)\|]*')
+        strip_characters = re.compile(r'[a-zA-Z\(\);,]*')
+
+        for i in probability_starts:
+            probability_end = self.network.find('}\n',i)
+            probability_block.append(self.network[i:probability_end])
+
+        variable_cpds = {}
+
+        for block in probability_block :
+            block = block.split('\n')
+            block[0] = strip_parenthesis.sub('', block[0])
+            block[0] = block[0].strip()
+            name = block[0].split()[0]
+            block = block[1:]
+            cpd = []
+
+            for line in block:
+                line = line.strip()
+                line = strip_characters.sub('', line)
+                cpd.extend([float(x) for x in line.split() if x!= ''])
+
+            arr = np.array(cpd)
+            arr = arr.reshape(len(self.variable_states[name])
+                            ,arr.size//len(self.variable_states[name]))
+            variable_cpds[name] = arr
+
+        self.variable_cpds = variable_cpds
+        return variable_cpds
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -63,9 +63,9 @@ class BIFReader(object):
         self.variable_properties = self.get_property()
         self.variable_cpds = self.get_cpd()
         self.variable_parents = self.get_parents()
-        self.variable_edges = self.get_edges()
-        self.model = self.get_model() 
-
+        self.variable_edges = self.get_edges() 
+        self.get_model()
+ 
     def get_network_name(self):
 
         """
@@ -111,7 +111,7 @@ class BIFReader(object):
             variable_properties = {}
             # Defining a expression for valid word
             word_expr = Word(alphanums+'_'+'-')
-            name_expr = Suppress('variable')+ word_expr + Suppress('{')
+            name_expr = Suppress('variable') + word_expr + Suppress('{')
             state_expr = ZeroOrMore(word_expr + Optional(Suppress(",")))
             # Defining a variable state expression
             variable_state_expr = Suppress('type') + Suppress(word_expr) + Suppress('[') + Suppress(Word(nums)) + \
@@ -149,7 +149,7 @@ class BIFReader(object):
             probability_expr = Suppress('probability') + Suppress('(') + OneOrMore(word_expr) + Suppress(')')
             optional_expr = Suppress('(') + Suppress(OneOrMore(word_expr)) + Suppress(')')
             probab_attributes = optional_expr | Suppress('table')
-            cpd_expr = probab_attributes + OneOrMore( num_expr)
+            cpd_expr = probab_attributes + OneOrMore(num_expr)
 
             variable_parents = {}
             variable_cpds ={}
@@ -310,18 +310,37 @@ class BIFReader(object):
 
         """
         Returns the fitted bayesian model
+        
+        Example
+        ----------
+        >>> from pgmpy.readwrite import BIFReader
+        >>> reader = BIFReader("bif_test.bif")
+        >>> reader.get_model()
+        'Dog-Problem'
         """
         try:
             model = BayesianModel(self.variable_edges)
             model.name = self.network_name
 
             tabular_cpds = []
-            for var, values in self.variable_cpds.items():
-                cpd = TabularCPD(var, len(self.variable_states[var]), values,
-                                evidence = self.variable_parents[var],
-                                evidence_card = [len(self.variable_states[evidence_var])
-                                                for evidence_var in self.variable_parents[var]])
-                tabular_cpds.append(cpd)
+            count_dict={}
+            for var1,var2 in self.variable_edges:
+                if count_dict.get(var1,0) == 0:
+                    values = self.variable_cpds[var1]
+                    cpd = TabularCPD(var1, len(self.variable_states[var1]), values,
+                                    evidence = self.variable_parents[var1],
+                                    evidence_card = [len(self.variable_states[evidence_var])
+                                                    for evidence_var in self.variable_parents[var1]])
+                    count_dict[var1] = 1
+                    tabular_cpds.append(cpd)
+                if count_dict.get(var2,0) == 0:
+                    values = self.variable_cpds[var2]
+                    cpd = TabularCPD(var2, len(self.variable_states[var2]), values,
+                                    evidence = self.variable_parents[var2],
+                                    evidence_card = [len(self.variable_states[evidence_var])
+                                                    for evidence_var in self.variable_parents[var2]])
+                    count_dict[var2] = 1
+                    tabular_cpds.append(cpd)
 
             model.add_cpds(*tabular_cpds)
 
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -1,5 +1,5 @@
 import numpy as np
-from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nums, ZeroOrMore, OneOrMore
+from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nums, ZeroOrMore, OneOrMore, cppStyleComment
 from pgmpy.models import BayesianModel
 from pgmpy.factors import TabularCPD, State
 import re
@@ -27,10 +27,6 @@ class BifReader(object):
         # http://www.cs.cmu.edu/~javabayes/Examples/DogProblem/dog-problem.bif
         >>> reader = BifReader("bif_test.bif")
         """
-        # Defining some regular expressions
-
-        cpp_comment_regex = re.compile('/\*[^/]*/')                                     # A regular expression to check for block comments
-        line_comment_regex = re.compile(r'//[^"\n"]*')                                  # A regular expression to check for line comments
         remove_multipule_spaces = re.compile(r'[" ""\t""\r""\f"][" ""\t""\r""\f"]*')    # A regular expression to check for multiple spaces
 
         if path:
@@ -45,8 +41,7 @@ class BifReader(object):
                 """
             path = remove_multipule_spaces.sub(' ', path)                               # replacing multiple spaces or tabs by one space
             if '/*' or '//' in path:
-                path = cpp_comment_regex.sub('', path)
-                path = line_comment_regex.sub('', path)                                 # Striping comments off both type of comments
+                path = cppStyleComment.suppress().transformString(path)                 # removing comments from the file
 
             self.network = path
         elif string:
@@ -55,8 +50,7 @@ class BifReader(object):
 
             string = remove_multipule_spaces.sub(' ', string)                           # replacing mulitple spaces or tabs by one space
             if '/*' or '//' in string:
-                string = cpp_comment_regex.sub('', string)
-                string = line_comment_regex.sub('', string)                             # Striping comments off both types of comments
+                string = cppStyleComment.suppress().transformString(string)             # removing comments from the file
 
             self.network = string
         else:
@@ -141,7 +135,7 @@ class BifReader(object):
             temp = variable_state_expr.searchString(block)                                  # Finding the variable states
             variable_states[name] = list(temp[0][0])                                        # Assigning the variable states in form of list
             properties = property_expr.searchString(block)                                  # Getting the properties of variable                
-            variable_properties[name] = [ x[0].strip() for x in properties ]
+            variable_properties[name] = [x[0].strip() for x in properties ]
 
         self.variable_names = variable_names
         self.variable_states = variable_states
@@ -239,7 +233,7 @@ class BifReader(object):
         word_expr = Word(alphanums + '-' + '_') + Suppress(Optional("|")) + Suppress(Optional(","))
         num_expr = Word(nums + '-' + '+' + 'e' +'E' +'.')+ Suppress(Optional(","))
         probability_expr = Suppress('probability') + Suppress('(') + OneOrMore(word_expr) + Suppress(')')
-        cpd_expr = Suppress( Optional('('))+ Suppress(Group( OneOrMore( word_expr))) + Suppress( Optional(')')) + OneOrMore( num_expr)
+        cpd_expr = Suppress(Optional('('))+ Suppress(Group(OneOrMore(word_expr))) + Suppress(Optional(')')) + OneOrMore(num_expr)
 
         variable_parents = {}
         variable_cpds ={}
@@ -251,7 +245,7 @@ class BifReader(object):
             names = names[1:]
             variable_parents[variable] = names
             temp = cpd_expr.searchString( block )
-            arr = [ float(j) for i in temp for j in i ]
+            arr = [float(j) for i in temp for j in i]
             arr = np.array(arr)
             arr = arr.reshape((len(self.variable_states[variable]), arr.size//len(self.variable_states[variable])))
             variable_cpds[variable] = arr
@@ -291,8 +285,8 @@ class BifReader(object):
          ['bowel-problem', 'dog-out'],
          ['dog-out', 'hear-bark']]
         """
-        self.edges = [ [value, key] for key in self.variable_parents.keys()
-                     for value in self.variable_parents[key] ]
+        self.edges = [[value, key] for key in self.variable_parents.keys()
+                     for value in self.variable_parents[key]]
         return self.edges
 
     def get_model(self):
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -127,9 +127,11 @@ class BIFReader(object):
         >>> reader.network_name()
         'Dog-Problem'
         """
+        start = self.network.find('network')
+        end = self.network.find('}\n', start)
         # Creating a network attribute
         network_attribute = Suppress('network') + Word(alphanums + '_' + '-') + '{'
-        network_name = network_attribute.searchString(self.network)[0][0]
+        network_name = network_attribute.searchString(self.network[start:end])[0][0]
 
         return network_name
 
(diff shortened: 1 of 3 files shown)


khalibartan:
@ankurankan look at these commits and tell me for improvement . I have tried to make it robust. I think I can make it more robust to cases where there are multiple whitespaces together. Should I add this feature.

khalibartan:
I have some doubts as how to find parents and edges using the conditional probability.
If this doubt is cleared I'll be able to complete the reader function.
Should I implement writer class also?

ankurankan:
@khalibartan I feel like using pyparsing would make this all much more cleaner.

khalibartan:
@ankurankan Yeah it would have made it cleaner but I don't know how to use pyparsing .I also first thought of using it but there documentation was not good . So i switched back to regex.
I will try to rewrite reader using pyparsing but I'll use it along with regex.So ignore the commits of BifReader. Also I just now realised that I forgot to include exponential notation for small probabilities  and I didn't knew that default keyword is used. Also there are many issues still if you go through this [link](http://www.cs.cmu.edu/~fgcozman/Research/InterchangeFormat/Old/xmlbif02.html) you will find. There are many things they have mentioned which I couldn't understand. Here it is :-
What is meant by noisy OR , noisy AND and noisy ADDER. Also how I can identify parent and edges. Also here I have a doubt

```
probability(GasGauge | Gas, BatteryPower) {              
           (yes, high) 0.999 0.001;          
           (yes, low) 0.850 0.150;          
           (yes, medium) 0.000 1.000;          
           (no, high) 0.000 1.000;          
           (no, low) 0.000 1.000;          
           (no, medium) 0.000 1.000;  
}
```

In this example the yes, high, medium .. etc attributes are significant or not while making the cpd. I couldn't come up with the answer while writing so I imitated what was being done in XMLBIF .
Reply to this as early as possible so that I can get working.
PS: The falling of test on new commit is due to network error otherwise it would have passed and decrease in health is because of cyclic import in MarkovChain.py .

khalibartan:
@ankurankan the failing of test and that many code smells are not because of my commit. The internet connection failed while the last test. you can run nosetests on your system. And those code smells are from MarkovChain.

ankurankan:
@khalibartan Sorry for the late reply. I would still prefer the implementation by using pyparsing completely as it will make things much more cleaner and easy to understand in future. And for now leave things that are specific to Noisy OR, AND etc models (We currently don't have specific implementations for them so doesn't make much sense to support them right now). 
Each probability block is of the form of `probability( var | var1, var2, ... )`. So in the case of a Bayesian Network `var1, var2, ...` will be parents of `var` in the network.

khalibartan:
@ankurankan How are edges identified.
```
probability(GasGauge | Gas, BatteryPower) {              
           (yes, high) 0.999 0.001;          
           (yes, low) 0.850 0.150;          
           (yes, medium) 0.000 1.000; 
}
```
Should I take the yes, high etc into account while creating the cpd or ignore them and just do what is being done for `table` keyword.

ankurankan:
@khalibartan You can get the edges from the probability blocks. Each block is in the form of `probability(var | var1, var2, ...)`, and since `var1, var2, ...` are parents of `var`, there will be edges `var1 --> var`, `var2 --> var` etc. For now ignore the state names (ie high, low, yes etc). We still don't have the feature to store the state names.

khalibartan:
@ankurankan  I'm having a problem while replacing things using pyparsing. So I have implemented these using regex , otherwise pyparsing is being used. I think you should go through the code and comment if you find regex difficult to understand. I'll try to think pyparsing implementation of it.

ankurankan:
@khalibartan We would like to do it purely with pyparsing.

khalibartan:
@ankurankan I'm not able to remove comments using pyparsing. Can you suggest how I can do so.

khalibartan:
@ankurankan cpd and related functions are implemented.

khalibartan:
@ankurankan cpd and related functions are implemented. You can now give it a look , specifically the get_model method. I couldn't write the test for it. Though as per my understanding implementation is correct. Also now only one regex is being used which I'll remove when I'll find its alternative in pyparsing *(though I can remove it without any alternative but it tries to remove some outliers sensitivity )*. Apart form if you find any other issue you can comment.
Also how coverage can be increased .

khalibartan:
All the read functions are implemented. @ankurankan  Now you can take a look. All the regex are removed and it is being done using pyparsing. I want you to take a good look at `get_model()` method. I have not written its test function . Though I think its implementation is correct.
Any other things that I have skipped while considering the cases please also comment them.
Why health decreased so much. I didn't even changed the code much.

ankurankan:
@khalibartan For checking the reasons for decrease in the repo health you can click on the link in the comment by landscape-bot. And go through the issues there and try to fix those which look relevant to your changes.
And also please implement the writer class and add tests for `get_model`.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
+import numpy as np
```
ankurankan: @khalibartan Please check PEP8 way of ordering and grouping imports. https://www.python.org/dev/peps/pep-0008/#imports

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nums, ZeroOrMore, OneOrMore, cppStyleComment, Literal
+from pgmpy.models import BayesianModel
+from pgmpy.factors import TabularCPD
+import re
+
+
+class BifReader(object):
```
ankurankan: `Should be BIFReader`

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        ----------------
+        path : file or str
+                File of bif data
+        string : str
+                String of bif data
+        Examples
+        -----------------
+        # dog-problem.bif file is present at
+        # http://www.cs.cmu.edu/~javabayes/Examples/DogProblem/dog-problem.bif
+        >>> reader = BifReader("bif_test.bif")
```
ankurankan: @khalibartan Should be a working example. So add import statement and maybe fetch the file and then read it.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        Initialisation of BifReader object
+
+        Parameters
+        ----------------
+        path : file or str
+                File of bif data
+        string : str
+                String of bif data
+        Examples
```
ankurankan, line 30: Add an empty line above this

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+            """
+            Replacing quotes by spaces to remove case sensitivity like:
+            "Dog-Problem" and Dog-problem
+            or "true""false" and "true" "false" and true false
+            """
+            self.network = self.network.replace('"', ' ')
+
+        self.network = multi_space.setParseAction(lambda t:' ').transformString(self.network)   # replacing mulitple spaces or tabs by one space
+
+        if '/*' or '//' in self.network:
```
ankurankan: @khalibartan I don't think this is doing what you are expecting because this will always return true.

``` python
In [7]: if 1 or 2 in [1, 2, 3]:
   ...:     print (True)
   ...:     
True

In [8]: if 1 or 2 in [1, 3]:
   ...:     print (True)
True
```

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+            self.network = cppStyleComment.suppress().transformString(self.network)             # removing comments from the file
+
+        self.get_network_name()
+        self.get_variables_info()
+        self.get_cpd()
+        self.get_edges()
+        self.model = self.get_model()
```
ankurankan: @khalibartan Also please keep the API same as in other reader classes. Don't have a `model` property. Let the user do call `get_model` to get the model.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+    def get_network_name(self):
+
+        """
+        Retruns the name of the network
+
+        network attributes are of the format
+            network <name_of_network> {
+                attirbutes
+            }
+        Example of network attribute
```
ankurankan: @khalibartan Keep all the docstrings in the format that we are using everywhere. We use these docstrings to automatically generate our documentation, so we need it in that format only.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        """
+        network_attribute = Suppress('network') + Word(alphanums+'_'+'-') + '{'                            # Creating a network attribute 
+        temp = network_attribute.searchString(self.network)
+        self.network_name = temp[0][0]
+        return self.network_name
+
+    def get_variables_info(self):
+
+        """
+        Functions gets all type of variable information
```
ankurankan: Same here

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+
+        variable block is of format
+
+        variable <name_of_variable> {
```
ankurankan: Ideally this should be shown in code blocks on the documentation webpage. We use sphinx for generating documentation so check how to put something in code blocks.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        return
+
+    def get_variables(self):
+
+        """
+        Returns list of variables of the network
+
+        Sample run
+        -------------
+        >>> reader = BIF.BifReader("bif_test.bif")
```
ankurankan: Same here about formatting and have a complete working example.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        Examples
+        -----------------
+        # dog-problem.bif file is present at
+        # http://www.cs.cmu.edu/~javabayes/Examples/DogProblem/dog-problem.bif
+        >>> reader = BifReader("bif_test.bif")
+        """
+        multi_space = OneOrMore(Literal(' ')|'\t'|'\r'|'\f')                                    # An expression for matching multi spaces
+
+        if path:
+            self.network = open(path, 'r').read()
```
ankurankan: Use `with open(path, 'r') as ....`

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+
+        Sample run
+        ---------------
+        >>> reader = BIF.BifReader("bif_test.bif")
+        >>> reader.network_name()
+        'Dog-Problem'
+        """
+        network_attribute = Suppress('network') + Word(alphanums+'_'+'-') + '{'                            # Creating a network attribute 
+        temp = network_attribute.searchString(self.network)
+        self.network_name = temp[0][0]
```
ankurankan: @khalibartan Try to avoid defining properties in methods other than `__init__`. It helps in keeping things clear as you will be able to just see the `__init__` method and can tell what are the properties.

So here rather than assigning value to `self.network_name` just simply return `temp[0][0]` and in the `__init__` method rather that simply calling `self.get_network_name()` do something like `self.network_name = self.get_network_name()`. 
The end result is the same in both cases but the second method will make the code much more cleaner.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        }
+
+        Sample run
+        ---------------
+        >>> reader = BIF.BifReader("bif_test.bif")
+        >>> reader.network_name()
+        'Dog-Problem'
+        """
+        network_attribute = Suppress('network') + Word(alphanums+'_'+'-') + '{'                            # Creating a network attribute 
+        temp = network_attribute.searchString(self.network)
```
ankurankan: Don't use variable names like `temp`. Use something that tells about the variable. For example here you could have done something like this: `network_name = network_attribute.searchString(self.network)[0][0]`

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+	        type discrete[2] {  "true"  "false" };
+            property "position = (218, 195)" ;
+        }
+        """
+        variable_block_starts = [x.start() for x in re.finditer('variable', self.network)]  # Finding the beginning of variable block
+        variable_block = []
+
+        for i in variable_block_starts:
+            variable_block_end = self.network.find('}\n', i)                                # Finding the end of variable block
+            variable_block.append( self.network[ i : variable_block_end ] )                 # Appending the complete variable block in a list
```
ankurankan: @khalibartan Add comments only if you are doing something complex and the comment would actually help in understanding it quickly.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+
+        # Defining a variable state expression
+        variable_state_expr = Suppress('type') + Suppress(word_expr) + Suppress('[') + Suppress(Word(nums)) + \
+        Suppress(']') + Suppress('{') + Group(state_expr) + Suppress('}') + Suppress(';')
+        # variable states is of the form type description [args] { val1, val2 }; (comma may or may not be present)
+
+        property_expr = Suppress('property') + CharsNotIn(';') + Suppress(';')              # Creating a expr to find property
+
+        for block in variable_block:
+            temp = name_expr.searchString(block)                                            # Finding the string of the format name_expr
```
ankurankan: Don't use `temp` as variable name.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+            name = temp[0][0]                                                               # Assigning name the value of variable name
+            variable_names.append(name) 
+            temp = variable_state_expr.searchString(block)                                  # Finding the variable states
+            variable_states[name] = list(temp[0][0])                                        # Assigning the variable states in form of list
+            properties = property_expr.searchString(block)                                  # Getting the properties of variable                
+            variable_properties[name] = [x[0].strip() for x in properties ]
+
+        self.variable_names = variable_names
+        self.variable_states = variable_states
+        self.variable_properties = variable_properties
```
ankurankan: Same here. Return all these values and assign it in the `__init__` method.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        >>> reader = BIF.BifReader("bif_test.bif")
+        >>> reader.get_states()
+        {'bowel-problem': ['true','false'],
+         'dog-out': ['true','false'],
+         'family-out': ['true','false'],
+         'hear-bark': ['true','false'],
+         'light-on': ['true','false']}
+        """
+
+        return self.variable_states
```
ankurankan: This would throw an error until the user has called `get_variable_info` before calling this method as you are assigning `self.variable_states` in that method. So add a check here for that.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,312 @@
...
+        >>> reader = BIF.BifReader("bif_test.bif")
+        >>> reader.get_property()
+        {{'bowel-problem': ['position = (335, 99)'],
+          'dog-out': ['position = (300, 195)'],
+          'family-out': ['position = (257, 99)'],
+          'hear-bark': ['position = (296, 268)'],
+          'light-on': ['position = (218, 195)']}
+        """
+
+        return self.variable_properties
```
ankurankan: Same as above.

ankurankan:
@khalibartan Also I tried to read some other files and got error for these:
1. http://www.bnlearn.com/bnrepository/andes/andes.bif.gz
2. http://www.bnlearn.com/bnrepository/child/child.bif.gz

And any particular reason why are we not defining a grammar for the format and then directly parse everything using it as we have done in UAI Reader: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/readwrite/UAI.py#L50

Also can you check if there is some space for performance optimization. Some bigger networks like diabetes takes several minutes to read.

khalibartan:
@ankurankan No there is no particular reason for using a grammar method will do so. I'll Improve all the things as mentioned.
There seems to be some problem in get_model method , otherwise it was working fine. I'll take a look at that.

khalibartan:
@ankurankan I tried to check the error , but I couldn't find the problem. get_model method was working fine . I tried to run on other datasets apart from dog-problem , it worked fine on http://www.bnlearn.com/bnrepository/alarm/alarm.bif.gz and http://www.bnlearn.com/bnrepository/munin4/munin2.bif.gz to check if my methods were generalised . But when I ran on your mentioned dataset they threw error. I tried to debug it but I couldn't . Can you take a look? I can tell you that states, properties, variable names returned by the reader are on these datasets are correct. Cannot say it for cpd.

khalibartan:
@ankurankan I tried to run on other datasets apart from dog-problem , it worked fine on http://www.bnlearn.com/bnrepository/alarm/alarm.bif.gz and http://www.bnlearn.com/bnrepository/munin4/munin2.bif.gz to check if my methods were generalised . But when I ran on your mentioned dataset they threw error. I tried to debug it but I couldn't . **Can be problem caused while creating Bayesian model when a particular variable name is not present in any of the edges but has it cpd. It seems that in andes.bif `SNode_14` has CPD and has been declared as variable but doesn't contain any edges from it or into it.** The new commit worked fine on your mentioned data set.

ankurankan:
@khalibartan Still getting error for the child network:

``` python

In [8]: reader = BIFReader('child.bif')
<ipython-input-8-826ca7999b85> in <module>()
----> 1 reader = BIFReader('child.bif')

/home/ankur/pgmpy/pgmpy/readwrite/BIF.py in __init__(self, path, string)
     55             self.network = cppStyleComment.suppress().transformString(self.network)  # removing comments from the file
     56 
---> 57         self.grammar = self.Get_grammar(self.network)
     58         self.network_name = self.get_network_name()
     59         self.variable_names = self.get_variables()

/home/ankur/pgmpy/pgmpy/readwrite/BIF.py in __init__(self, network)
     88 
     89             self.net = network
---> 90             self.variable_names, self.variable_states, self.variable_properties = self.get_variable_grammar()
     91             self.variable_parents, self.variable_cpds = self.get_probability_grammar()
     92 

/home/ankur/pgmpy/pgmpy/readwrite/BIF.py in get_variable_grammar(self)
    120                 name = name_expr.searchString(block)[0][0]
    121                 variable_names.append(name)
--> 122                 variable_states[name] = list(variable_state_expr.searchString(block)[0][0])
    123                 properties = property_expr.searchString(block)
    124                 variable_properties[name] = [x[0].strip() for x in properties]

/home/ankur/anaconda3/envs/py3/lib/python3.4/site-packages/pyparsing.py in __getitem__(self, i)
    318     def __getitem__( self, i ):
    319         if isinstance( i, (int,slice) ):
--> 320             return self.__toklist[i]
    321         else:
    322             if i not in self.__accumNames:

IndexError: list index out of range
```

khalibartan:
@ankurankan Okay. Will check for the cause of error.
Also have you read the above comment. I'm not considering the nodes which don't have any edges while creating the model. Is that right ?

ankurankan:
@khalibartan No. You should add all the nodes even if it doesn't have an edge associated with it.

khalibartan:
@ankurankan Then bayesian model is giving an error. That what caused error in andes.bif . How I can add a node without any edge in bayesian model?

ankurankan:
@khalibartan Can you please explain what you mean by bayesian model is giving an error ?

khalibartan:
```
 File "BIF.py", line 341, in <module>
    reader.get_model()
  File "BIF.py", line 329, in get_model
    model.add_cpds(*tabular_cpds)
  File "/home/utgup/pgmpy/pgmpy/models/BayesianModel.py", line 155, in add_cpds
    raise ValueError('CPD defined on variable not in the model', cpd)
ValueError: ('CPD defined on variable not in the model', <TabularCPD representing P(SNode_19:2) at 0x7fa390026f60>)
```

here is that error

ankurankan:
@khalibartan Did you add that node to the model ?

khalibartan:
Nope , that what I'm asking how can I add nodes. I only added the edges .

ankurankan:
@khalibartan Check add_node method in BayesianModel

khalibartan:
@ankurankan  Okay. Will send you a commit in a few moments. I have also found what cause error in child.bif . Variable states in child.bif are inequalities . I didn't created an expression for parsing it.

khalibartan:
@ankurankan Now you can take a look. Fixed all the mentioned things now.Code is working on child.bif.
 Also this decrease in Health is not due to my code. It is from other modules you can check that.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,340 @@
+import numpy
+from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nums, ZeroOrMore, OneOrMore, cppStyleComment, Literal, printables
+import re
```
ankurankan, line 1: @khalibartan I think the import order according to PEP8 should be builtin, then third party. And add a blank line after each block of imports.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,340 @@
+import numpy
+from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nums, ZeroOrMore, OneOrMore, cppStyleComment, Literal, printables
```
ankurankan: @khalibartan More than 120 chars.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,340 @@
...
+        >>> from pgmpy.readwrite import BIFReader
+        >>> reader = BIF.BifReader("bif_test.bif")
+        >>> reader.network_name()
+        'Dog-Problem'
+        """
+        network_attribute = Suppress('network') + Word(alphanums+'_' + '-') + '{'   # Creating a network attribute 
+        network_name = network_attribute.searchString(self.network)[0][0]
+        return network_name
+
+    class Get_grammar(object):
```
ankurankan: @khalibartan Do we really need a nested class ? And never name classes starting with `get` etc. If you really feel like you need to name a class starting with `get`, `add` etc. then maybe you don't need to define a class.

ankurankan:
@khalibartan pep8 on BIF.py:

``` bash
ankur@ankur-pc:~/pgmpy (pr/512 $%)$ pep8 pgmpy/readwrite/BIF.py --max-line-length=120
pgmpy/readwrite/BIF.py:2:121: E501 line too long (143 > 120 characters)
pgmpy/readwrite/BIF.py:34:121: E501 line too long (124 > 120 characters)
pgmpy/readwrite/BIF.py:64:47: W291 trailing whitespace
pgmpy/readwrite/BIF.py:65:1: W293 blank line contains whitespace
pgmpy/readwrite/BIF.py:78:115: W291 trailing whitespace
pgmpy/readwrite/BIF.py:110:70: E231 missing whitespace after ','
pgmpy/readwrite/BIF.py:110:74: E231 missing whitespace after ','
pgmpy/readwrite/BIF.py:110:78: E231 missing whitespace after ','
pgmpy/readwrite/BIF.py:110:84: W291 trailing whitespace
pgmpy/readwrite/BIF.py:115:35: E127 continuation line over-indented for visual indent
pgmpy/readwrite/BIF.py:122:44: W291 trailing whitespace
pgmpy/readwrite/BIF.py:128:1: W293 blank line contains whitespace
pgmpy/readwrite/BIF.py:137:60: E231 missing whitespace after ','
pgmpy/readwrite/BIF.py:141:65: W291 trailing whitespace
pgmpy/readwrite/BIF.py:143:70: E231 missing whitespace after ','
pgmpy/readwrite/BIF.py:143:74: E231 missing whitespace after ','
pgmpy/readwrite/BIF.py:143:78: E231 missing whitespace after ','
pgmpy/readwrite/BIF.py:146:53: E225 missing whitespace around operator
pgmpy/readwrite/BIF.py:146:58: E225 missing whitespace around operator
pgmpy/readwrite/BIF.py:164:72: W291 trailing whitespace
pgmpy/readwrite/BIF.py:303:22: E127 continuation line over-indented for visual indent
pgmpy/readwrite/BIF.py:310:1: W293 blank line contains whitespace
pgmpy/readwrite/BIF.py:322:1: W293 blank line contains whitespace
pgmpy/readwrite/BIF.py:326:33: E128 continuation line under-indented for visual indent
pgmpy/readwrite/BIF.py:326:41: E251 unexpected spaces around keyword / parameter equals
pgmpy/readwrite/BIF.py:326:43: E251 unexpected spaces around keyword / parameter equals
pgmpy/readwrite/BIF.py:327:33: E128 continuation line under-indented for visual indent
pgmpy/readwrite/BIF.py:327:46: E251 unexpected spaces around keyword / parameter equals
pgmpy/readwrite/BIF.py:327:48: E251 unexpected spaces around keyword / parameter equals
pgmpy/readwrite/BIF.py:328:49: E128 continuation line under-indented for visual indent
```

Please fix all these.

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,340 @@
...
+        >>> reader = BIFReader("bif_test.bif")
+        >>> reader.get_states()
+        {'bowel-problem': ['true','false'],
+         'dog-out': ['true','false'],
+         'family-out': ['true','false'],
+         'hear-bark': ['true','false'],
+         'light-on': ['true','false']}
+        """
+
+        try:
```
ankurankan: @khalibartan Not a very good way to use try and except everywhere. Better do something that makes sure that you have assigned `self.grammar` before using it.

ankurankan:
@khalibartan I am still not happy with the way the grammar is implemented. The grammar is something that just defines the structure of something but doesn't do any parsing. So, the whole idea is to keep the grammar separate from the parsing parts. It will make things much easier in terms of code maintenance.

khalibartan:
@ankurankan Yeah you are right. I also thought this later after committing. I'll send a new commit. But that will take time. My end sem exams are about to begin. So I'll do it later.

khalibartan:
@ankurankan Applied the mentioned changes, there won't be a need for `try` and `except` , because everything is done when class is being initialized like in all other readwrite modules.

khalibartan:
@ankurankan Applied the mentioned changes, there won't be a need for `try` and `except` , because everything is done when class is being initialized like in all other readwrite modules.
Don't know for decrease in health. You can run pep8 . Here are my pep8 results
```
➜  pgmpy git:(dev) ✗ pep8 pgmpy/readwrite/BIF.py --max-line-length=120      
➜  pgmpy git:(dev) ✗ 
```
Are there any other issues, you can mention that, but I cannot assure you that I'll be able to do take a look at them anytime soon. My end sem exams are about to begin.
You can merge this pull request if they aren't I'll add the writer class after my exams are over.

khalibartan:
@ankurankan Now you can take a look. If there are any other issues with the code, mention them. While writing I'm using variables states as variable_name_1 or variable_name_2 (replace variable name by name of variable) instead of 1,2 etc. I have tried the writer and reader classes on nearly all the files given in [bnlearn](http://www.bnlearn.com/bnrepository/) and they worked fine. 
In reader class grammar is defined separately and in writing class I have defined string templates separately for writing to improve readability

pgmpy/readwrite/BIF.py
```diff
@@ -0,0 +1,564 @@
...
+from pgmpy.extern.six.moves import map, range
+
+import numpy
+
+from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nums, ZeroOrMore, OneOrMore,\
+        cppStyleComment, Literal, printables
+
+import re
+
+from string import Template
```
ankurankan: @khalibartan Check for imports in other modules. Don't need so many blank lines

ankurankan:
@khalibartan Can you please spend some time on optimizing this? Reading diabetes network takes like more than 2 mins.

pgmpy/readwrite/BIF.py
khalibartan: @ankurankan You asked me to leave a blank line after imports (pep8 style) [here](https://github.com/pgmpy/pgmpy/pull/512#discussion_r44890536)
Also I couldn't understand

> Check for imports in other modules

Could you explain what you mean by it.

khalibartan:
@ankurankan Yeah I know that, it is quite slow right now, I tried it on [munin2.bif](http://www.bnlearn.com/bnrepository/munin4/munin2.bif.gz) it took 1.5 mins like that, but I couldn't optimize it. Slowest routine are caused by pyparsing, I'm parsing things block by block to optimize its speed. I'll look for optimizations , but at the moment I'm not able to find.

pgmpy/readwrite/BIF.py
ankurankan: @khalibartan I didn't mean to put blank lines after each import. You basically have to categorize the imports into 3 categories (check PEP8) and have a blank line in between each of these three blocks (check other modules).

khalibartan: @ankurankan Okay. I'll change it.

khalibartan:
@ankurankan I ran the reader on diabetes.bif and really it was quite slow on that one took around 4 minutes to run.

 I think I can think of an optimization, I was going through the code and pyparsing expression for creating cpd is quite complex. Its complexity can be reduced. here is an example of probability expression
```
(2_8mmol_kg_h, 0_8mmol_kg_h) 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0;
````
If I reaplce the parenthesis and its inside content by null string before parsing it using pyparsing , I think complexity of expression will be reduced. Can I use regex for this replacement, because pyparsing for this replacement won't optimize it much . You can give me your views I'll implement accordingly, using pyparsing or regex which ever you feel is better.

**Comment by khalibartan** [comment]
@ankurankan I tried some optimizations like checking if property attribute exists or not and then acting accordingly, also the above mentioned change and that too with regex and many other changes like these but run time didn't change. I cannot think of any other optimizations .

**Comment by khalibartan** [comment]
@ankurankan It seems that a routine to remove multiple spaces was extremely slow, It wasn't that necessary also so I removed it. Now it took 97  from previous 217 seconds on diabetes.bif 
Also I'm now using some generators , though they might be slow slightly ( haven't checked ) but they are memory efficient than before .
Can I use multiprocessing for optimizations ?

**Comment by khalibartan** [comment]
@ankurankan It seems that a routine to remove multiple spaces was extremely slow, It wasn't that necessary also so I removed it.Also some change in self.get_network_name() and now it took **61 secs** from previous **217 secs** on **diabetes.bif**
Also I'm now using some _generators_ , so code is _memory efficient_ than before .
I also found that get_cpd() is most time consuming processes and cannot find any useful optimizations for it.
Though I was toying with idea of using split directly but will harm the robustness .
Look into the code, if you find any other issues apart from speed , I'll look dig in. If you still find speed to be slow I cannot really find any optimizations for get_cpd() really .
Also I was able to get significant performance boost by **parallel computing** using **3 pools**. Run time further reduced to **39 secs** on diabetes.bif . Should I Send that patch ? 
Here is a running time of codes on different data sets (these don't include import time of pgmpy)

```
                  parallel  non-parallel
dog-problem.bif   0.036433      0.024985
child.bif         0.190910      0.147521
alarm.bif         0.203751      0.274170
munin.bif         9.905248     12.910666
diabetes.bif     39.361769     59.143942


```

ankurankan:
@khalibartan I am merging this for now. But send a PR for the parallel implementation.

khalibartan:
@ankurankan Okay I'll.

PR #521, opened 2015-11-30
Naive Bayes model

yashu-seth:


ankurankan:
@yashu-seth We don't need all this. It should simply inherit from `BayesianModel` and the user should be able to simply call the `fit` and `predict` method by just specifying the feature vectors and the variable
to predict.

yashu-seth:
@ankurankan I also initially planned to focus only on fit and predict method. But then i thought that there are certain methods that can have almost direct results in Naive Bayes case like for local dependencies we really don't need to find the descendents. So I changed some methods. Please tell me if you feel otherwise.

yashu-seth:
@ankurankan In fit method mostly the BaseEstimator class is being used. Please tell me what changes are required for Naive Bayes.

ankurankan:
@yashu-seth For `NaiveBayes` I wanted an API which is between scikit-learn and what we currently have. So, something like this:

``` python
from pgmpy.models import NaiveBayes
model = NaiveBayes(some_parameters='something')
model.fit(pandas_dataframe, extra_params='something')
model.predict(pandas_dataframe_with_some_missing_column, extra_params='something')
```

So, you will just need to define a `fit` and `predict` method in `NaiveBayes`. `fit` should be doing all the work of creating the model and then calling the `fit` method of `BayesianModel`.

And since we will be creating the model inside the fit method, I am not very sure about having all those extra methods. What do you think?

yashu-seth:
@ankurankan so can't we give the users an option to create a custom Naive Bayes model too? I think fit method can do all this along with an additional option of custom Naive Bayes model too.

ankurankan:
@yashu-seth Can you like share the API that you have in mind? It will be much easier to understand.

yashu-seth:
@ankurankan

``` python
>>> from pgmpy.models import NaiveByaesModel
>>> model = NaiveBayesModel()
>>> model.add_edge('a', 'b')
>>> model.add_edge('a', 'c')
>>> model.active_trail_nodes('b')
{'b', 'a', 'c'}
>>> model.active_trail_nodes('b', 'a')
{'b'}
>>> model.local_independencies('b')
[(b _|_ c | a)]
>>> values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)), columns=['a', 'b', 'c', 'd', 'e'])
```

We have five variables in DataFrame hence the fit method will automatically add the missing two variables.

``` python
>>> models.fit(values, 'a')
>>> model.cpds
[<TabularCPD representing P(e:2 | a:2) at 0x52e2bd0>, <TabularCPD representing P(d:2 | a:2) at 0x52e2cf0>, <TabularCPD representing P(b:2 | a:2) at 0x52e2ef0>, <TabularCPD representing P(a:2) at 0x52e2fb0>, <TabularCPD representing P(c:2 | a:2) at 0x52e2e50>]
```

The NaiveBayesModel inherits from BayesianModel hence a user can use methods like active_trail_nodes, local independencies. I have simplified some of these methods for the Naive Bayes case. 
Edges can be added by the user specifically as well as automatically when fit method is used.

ankurankan:
@yashu-seth This is too much work for a user to create a model. Let do this, inherit from `BayesianModel` and add the methods so that it restricts the user for just creating a `NaiveBayes` model. And make the fit method so that the user can directly do something like this:

``` python
from pgmpy.models import NaiveBayes
model = NaiveBayes()
model.fit(data, params=something)
```

I think `BayesianModel` already supports calling fit in this way (just check once).

yashu-seth:
@ankurankan But in other models too the user has been allowed to create the model this way. It is too much work but then we can keep both the options available. Can't we? 
and if not, then we don't want the user to access methods like active_trail_node and local independencies etc. for NaiveBayesModel and only allow creation of the model. Is this what you are suggesting? Do you want these methods to be particularly restricted? And only let the user have access to predict and fit?

ankurankan:
@yashuseth

But in other models too the user has been allowed to create the model this
way. It is too much work but then we can keep both the options available.
Can't we?

Because Naive Bayes doesn't allow that much flexibility for creating
models. Once you know the feature variables and the variables you want to
predict you can create the model.

and if not, then we don't want the user to access methods like
active_trail_node and local independencies etc. for NaiveBayesModel and
only allow creation of the model. Is this wat you are suggesting? Do you
want these methods to be particularly restricted? And only let the user
have access to predict and fit?
I didn't say we won't be having these methods. Since we are inheriting from
BayesianModel, it will have all these methods (and change these in
NaiveBayes if you can think of any optimizations in the case of Naive
Bayes).

What I meant was we will have all the features of Bayesian Model in Naive
Bayes plus a way to directly fit data (and I think we already have this).

yashu-seth:
@ankurankan okay. I understood. I will do the same.

yashu-seth:
@ankurankan In BayesianModel the fit method takes only pandas dataframe object as a parameter.
In NaiveBayes we are allowing creation of the model with the method too so I think we will have to give an extra parameter of the parent_node to the fit method.

ankurankan:
@yashu-seth Yes. In `BayesianModel` directly calling `fit` would be the case of structure learning. So, yes we will need to pass an extra parameter in the case of NaiveBayes.

yashu-seth:
@ankurankan okay.

yashu-seth:
@ankurankan please have a look at this. I have added the fit method and also some other methods which had simple results for Naive Bayes.

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,233 @@
...
+
+        The nodes u and v will be automatically added if they are
+        not already in the graph
+
+        Parameters
+        ----------
+        u,v : nodes
+              Nodes can be any hashable python object.
+
+        EXAMPLE
```
ankurankan: @yashu-seth Replace with `Examples`

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,233 @@
...
+    True
+    >>> len(G)  # number of nodes in graph
+    3
+    """
+
+    def __init__(self, ebunch=None):
+        super(NaiveBayesModel, self).__init__(ebunch)
+        self.parent_node = None
+        self.children_nodes = []
+        if ebunch:
```
ankurankan: @yashu-seth This doesn't look a very good way of checking for the model structure. Can you think of some other cleaner or optimized way ?

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,233 @@
...
+        EXAMPLE
+        -------
+        >>> from pgmpy.models import NaiveBayesModel
+        >>> G = NaiveBayesModel()
+        >>> G.add_nodes_from(['a', 'b', 'c'])
+        >>> G.add_edge('a', 'b')
+        >>> G.add_edge('a', 'c')
+        """
+
+        if hasattr(self, 'parent_node'):
```
ankurankan: @yashu-seth Why this check ? Wouldn't the object always have `parent_node` attribute since you are defining it in the `__init__` method ?

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,233 @@
...
+        [<TabularCPD representing P(D:2 | A:2) at 0x4b72870>,
+         <TabularCPD representing P(E:2 | A:2) at 0x4bb2150>,
+         <TabularCPD representing P(A:2) at 0x4bb23d0>,
+         <TabularCPD representing P(B:2 | A:2) at 0x4bb24b0>,
+         <TabularCPD representing P(C:2 | A:2) at 0x4bb2750>]
+        >>> model.edges()
+        [('A', 'D'), ('A', 'E'), ('A', 'B'), ('A', 'C')]
+        """
+        if parent_node not in data.columns:
+            raise ValueError("parent node {node} is not present in the given data".format(node=parent_node))
```
ankurankan: @yashu-seth Just add a `:` after `parent node` in the error message

ankurankan:
@yashu-seth This looks fine. Just add the `predict` method and some tests.

yashu-seth:
@ankurankan Should the `predict` method take both the training data as well as the data_to_predict as input so that it can call the `fit` method on its own. Or should we let the user first call the `fit` method and then run the `predict` method? If latter is the case I don't think we will have to add the `predict` method here, it can directly be inherited from the `BayesianModel`.

ankurankan:
@yashu-seth We want it same as that of Bayesian Model. So, we won't need to add it here, my bad.

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,224 @@
...
+    >>> 'a' in G     # check if node in graph
+    True
+    >>> len(G)  # number of nodes in graph
+    3
+    """
+
+    def __init__(self, ebunch=None):
+        self.parent_node = None
+        self.children_nodes = []
+        super(NaiveBayesModel, self).__init__(ebunch)
```
ankurankan: @yashu-seth How are you checking for the structure constraints ?

yashu-seth: @ankurankan 
`__init__` calls `add_edges_from` method which calls  `add_edge` method and I have ensured that the structure constraints are checked in `add_edge` method of `NaiveBayes`.

``` python
>>> from pgmpy.models import  NaiveBayesModel
>>> model = NaiveBayesModel([('a','b'), ('a','c'), ('b', 'c')])
  File "<pyshell#37>", line 1, in <module>
    model = NaiveBayesModel([('a','b'), ('a','c'), ('b', 'c')])
  File "D:\PC\t\naive_test.py", line 79, in __init__
    super(NaiveBayesModel, self).__init__(ebunch)
  File "C:\Python34\lib\site-packages\pgmpy-0.1.0-py3.4.egg\pgmpy\models\BayesianModel.py", line 85, in __init__
    self.add_edges_from(ebunch)
  File "C:\Python34\lib\site-packages\pgmpy-0.1.0-py3.4.egg\pgmpy\base\DirectedGraph.py", line 150, in add_edges_from
    self.add_edge(*edge, **kwargs)
  File "D:\PC\t\naive_test.py", line 103, in add_edge
    raise ValueError("Model can have only one parent node.")
ValueError: Model can have only one parent node.
```

ankurankan:
@yashu-seth Some more minor changes that are needed:
1. I think keeping the name `NaiveBayes` would be sufficient. Don't think there's a need for `NaiveBayesModel`.
2. Add `NaiveBayesModel` to `__init__.py` of `models`, so that we could simply import it like: `from pgmpy.models import NaiveBayesModel`.
3. `parent_node` should be an optional parameter in `fit`. Think about the case where the user creates the model structure, in that case we will already be knowing the `parent_node`.

Rest everything looks fine. Fix these and it should be ready to merge.

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,224 @@
...
+        u,v : nodes
+              Nodes can be any hashable python object.
+
+        Examples
+        --------
+        >>> from pgmpy.models import NaiveBayesModel
+        >>> G = NaiveBayesModel()
+        >>> G.add_nodes_from(['a', 'b', 'c'])
+        >>> G.add_edge('a', 'b')
+        >>> G.add_edge('a', 'c')
```
ankurankan: @yashu-seth Show some outputs also. Like do `G.edges()` etc.

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,224 @@
...
+        super(NaiveBayesModel, self).add_edge(u, v, *kwargs)
+
+    def _get_ancestors_of(self, obs_nodes_list):
+        """
+        Returns a list of all ancestors of all the observed nodes.
+
+        Parameters
+        ----------
+        obs_nodes_list: string, list-type
+            name of all the observed nodes
```
ankurankan: @yashu-seth Add examples here.

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,224 @@
...
+        return set(obs_nodes_list) | set(self.parent_node)
+
+
+    def active_trail_nodes(self, start, observed=None):
+        """
+        Returns all the nodes reachable from start via an active trail.
+
+        Parameters
+        ----------
+
```
ankurankan: @yashu-seth No need for blank line here.

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,224 @@
...
+        start: Graph node
+
+        observed : List of nodes (optional)
+            If given the active trail would be computed assuming these nodes to be observed.
+
+        Examples
+        --------
+
```
ankurankan: Same here

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,224 @@
...
+        """
+        Computes the CPD for each node from a given data in the form of a pandas dataframe.
+        If a variable from the data is not present in the model, it adds that node into the model. 
+
+        Parameters
+        ----------
+        data : pandas DataFrame object
+            A DataFrame object with column names same as the variable names of network
+
+        parent_node: str
```
ankurankan: @yashu-seth `parent_node` can be any hashable python object not just `str`

pgmpy/models/NaiveBayesModel.py
```diff
@@ -0,0 +1,224 @@
...
+            parent node of the model
+
+        estimator: Estimator class
+            Any pgmpy estimator. If nothing is specified, the default Maximum Likelihood
```
ankurankan: @yashu-seth Replace `Maximum Likelihood estimator` with "`MaximumLikelihoodEstimator`" (notice the backtick). Using backtick will put the name in code snippet in out online docs.

pgmpy/models/NaiveBayesModel.py
yashu-seth: @ankurankan I think the `_get_ancestors_of` method is not meant for the users. In `BayesianModel` too I found that this method does not actually return the ancestors, it returns a set of node and ancestors. Should the examples be put here?

ankurankan:
@yashu-seth Sorry, I forgot to mention about tests. Please add tests for this.

yashu-seth:
@ankurankan I have added the necessary tests.

pgmpy/tests/test_models/test_NaiveBayes.py
```diff
@@ -0,0 +1,193 @@
+import unittest
+import networkx as nx
+import pandas as pd
```
ankurankan, line 4: @yashu-seth Fix the import order according to PEP8

pgmpy/tests/test_models/test_NaiveBayes.py
```diff
@@ -0,0 +1,193 @@
...
+                            [Independencies(['b', ['e','c','d'], 'a'])])
+        self.assertListEqual(self.G.local_independencies('c'), 
+                            [Independencies(['c', ['e','b','d'], 'a'])])
+        self.assertListEqual(self.G.local_independencies('d'),
+                            [Independencies(['d', ['b','c','e'], 'a'])])
+        
+    def tearDown(self):
+        del self.G
+        
+class TestNaiveBayes_active_trail_nodes(unittest.TestCase):
```
ankurankan: Can't you put these in `TestNaiveBayesMethods`  since it's `active_trail_nodes`?

pgmpy/tests/test_models/test_NaiveBayes.py
```diff
@@ -0,0 +1,193 @@
...
+        values2 = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 3)),
+                                            columns=['C', 'D', 'E'])
+
+        self.assertRaises(ValueError, self.model1.fit, values)
+        self.assertRaises(ValueError, self.model1.fit, values2)
+        self.assertRaises(ValueError, self.model2.fit, values2, 'A')
+
+    def tearDown(self):
+        del self.model1
+        del self.model2
```
ankurankan: @yashu-seth PEP8. Leave a blank line at the end of the file.

PR #532, opened 2015-12-11
Hash function for IndepedenceAssertion class

### LINKED ISSUE #531: Making IndependenceAssertion object hashable
The current `IndepedenceAssertion` object has the `__eq__` method and adding a suitable `__hash__` function is needed in order to make it a hashable object.

kshitij10496:
- [x] Added `__hash__()` to the `IndependenceAssertion` class.
- [x] Added `__hash__()` to the `Independencies` class.  
  Fixes #531  
  @ankurankan.
  Have a look into this.  
  I checked the [Factor](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/factors/Factor.py#L747-L749) class for a similar `__hash__()` and found that there were no tests implemented for this function.
  So, just let me know if we need to add tests here.

### DIFF
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -197,6 +197,10 @@ class IndependenceAssertion(object):
     def __eq__(self, other):
         return self.get_assertion() == other.get_assertion()
 
+    def __hash__(self):
+        return hash(str(sorted(self.event1)) + str(sorted(self.event2)) +
+                    str(sorted(self.event3)))
+
     @staticmethod
     def _return_list_if_str(event):
         """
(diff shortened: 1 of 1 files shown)


pgmpy/independencies/Independencies.py
```diff
@@ -55,6 +55,9 @@ def __eq__(self, other):
         other_assertions = other.get_assertions()
         return all(self_independency in other_assertions for self_independency in self.get_assertions())
 
+    def __hash__(self):
+        return sum(hash(assertion) for assertion in self.get_assertions())
```
ankurankan, line 59: @kshitij10496 I am not very sure this will be unique in each case. Why don't you simply convert `self.assertions` to a `str` and return the hash of that string.

kshitij10496, line 59: @ankurankan Consider the case of two `Independencies` objects with same `assertions` but these `assertions` are misordered. The `str` will result in different hashes for both.

PR #533, opened 2015-12-11
Fixed issue where networkx structure was not updated remove_cpds after cpds are r…

Erotemic:
There was an issue where passing a model to an Inference structure failed after model.remove_cpds was called. This was because the .node and .edge attributes of the networkx base class are not updated accordingly. Thus clique computation fails.


Erotemic:
Closing because I had a misconception about what this function was trying to do.

Issue #525, opened 2015-12-02
add more independencies in get_independencies of Bayesian Model


kshitij10496:
The `get_independencies` method should compute all the independencies present in the Bayesian network.  It should also include the local independencies given by the `local_independencies` as well.

```
[In] : from pgmpy.models import BayesianModel
[In] : G = BayesianModel()
[In] : G.add_edges_from([('D','G'), ('I','G'), ('G','L'), ('I','S')])
[In] : G.get_independencies()
[Out]: (S _|_ I, D | G)
           (S _|_ G, I, L | D)
           (S _|_ G, I, D | L)
           (G _|_ I, D, L | S)
           (G _|_ D, L | I)
           (G _|_ S, I, L | D)
           (G _|_ S, I, D | L)
           (I _|_ G, L | S)
           (I _|_ S, D | G)
           (I _|_ S, G, L | D)
           (I _|_ S, G, D | L)
           (D _|_ G, L | S)
           (D _|_ S, I | G)
           (D _|_ G, L | I)
           (D _|_ S, G, I | L)
           (L _|_ G, I, D | S)
           (L _|_ G, D | I)
           (L _|_ S, G, I | D)  
[In] : G.local_independencies('S')  # not present in `get_independencies`
[Out]: (S _|_ G, D, L | I)
```

kshitij10496:
I am working on this

chrisittner:
See #607. Once the d-seperation check in `get_independencies` works as it should, local independencies will be included automatically. This issue can possibly be closed already.

PR #536, opened 2015-12-12
Added Independencies to get_independencies

### LINKED ISSUE #523: add tests for get_independencies method in BayesianModel

kshitij10496:
- [x] added local independencies to `get_independencies`
- [ ] added tests

fixes #523 #525

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -114,7 +114,7 @@ class BayesianModel(DirectedGraph):
                  'Loops are not allowed. Adding the edge from (%s->%s) forms a loop.' % (u, v))
         else:
             super(BayesianModel, self).add_edge(u, v, **kwargs)
-        
+
     def add_cpds(self, *cpds):
         """
         Add CPD (Conditional Probability Distribution) to the Bayesian Model.
@@ -442,6 +442,10 @@ class BayesianModel(DirectedGraph):
                     if independent_variables:
                         independencies.add_assertions([start, independent_variables,
                                                        observed])
+            local_independence = self.local_independencies(start)
+            for item in set(local_independence.get_assertions()):
+                if item not in set(independencies.get_assertions()):
+                    independencies.add_assertions(item)
 
         independencies.reduce()
 
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -57,6 +57,9 @@ class Independencies(object):
         other_assertions = other.get_assertions()
         return all(self_independency in other_assertions for self_independency in self.get_assertions())
 
+    def __hash__(self):
+        return sum(hash(assertion) for assertion in self.get_assertions())
+
     def get_assertions(self):
         """
         Returns the independencies object which is a set of IndependenceAssertion objects.
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -206,6 +206,10 @@ class IndependenceAssertion(object):
         return all(sorted(self_event)==sorted(other_event) for self_event,other_event
                     in zip(self_assertions,other_assertions))
 
+    def __hash__(self):
+        return hash(str(sorted(self.event1)) + str(sorted(self.event2)) +
+                    str(sorted(self.event3)))
+
     @staticmethod
     def _return_list_if_str(event):
         """
--- a/pgmpy/tests/test_models/test_BayesianModel.py
+++ b/pgmpy/tests/test_models/test_BayesianModel.py
@@ -127,6 +127,19 @@ class TestBayesianModelMethods(unittest.TestCase):
         self.assertEqual(self.G.local_independencies('e'), Independencies(['e',['c','b','a'],'d']))
         self.assertEqual(self.G.local_independencies('b'), Independencies(['b','a']))
 
+    def test_get_independencies(self):
+        test_independencies = Independencies(['a', ('b', 'c'), 'd'], ['a', ('d', 'e'), 'b'], ['a', ('e', 'd'), 'c'],
+                                             ['a', ('b', 'c', 'd'), 'e'], ['a', 'b', 'c'], ['b', ('a', 'c'), 'd'],
+                                             ['b', ('d', 'e'), 'c'], ['b', ('c', 'e', 'd'), 'a'],
+                                             ['b', ('c', 'd', 'a'), 'e'], ['b', 'a'], ['c', ('a', 'b'), 'd'],
+                                             ['c', ('b', 'e', 'd'), 'a'], ['c', ('b', 'd', 'a'), 'e'],
+                                             ['c', ('a', 'd', 'e'), 'b'], ['d', ('a', 'e', 'b'), 'c'],
+                                             ['d', ('a', 'e'), 'b'], ['d', ('b', 'e', 'c'), 'a'],
+                                             ['d', ('a', 'b', 'c'), 'e'], ['d', 'c', ('b', 'a')],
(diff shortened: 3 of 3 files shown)


pgmpy/models/BayesianModel.py
```diff
@@ -442,6 +442,12 @@ def get_independencies(self, latex=False):
                     if independent_variables:
                         independencies.add_assertions([start, independent_variables,
                                                        observed])
+            local_independence = self.local_independencies(start)
+            common_assertions = set(local_independence.get_assertions())
+            & set(independencies.get_assertions())
```
khalibartan: @kshitij10496 This is not a valid syntax. travis CI was failing because of this.

kshitij10496: @khalibartan Thanks. I have made an amend

ankurankan:
@kshitij10496 Let me know when this is ready for review. And also changes in #532 overlap with the changes in this PR. Why don't you close one ?

kshitij10496:
@ankurankan I have closed #532 .
I would like to discuss on the `hash` function

ankurankan:
@kshitij10496 What about the `hash` function?

kshitij10496:
I wanted to discuss about the `hash` function in the `Independencies` class.
You had suggested in the other PR that we should use `str` instead.

ankurankan:
@kshitij10496 Yes, I meant use the inbuilt `hash` function on some unique string. So, you will have to have a way of constructing a string so that it's unique for each different object and then use the `hash` function on that.

kshitij10496:
I tried to hash the object using its `__str__` before hand.
But the hash would depend on the order of assertions in the `Independencies` object.

Hence, I used adding the hash of each assertion in the object.
I know, this might not be the best way to hash but the probability of collisions is very small.

Do consider it and any advice on how can we do better ?

khalibartan:
@kshitij10496 Could you replace `_|_` in `__str__` method by unicode `u"\u27C2"` . It is unicode for perpendicular ;) . Also the hash function using addition has good chances of collision and `__str__` will depend on order. Since get_assertion returns a list we can sort that list and convert it into a tuple and apply a tuple hash on it (built in tuple hash). Or is there a catch with this hash method ? But I don't think there is .

kshitij10496:
@khalibartan Sounds cool.
If @ankurankan has no problem with it, I would be happy to make the necessary changes.

kshitij10496:
@khalibartan What are your thoughts on the `hash` function in the `Independencies` class ?

khalibartan:
@kshitij10496 I think it is good, we have already discussed about this on issue . It is slow , apart from it there is no issue . If I find any better implementation will definitely tell you . 

Also can you take a look at the local_independencies in BayesianModel , since you are working on independecies class there might be some changes that we need in it .

kshitij10496:
@khalibartan About the `hash` function:
Initially, I attempted that approach.
However, I was unable to `sort` the list of `independencies` of the obtained by the `get_assertions()` method of the `Independencies` object.

khalibartan:
@kshitij10496 Yeah It would happen actually since, it is a list of IndependenceAssertion objects. To sort it you have to do a bit of hard work, by iterating over each assertion object and converting them to inbuilt data type that can be sorted (str maybe) and then apply the method .
But now method seems a little vague , but it is as good collision free as inbuilt hash .
I'm thinking of some new method but you can try this method for now .

khalibartan:
Actually I couldn't understand why we need `__hash__` for independencies , since self.independencies is a list of IndependenceAssertion , and IndependenceAssertion objects are hashable(after adding `__hash__` method to it , so can't we directly convert self.independencies to a set ?

kshitij10496:
@khalibartan Yeah.
We can do that obviously.
For this issue, we dont need to hash the `Independencies` object.
However, I think we should add a `hash` function for code completeness.

khalibartan:
I doesn't really makes sense to me. If you want to do so , you can go with the method I have [explained](https://github.com/pgmpy/pgmpy/pull/536#issuecomment-166674959).

kshitij10496:
Basically, our motive is to add `hash` function and make the `Independencies` object hashabe.
Along the way, we want to solve the related issues.
And not the other way round.

khalibartan:
Okay, things could have worked without it . So simply you can return the hash value of the by converting self.independencies to a set.

kshitij10496:
I am getting your point here.
Are you suggesting to hash a set ?

khalibartan:
Hmm.. interesting. I always thought `sets` were **hashable** , but actually `frozensets` are, so frozensets instead.

kshitij10496:
Exactly !
Hence, I find it better to add a `hash` function

ankurankan:
@kshitij10496 @khalibartan What's the status of this PR ?

kshitij10496:
@ankurankan If you are fine with the methods, I will add the necessary tests.

ankurankan:
@kshitij10496 Yes, please add the tests.

kshitij10496:
Cool.

I am adding them :)

khalibartan:
@kshitij10496 Tests are failing please give it a look.

khalibartan:
@kshitij10496 Can fix the failing test. I want to work with these changes .

kshitij10496:
@khalibartan I am looking into them.

kshitij10496:
I am not entirely clear why the tests are failing.
The code runs perfectly fine on the local.

khalibartan:
Try creating a separate `conda-env` for pgmpy and then run the tests. They are falling. I pulled in you branch and ran nosetests they failed . I haven't gone through the changes but I believe that order of assertions might be cause of failures, but it shouldn't have happened .

kshitij10496:
Yeah. 
The tests are failing locally
However, the result seems to be correct.

Even I think the tests are failing either because of the order of assertions or the order of items in `event` of the assertions.

khalibartan:
@kshitij10496 I have send PR to resolve your merge conflicts . You can pull changes of `pgmpy-dev` into your branch and your tests will pass then . :)
I found a bug in `__eq__` method. It was taking order into account.

ankurankan:
@kshitij10496 @khalibartan Don't merge dev into this branch. Rather rebase your commits over the current dev. This will make the git history linear and much more cleaner.

khalibartan:
Okay . My bad.

khalibartan:
@kshitij10496 are you still working on this ?

kshitij10496:
@khalibartan Yeah
I will fix this by tomorrow

khalibartan:
@kshitij10496  Can you add a new implementation of the things, in IndependenceAssertion class by converting events from sets to frozensets and then returning their hash value. I did opened an issue for this but closed it later #575 .

kshitij10496:
@khalibartan Yeah. Sure
I read that issue

khalibartan:
@kshitij10496 Can you fix the merge conflicts ? I want to work with is_iequivalence class in BayesianModel

ankurankan:
@kshitij10496 Tests are failing even after resolving the merge conflicts. Can you check?

khalibartan:
@kshitij10496 are you still working on it ?

yashu-seth:
@kshitij10496 Are you still working on this? The PR has been around for some time now. If you are unable to work on this for now, kindly let us know.

chrisittner:
Regarding the change in `get_independencies`: I think `get_independencies` currently tries to compute all independencies by checking active trails/d-seperation. If done correctly, this should already include any dependencies found by `local_dependencies`, d-sep is a more general method. Thus adding the local independencies for each variable should be redundant once the issues (#607) in the algorithm are fixed.

khalibartan:
@ankurankan I think you can close this PR , it has been here for quite a long time.

PR #539, opened 2015-12-13
Fixed bug in factor reduce, marginalize, and maximize where the cardi…

Erotemic:
…nalities and values became missaligned. Added corresponding tests.

I encountered an error where calling reduce on a Factor would produce unexpected results. I tracked the issue down to the ordering of index_to_keep in the reduce method. This variable is defined as the result of a set difference. Set difference in Python is not guaranteed to maintain any specific ordering, but often times it will return an ordered list. In the case that it doesn't phi.variables and phi.cardinality will  be in a different order than phi.values. 

This problem also existed in maximize and marginalize. 

To fix the issue I simply sorted the output of the set difference. 

I added an example of a factor that caused this issue to pgmpy/tests/test_factors/test_Factor.py
With the fix the new test passes for each case.

### DIFF
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -248,6 +248,7 @@ class Factor(object):
         var_indexes = [phi.variables.index(var) for var in variables]
 
         index_to_keep = list(set(range(len(self.variables))) - set(var_indexes))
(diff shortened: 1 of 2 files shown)


ankurankan:
@Erotemic Thanks for the fix.

Issue #538, opened 2015-12-13
Confusing error in fit(): IndexError index 0 is out of bounds


kmbnw:
EDIT: I am using the book/v0.1 branch.

My apologies if this is in the bug list somewhere; I looked through the active bugs and searched, but I couldn't find a duplicate.

I'm experimenting with the examples from the book "Mastering Probabilistic Graphical Models in Python".  In setting up what seemed like some fairly simple models, I discovered what looks like a bug (occurs on both Python 2.7 and Python 3.4, on Ubuntu 14.04 LTS).  

I could be specifying the models wrong, but the error is not informative enough for me to tell what might be happening.

The following code works:

``` python
data = np.random.randint(low=0, high=2, size=(1000, 3))
data = pd.DataFrame(data, columns=[
    'cost',
    'quality', 
    'location'
    ])
train = data[:750]
restaurant_model = BayesianModel([
        ('location', 'cost')
    ])
restaurant_model.check_model()
restaurant_model.fit(train)
```

But when I remove the unused 'quality' column, I get an error:

``` python
data = np.random.randint(low=0, high=2, size=(1000, 2))
data = pd.DataFrame(data, columns=[
    'cost',
   # 'quality', 
    'location'
    ])
train = data[:750]
restaurant_model = BayesianModel([
        ('location', 'cost')
    ])
restaurant_model.check_model()
restaurant_model.fit(train)

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-31-fb9099559416> in <module>()
      6     ])
      7 restaurant_model.check_model()
----> 8 restaurant_model.fit(train)

/home/krystab/virtualenvs/compmodel3/lib/python3.4/site-packages/pgmpy-0.1.0-py3.4.egg/pgmpy/models/BayesianModel.py in fit(self, data, estimator_type)
    555         estimator = estimator_type(self, data)
    556 
--> 557         cpds_list = estimator.get_parameters()
    558         self.add_cpds(*cpds_list)
    559 

/home/krystab/virtualenvs/compmodel3/lib/python3.4/site-packages/pgmpy-0.1.0-py3.4.egg/pgmpy/estimators/MLE.py in get_parameters(self)
     70                 var_card = self.node_card[node]
     71                 state_counts = self.data.groupby([node] + self.model.predecessors(node)).count()
---> 72                 values = state_counts.iloc[:, 0].reshape(var_card,
     73                                                          np.product(parent_card))
[...]

PR #541, opened 2015-12-13
fit method bug

yashu-seth:


yashu-seth:
@ankurankan I will add the tests for this soon.

ankurankan:
@yashu-seth Can you please describe the bug ?

yashu-seth:
@ankurankan 
The failing case has been described in issue #538. In MLE.py 
what happens is in line 71 groupby method is used follwed by a count method. But  `df.groupby (list).count ()` gives an empty dataframe whenever `list` contains all the column variables. Hence line 72 throws an error because of this empty datarame.

pgmpy/estimators/MLE.py
```diff
@@ -68,10 +68,8 @@ def get_parameters(self):
             else:
                 parent_card = np.array([self.node_card[parent] for parent in parents])
                 var_card = self.node_card[node]
-                state_counts = self.data.groupby([node] + self.model.predecessors(node)).count()
-                values = state_counts.iloc[:, 0].reshape(var_card,
-                                                         np.product(parent_card))
-                cpd = TabularCPD(node, var_card, values,
+                values = self.data.groupby([node] + parents).size().unstack(parents).fillna(0)
```
ankurankan, line 71: @yashu-seth I can't think of a case when we will get a `NaN`. Do you have any case in mind ? Otherwise we won't be needing the `fillna` call.

yashu-seth:
@ankurankan We are generating the rows randomly. There can be a possibility where a particular row is not generated at all. Although it would be extremely rare with large number of columns but in that case `unstack` would give `NaN`. Or maybe when the dataframe is not being generated randomly.

``` python
>>> values = pd.DataFrame(np.random.randint(low=0, high=2, size=(5, 3)),columns=['A', 'B', 'C'])
>>> values
   A  B  C
0  0  0  0
1  0  0  1
2  0  1  1
3  1  1  0
4  1  0  1
>>> values.groupby(['A','B','C']).size().unstack(['A','B'])
A  0       1    
B  0   1   0   1
C               
0  1 NaN NaN   1
1  1   1   1 NaN
>>> values.groupby(['A','B','C']).size().unstack(['A','B']).fillna(0)
A  0     1   
B  0  1  0  1
C            
0  1  0  0  1
1  1  1  1  0
```

pgmpy/estimators/MLE.py
```diff
@@ -62,16 +62,14 @@ def get_parameters(self):
             if not parents:
                 state_counts = self.data.ix[:, node].value_counts()
                 cpd = TabularCPD(node, self.node_card[node],
-                                 state_counts.values[:, np.newaxis])
+                                 np.array(state_counts))
```
yashu-seth: @ankurankan I think the values can be directly converted into numpy arrays.

yashu-seth:
@ankurankan
For line 63,

``` python
>>> data = pd.DataFrame(columns=('a', 'b', 'c', 'd'))
>>> for i in range(250):
           data3.loc[i] = [0,1,0,0]
>>> for i in range(250,500):
           data3.loc[i] = [1,1,1,0]
>>> for i in range(500,1000):
           data3.loc[i] = [1,0,0,0]
>>> data3.ix[:,'a'].value_counts()
1    750
0    250
dtype: int64
>>> data3.ix[:,'c'].value_counts()
0    750
1    250
dtype: int64
```

When we convert these to numpy arrays both give the same numpy array - `array([[750], [250]], dtype=int64)`. Shouldn't this be a problem?

ankurankan:
@yashu-seth I get the issue with `unstack` now. For the second issue I have done reindexing to sort the index.

And I am not really happy with the way things are now. And I am thinking of moving to a pure numpy solution rather than using pandas. Would you like to work on this ?

yashu-seth:
@ankurankan yes the same problem is with unstack as well. I will look for a better implementation.

PR #542, opened 2015-12-14
Parallel implementation of get_cpds()

khalibartan:
@ankurankan You can take a look at this, though implementation isn't that good, but gets work done in time.

### DIFF
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -8,6 +8,24 @@ from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nu
 from pgmpy.models import BayesianModel
 from pgmpy.factors import TabularCPD
 from pgmpy.extern.six.moves import map, range
+from multiprocessing import Pool
+
+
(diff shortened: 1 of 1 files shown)


khalibartan:
@ankurankan Why coverage has decreased, It shows that call hasn't been made to search_cpds . Is this because the function was being run in pool

khalibartan:
@ankurankan Can you suggest me a way on how to terminate the pool. I just ran the code and found pools are still active in the background, searched for solutions but couldn't find one.

ankurankan:
@khalibartan Have a look at joblib (https://pypi.python.org/pypi/joblib). It's much simpler to work with. And also add an optional parameter `n_jobs` in the methods so that the user can specify on how many cores does he want to run it on.

khalibartan:
@ankurankan I'll look into it, but should I send a new patch because using joblib will add an extra dependency .

ankurankan:
@khalibartan Yeah you can.

khalibartan:
@ankurankan Cool then, will send a patch.

ankurankan:
@khalibartan I am closing this as you will be sending a new PR

PR #550, opened 2015-12-23
Adds naive bayes to inference and MLE

yashu-seth:


pgmpy/estimators/MLE.py
```diff
@@ -28,7 +29,7 @@ class MaximumLikelihoodEstimator(BaseEstimator):
     >>> estimator = MaximumLikelihoodEstimator(model, values)
     """
     def __init__(self, model, data):
-        if not isinstance(model, BayesianModel):
+        if not isinstance(model, (BayesianModel, NaiveBayes)):
```
ankurankan: @yashu-seth I don't think you will need to add `NaiveBayes` here. Since you are inheriting `NaiveBayes` from `BayesianModel`, isinstance will return true even if you compare it with `BayesianModel`.

``` python
In [5]: model = NaiveBayes([('A', 'B'), ('A', 'C')])

In [6]: model
Out[6]: <pgmpy.models.NaiveBayes.NaiveBayes at 0x7efc7f9496a0>

In [7]: isinstance(model, BayesianModel)
Out[7]: True
```

yashu-seth: @ankurankan Yes my bad. I'll remove the unnecessary checks.

PR #562, opened 2016-01-06
Added get_ancestors_of method to Bayesian Model

### LINKED ISSUE #522: add tests for _get_ancestors_of method in BayesianModel

### LINKED ISSUE #561: _get_ancestors_of in BayesianModel includes the concerned node
``` python
In [ ]: from pgmpy.models import BayesianModel

In [ ]: G = BayesianModel([('D', 'G'), ('I', 'G'), ('I', 'S'), ('G', 'L')])

In [ ]: G._get_ancestors_of('L')
Out[ ]: {'D', 'G', 'I', 'L'}

In [ ]: G._get_ancestors_of('G')
Out[ ]: {'D', 'G', 'I'}

In [ ]: G._get_ancestors_of('I')
Out[ ]: {'I'}
```

A node cannot be its own child or ancestor.

I will try to fix this and write the tests.

kshitij10496:
- [x] fixes #561 
- [ ] added tests
- [ ] closes #522

The current implementation is a bit misleading when finding the ancestors of a list of nodes.
For example:

``` python
 In[ ]: G._get_ancestors_of(['G', 'I'])
Out[ ]: {'D', 'G', 'I'}      # not specifying the ancestors of each node separately
```

However, I think it would be better if we can specify the ancestors of each node.

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -114,7 +114,7 @@ class BayesianModel(DirectedGraph):
                  'Loops are not allowed. Adding the edge from (%s->%s) forms a loop.' % (u, v))
         else:
             super(BayesianModel, self).add_edge(u, v, **kwargs)
-        
+
     def add_cpds(self, *cpds):
         """
         Add CPD (Conditional Probability Distribution) to the Bayesian Model.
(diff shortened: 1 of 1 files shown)


ankurankan:
@kshitij10496 I don't see the need of adding another method since we won't be using it anywhere. A better option is to just update the docstring of the existing method.

kshitij10496:
@ankurankan Ideally, a node should not be considered its own ancestor.
However, the `_get_ancestors_of` method includes the concerned node.
If you think this method won't be of much use, we can scrape it and make the docstrings more specific to refer to this behaviour.

ankurankan:
Fixed in f18b86d

PR #563, opened 2016-01-08
Fixed bug in __eq__ in IndependenceAssertion

khalibartan:
The current(after this PR) method doesn't take into account the order, that means
event1 , event2, event3 is equal to event2, event1, event3 which is the
correct behaviour . Also added the`__ne__` method.
Previous behaviour 

```
>>> from pgmpy.independencies import Independencies
>>> a = Independencies(['a', ('b', 'c'), 'd'], ['a', ('e', 'd'), 'b'])
>>> b = Independencies(['a', ('b', 'c'), 'd'], [('d', 'e'), 'a', 'b'])
>>> a == b
False
```

After this PR

```
>>> from pgmpy.independencies import Independencies
>>> a = Independencies(['a', ('b', 'c'), 'd'], ['a', ('e', 'd'), 'b'])
>>> b = Independencies(['a', ('b', 'c'), 'd'], [('d', 'e'), 'a', 'b'])
>>> a == b
True
```

Ping @ankurankan .

### DIFF
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -206,8 +206,9 @@ class IndependenceAssertion(object):
         other_assertions = other.get_assertion()
         if len(self_assertions) != len(other_assertions):
             return False
-        if (sorted(self_assertions[:2], key=lambda x: list(x)) == sorted(other_assertions[:2], key=lambda x: list(x))
-                and sorted(self_assertions[2]) == sorted(other_assertions[2])):
+        self_assertions = set(six.moves.map(frozenset, self_assertions))
+        other_assertions = set(six.moves.map(frozenset, other_assertions))
+        if self_assertions == other_assertions:
             return True
         return False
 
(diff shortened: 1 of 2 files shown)


khalibartan:
@ankurankan Don't know for the failing tests in python 3.4.3 for travis. In my system it has passes. Also Travis is passing for python 3.3 and python 2.7 . I even ran nosetests and even specifically ran the failing example, but it passed .

ankurankan:
@khalibartan Can you please check why coverage is decreasing and add tests for that ?

khalibartan:
@ankurankan I have added the necessary test. Problem is that in python 3 `__ne__` is not being called though in a test I have used `!=` .

khalibartan:
@ankurankan it look like `tearUp` is not being called. I'll remove that method in tests.

ankurankan:
@khalibartan Because it should be `tearDown` not `tearUp`.

khalibartan:
@ankurankan Okay. I just added the lines in the method. Will send a new commit in a moment .

khalibartan:
@ankurankan Now I think it is ready for merge.

Issue #353, opened 2015-03-28
Add method in TabularCPD to reorder the parent's order


ankurankan:

kislayabhi:
Please add some more info. I intend to work on this.

palashahuja:
@kislayabhi Suppose if we have a tabular CPD like this 

```
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_0 | C_0 | C_0 | C_1 | C_1 | C_1 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_0 | B_1 | B_1 | B_0 | B_0 | B_1 | B_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| A   | A_0 | A_1 | A_0 | A_1 | A_0 | A_1 | A_0 | A_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| J_0 | 0.9 | 0.3 | 0.9 | 0.3 | 0.8 | 0.8 | 0.4 | 0.4 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| J_1 | 0.1 | 0.7 | 0.1 | 0.7 | 0.2 | 0.2 | 0.6 | 0.6 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
```

then there should be a separate method to switch the tuples inside the tables(i.e for example if the user wants to switch B and A inside the table such that B is the most rapidly changing value)
Keep in mind the values also change accordingly. After switching A and B inside the table it should look something like this.

```
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_0 | C_0 | C_0 | C_1 | C_1 | C_1 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| A   | A_0 | A_0 | A_1 | A_1 | A_0 | A_0 | A_1 | A_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_1 | B_0 | B_1 | B_0 | B_1 | B_0 | B_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| J_0 | 0.9 | 0.9 | 0.3 | 0.3 | 0.8 | 0.4 | 0.8 | 0.4 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
[...]

PR #576, opened 2016-01-15
reorder_parents method and its tests implemented

anaviltripathi:
This implements #353 and its tests. I have implemented the basic structure but I have not yet completely written the docstring. If the method looks okay I will send a separate PR for that.

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 """Contains the different formats of CPDs used in PGM"""
+
 from __future__ import division
 
 from itertools import product
@@ -11,6 +12,7 @@ from pgmpy.factors import Factor
 from pgmpy.extern import tabulate
 from pgmpy.extern import six
 from pgmpy.extern.six.moves import range, zip
+from warnings import warn
 
 
 class TabularCPD(Factor):
@@ -367,6 +369,51 @@ class TabularCPD(Factor):
         <Factor representing phi(grade:3, evi1:2) at 0x7f847a4f2d68>
         """
         return Factor(self.variables, self.cardinality, self.values)
+
+
+    def reorder_parents(self,new_order,inplace = False):
+        '''
+
+        Parameters
+        ----------
+        new_order: list
+            list of new ordering of variables
+        inplace: boolean
+            If inplace == True it will modify the CPD itself
+            otherwise new value will be returned without affecting old values
+
+        Returns
+        -------
+
+        '''
+        if not self.evidence or (set(new_order) - set(self.evidence)) or (set(self.evidence) - set(new_order)):
+            raise ValueError("New order either has missing or extra arguments")
+        else:
+            # No need to do anything if the user provides the same order
+            if new_order != self.evidence[::-1]:
+
+                # Find cardinality mapping of variables
+                card_map = dict(zip(self.evidence, self.evidence_card))
+                # Find old position mapping of variables
+                old_pos_map = dict(zip(self.evidence, range(len(self.evidence))))
+                # Create new ordering of variables based on the given order and previous order
+                trans_ord = [0]+[(old_pos_map[letter]+1) for letter in new_order][::-1]
+                # Find the actual transpose
+                new_values = np.transpose(self.values, trans_ord)
+                # Update old values if it is inplace
+                if inplace:
+                    self.evidence[:] = new_order[::-1]
+                    self.evidence_card = [card_map[var] for var in new_order[::-1]]
+                    variables = [self.variables[0]] + self.evidence
+                    cardinality = [self.variable_card] + self.evidence_card
+                    super(TabularCPD, self).__init__(variables, cardinality, new_values.flatten('C'))
+                    return self.get_cpd()
+                else:
+                    return new_values.reshape(self.cardinality[0], np.prod([card_map[var] for var in new_order[::-1]]))
+                    #return new_values.reshape(self.variable_card, [card_map[var] for var in new_order[::-1]])
+            else:
+                warn("Same ordering provided as current")
+                return self.get_cpd()
 
 
 # Commenting out because not used anywhere for now and not implemented in a very good way.
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -2,7 +2,7 @@
 """Contains the different formats of CPDs used in PGM"""
 
 from __future__ import division
-
+from warnings import warn
 from itertools import product
 
 import numpy as np
@@ -12,7 +12,6 @@ from pgmpy.factors import Factor
 from pgmpy.extern import tabulate
 from pgmpy.extern import six
 from pgmpy.extern.six.moves import range, zip
-from warnings import warn
 
 
 class TabularCPD(Factor):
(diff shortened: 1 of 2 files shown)


ankurankan:
@anaviltripathi The coverage (https://coveralls.io/builds/4738098) is decreasing because of this PR. Can you please check what lines of code are not getting hit by the tests ? And if possible add a few more tests covering those.

pgmpy/factors/CPD.py
```diff
@@ -11,6 +12,7 @@
 from pgmpy.extern import tabulate
 from pgmpy.extern import six
 from pgmpy.extern.six.moves import range, zip
+from warnings import warn
```
ankurankan: @anaviltripathi Please refer to PEP8 (https://www.python.org/dev/peps/pep-0008/#imports) for how to group import statements. `warnings` is a standard library import so should be in the top group.

pgmpy/factors/CPD.py
```diff
@@ -369,6 +371,51 @@ def to_factor(self):
         return Factor(self.variables, self.cardinality, self.values)
 
 
+    def reorder_parents(self,new_order,inplace = False):
+        '''
+
```
ankurankan, line 375: @anaviltripathi Please refer to docstrings of other methods and try to follow that exact same format. Since our online docs are automatically generated from the docstrings, changing the format here will mess up the docs. Also please add an examples section and add a few lines describing what this method does.

pgmpy/factors/CPD.py
```diff
@@ -369,6 +371,51 @@ def to_factor(self):
...
+        else:
+            # No need to do anything if the user provides the same order
+            if new_order != self.evidence[::-1]:
+
+                # Find cardinality mapping of variables
+                card_map = dict(zip(self.evidence, self.evidence_card))
+                # Find old position mapping of variables
+                old_pos_map = dict(zip(self.evidence, range(len(self.evidence))))
+                # Create new ordering of variables based on the given order and previous order
+                trans_ord = [0]+[(old_pos_map[letter]+1) for letter in new_order][::-1]
```
ankurankan, line 412: PEP8. You can install pep8 (https://pypi.python.org/pypi/pep8/1.6.2) and try to run it on the file. It will show all the pep8 issues in the file.

pgmpy/factors/CPD.py
```diff
@@ -369,6 +371,51 @@ def to_factor(self):
...
+                # Find the actual transpose
+                new_values = np.transpose(self.values, trans_ord)
+                # Update old values if it is inplace
+                if inplace:
+                    self.evidence[:] = new_order[::-1]
+                    self.evidence_card = [card_map[var] for var in new_order[::-1]]
+                    variables = [self.variables[0]] + self.evidence
+                    cardinality = [self.variable_card] + self.evidence_card
+                    super(TabularCPD, self).__init__(variables, cardinality, new_values.flatten('C'))
+                    return self.get_cpd()
```
anaviltripathi, line 422: @ankurankan I need your input on the way I am updating the values if it is inplace. 
I am calling **init** method of super for updating other values but I don't really think that it is correct here.

anaviltripathi:
@ankurankan In the new order taken from the user, I have assumed that the leftmost variable will be fastest changing then the next one and so on. In the docs for the original CPD initialization same approach is mentioned. But I don't think it is followed. For example:

``` python
 cpd2 = TabularCPD('J', 2, [[0.9,0.3,0.9,0.3,0.8,0.8,0.4,0.4],
...                                         [0.1,0.7,0.1,0.7,0.2,0.2,0.6,0.6]],
...                                evidence=['C', 'B', 'A'], evidence_card= [2, 2, 2] )
>>> 
>>> print(cpd2)
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_0 | C_0 | C_0 | C_1 | C_1 | C_1 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_0 | B_1 | B_1 | B_0 | B_0 | B_1 | B_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| A   | A_0 | A_1 | A_0 | A_1 | A_0 | A_1 | A_0 | A_1 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| J_0 | 0.9 | 0.3 | 0.9 | 0.3 | 0.8 | 0.8 | 0.4 | 0.4 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
| J_1 | 0.1 | 0.7 | 0.1 | 0.7 | 0.2 | 0.2 | 0.6 | 0.6 |
+-----+-----+-----+-----+-----+-----+-----+-----+-----+
```

Considering #353 to be right about what fastest changing means I think the [docs](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/factors/Factor.py#L70) are a little misleading
Docs state that if given ['x1', 'x2', 'x3'] 'x1' would be the value cycling through its values most rapidly as it is left-most but just below that an example is given which suggests otherwise.
Or am I not understanding what fastest changing means?

In my approach I have taken order to ['B', 'A', 'C'] to denote that B is innermost (fastest changing) and C outermost. Should I change it to ['C', 'A', 'B'] to make it consistent with the initialization?

khalibartan:
@anaviltripathi Till ankur comments you can add tests for `raises` and `warn` and fix the pep8 issues , to increase the coverage.
Also in my perspective current implementation seems more intuitive. 

@ankurankan I agree with @anaviltripathi doc seems ambiguous , what we mean by fastest changing.

pgmpy/factors/CPD.py
```diff
@@ -368,6 +369,63 @@ def to_factor(self):
...
+        >>> cpd.get_cpd()
+        array([[ 0.9,  0.3,  0.9,  0.3,  0.8,  0.8,  0.4,  0.4],
+                [ 0.1,  0.7,  0.1,  0.7,  0.2,  0.2,  0.6,  0.6]])
+        >>> cpd.reorder_parents(['B', 'A', 'C'])
+        array([[ 0.9,  0.9,  0.3,  0.3,  0.8,  0.4,  0.8,  0.4],
+               [ 0.1,  0.1,  0.7,  0.7,  0.2,  0.6,  0.2,  0.6]])
+        """
+        if not self.evidence or (set(new_order) - set(self.evidence)) or (set(self.evidence) - set(new_order)):
+            raise ValueError("New order either has missing or extra arguments")
+        else:
```
khalibartan, line 403: No need for this. We should try to keep our code as less nested as possible.

khalibartan:
@anaviltripathi You have commented a lot of things only complex things should have been commented to help. http://blog.codinghorror.com/coding-without-comments/ .
Also don't implement to many methods to test a single method. Usually avoid that. 1 method does the job. Sometimes we may need 2 or 3 to keep things clean. I saw your tests, and I believe all the methods can be merged into one named `test_reorder` .

pgmpy/factors/CPD.py
anaviltripathi, line 403: @khalibartan Could you explain a little more?

khalibartan, line 403: @anaviltripathi Without having that `else` ,  code would have worked just fine. You are adding an another level of indentation using `else`. Though its not a problem to do things like this but we should avoid it.

anaviltripathi:
@khalibartan Yes, I wanted to keep all the tests in one single method as it doesn't make too much sense to keep all tiny tests apart in separate methods, but I followed suit of marginalize and reduce tests that had many tests written for small variations. I will merge all the tests if that's better approach.

khalibartan:
I have worked with `test_factor.py` when I was writing the test suits for JointProbabilityDistribution. I didn't like the fact that CPD classes and JointProbabilityDistribution classes have their tests written in it.

Ping @ankurankan Why did we wrote the tests like this ?

@anaviltripathi  For me `inplace=True` seems to be more intuitive than `inplace=False`. In your docstring you should add a line about the default value of parameter and should demonstrate in `Example` this behaviour .

pgmpy/factors/CPD.py
```diff
@@ -368,6 +369,63 @@ def to_factor(self):
...
+            If inplace=True it will modify the CPD itself and return it,
+            otherwise new value will be returned without affecting old values
+
+        Returns
+        -------
+        Returns the new reordered CPD
+
+        Examples
+        --------
+        >>>from pgmpy.factors.CPD import TabularCPD
```
ankurankan, line 390: @anaviltripathi Need a space after `>>>`

ankurankan:
@anaviltripathi Let's hold this PR for a while. I feel like a few things need to be changed in `TabularCPD` to make it more clearer. Right now it feels like there are a lot of redundant properties which we are defining. I will try to find a more clearer way to implement this.

anaviltripathi:
@ankurankan Let me know if I can be of some help.

Issue #489, opened 2015-10-09
Add copy methods to all models and Factor class


ankurankan:
- [x] BayesianModel
- [x] TabularCPD
- [x] ClusterGraph
- [ ] DynamicBayesianNetwork
- [ ] FactorGraph
- [x] JunctionTree
- [x] MarkovChain
- [x] MarkovModel

souravsingh:
I would like to work on this.How do I start?

ankurankan:
@souravsingh We just need `copy` methods in these models/classes which should return an instance which is a copy of the instance on which it was called. For examples, you can check Factor's copy method here: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/factors/Factor.py#L639 , `BayesianModel`'s copy method: https://github.com/pgmpy/pgmpy/pull/488/files#diff-5486bc43772bb0d18a7231f9496abcb7R658

khalibartan:
@souravsingh  If you haven't gain any progress I can work on this issue.

souravsingh:
@khalibartan I have made Progress and Currently Testing the functions.

kshitij10496:
@souravsingh Are you still working on this ?

souravsingh:
@kshitij10496 Yes,I am.It will take some time to make a PR since I am testing the methods.

khalibartan:
@souravsingh you should send a patch. Testing can be done along. Small commits help in better review .

khalibartan:
@souravsingh Can you send in a patch. I need these changes for some work . If you'll send a patch then I'll work with your branch, otherwise I'll  send a new PR for this because I need these for work for some fixes that needs to be done .

khalibartan:
@ankurankan the copy method in Factors class is not actually returning a copy I think, numpy arrays are mutable objects and also a list return by a function.
```
In [1]: def x(i):
   ...:     return i
   ...: 

In [2]: a =[1,2,3]

In [3]: b = x(a)

In [4]: b[0]=2

In [5]: a
Out[5]: [2, 2, 3]

In [6]: b
Out[6]: [2, 2, 3]
```
Function is not doing what you are actually thinking .

kshitij10496:
Ping @souravsingh

Pratik151:
@souravsingh  If you are not working on this can I work on this?

souravsingh:
[...]

PR #584, opened 2016-01-17
Added copy method for FactorGraph

Pratik151:
I will add more tests for this soon. I was not sure if I was doing it right, so I made PR.

I made copy of edges because the nodes are factors, so I need to also make factors copy before I can make copy graph.
#489

### DIFF
--- a/pgmpy/models/FactorGraph.py
+++ b/pgmpy/models/FactorGraph.py
@@ -392,3 +392,42 @@ class FactorGraph(UndirectedGraph):
             raise ValueError('Factor for all the random variables not defined.')
 
         return np.sum(factor.values)
+
+    def copy(self):
+        """
+        Returns a copy of FactorGraph.
+
+        Returns:
+        -------
+        FactorGraph : Copy of FactorGraph
+
+        Example
+        -------
+        >>> from pgmpy.models import FactorGraph
+        >>> from pgmpy.factors import Factor
+        >>> import numpy as np
+        >>> G = FactorGraph()
+        >>> phi1 = Factor(['a', 'b'], [2, 2], range(4))
+        >>> phi2 = Factor(['b', 'c'], [2, 2], range(4))
+        >>> G.add_edges_from([('a', phi1), ('b', phi1),
+                              ('b', phi2), ('c', phi2)])
+        >>> G.add_factors(phi1, phi2)
+        >>> G_copy = G.copy()
+        >>> G_copy.nodes()
+        [<Factor representing phi(b:2, c:2) at 0xb4badd4c>, 'b', 'c',
+          'a', <Factor representing phi(a:2, b:2) at 0xb4badf2c>]
+
+        """
+        edges_copy = []
+        for node1, node2 in self.edges():
+            if isinstance(node1, Factor):
+                edges_copy.append((node1.copy(), node2))
+            elif isinstance(node2, Factor):
+                edges_copy.append((node1, node2.copy()))
+            else:
+                edges_copy.append((node1, node2))
+        copy = FactorGraph(edges_copy)
+        if self.factors:
+            factors_copy = [factor.copy() for factor in self.factors]
+            copy.add_factors(*factors_copy)
+        return copy
--- a/pgmpy/tests/test_models/test_FactorGraph.py
+++ b/pgmpy/tests/test_models/test_FactorGraph.py
@@ -287,5 +287,20 @@ class TestFactorGraphMethods(unittest.TestCase):
         self.graph.add_edges_from([('a', phi3), ('c', phi3)])
         self.assertRaises(ValueError, self.graph.check_model)
 
+    def test_copy(self):
+        self.graph.add_nodes_from(['a', 'b', 'c'])
+        phi1 = Factor(['a', 'b'], [2, 2], np.random.rand(4))
+        phi2 = Factor(['b', 'c'], [2, 2], np.random.rand(4))
+        self.graph.add_nodes_from([phi1, phi2])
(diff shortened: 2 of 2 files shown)


pgmpy/models/FactorGraph.py
```diff
@@ -392,3 +392,42 @@ def get_partition_function(self):
...
+            if isinstance(node1, Factor):
+                edges_copy.append((node1.copy(), node2))
+            elif isinstance(node2, Factor):
+                edges_copy.append((node1, node2.copy()))
+            else:
+                edges_copy.append((node1, node2))
+        copy = FactorGraph(edges_copy)
+        if self.factors:
+            factors_copy = [factor.copy() for factor in self.factors]
+            copy.add_factors(*factors_copy)
```
yashu-seth, line 432: @Pratik151 Is it necessary to make a copy of edges and factors? Can't they be directly added?

Pratik151, line 432: First I thought of adding edges directly using `FactorGraph(self.edges())` but then both graphs will have same factor instance. So i thought of making copy of factors so that copy graph will have its own copy of factors. Was the copy of factors not needed?

yashu-seth, line 432: @Pratik151 When a copy of the model is made it is required that both the graphs should have same factor instances. If you make a change in the factors after adding them in the model, the factors added in the model does not change. You will have to update the factors in the model too. Hence I think copy of factors would not be required. What do you think?

pgmpy/models/FactorGraph.py
```diff
@@ -392,3 +392,42 @@ def get_partition_function(self):
...
+                              ('b', phi2), ('c', phi2)])
+        >>> G.add_factors(phi1, phi2)
+        >>> G_copy = G.copy()
+        >>> G_copy.nodes()
+        [<Factor representing phi(b:2, c:2) at 0xb4badd4c>, 'b', 'c',
+          'a', <Factor representing phi(a:2, b:2) at 0xb4badf2c>]
+
+        """
+        edges_copy = []
+        for node1, node2 in self.edges():
```
ankurankan, line 422: @Pratik151 Why aren't you iterating over `self.nodes()` ? I think that will save you one if-else condition.

pgmpy/models/FactorGraph.py
```diff
@@ -392,3 +392,42 @@ def get_partition_function(self):
...
+            if isinstance(node1, Factor):
+                edges_copy.append((node1.copy(), node2))
+            elif isinstance(node2, Factor):
+                edges_copy.append((node1, node2.copy()))
+            else:
+                edges_copy.append((node1, node2))
+        copy = FactorGraph(edges_copy)
+        if self.factors:
+            factors_copy = [factor.copy() for factor in self.factors]
```
ankurankan, line 431: @Pratik151 Can you think of a way to reuse the copies of factors that you have already made ?

pgmpy/tests/test_models/test_FactorGraph.py
```diff
@@ -287,5 +287,20 @@ def test_check_model4(self):
...
+        phi2 = Factor(['b', 'c'], [2, 2], np.random.rand(4))
+        self.graph.add_nodes_from([phi1, phi2])
+        self.graph.add_edges_from([('a', phi1), ('b', phi1), 
+                                   ('b', phi2), ('c', phi2)])
+        self.graph.add_factors(phi1, phi2)
+        graph_copy = self.graph.copy()
+        self.assertIsInstance(graph_copy, FactorGraph)
+        self.assertTrue(graph_copy.check_model())
+        self.assertListEqual(graph_copy.factors[0].scope(), ['a', 'b'])
+        self.assertListEqual(graph_copy.factors[1].scope(), ['b', 'c'])
```
ankurankan, line 302: @Pratik151 Add more tests. Try to change  something in the actual object and check if it also gets changed in the copy.

pgmpy/models/FactorGraph.py
ankurankan, line 432: @Pratik151 @yashu-seth No we will need to make copies of Factors also. We don't want any change in the original object to change anything in the copy.

yashu-seth, line 432: @Pratik151 @ankurankan Sorry my bad making factor copy is necessary. I just checked changes in factors of one model do get reflected in the other.

ankurankan:
@Pratik151 What's the status of this ?

Pratik151:
@ankurankan sorry I didn't work on this still now as I was busy with college fest and all. I think I still won't have time to work on this for couple of days but I will try to look in this tomorrow.
In that time If someone wants to work on this ping me once.

pgmpy/models/FactorGraph.py
Pratik151, line 422: @ankurankan  Sorry for being late, I can work on this now.
I will need to make copy of edges to make new `FactorGraph`from the `edges_copy` I did this. Can you  help with your idea of iterating over `self.nodes()` and how can I make edges from there?

pgmpy/models/FactorGraph.py
Pratik151, line 431: I think this can be done by making `factors_copy` before I make copy of factor nodes and when I am making copy of the factor node I will append that to `factors_copy` and also add it to edge.

Something like : 

```
if isinstance(node1, Factor):
    factors_copy.append(node1.copy())
    edge_copy.append((factors_copy[-1], node2)
```

lohani2280:
@ankurankan This issue has been solved in PR https://github.com/pgmpy/pgmpy/pull/839 . I think we should close this now.

Issue #579, opened 2016-01-15
Bugs in MaximumLikelihoodEstimator.get_parameters()


yashu-seth:
The `MaximumLikelihoodEstimator.get_parameters()` is not properly working. Work needs to be done to make it bug free.

yashu-seth:
@ankurankan You once suggested that we should move to a complete numpy implementation. Should input as pandas dataframe objects be allowed or that should also be done using numpy arrays?

khalibartan:
@yashu-seth Can you elaborate the bug in it.

yashu-seth:
```python
>>> data = pd.DataFrame(columns=('a', 'b', 'c', 'd'))
>>> for i in range(250):
           data3.loc[i] = [0,1,0,0]
>>> for i in range(250,500):
           data3.loc[i] = [1,1,1,0]
>>> for i in range(500,1000):
           data3.loc[i] = [1,0,0,0]
>>> data3.ix[:,'a'].value_counts()
1    750
0    250
dtype: int64
>>> data3.ix[:,'c'].value_counts()
0    750
1    250
dtype: int64
```
When we convert these to numpy arrays both give the same numpy array - ```array([[750], [250]], dtype=int64)```. It should not happen.

yashu-seth:
@khalibartan Working with pandas dataframes at times does not give sorted results which gives incorrect values for variables in CPD. Also there is issue #573 .

yashu-seth:
@khalibartan I am planning to use a completely numpy based soulution.

yashu-seth:
@ankurankan What do you feel about this?

ankurankan:
@yashu-seth Can you elaborate on the bugs that you found ?

yashu-seth:
@ankurankan 

``` python
>>> model = BayesianModel([('A', 'B'), ('C', 'B'), ('C', 'D'), ('B', 'E')])
>>> values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)),
                              columns=['A', 'B', 'C', 'D', 'E'])
>>> fit_data = values[:800]
>>> model.fit(fit_data)

Traceback (most recent call last):
  File "<pyshell#12>", line 1, in <module>
    model.fit(fit_data)
  File "c:\work\pgmpy\pgmpy\models\BayesianModel.py", line 557, in fit
    cpds_list = estimator.get_parameters()
[...]

PR #596, opened 2016-01-21
creates DataFrame object to remove pandas dependency

### LINKED ISSUE #594: Remove pandas dependency
We are using pandas just to keep a track on which column of the array correspond to which variable. 
For example in the case of sampling we return a pandas dataframe with the samples generated. Something like this:

``` python
    A    B    C
0   1    0    1
1   0    0    1
2   1    1    1
.....
```

We should replace it with just a numpy array and a list of variable names. We can define an object which has two properties `variables` and `values`. Something like this:

``` python
sampling_results.variables = ['A', 'B', 'C']
sampling_results.values = [[1, 0, 1],
                           [0, 0, 1],
                           [1, 1, 1]
                           ....    ]
```

yashu-seth:
Fixes #594 #579

### DIFF
--- a/pgmpy/factors/DataFrame.py
+++ b/pgmpy/factors/DataFrame.py
@@ -3,9 +3,10 @@ import numpy as np
 from pgmpy.extern import six
 from pgmpy.extern import tabulate
 
+
 class DataFrame(object):
 
-	#TODO: add docstring with examples
+    # TODO: add docstring with examples
     def __init__(self, variables=[], values=[]):
         """
         Initializes a DataFrame Class.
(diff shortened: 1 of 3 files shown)


yashu-seth:
@ankurankan Please have a look at this. If things are fine I will start replacing pandas dataframe with pgmpy DataFrame.

yashu-seth:
@ankurankan Please have a look at this. If things are fine I will start replacing pandas dataframe with pgmpy.DataFrame.

yashu-seth:
@ankurankan I'll add an add_samples method and any other method that would be necessary while replacing panadas dataframe.

ankurankan:
@yashu-seth Give me some time on this. Need to think the best way for doing this.

yashu-seth:
Ok.

yashu-seth:
@ankurankan Can you tell me a bit about what you have in mind with this?

khalibartan:
@ankurankan What implementation you have in mind ?

ankurankan:
Ref #703

Issue #602, opened 2016-01-26
TypeError raised on my windows 64 platform with python3.4


pheman:
I test the simple example on the tutorials, however a fatal TypeError raised, associated with int int32 and int64. 
If I comment the corresponding TypeError check in Factor.py Line 99, no errors will occur although the final results is completely wrong. I guess there is some serious bugs behind this simple TypeError.

Same error can be also find in 
Bugs in MaximumLikelihoodEstimator.get_parameters() #579
Any comments is welcome!

My environment is:
Windows 7 64
python 3.4
numpy 1.10,1
pandas 0.16.2

The traceback:
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "D:\Anaconda3\lib\site-packages\spyderlib\widgets\externalshell\sitecustomize.py", line 705, in debugfile
    debugger.run("runfile(%r, args=%r, wdir=%r)" % (filename, args, wdir))
  File "D:\Anaconda3\lib\bdb.py", line 431, in run
    exec(cmd, globals, locals)
  File "<string>", line 1, in <module>
  File "D:\Anaconda3\lib\site-packages\spyderlib\widgets\externalshell\sitecustomize.py", line 685, in runfile
    execfile(filename, namespace)
  File "D:\Anaconda3\lib\site-packages\spyderlib\widgets\externalshell\sitecustomize.py", line 85, in execfile
    exec(compile(open(filename, 'rb').read(), filename, 'exec'), namespace)
  File "d:/onebox/pythonprojects/bayesian/bayesian_test.py", line 18, in <module>
    model.fit(values)
  File "D:\Anaconda3\lib\site-packages\pgmpy\models\BayesianModel.py", line 570, in fit
    cpds_list = estimator.get_parameters()
  File "D:\Anaconda3\lib\site-packages\pgmpy\estimators\MLE.py", line 75, in get_parameters
    evidence_card=parent_card.astype('int'))
  File "D:\Anaconda3\lib\site-packages\pgmpy\factors\CPD.py", line 139, in **init**
    super(TabularCPD, self).**init**(variables, cardinality, values.flatten('C'))
  File "D:\Anaconda3\lib\site-packages\pgmpy\factors\Factor.py", line 100, in **init**
    raise TypeError("Values: Expected type int or type float, got ", values.dtype)
TypeError: ('Values: Expected type int or type float, got ', dtype('int64'))

khalibartan:
[...]

PR #604, opened 2016-01-27
Fix the data type bugs for windows platform. Update MLE.py

pheman:
fix the data type bugs for windows platform. Ref #602

### DIFF
--- a/pgmpy/estimators/MLE.py
+++ b/pgmpy/estimators/MLE.py
@@ -62,6 +62,7 @@ class MaximumLikelihoodEstimator(BaseEstimator):
             if not parents:
                 state_counts = self.data.ix[:, node].value_counts()
                 state_counts = state_counts.reindex(sorted(state_counts.index))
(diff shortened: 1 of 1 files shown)


khalibartan:
@pheman Add a comment there, what was the earlier issue. Why error was being raised for windows platform only ?

yashu-seth:
@ankurankan Can you please have a look at this? This fixes the windows platform bug.

pgmpy/estimators/MLE.py
```diff
@@ -62,6 +62,7 @@ def get_parameters(self):
             if not parents:
                 state_counts = self.data.ix[:, node].value_counts()
                 state_counts = state_counts.reindex(sorted(state_counts.index))
+                state_counts = state_counts.astype(np.int, copy=False)
```
ankurankan, line 65: @pheman I think a better way of doing this would be to check for the type when defining `self.data` in `BaseEstimator`.

PR #605, opened 2016-01-27
Fixed a bug in BIFReader class

khalibartan:
Actually there was an issue with the semantics of BIF file types,
when table argument is present in cpd and when not, we have to reshape
the things differently. This behaviour wasn't mentioned and also I didn't looked at it carefully.
For more reference refer to
http://www.cs.cmu.edu/~javabayes/Home/node7.html#SECTION00730000000000000000
Also I ran autopep8 to remove small pep8 issues
Ping @ankurankan .
After merging this re-run the sphinx to build the documentation. Documentation at pgmpy.org is not up to date.

### DIFF
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -3,7 +3,7 @@ from string import Template
 
 import numpy
 from pyparsing import Word, alphanums, Suppress, Optional, CharsNotIn, Group, nums, ZeroOrMore, OneOrMore,\
-        cppStyleComment, Literal, printables
+    cppStyleComment, Literal, printables
 
 from pgmpy.models import BayesianModel
 from pgmpy.factors import TabularCPD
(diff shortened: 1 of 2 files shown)


pgmpy/readwrite/BIF.py
```diff
@@ -248,15 +239,22 @@ def get_cpd(self):
...
-                              arr.size//len(self.variable_states[name])))
+            if 'table' in block:
+                arr = numpy.array(arr)
+                arr = arr.reshape((len(self.variable_states[name]),
+                                   arr.size // len(self.variable_states[name])))
+            else:
+                length = len(self.variable_states[name])
+                reshape_arr = [[] for i in range(length)]
+                for i, val in enumerate(arr):
+                    reshape_arr[i % length].append(val)
```
khalibartan, line 250: I'm manually reshaping the array, because values were given like this

```
() v1 1-v1
() v2 1-v2
() v3 1-v3
```

and were to be converted like

```
[ v1, v2, v3]
[1-v1, 1-v2, 1-v3]
```

Couldn't find another method, for this.

PR #606, opened 2016-01-27
[Help!] Implemented Pmap method for JointProbabilityDistribution

khalibartan:
@ankurankan I have implemented `_mark_immoralities` and `_pmap_skeleton` method , and both are working as intended. I wasn't able to write `pmap` method that will actually create the pmap using these methods.
I have tested methods using cancer.bif at Constructed using cancer.bif at http://www.bnlearn.com/bnrepository/cancer/cancer.bif.gz
There was no particular reason to chose it, just it has small number of parameters to represent JointProbabilityDistribution.
Help wanted !

### DIFF
--- a/pgmpy/factors/JointProbabilityDistribution.py
+++ b/pgmpy/factors/JointProbabilityDistribution.py
@@ -2,12 +2,13 @@ import itertools
 from operator import mul
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


khalibartan:
Not able to find any graphs to test.

Issue #615, opened 2016-01-31
Unable to run `nosetests` inside a `virtualenv` because of missing dependencies in requirements.txt


liquidmetal:
To enable running `nosetests` within virtual environments, we need to add new dependencies to requirements.txt. Here is the list:
- nose==1.3.1
- coverage==4.0.3 (if you want to generate coverage reports)

This would ensure that running `nosetests` will run nose from the current virtualenv.

However, I'm not sure if we should have dev dependencies within `requirements.txt`. I can see `mock==1.3.0` in there which does not seem appropriate. I think we should have a separate file called `test-requirements.txt` that contains such secondary requirements[1].

[1] http://nvie.com/posts/pin-your-packages/

ankurankan:
Fixed.

PR #616, opened 2016-01-31
Testing changes for requirements.txt and test-requirements.txt

liquidmetal:
Addresses #615 to get nosetests running from within a `virtualenv`

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,11 +42,11 @@ In the directory run either of the following:
 
 Using pip
 ```
-$ pip install -r requirements.txt
+$ pip install -r requirements.txt  # or requirements-dev.txt if you want to run unittests
 ```
 or conda
 ```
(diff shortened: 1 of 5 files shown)


ankurankan:
@liquidmetal I think better name would be `requirements-dev.txt`. And it should be a super set of `requirements.txt`. We wouldn't want someone to run installation on both the files.

And you might need to update docs according to this.

liquidmetal:
I've changed the name to `requirements-dev.txt` - however, I did find some references to `test-requirements.txt` in the `pbr` module (also - https://github.com/openstack-dev/pbr/blob/65f83aef8337c36a081240c4a4db671135bd4a9f/doc/source/testing.rst). This doesn't seem to be affecting us right now, though.

PR #623, opened 2016-02-07
created a readme file to describe the folders contents

AMAN3003:
working on the real world example of pgmpy


ankurankan:
@AMAN3003 Don't think we need this. Closing.

Issue #581, opened 2016-01-16
Doctests for pgmpy


khalibartan:
Currently pgmpy have unitest to test our methods but no doctests to test the docstrings. Having doctests can help us locate broken docstrings and related issues.
@ankurankan Label this an easy issue .

pdoongarwal:
@ankurankan @khalibartan I would like to work on this .

khalibartan:
@pdoongarwal  Let @ankurankan reply , and then you can get going.

ankurankan:
@pdoongarwal @khalibartan Not really sure if we need this. Any particular reason for having this ?

khalibartan:
@ankurankan @yashu-seth  and I discussed on gitter channel. Having doctests can help us locate broke docstrings easily if we change API, or even can help us in locating bugs in the docstring. Bugs in docstrings can cause difficulty to beginners . And because sometimes some bugs might slip in the documentation, doctests can be of real help.

liquidmetal:
+1 for this. Pgmpy has complex parameters (factors, models, edges, etc). It might be worth it to have these tests. New users would be able to copy-paste example code and get started. Also, we might be able to incorporate these into the docs (examples: [like here](https://clojuredocs.org/clojure.data/diff))

However, this may not be as straightforward as it seems. I ran a quick test to see the current state of doctests. Most of the examples [currently fail](https://travis-ci.org/pgmpy/pgmpy/jobs/106115126#L460) `doctest`. This might entail rewriting a lot of the docstrings.

khalibartan:
That's what I'm talking about. Also a new contributor can work on this, and along get familiar with our code base. :)

nishnik:
Hey, I have started working on this, along with it I am taking care of pep8 errors.
I want you to look over this [line](https://github.com/pgmpy/pgmpy/pull/625/files#diff-db233cb5229ad26c8f3fc2773510a40eR176), As the output can varyt at run-time, is it fine to ignore it or is there any better way ?

Vanshitaaa20:
> Currently pgmpy have unitest to test our methods but no doctests to test the docstrings. Having doctests can help us locate broken docstrings and related issues. [@ankurankan](https://github.com/ankurankan) Label this an easy issue .

I'm working on this issue along with #757

ankurankan:
[...]

PR #625, opened 2016-02-10
Correcting docstrings

nishnik:
As per issue #581 .

### DIFF
--- a/pgmpy/base/DirectedGraph.py
+++ b/pgmpy/base/DirectedGraph.py
@@ -25,7 +25,6 @@ class DirectedGraph(nx.DiGraph):
     --------
     Create an empty DirectedGraph with no nodes and no edges
 
-    >>> from pgmpy.base import DirectedGraph
     >>> G = DirectedGraph()
 
     G can be grown in several ways
@@ -81,7 +80,6 @@ class DirectedGraph(nx.DiGraph):
 
         Examples
         --------
-        >>> from pgmpy.base import DirectedGraph
         >>> G = DirectedGraph()
         >>> G.add_node('A')
         """
@@ -98,7 +96,6 @@ class DirectedGraph(nx.DiGraph):
 
         Examples
         --------
-        >>> from pgmpy.base import DirectedGraph
         >>> G = DirectedGraph()
         >>> G.add_nodes_from(['A', 'B', 'C'])
         """
@@ -119,7 +116,6 @@ class DirectedGraph(nx.DiGraph):
 
         Examples
         --------
-        >>> from pgmpy.base import DirectedGraph
         >>> G = DirectedGraph()
         >>> G.add_nodes_from(['Alice', 'Bob', 'Charles'])
         >>> G.add_edge('Alice', 'Bob')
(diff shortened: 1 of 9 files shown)


khalibartan:
We haven't decided if we need doctests right now. So there is no need to change it actually, we want our each example to be working itself.

ankurankan:
@khalibartan @nishnik We should have doctests also. And @nishnik are you fixing other doctests also or should I merge this ?

nishnik:
@ankurankan I am fixing others too. But I want you to look at this method, I am using in [link](https://github.com/pgmpy/pgmpy/pull/625/files#diff-db233cb5229ad26c8f3fc2773510a40eR176), As the output can vary at run-time, is it fine to ignore it or is there any better way ?

pgmpy/base/DirectedGraph.py
```diff
@@ -177,12 +171,11 @@ def moralize(self):
 
         Examples
         --------
-        >>> from pgmpy.base import DirectedGraph
         >>> G = DirectedGraph([('diff', 'grade'), ('intel', 'grade')])
         >>> moral_graph = G.moralize()
-        >>> moral_graph.edges()
+        >>> moral_graph.edges() # doctest: +IGNORE_RESULT
```
khalibartan, line 176: Don't ignore the result,sort instead.

nishnik, line 176: But adding `sort` there will confuse the user, and furthermore we will have to do it anyways for results having object address, like
`<Factor representing phi(x1:2, x2:2, x3:2) at 0x7f8188fcaa90>`
Because they are always going to be variable.

nishnik:
Sorry I have to close this PR, I had deleted my local branch as I thought no work was to be done, then now somehow I mixed up and there are merge conflicts.
Will create a new branch and expect a PR by midnight (IST).

khalibartan:
@nishnik These link may help, https://docs.python.org/2/library/doctest.html#unittest-api 
https://docs.python.org/2/library/doctest.html#soapbox

pgmpy/base/DirectedGraph.py
yashu-seth, line 176: @khalibartan @nishnik can you please tell me why exactly is sorting required?

khalibartan, line 176: @yashu-seth Actually sorting is required because output may not be same each time. But we can use doctests with unittests to overcome this.( as mentioned in below comment)

yashu-seth, line 176: @khalibartan cool. I was thinking the same.

nishnik, line 176: @khalibartan But for outputs having object address ?

khalibartan, line 176: @nishnik Did you went through this link https://docs.python.org/2/library/doctest.html#unittest-api , I believe python provides a good doctest API, I'll also talk to someone at sympy and see how do they deal with doctests, will let you know on this thread. Till then you should go through the mentioned link.

pgmpy/factors/JointProbabilityDistribution.py
```diff
@@ -376,11 +374,10 @@ def to_factor(self):
         Examples
         --------
         >>> import numpy as np
-        >>> from pgmpy.factors import JointProbabilityDistribution
         >>> prob = JointProbabilityDistribution(['x1', 'x2', 'x3'], [2, 3, 2], np.ones(12)/12)
         >>> phi = prob.to_factor()
         >>> type(phi)
-        pgmpy.factors.Factor.Factor
+        <class 'pgmpy.factors.Factor.Factor'>
```
khalibartan: Previous output was correct.

yashu-seth:
@nishnik What is the status of this?

nishnik:
Sorry, I am having exams. I will do it the next week.

Issue #627, opened 2016-02-11
Bug in TabularCPD 's marginalize and reduce methods


ankurankan:
``` python
In [6]: >>> from pgmpy.factors import TabularCPD
   ...: >>> cpd_table = TabularCPD('grade', 2,
   ...: ...                        [[0.7, 0.6, 0.6, 0.2],[0.3, 0.4, 0.4, 0.8]],
   ...: ...                        ['intel', 'diff'], [2, 2])
   ...: >>> cpd_table.marginalize(['diff'])

In [7]: cpd_table.evidence
Out[7]: ['intel', 'diff']

In [8]: cpd_table.evidence_card
Out[8]: [2, 2]
```

The attributes `evidence` and `evidence_card` are not being updated after the operation.

Should be automatically fixed if changes proposed in #595 are implemented.

khalibartan:
Yeah I saw this behaviour.

anaviltripathi:
@ankurankan Can be closed now. PR #643 fixes this.

PR #628, opened 2016-02-14
Refactor cpd

### LINKED ISSUE #595: Refactor TabularCPD
Currently `TabularCPD` has the these attributes: `variable`, `variable_card`, `evidence`, `evidence_card`.
And since it inherits `Factor`, it also has: `variables`, `cardinality` and `values`. 

`evidence` and `evidence_card` can be easily derived if we know other attributes. And since we are not directly using `evidence` or `evidence_card` too many times, we can remove these two and compute these when needed. Right now `evidence` and `evidence_card` is adding overhead in a lot of operations because we need to update these as well as `variables` and `cardinality`.

anaviltripathi:
I have removed `evidence` and `evidence_card` as was discussed in #595. I have also modified the tests accordingly. Let me know if there are some issues. 
Fixes #595 and #627.

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 """Contains the different formats of CPDs used in PGM"""
+
 from __future__ import division
 
 from itertools import product
@@ -11,6 +12,7 @@ from pgmpy.factors import Factor
 from pgmpy.extern import tabulate
 from pgmpy.extern import six
 from pgmpy.extern.six.moves import range, zip
+from warnings import warn
 
 
 class TabularCPD(Factor):
@@ -367,6 +369,51 @@ class TabularCPD(Factor):
         <Factor representing phi(grade:3, evi1:2) at 0x7f847a4f2d68>
         """
         return Factor(self.variables, self.cardinality, self.values)
+
+
+    def reorder_parents(self,new_order,inplace = False):
+        '''
+
+        Parameters
+        ----------
+        new_order: list
+            list of new ordering of variables
+        inplace: boolean
+            If inplace == True it will modify the CPD itself
+            otherwise new value will be returned without affecting old values
+
+        Returns
+        -------
+
+        '''
+        if not self.evidence or (set(new_order) - set(self.evidence)) or (set(self.evidence) - set(new_order)):
+            raise ValueError("New order either has missing or extra arguments")
+        else:
+            # No need to do anything if the user provides the same order
+            if new_order != self.evidence[::-1]:
+
+                # Find cardinality mapping of variables
+                card_map = dict(zip(self.evidence, self.evidence_card))
+                # Find old position mapping of variables
+                old_pos_map = dict(zip(self.evidence, range(len(self.evidence))))
+                # Create new ordering of variables based on the given order and previous order
+                trans_ord = [0]+[(old_pos_map[letter]+1) for letter in new_order][::-1]
+                # Find the actual transpose
+                new_values = np.transpose(self.values, trans_ord)
+                # Update old values if it is inplace
+                if inplace:
+                    self.evidence[:] = new_order[::-1]
+                    self.evidence_card = [card_map[var] for var in new_order[::-1]]
+                    variables = [self.variables[0]] + self.evidence
+                    cardinality = [self.variable_card] + self.evidence_card
+                    super(TabularCPD, self).__init__(variables, cardinality, new_values.flatten('C'))
+                    return self.get_cpd()
+                else:
+                    return new_values.reshape(self.cardinality[0], np.prod([card_map[var] for var in new_order[::-1]]))
+                    #return new_values.reshape(self.variable_card, [card_map[var] for var in new_order[::-1]])
+            else:
+                warn("Same ordering provided as current")
+                return self.get_cpd()
 
 
 # Commenting out because not used anywhere for now and not implemented in a very good way.
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -214,12 +214,9 @@ class TabularCPD(Factor):
         evidence_card = self.cardinality[1:]
 
         if evidence:
-            #evidence_card.reverse()
             evidence_card.insert(0, 1)
             cum_card = np.cumprod(evidence_card)
             max_card = cum_card[-1]
-
-            #evidence.reverse()
 
             for i in range(len(evidence)):
                 var = str(evidence[i])
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -98,8 +98,6 @@ class TabularCPD(Factor):
 
         self.variable = variable
         self.variable_card = None
-        self.evidence = None
-        self.evidence_card = None
 
         variables = [variable]
 
@@ -109,27 +107,25 @@ class TabularCPD(Factor):
 
         cardinality = [variable_card]
         if evidence_card is not None:
-            if not isinstance(evidence_card, (list, set, tuple)):
+            if not isinstance(evidence_card, (list, tuple)):
                 if isinstance(evidence_card, np.ndarray):
                     evidence_card = evidence_card.tolist()
                 elif isinstance(evidence_card, (int, float)):
                     evidence_card = [evidence_card]
                 else:
-                    raise TypeError("Evidence cardinality must be list, set, "
+                    raise TypeError("Evidence cardinality must be list, "
                                     "tuple, numpy ndarray or simply a number.")
-            self.evidence_card = evidence_card
             cardinality.extend(evidence_card[::-1])
 
         if evidence is not None:
-            if not isinstance(evidence, (list, set, tuple)):
+            if not isinstance(evidence, (list, tuple)):
                 if isinstance(evidence, np.ndarray):
                     evidence = evidence.tolist()
                 elif isinstance(evidence, six.string_types):
                     evidence = [evidence]
                 else:
-                    raise TypeError("Evidence must be list, set, tuple or array"
+                    raise TypeError("Evidence must be list, tuple or array"
                                     " of strings.")
-            self.evidence = evidence
             variables.extend(evidence[::-1])
             if not len(evidence_card) == len(evidence):
                 raise exceptions.CardinalityError("Cardinality of all "
@@ -144,9 +140,11 @@ class TabularCPD(Factor):
         var_str = '<TabularCPD representing P({var}:{card}'.format(
                             var=self.variable, card=self.variable_card)
 
-        if self.evidence:
+        evidence = self.variables[:0:-1]
+        evidence_card = self.cardinality[:0:-1]
+        if evidence:
             evidence_str = ' | ' + ', '.join(['{var}:{card}'.format(var=var, card=card)
-                                              for var, card in zip(self.evidence, self.evidence_card)])
+                                              for var, card in zip(evidence, evidence_card)])
         else:
             evidence_str = ''
 
@@ -183,10 +181,13 @@ class TabularCPD(Factor):
     def _make_table_str(self, tablefmt="fancy_grid"):
         headers_list = []
         # build column headers
-        if self.evidence is not None:
-            col_indexes = np.array(list(product(*[range(i) for i in self.evidence_card])))
-            for i in range(len(self.evidence_card)):
-                column_header = [self.evidence[i]] + ['{s}_{d}'.format(s=self.evidence[i], d=d) for d in col_indexes.T[i]]
+
+        evidence = self.variables[1:]
+        evidence_card = self.cardinality[1:]
+        if evidence:
+            col_indexes = np.array(list(product(*[range(i) for i in evidence_card])))
+            for i in range(len(evidence_card)):
+                column_header = [evidence[i]] + ['{s}_{d}'.format(s=evidence[i], d=d) for d in col_indexes.T[i]]
                 headers_list.append(column_header)
 
         # Build row headers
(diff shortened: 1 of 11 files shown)


pgmpy/factors/CPD.py
```diff
@@ -11,6 +12,7 @@
 from pgmpy.extern import tabulate
 from pgmpy.extern import six
 from pgmpy.extern.six.moves import range, zip
+from warnings import warn
```
ankurankan, line 15: @anaviltripathi Check PEP 8 for grouping import statements: https://www.python.org/dev/peps/pep-0008/#imports

pgmpy/factors/CPD.py
```diff
@@ -107,27 +107,25 @@ def __init__(self, variable, variable_card, values,
...
         if evidence is not None:
-            if not isinstance(evidence, (list, set, tuple)):
+            if not isinstance(evidence, (list, tuple)):
                 if isinstance(evidence, np.ndarray):
                     evidence = evidence.tolist()
                 elif isinstance(evidence, six.string_types):
                     evidence = [evidence]
                 else:
-                    raise TypeError("Evidence must be list, set, tuple or array"
+                    raise TypeError("Evidence must be list, tuple or array"
```
ankurankan, line 127: @anaviltripathi Must be a list, tuple or array of variable names and variable name can be any hashable python object.

pgmpy/factors/CPD.py
```diff
@@ -107,27 +107,25 @@ def __init__(self, variable, variable_card, values,
...
                 if isinstance(evidence_card, np.ndarray):
                     evidence_card = evidence_card.tolist()
                 elif isinstance(evidence_card, (int, float)):
                     evidence_card = [evidence_card]
                 else:
-                    raise TypeError("Evidence cardinality must be list, set, "
+                    raise TypeError("Evidence cardinality must be list, "
                                     "tuple, numpy ndarray or simply a number.")
-            self.evidence_card = evidence_card
             cardinality.extend(evidence_card[::-1])
```
ankurankan, line 118: @anaviltripathi Why are you reversing the lists ? Wouldn't it assign wrong values to different states ?

anaviltripathi, line 118: @ankurankan Actually, this is part of the old code. Previous implementation used reversing the lists so I didn't modify it. I believe it has to do with [reshaping](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/factors/Factor.py#L113) the array according to cardinalities, and I think it is correct but counter-intuitive at first glance. Let me know if I am wrong.

pgmpy/factors/CPD.py
anaviltripathi, line 127: @ankurankan Ok, I will add this sentence.

pgmpy/factors/CPD.py
anaviltripathi, line 15: @ankurankan Yes, about this, actually there is one more issue you might have noticed. Due to carelessness I created a branch that had previous commits with regards to other PRs I had sent and failed to create a fresh branch. Although I have only kept the features pertaining to this particular feature, I missed this line. If you want, I could start afresh on a new branch that doesn't have irrelevant commit history.

pgmpy/factors/CPD.py
ankurankan, line 118: @anaviltripathi Actually we need to just make this point clear in our docs. For example with the implementation that you have done:

``` python
 cpd = TabularCPD('A', 2, [[0.1, 0.2, 0.3, 0.4], [0.9, 0.8, 0.7, 0.6]], ['B', 'C'], [2, 2])
```

this results in a CPD like this:

```
╒═════╤═════╤═════╤═════╤═════╕
│ C   │ C_0 │ C_0 │ C_1 │ C_1 │
├─────┼─────┼─────┼─────┼─────┤
│ B   │ B_0 │ B_1 │ B_0 │ B_1 │
├─────┼─────┼─────┼─────┼─────┤
│ A_0 │ 0.1 │ 0.2 │ 0.3 │ 0.4 │
├─────┼─────┼─────┼─────┼─────┤
│ A_1 │ 0.9 │ 0.8 │ 0.7 │ 0.6 │
╘═════╧═════╧═════╧═════╧═════╛
```

And without the reversing it would have been:

```
╒═════╤═════╤═════╤═════╤═════╕
│ B   │ B_0 │ B_0 │ B_1 │ B_1 │
├─────┼─────┼─────┼─────┼─────┤
│  C   │ C_0 │ C_1 │ C_0 │ C_1 │
├─────┼─────┼─────┼─────┼─────┤
│ A_0 │ 0.1 │ 0.2 │ 0.3 │ 0.4 │
├─────┼─────┼─────┼─────┼─────┤
│ A_1 │ 0.9 │ 0.8 │ 0.7 │ 0.6 │
╘═════╧═════╧═════╧═════╧═════╛
```

So, we need to make this clear in the docs that how does the order of evidence affect the parents of the CPD.

And the previous behavior was the latter one. I think we should keep that only as that's more intuitive. What do you think ?

pgmpy/factors/CPD.py
ankurankan, line 15: @anaviltripathi Please start on a fresh branch. And you can [`cherry-pick`](https://git-scm.com/docs/git-cherry-pick) the relevant commits from this branch.

pgmpy/factors/CPD.py
anaviltripathi, line 118: @ankurankan I am a little confused here. As I have already mentioned I didn't change these lines and removing `evidence` and `evidence_card` shouldn't have affected this as values being entered in the CPD are actually entered by calling its super class `Factor`. 
Also, although when using `print(cpd)` to print the cpd in my case gives weird output that I don't really understand why, below are some other things that I will need clarification upon.

Previous version, which is currently in use:

``` python
>>> cpd = TabularCPD('A', 2, [[0.1, 0.2, 0.3, 0.4, 0.01, 0.02], [0.9, 0.8, 0.7, 0.6, 0.99, 0.98]], ['B', 'C'], [3, 2])
>>> cpd.values
array([[[ 0.1 ,  0.2 ,  0.3 ],
        [ 0.4 ,  0.01,  0.02]],

       [[ 0.9 ,  0.8 ,  0.7 ],
        [ 0.6 ,  0.99,  0.98]]])
>>> 
KeyboardInterrupt
>>> print(cpd)
+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_0 | B_1 | B_1 | B_2 | B_2 |
+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_1 | C_0 | C_1 | C_0 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+
| A_0 | 0.1 | 0.2 | 0.3 | 0.4 | 0.0 | 0.0 |
+-----+-----+-----+-----+-----+-----+-----+
| A_1 | 0.9 | 0.8 | 0.7 | 0.6 | 0.9 | 0.9 |
+-----+-----+-----+-----+-----+-----+-----+
>>> cpd.variables
['A', 'C', 'B']
>>> cpd.cardinality
array([2, 2, 3])
```

My current version which reverses the list (as I believe was done previously too):

``` python
>>> cpd = TabularCPD('A', 2, [[0.1, 0.2, 0.3, 0.4, 0.01, 0.02], [0.9, 0.8, 0.7, 0.6, 0.99, 0.98]], ['B', 'C'], [3, 2])
>>> cpd.values
array([[[ 0.1 ,  0.2 ,  0.3 ],
        [ 0.4 ,  0.01,  0.02]],

       [[ 0.9 ,  0.8 ,  0.7 ],
        [ 0.6 ,  0.99,  0.98]]])
>>> print(cpd)
+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_0 | C_0 | C_1 | C_1 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_1 | B_2 | B_0 | B_1 | B_2 |
+-----+-----+-----+-----+-----+-----+-----+
| A_0 | 0.1 | 0.2 | 0.3 | 0.4 | 0.0 | 0.0 |
+-----+-----+-----+-----+-----+-----+-----+
| A_1 | 0.9 | 0.8 | 0.7 | 0.6 | 0.9 | 0.9 |
+-----+-----+-----+-----+-----+-----+-----+
>>> cpd.variables
['A', 'C', 'B']
>>> cpd.cardinality
array([2, 2, 3])

```

If don't reverse the list, as suggested by you:

``` python
>>> cpd = TabularCPD('A', 2, [[0.1, 0.2, 0.3, 0.4, 0.01, 0.02], [0.9, 0.8, 0.7, 0.6, 0.99, 0.98]], ['B', 'C'], [3, 2])
>>> cpd.values
array([[[ 0.1 ,  0.2 ],
        [ 0.3 ,  0.4 ],
        [ 0.01,  0.02]],

       [[ 0.9 ,  0.8 ],
        [ 0.7 ,  0.6 ],
        [ 0.99,  0.98]]])
>>> print(cpd)
+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_0 | B_1 | B_1 | B_2 | B_2 |
+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_1 | C_0 | C_1 | C_0 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+
| A_0 | 0.1 | 0.2 | 0.3 | 0.4 | 0.0 | 0.0 |
+-----+-----+-----+-----+-----+-----+-----+
| A_1 | 0.9 | 0.8 | 0.7 | 0.6 | 0.9 | 0.9 |
+-----+-----+-----+-----+-----+-----+-----+
>>> cpd.variables
['A', 'B', 'C']
>>> cpd.cardinality
array([2, 3, 2])

```

Notice that in my case `cpd.values`, `cpd.variables`, `cpd.cardinality` match with the previous implementation but not the output with `print(cpd)`. 
So, I am reversing the lists as I think that values should be in the same format as they were previously for expected behavior of TabularCPD. 
I maybe completely wrong here. Let me know.

ankurankan, line 118: @anaviltripathi What I understand now (not very sure) is that the earlier representation was considering the fastest changing parent first. So, if the evidence was `['B', 'C']` then the table should have been something like this:

```
+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_1 | C_0 | C_1 | C_0 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_0 | B_1 | B_1 | B_2 | B_2 |
+-----+-----+-----+-----+-----+-----+-----+
+-----+-----+-----+-----+-----+-----+-----+
```

But the `__str__` method wasn't considering this.

Anyways let's fix this all now. I think the earlier representation of evidence was not that intuitive. And intuitive way for defining evidence for the above table should be `['C', 'B']`. What do you think ?

ankurankan, line 118: @anaviltripathi What I understand now (not very sure) is that the earlier representation was considering the fastest changing parent first. So, if the evidence was `['B', 'C']` then the table should have been something like this:

```
+-----+-----+-----+-----+-----+-----+-----+
| C   | C_0 | C_1 | C_0 | C_1 | C_0 | C_1 |
+-----+-----+-----+-----+-----+-----+-----+
| B   | B_0 | B_0 | B_1 | B_1 | B_2 | B_2 |
+-----+-----+-----+-----+-----+-----+-----+
+-----+-----+-----+-----+-----+-----+-----+
```

But the `__str__` method wasn't considering this.

Anyways let's fix this all now. I think the earlier representation of evidence was not that intuitive. And intuitive way for defining evidence for the above table should be `['C', 'B']`. What do you think ?

anaviltripathi, line 118: @ankurankan I agree with you. So, I am not reversing the list finally and the output would be as in the third case. Also, in many other places, just like in `__str__` method, `evidence` was used directly, so the method implementation and their tests were designed that way only. I may need to look into those carefully and I will send in patches as I keep working on those to make sure new bugs are not introduced. 
In short, new implementation will be they way you suggested above.

anaviltripathi, line 118: I have one more doubt. Should we allow reducing or marginalizing the variable itself in CPD, especially inplace. If it were a JPD then marginalizing wrt any variable would be correct, but I suspect that it is not the case here. 
Currently reducing and marginalizing w.r.t. `variable` is allowed and it gives all `1.` output as can be expected if this behavior is to be allowed. But, `print(cpd)` doesn't work as can also be expected. 
But, in my view this should not be allowed inplace. If inplace is chosen to `False` then it can be allowed I guess. 
Also, docstrings for marginalize and reduce are not correct, errors pop up if you follow the directions exactly.

yashu-seth:
@anaviltripathi You mean error pops up in this branch or in the dev for marginalize and reduce?

anaviltripathi:
@yashu-seth In this branch, and in dev.

pgmpy/factors/CPD.py
anaviltripathi, line 118: Update: I ran tests by not reversing list of evidence and a lot of tests are failing (30). Should we really opt for not reversing the lists, considering that it has repercussions in many other methods?

yashu-seth:
@anaviltripathi Okay. I think you have opened issue #635 for it.

anaviltripathi:
@yashu-seth No, that is different. What I am talking about here is whether we should even allow reducing and marginalizing with respect to the variable under consideration in a TabularCPD. Present implementation allows this, but I think that we should not allow it inplace, and maybe give a warning even for case of non-inplace. 
Also, even though current implementation allows this, after reducing w.r.t. the variable `print` command doesn't work and produces errors. I didn't think necessary to open a separate issue for this. Should I?

yashu-seth:
@anaviltripathi I also got errors for the cases you mentioned. I think issue #635 is similar to this problem because I get similar error for both. We will have to look into it to be sure. I think if not a separate issue we must mention our findings there itself.

anaviltripathi:
This branch had some irrelevant commit history so I am closing this. Refer new PR #643.

PR #636, opened 2016-02-18
Feature: Independency equivalence & entailment

chrisittner:
I wrote three methods for `Independencies` to check (1) whether one `Independencies`-object implies another, (2) whether two `Independencies`-objects are equivalent, and (3) to add all implied `IndependenceAssertion`s to a `Independencies`-object. 

**Background:**
If you know one (conditional) independency then you can infer others. For example,  `X ⟂ Y,W | Z` implies that `X ⟂ Y | Z`. Whenever `X` is (conditionally) independent of `Y` and `W` given `Z`, `X` must also be independent of just `Y` given `Z`. Because of this, different-looking sets of independency statements can be equivalent (= whenever the one is satisfied, the other is too).
For example, any Bayesian Network with `X ⟂ W | Y,Z` and `X ⟂ Y | Z` also satisfies `X ⟂ W,Y | Z` and vice versa.

**How can you check for equivalence?**
If you read off your independencies from two Bayesian networks, you can just check whether the BNs are I-equivalent. But if you don't already know the networks and just want to work with independencies, you could do this: Given two sets of independencies `A` and `B`, compute all implications of `A` and all implications of `B` and then check if they are equal. To compute the implied independencies, one can use the semi-graphoid axioms ([described here on page 441f](https://books.google.de/books?id=yQVGAAAAQBAJ&lpg=PA442&ots=wRu_v5f-9b&dq=semi%20graphoid%20axioms%20d%20separation%20completeness&pg=PA441#v=onepage&q=semi%20graphoid%20axioms%20d%20separation%20completeness&f=false)). I used them in the `closure()`-method, it returns a new `Independencies`-object that is closed under the axioms (= contains its implied independencies). The other two methods are simple from there. I put examples in the docstring.

If you are interested in merging them, I'll add tests and can try to improve performance (the complexity of the algorithm is quite bad, but I'm not sure if you can do better).

### DIFF
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -1,3 +1,6 @@
+# -*- coding: utf-8 -*-
+
+import itertools
 from pgmpy import exceptions
 from pgmpy.extern import six
 
@@ -40,6 +43,9 @@ class Independencies(object):
     add_assertions
     get_assertions
     get_factorized_product
+    closure
+    entails
+    is_equivalent
     """
     def __init__(self, *assertions):
         self.independencies = []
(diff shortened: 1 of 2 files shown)


pgmpy/independencies/Independencies.py
```diff
@@ -215,6 +363,9 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    def __hash__(self):
+        return hash(str(sorted(self.event1)) + str(sorted(self.event2)) + str(sorted(self.event3)))
+
```
khalibartan: There is a PR that deals with this issue.

pgmpy/independencies/Independencies.py
```diff
@@ -60,6 +66,9 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    def __hash__(self):
+        return sum(hash(assertion) for assertion in self.get_assertions())
+
```
khalibartan: There is a PR that deals with this issue.

pgmpy/independencies/Independencies.py
```diff
@@ -97,6 +106,143 @@ def add_assertions(self, *assertions):
...
+            # we need to apply axioms sg1-sg3 to both X_|_Y | Z and Y_|_X | Z
+            inds_flipped = set(map(sg0, inds))
+            inds_both = inds | inds_flipped
+            ind_pairs = itertools.permutations(inds_both, 2)
+
+            inds |= set(sum(map(sg1, inds_both), ()))
+            inds |= set(sum(map(sg2, inds_both), ()))
+            inds |= set(sum(map(lambda i: sg3(*i), ind_pairs), ()))
+
```
khalibartan, line 209: Use `six.moves.map` .

pgmpy/independencies/Independencies.py
```diff
@@ -97,6 +106,143 @@ def add_assertions(self, *assertions):
...
+        def single_var(var):
+            "Checks if var represents a single variable"
+            if not hasattr(var, '__iter__'):
+                return True
+            else:
+                return len(var)==1
+
+        def sg0(ind):
+            "Symmetry: 'X ⟂ Y | Z' -> 'Y ⟂ X | Z'"
```
khalibartan: I would recommend to use `.. math:: X \perp Y | Z` instead of `u"X \u27C2 Y | Z "`

khalibartan:
I would like if you could modify how independence assertion objects are stored internally too. Frozensets seems more generic solution. Ping @ankurankan . I would like if we could have this implementation instead of set representation,  `__eq__` method and`__ hash__` method would be cleaner this way. This issue has been around for quite a time.

khalibartan:
I would be great if you could squash your commits, commit history remains more clean this way.

chrisittner:
Cool, I will make the fixes + write tests, if I get green light that the `closure`-, `entails`\- and `is_equivalent`-methods seem in principle useful to you and might get merged in some way. 

My use-case is that I wanted some scripts to help construct Bayesian networks from a list of (conditional) independence judgements (For example, an expert tells you what depends or does not depend on what and now you want to know how to put the arrows in your BN). Then you need to reason just with the independencies.

Regarding `IndependenceAssertion.__hash__`:  It's probably best to wait with this PR until `IndependenceAssertion` is nice and hashable, then I can rebase. There was PR #536 but I'm not sure about its state, so I created a seperate pull request for that, PR #640.

khalibartan:
For the `closure` and `entails` I have to go through the link mentioned by you and others resources too. Right now I'm busy with my mid terms. Will tell you soon, after exams are over. Let see what @ankurankan has to think about it.

ankurankan:
@chrisittner This will certainly be helpful. Can you please add tests for these ?

aelnouby:
@ankurankan  @khalibartan @chrisittner  what the status of this issue , i can help finish it, if it is OK ?

chrisittner:
@aelnouby hm, i do not think this is a good issue to join in on (it isn't even in the list of issues and i need to rewrite parts of it). I didn't get to it in the past week, but i will finish it on the weekend, so that it's out of the way.

aelnouby:
@chrisittner Great, good luck :D

chrisittner:
@ankurankan @khalibartan I reworked this PR, I'd be happy if you could take a look.

I rewrote the `closure()` method in the hope to make in more readable (and less inefficient) and added tests for all the methods. The docstring now contains two reference links for the semi-graphoid rules, one to [wikipedia](https://en.wikipedia.org/w/index.php?title=Conditional_independence&oldid=708760689#Rules_of_conditional_independence) and the other to the [paper](http://www.cs.technion.ac.il/~dang/journal_papers/pearl1989conditional.pdf) .

I also tested the speed on randomly generated lists of independence jugdements (and on independencies obtained from randomly generated directed acyclic graphs) with [this script](https://gist.github.com/chrisittner/341fb79c5d5deeda1752). Unfortunaly determining closure/entailment/equivalence is very expensive (in the worst-case it is exponential in the number of variables, i think) and in the tests it won't work in reasonable time as soon as 7 or more variables are involved. I added that as a note in the docstring.

chrisittner:
regarding the missing `reduce()`-method (discussed over at #641, referenced above): reduction using the `entails`-function in any way is clearly too slow, as `reduce()` is called in more efficient functions like `BayesianModel.get_independencies()`.

I tested the `get_independencies()`-method with the above script (by running `test2()`, but with line 60 commented out) and it can read off the independencies from models with up to 14 variables, before getting really slow (on my machine). With 14 variables and not too many edges, it returns ~100k indepenence assertions, so i'd say the `reduce` method should be able to deal with such an amount quickly, even if that means leaving some redundancy in the list. For example, one could (1) remove symmetries (`X ⟂ Y | Z` ~ `Y ⟂ X | Z`), then (2) do subset-checking to remove things like `X  ⟂ Y | Z` and `X ⟂ W | Z`, if `X ⟂ Y,W | Z` is already in the list, and finally (3) remove assertions like `X ⟂ Y | W, Z`, if the list already contains the stronger statement `X ⟂ Y,W | Z`. This would remove duplicates that have the form of the first three of [these](https://en.wikipedia.org/wiki/Conditional_independence#Rules_of_conditional_independence) rules. 
@aelnouby if you are still looking for work and want to do the `reduce()`-method, feel free. The implementation will be independent from the entailment stuff after all, I won't include it in this PR (but might o/w do it sometime in the future if its still open).

Issue #639, opened 2016-02-19
IndependenceAssertion('X','Y','Z') == IndependenceAssertion('X','Z','Y')


chrisittner:
```
>>> from pgmpy.independencies import IndependenceAssertion
>>> IndependenceAssertion('X','Y','Z') == IndependenceAssertion('X','Z','Y')
True
```

This should evaluate to `False`.

khalibartan:
Yeah, bug is at here https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/independencies/Independencies.py#L209 , we should only consider `self.event1` and `self.event2` in unordered form, not `self.event3` too. Can you fix it. Fix is fairly easy.

chrisittner:
Yep :).

khalibartan:
Close this issue.

Issue #638, opened 2016-02-18
Making IndependenceAssertion Hashable


aelnouby:
Hi all,

I am a newbie here , I was going through the TODOs in #548 , i started investigating the first issue
 independencies/Independencies.py: # TODO: write reduce function.

I think making the IndependenceAssertion hashable would be a good idea , it will solve this issue by storing the objects in a set eliminating any duplication from the beginning , also making it hashable can be useful in the future 

Please tell me what you think , and if this might cause any problems.

khalibartan:
@aelnouby Yeah you are right. But there is a PR that deals with this issue and issue #531

chrisittner:
Now there are two :D (Also, this issue is duplicate of #531 )

khalibartan:
@ankurankan Fixed. Can be closed now and #531 too.

Issue #575, opened 2016-01-15
In IndependenceAssertion Class events should converted to frozenset instead of set


khalibartan:
Having `self.event1`,`self.event2` etc as frozenset instead of set (see [here](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/independencies/Independencies.py#L189)) have advantage because they are hashable , moreover `__eq__` method that I have implemented can be simplified and will also lead to simplification of `__hash__` method for IndependenceAssertion and Independencies class.Since there is no notion for adding more assertion to a IndependenceAssertion, frozenset are better to use than sets because of above mentioned advantages.
Ping @ankurankan What are your views on it ?
@kshitij10496 Also since you are working with `__hash__` you might also want these changes.

PR #640, opened 2016-02-19
Make IndependenceAssertion immutable + hashable

chrisittner:
Issues #531, #575 and #638 suggest that `IndependenceAssertion` should be hashable. There is PR #536 from December last year that contains a fix among other changes, but some of the other parts of that PR seem outdated, so it might not get merged anytime soon.

So I wanted to propose a fix just for that.
(because I used that they are hashable in PR #636)

**Changes:**
    - use `frozenset` instead of `set` inside `IndependenceAssertion`
    - remove `IndependenceAssertion.set_assertions` (never used in remaining codebase, but has tests)
    - remove tests for `set_assertion`.
    - add `IndependenceAssertion.__hash__`-method + simplify `__eq__`

Also fixes bug #639 in `IndependenceAssertion.__eq__`.

### DIFF
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -163,7 +163,6 @@ class IndependenceAssertion(object):
     Public Methods
     --------------
     get_assertion
-    set_assertion
     """
     def __init__(self, event1=[], event2=[], event3=[]):
         """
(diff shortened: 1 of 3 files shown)


pgmpy/independencies/Independencies.py
```diff
@@ -42,7 +42,7 @@ class Independencies(object):
     get_factorized_product
     """
     def __init__(self, *assertions):
-        self.independencies = []
+        self.independencies = set()
```
khalibartan: Let it be a list, maybe if user want to have assertions like `(X_|_ Y,Z | W)` or `(Z,Y_|_X | W)`, we cannot just remove them.Let reduce function remove them and also, case like `(X _|_ Y | W)`

chrisittner: Convinced. I'll remove the last two commits.

Issue #557, opened 2015-12-25
Bug in DynamicBayesianNetwork


khalibartan:
It is not clear that this is bug in DirectedGraph or DynamicBayesianNetwork. While working with DynamicBayesianNetwork there is an error while creating a Digraph.( similar results were not reproduced in BayesianModel . both times it gave error).

```
In [54]: from pgmpy.models import DynamicBayesianNetwork as dbn

In [55]: model = dbn()

In [56]: model.add_edges_from([(('a',0),('b',1)),(('c',0),('b',1))])

In [57]: model.get_parents(('b', 0))
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/home/utgup/anaconda3/lib/python3.4/site-packages/networkx/classes/digraph.py in predecessors_iter(self, n)
    754         try:
--> 755             return iter(self.pred[n])
    756         except KeyError:

KeyError: ('b', 0)

During handling of the above exception, another exception occurred:

NetworkXError                             Traceback (most recent call last)
<ipython-input-57-b0c67100eefd> in <module>()
----> 1 model.get_parents(('b', 0))

/home/utgup/pgmpy/pgmpy/base/DirectedGraph.py in get_parents(self, node)
    166         ['diff', 'intel']
    167         """
--> 168         return self.predecessors(node)
    169 
    170     def moralize(self):

/home/utgup/anaconda3/lib/python3.4/site-packages/networkx/classes/digraph.py in predecessors(self, n)
    766     def predecessors(self, n):
    767         """Return a list of predecessor nodes of n."""
--> 768         return list(self.predecessors_iter(n))
    769 
    770 

/home/utgup/anaconda3/lib/python3.4/site-packages/networkx/classes/digraph.py in predecessors_iter(self, n)
    755             return iter(self.pred[n])
    756         except KeyError:
--> 757             raise NetworkXError("The node %s is not in the digraph."%(n,))
    758 
    759     def successors(self, n):

NetworkXError: The node ('b', 0) is not in the digraph.

```

Now if I do this:

```
In [58]: from pgmpy.models import DynamicBayesianNetwork as dbn

In [59]: model = dbn()

In [60]: model.add_edges_from([(('a',0),('b',1)),(('c',0),('b',1))])

In [61]: model.add_nodes_from(model.nodes())

In [62]: model.get_parents(('b', 0))
[ ]
```

There is no error.
[...]

Issue #452, opened 2015-07-22
Dynamic Bayesian Network bug 


palashahuja:
```
from pgmpy.models import DynamicBayesianNetwork as dbn
from pgmpy.factors import TabularCPD
import numpy as np
model = dbn()
model.add_edges_from([(('a',0),('b',1)),(('c',0),('b',1))])
values = np.split(np.random.rand(4),2)
a_cpd = TabularCPD(('b',1), 2, values, [('a',0)], 2)
c_cpd = TabularCPD(('b',1), 2, values, [('c',0)], 2)
model.add_cpds(a_cpd, c_cpd)
model.initialize_initial_state()
```

Shekharrajak:
Can you please explain little about the bug for better understanding ?

palashahuja:
The above example looks like this, if considered of only the first two time slices.
![example](https://cloud.githubusercontent.com/assets/8504351/11900053/5a57a8cc-a5c9-11e5-8a97-ed2f32d5b57d.png)
The nodes `a_0` and `c_0` are the nodes which don't have an intra-connection(i.e. connection among the nodes inside the time slice that they belong to). These are nodes that provide a constant noise to the next time slice. 
 The problem is that for every `intra-connection` or `inter-connection` in the 0th timeslice, it is replicated for the same in the 1st timeslice in the method `initialize_initial_state`. The nodes having a constant noise aren't taken care of, and thus the bug.
The rectification is fairly simple inside the method of `DynamicBayesianNetwork`. But the problem arises when you consider the same for inference inside `dbn_inference.py`. I suggest you work out the details accordingly. Let me know if you have any further doubts.

khalibartan:
@ankurankan Can you take a look at this bug. Its fairly easy to know what caused this error, but I don't understand algorithm so I think I cannot work . Also the example in docstring of initialize_inital_state is not reproducible please see for that too .

palashahuja:
@khalibartan , I have looked into the `initialize_initial_state` method and it seems fine to me.(i.e. the example from the docstring). I could try to explain the algorithm further, but could you please refer to the algorithm that you seek to understand ?

khalibartan:
@palashahuja here is the bug I got , when I tried to reproduce the code.

```
In [2]: %paste
        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
        >>> from pgmpy.factors import TabularCPD
        >>> student = DBN()
[...]

Issue #65, opened 2014-02-17
Implementing reading from and writing to various file formats


ankurankan:
- [x] **ProbModelXML:** http://leo.ugr.es/pgm2012/submissions/pgm2012_submission_43.pdf
- [ ] **DNET_1:** https://www.norsys.com/downloads/DNET_File_Format.txt
- [ ] **Elvira:** http://leo.ugr.es/elvira/devel/Formato/formato.html
- [x] **XML BIF:** http://www.cs.cmu.edu/~fgcozman/Research/InterchangeFormat/
- [ ] **MSBNx:** http://research.microsoft.com/en-us/um/redmond/groups/adapt/msbnx/
- [x] **XDSL:**
- [ ] **Cassandra's Format:** http://www.pomdp.org/code/pomdp-file-grammar.shtml
- [ ] **SPUDD:** 
- [x] **pomdpX:** http://bigbird.comp.nus.edu.sg/pmwiki/farm/appl/index.php?n=Main.PomdpXDocumentation
- [ ] **PPDDL:**
- [ ] **RDDL:**
- [x] **UAI**: http://graphmod.ics.uci.edu/uai08/FileFormat
- [x] NET
- [ ] DSC

jp111:
Which of these are preferred? Also please give a rough idea of how much time will be consumed in one so that I can better plan schedule and decide time-line for project.

ankurankan:
We don't have any preference over the formats. You can look at the
implementation of ProbModelXML to get a basic idea of what you will have to
do.

I would suggest you to select 3 file formats and allocate 4 weeks on each.
One week for understanding the format and discussing it's implementation,
one week for implementing its reader class, one week for implementing its
writer class and then one week for writing tests.

ankurankan:
Git markdown messed with the xml file. The correct image and file is follows

https://lh5.googleusercontent.com/-JaAZWwk0yfU/UzpwyeKVZMI/AAAAAAAACDc/j_RH6rHzh1k/s1600/CANCER.GIF

<?XML VERSION="1.0"?>





<!DOCTYPE BIF [
<!ELEMENT BIF ( NETWORK )_>
<!ELEMENT PROPERTY (#PCDATA)>
<!ELEMENT TYPE (#PCDATA)>
<!ELEMENT VALUE (#PCDATA)>
<!ELEMENT NAME (#PCDATA)>
<!ELEMENT NETWORK
[...]

PR #641, opened 2016-02-19
Implementing Hash for IndependenceAssertion and reduce method

aelnouby:
Used different Hash function than that in #640 
`return (hash(frozenset(self.event1))<<1)^(hash(frozenset(self.event2))>>1)^hash(frozenset(self.event3))
`
 Now the order of the events will produce different hash and tested it against the case @khalibartan proposed ( X_|_ Y,Z | W) and (Z,Y_|_X | W)

### DIFF
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -42,7 +42,7 @@ class Independencies(object):
     get_factorized_product
     """
     def __init__(self, *assertions):
-        self.independencies = []
+        self.independencies = set()
         self.add_assertions(*assertions)
 
     def __str__(self):
@@ -90,19 +90,12 @@ class Independencies(object):
         """
         for assertion in assertions:
             if isinstance(assertion, IndependenceAssertion):
-                self.independencies.append(assertion)
+                self.independencies.add(assertion)
             else:
                 try:
-                    self.independencies.append(IndependenceAssertion(assertion[0], assertion[1], assertion[2]))
+                    self.independencies.add(IndependenceAssertion(assertion[0], assertion[1], assertion[2]))
                 except IndexError:
-                    self.independencies.append(IndependenceAssertion(assertion[0], assertion[1]))
-
-        # TODO: write reduce function.
-    def reduce(self):
-        """
-        Add function to remove duplicate Independence Assertions
-        """
-        pass
+                    self.independencies.add(IndependenceAssertion(assertion[0], assertion[1]))
 
     def latex_string(self):
         """
@@ -214,6 +207,9 @@ class IndependenceAssertion(object):
 
     def __ne__(self, other):
         return not self.__eq__(other)
+
+    def __hash__(self):
+        return hash(frozenset(self.event1))^hash(frozenset(self.event2))^hash(frozenset(self.event3))
 
     @staticmethod
     def _return_list_if_str(event):
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -42,7 +42,7 @@ class Independencies(object):
     get_factorized_product
     """
     def __init__(self, *assertions):
-        self.independencies =set()
+        self.independencies =[]
         self.add_assertions(*assertions)
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


aelnouby:
There is a problem , currently fixing

khalibartan:
@aelnouby This is not good, there are already PR that deals with this issue, #536 #640 .

aelnouby:
@khalibartan I am sorry if this caused any problems , but things aren't very clear #536 has been open for a long time and your last comment was a recommendation to close it , #640 i have just changed the hashing function in order to get the desired behaviour  , so please tell me what is the best practise in cases such like this.

khalibartan:
@aelnouby If you want to have a particular implementation , you can discuss it with the user, who is currently working on it. Like if you wanted to be hash function like this it would have been good if you discussed it over #640 thread . Let see what @ankurankan has to say about it. But in my view it is waste of efforts of both developers.

ankurankan:
@aelnouby Can you please create an issue for the problem you were facing when you were working with changes from #640 ?

aelnouby:
@ankurankan i just want to understand a few things , first right now the merged #640 , i can't see how the duplication problem was solved.

Secondly, right now what i implemented is a hash method that will produce different hashes to different independencies whatever the order is , and implemented the reduce method (using a set) to remove duplication.

The problems i faced happened when i tried to get rid of the reduce method and prevent duplication form the beginning by making Independencies a set (you can imagine this might need some modifications to happen in many places ), when i referred to #640 in this problem, i  was referring to a commit that dealt with a similar problem but it was removed later.

I didn't spend much time trying to solve this problem , but if making independencies a set is required, i will work on it and i will report if i faced any problems.

khalibartan:
@aelnouby The `__hash__` should return similar hash values for `(X _|_ Y,Z | W)` and `(Z,Y _|_ X | W)`, I never meant that they should return different hash values. If you properly read the context at [here](https://github.com/pgmpy/pgmpy/pull/640#discussion_r53423413) you will understand that I was implying that if represent Independencies (not IndependenceAssertion) as a set than we will remove such cases without users knowledge. It should be reduce function that removes all the similar assertions.

aelnouby:
@khalibartan Thanks i get it now , i misunderstood what you meant. 

So is there is anything more to do concerning this issue , or should it be closed.

khalibartan:
@aelnouby First @ankurankan needs to define what he wants from reduce method. Two things come in my mind:
- When independencies are `(X _|_ Y,Z | W )` , `(Y,Z _|_ X | W)` and `(X _|_ Y | W )`, should it return `(X _|_ Y,Z| W)` because first and second are equivalent and third one assert a independence that is asserted by first and second.
- Or should it return `(X _|_ Y,Z | W)` and `(X _|_ Y | W )`.

If it is the second case then nothing needs to be done.

ankurankan:
@aelnouby @khalibartan I was thinking that reduce should return the least number of assertions. So reducing these: `(X _|_ Y,Z | W )` , `(Y,Z _|_ X | W)` and `(X _|_ Y | W )` should return `(X _|_ Y, Z | W)`.

khalibartan:
@ankurankan I think we can either have a naive level implementation that utilises subset checking for events or we can use the changes of #636 . What would you prefer ? Or any other implementation.

chrisittner:
(With the methods from #636 you can implement `reduce` to remove all redundancy, but it is probably too slow. I will work on #636 some more in the next days and check performance on larger models + asymptotic complexity, for now i don't know. But `reduce` should be fast, because it gets called by `BayesianModel.get_independencies()` etc.)

khalibartan:
I believe this can be closed for now.

aelnouby:
fixed in #636

khalibartan:
@aelnouby Since you are looking to work with some thing, may be you can work with some read write classes #65 .Parsing formats is pretty easy task, just search for the semantics for particular file format and ping ankur before working on it. If you find any issues while working ping me over the thread.

aelnouby:
@khalibartan Great i will start with the DBN copy method in #489 and then i will have a look at this. I will also wait for ankur's response concerning the score and cross validation additions.

Also may you recommend for me issues that might be related or can help me with the ADD project , as an entrance task for this project ?

Thanks a lot for your help :D

khalibartan:
@aelnouby I cannot comment on this matter.
Since you are working with DBN class, I would like if you could do some cleaning, also there are quite some `#TODO:` marked in it and some issues at here #557 #452 , fix them too.

Issue #232, opened 2014-12-03
Add feature for storing the state name of the random variables.


ankurankan:
Currently we store the state names of a variables as `variable_0`, `variable_1` etc. But this can get confusing for a user so add the optional feature to add the states of the variables in a model and instead of showing `variable_0` or `variable_1`, display the actual names of the states.

yashu-seth:
Please tell me if this is still open. Is someone working on this?

ankurankan:
@yashu-seth Yes, this is still open. But this a fairly big change so will need a lot of work. 
The way I have thought of implementing this is to add the functionality to store state names to the `DirectedGraph` and `UndirectedGraph` classes. And then write a decorator, such that it replaces the state names from function arguments to the internal representation. And also to replace the internal state names returned by the methods to the actual state name.

yashu-seth:
@ankurankan Can you please recommend an issue which would be relatively easy to handle and is still open.

khalibartan:
@ankurankan Instead of replacing the internal architecture I think we can have a one to one mapping of variables name to the current way we are storing the things and whenever a query is raised, we can use this mapping to get to the internal representation . Also the current implementation has its own ease because user don't have to explicitly name the states. So we don't have to write two different implementations for this. We can run one above the other,

yashu-seth:
+1 for @khalibartan 's suggestion. @ankurankan  What do you think about this?

ankurankan:
@yashu-seth @khalibartan Yeah, even I want to keep this completely separate. I was thinking of a writing a decorator for this which would do all the mapping for the arguments passed to a method or returned result.

khalibartan:
@ankurankan  Yeah cool. +1 Decorator will keep things much cleaner.

yashu-seth:
@ankurankan I would like to work on this. 
From what I infer we will have to give an optional parameter that stores variable state names in Factor. Then we can have an attribute that stores a one to one mapping which could be used by the decorator that replaces state names to internal representation and vice versa.

ankurankan:
@yashu-seth Sure go ahead. Yes, but I don't want any part of code related to state names in any of the existing classes. So, I would suggest you to come up with exact implementation details and discuss it before starting to writing code.

yashu-seth:
@ankurankan Okay. Here is what I have planned.
[...]

Issue #423, opened 2015-06-17
Cache reduce and marginalize


ankurankan:
There are many instances (ex: while generating samples) when reduce or marginalize if called multiple times with the same arguments. Therefore caching the results is a good idea.

khalibartan:
@ankurankan I would like to work on this. I have a very naive implementation, I was thinking of maintaining an additional dictionary attribute in factor class and storing the results in them. Do you have any other implementation in mind.

ankurankan:
@khalibartan I want a more generic solution. Something like a decorator which we can put on any method that we want to cache. Something like `lru_cache` (https://docs.python.org/3/library/functools.html#functools.lru_cache)

khalibartan:
@ankurankan I'll create my own decorator, since `lru_cache`, don't work with non hashable type of arguments and in our case arguments are mostly lists. Also I wasn't able to find `lru_cache` implementation in python2.7

khalibartan:
@ankurankan I have implemented a decorator for `pseudo lru cache` but I wasn't able to work around with `inplace` argument.

yashu-seth:
@khalibartan Can you tell me what was the problem you faced with the inplace argument? Were you trying to write a generic decorator for this?

khalibartan:
@yashu-seth Yes, I was trying to write a generic decorator for this. I cannot remember what was issue with the inplace argument. here is the link to pull https://github.com/pgmpy/pgmpy/pull/598

yashu-seth:
@khalibartan Do you still plan to work on this or should I look into this?

khalibartan:
@ankurankan Told me to leave this issue, as it is of no use now.

yashu-seth:
@khalibartan @ankurankan But wouldn't this save a lot of extra calls?

khalibartan:
@yashu-seth Now I remember what was the issue with `inplace` argument . If `inplace = True` than hash value would be changed and we will not be able to use the cached value.

PR #648, opened 2016-02-24
Variable State Name Feature

yashu-seth:
Fixes issue #232


ankurankan:
@yashu-seth The decorators should not be in `Factor.py`. Move it to a new file in `utils`.

yashu-seth:
@ankurankan Okay. I also thought the same. Will do it.

ankurankan:
@yashu-seth I don't think we should have a separate decorator for each method. I would like to have a single decorator which works on all the methods. So, basically what you can do is that you can pass the argument name (which has the state names) to the decorator and it should be able to handle it. And it should also deal with the case when no state names are given.

ankurankan:
@yashu-seth: So the final thing should be looking something like:

``` python
@deco(argument=variables, return=None)   # return is for the case when there is some return value which needs to be converted.
def __init__(self, variables, cardinality, values):
     """
     Initialize a factor class.
```

And this way you should be able to apply the same decorator to all the methods. But also keep in mind that the arguments or return values might be in different formats.

I would suggest you to come up with different formats in which the argument/return can be and they accordingly handle all those cases.

yashu-seth:
@ankurankan Ok. I will create a single decorator that will handle all the cases.

yashu-seth:
@ankurankan Somthing like this - 

``` python
@stateNameDecorator('init')
def __init__(self, variables, cardinality, values):
         """
         Initialize a factor class.
......

@stateNameDecorator('reduce')
def reduce((self, values, inplace=True):
        """
        Reduces the factor to the context of given variable values.
......

@stateNameDecorator('str')
def _str(self, phi_or_p="phi", tablefmt="fancy_grid"):
        """
        Generate the string from `__str__` method.
........
```

The decorator will handle the different cases using if else.
For the same argument type the decorator input will be same.

ankurankan:
@yashu-seth No something like this:

``` python
@stateNameDecorator('variables')
def __init__(self, variables, cardinality, values):
         """
         Initialize a factor class.
......

@stateNameDecorator('values')
def reduce((self, values, inplace=True):
        """
        Reduces the factor to the context of given variable values.
......

@stateNameDecorator('str') # Don't need a decorator here since we don't have any argument which needs to be converted
def _str(self, phi_or_p="phi", tablefmt="fancy_grid"):
        """
        Generate the string from `__str__` method.
........
```

yashu-seth:
@ankurankan But the variables or the values input type may not always be the same. It would mostly depend on the method type. Hence I am suggesting something similar only the decorator should get method type string as input. I will ensure that the same mapping format will be there for same argument type if such a case arises. What do you think?

yashu-seth:
@ankurankan And in `_str` method there is no argument to be mapped but string that is returned should be changed. I think this will also be done by the decorator.

yashu-seth:
@ankurankan I have named the file VariableStateNameDecorator.py in pgmpy/utils. Please tell me if the file needs to be renamed. I don't why but somehow I am not good with naming things.

pgmpy/utils/VariableStateNameDecorators.py
```diff
@@ -0,0 +1,119 @@
...
+    {'x1': ('on', 'off'), 'x2': ('low', 'medium', 'high'), 'x3': ('day', 'night')}
+    >>> phi.reduce([('x1', 'on'), ('x2', 'low')])
+    >>> print(phi)
+    ╒═══════════╤═══════════╕
+    │ x3        │   phi(x3) │
+    ╞═══════════╪═══════════╡
+    │ x3(day)   │    1.0000 │
+    ├───────────┼───────────┤
+    │ x3(night) │    1.0000 │
+    ╘═══════════╧═══════════╛
```
khalibartan: Use python 2 output here, instead of python3

yashu-seth: @khalibartan These are just temporary docs that I put for clarity. Besides,
the entire class will be reimplemented.

khalibartan: @yashu-seth  Yeah read the above discussion .

khalibartan:
@ankurankan I agree with @yashu-seth about `'str'`. IMO also string representation has to be changed even if there is no mapping to be done.

ankurankan:
@khalibartan @yashu-seth I forgot to mention this in my previous comment. The decorator should be taking in two arguments: the argument name which needs to be converted and the variable which is getting returned and needs to be converted. In the case of `__str__` you should be able to write a separate intermediate method which just deals with returning the states and put this decorator on that so that it converts it into state names and then use this returned value in the `__str__` method to get the desired result.

khalibartan:
Just I wanted to ask, how will we deal with performance issues in case of inference algorithm ( variable elimination ) where reduce is repeatedly called. If for each reduce the decorator is called it will add an additional overhead.

yashu-seth:
@ankurankan But anyway the decorator has to behave differently for different methods. So why keep the two arguments? Can't we just tell the decorator which method it will currently decorate because anyway the arguments are in different formats. So instead of checking various argument types, can't we just check which method needs to be decorated and make the decorator behave accordingly?

ankurankan:
@khalibartan In python3 we can directly access the wrapped method. But am not sure about the performance difference in python 2. But shouldn't be much and ideally it should be just an extra check.

@yashu-seth No, it's not the same. What I meant is that the arguments for states in any method would either be just a state in the form of `(var, state)` (I don't think we are passing a single state anywhere but would be good to have) or a list of states in the form of `[(var1, state1), (var2, state2)]`, so we can simply handle the conversion based on whether we have a single state or a list of states. For example in the case of `reduce` your argument will be a list of states, similarly in the case of `query` method of inference algorithms you will be getting a list of states.

Also if we implement it in this way we won't need to change the decorator in future and it should work on any new method that we add.

yashu-seth:
@ankurankan And what about the **init** methods we need to make the state names dict as an attribute.

@ankurankan Then can't we just give strings like 'list of states' or 'dict of states' as arguments to the decorator? If we add a new method we won't have to change the decorator then.

ankurankan:
@yashu-seth You can write a separate decorator for that or add an extra argument for that. Something like:

``` python
@decorator(set_values=this_is_for_the_init_method, argument=arguments, return=return_var)
```

I am unable to understand what you mean by giving strings as argument, can you please elaborate?
Also since we are on the topic of decorators, I would highly recommend you to go through this blog series: http://blog.dscpl.com.au/search/label/decorators.

yashu-seth:
@ankurankan Because any new method in future may have different parameter names, but the parameter types handling state names would be one of them (list of states or dict of states etc.). And of course it will also have a second argument to describe return type.
Something like this - 
```python
@stateNameDecorator('list of states', None)
def reduce(...
....
@stateNameDecorator(None, 'list of states')
def assignment(...
......
```

yashu-seth:
@ankurankan I will go through that blog. Just to be sure, the '@ statement' that we write is similar to function calls, hence we will have to give actual parameters to it.  Isn't it? Hence I was talking about giving string as arguments. Please tell me if I have a wrong idea of using the '@ statements' ?

Can we use this - 
```python
@decorator(set_values=this_is_for_the_init_method, argument=arguments, return=return_var)
```
I think we will have to use it like this - 
```pytohn
@decorator('this_is_for_the_init_method', 'arguments', 'return_var')
```

yashu-seth:
@ankurankan Please ignore my previous comment. I have deleted it.
In

``` python
@decorator(set_values=this_is_for_the_init_method, argument=arguments, return=return_var)
```

this_is_for_the_init_method , arguments and return_var should be strings right?

yashu-seth:
@ankurankan What I am suggesting is instead of only checking the type of argument for states or list of states etc. in the decorator, we should provide the type of input as string along with the argument.
Something like - 

``` python
@stateNameDecorator(arg='variables', arg_type='list of states', return_val=None)
```

or

``` python
@stateNameDecorator(arg='values', arg_type='states', return_val=None)
```

or

``` python
@stateNameDecorator(arg=None, arg_type=None, return_val='list of states')
```

yashu-seth:
@ankurankan I will write a separate decorator for the **init** methods.

yashu-seth:
@ankurankan Giving the arg_type to the deocrator will save checking the argument type every single time.

ankurankan:
@yashu-seth I don't think it will. You will anyways be checking what is the arg_type.

yashu-seth:
@ankurankan Okay, I will go ahead with this the way you have suggested.

yashu-seth:
@ankurankan Please have a look at this. 
I will soon add the wrapper condition for _str methods.

pgmpy/factors/Factor.py
```diff
@@ -159,6 +161,7 @@ def get_cardinality(self, variables):
 
         return {var: self.cardinality[self.variables.index(var)] for var in variables}
 
+    @stateNameDecorator(argument=None, return_val=True)
```
ankurankan: @yashu-seth It's also possible that in some case we return more than one values and out of them one of them is a state. You will need to handle that case.

yashu-seth:
@ankurankan You mean one of the returned objects could be either list of states or dict of states etc.?

yashu-seth:
@ankurankan And I guess I have missed one case for list of list of states. I'll add that too.

yashu-seth:
@ankurankan Because of a number of internal calls of `reduce`, can we have a special argument for reduce that can deal with this case and skip the mapping?
Something like - 

``` python
@stateNameDecorator(argument='reduce', return_val=None)
    def reduce(self, values, inplace=True):
```

yashu-seth:
@ankurankan Because of a number of internal calls of ```reduce```, can we have a special argument for reduce that can deal with this case and skip the mapping?
Something like - 
```python
@stateNameDecorator(argument='reduce', return_val=None)
    def reduce(self, values, inplace=True):
```

ankurankan:
@yashu-seth I am not getting what you are proposing. Can you please elaborate a bit ?

pgmpy/utils/VariableStateNameDecorators.py
```diff
@@ -0,0 +1,171 @@
+"""
+The file contains decorators to manage the user defined
+variable state names. It maps the internal representaion
+of the varibale states to the user defined state names and
+vice versa.
+"""
+
+class stateNameInit():
```
ankurankan: @yashu-seth https://www.python.org/dev/peps/pep-0008/#class-names

yashu-seth:
@ankurankan I have added the case that you suggested. Please have a look.

yashu-seth:
@ankurankan For `str` methods I think that it would be better to add a condition for state names in the original method itself. I think this would be more feasible because,
- a simple condition in the original method can accomplish this.
- since the output uses the `tabulate` function, I don't think it would be easy to map the output to user defined state names directly. This can be achieved by the decorator but would result in a lot of code duplication. 
  What do you think about this?

ankurankan:
@yashu-seth We can keep the default value of `__str__` to show state names and add an extra argument if we want to see the current representation.

yashu-seth:
@ankurankan I have added the state names condition in the `__str__` methods as suggested. Should I proceed to writing tests?

ankurankan:
@yashu-seth Yes please go ahead.

yashu-seth:
@ankurankan I have added the required tests. This is ready for a final review.

pgmpy/utils/state_name.py
```diff
@@ -0,0 +1,201 @@
...
+    {'x1': ['on', 'off']}
+    where, 'x1' is a variable and 'on', 'off' are its correspinding
+    states.
+
+    Example
+    -------
+    >>> import numpy as np
+    >>> from pgmpy.factors import Factor
+    >>> sn = {'speed': ['low', 'medium', 'high'],
+             'switch': ['on', 'off'],
```
ankurankan: @yashu-seth Fix this please. You need to start a continuation line with triple dots in doctests. Check other files for example.

yashu-seth:
@ankurankan I have made the suggested changes in the docs.

yashu-seth:
@ankurankan I have made the necessary changes. This is ready for a final review. Please have a look at this whenever you get time. 
It would be great if you could provide your input on issue #423

khalibartan:
Great to see getting this PR merged.

Issue #652, opened 2016-02-27
Another Marginalize and reduce issue.


anaviltripathi:
As I have mentioned this particular issue in some other PRs. I think marginalize and reduce should not be allowed to modify the CPD inplace if the user wants to reduce/marginalize with respect to the main variable itself.
I think so because after such operation I am not able to comprehend the meaning of the new CPD. Also, it would give errors on further operations if this allowed. I think @khalibartan agrees with me on this.
@ankurankan If you could give your views on this.

ankurankan:
@anaviltripathi Yes you are right we shouldn't allow operations on the variable on which it's defined as it makes no sense. Please go ahead with this.

PR #654, opened 2016-03-03
Fixed reduce, marginalize and tests

### LINKED ISSUE #651: Marginalize and reduce can be improved using new copy method of CPD
@ankurankan This is very easy and I would like to add this.

anaviltripathi:
Fixes #651, #652. I have also fixed tests.

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -327,19 +327,15 @@ class TabularCPD(Factor):
         >>> cpd_table = TabularCPD('grade', 2,
         ...                        [[0.7, 0.6, 0.6, 0.2],[0.3, 0.4, 0.4, 0.8]],
         ...                        ['intel', 'diff'], [2, 2])
-        >>> cpd_table.marginalize('diff')
+        >>> cpd_table.marginalize(['diff'])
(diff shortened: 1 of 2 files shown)


pgmpy/factors/CPD.py
```diff
@@ -327,19 +327,15 @@ def marginalize(self, variables, inplace=True):
...
         """
-        if inplace:
-            tabular_cpd = self
-        else:
-            evidence = self.variables[1:]
-            evidence_card = self.cardinality[1:]
-            tabular_cpd = TabularCPD(self.variable, self.variable_card,
-                                     self.get_cpd(), evidence,
-                                     evidence_card)
+        if inplace and self.variable in variables:
```
ankurankan: @anaviltripathi I think we shouldn't allow it at all, no matter if it's inplace or not. What do you think?

anaviltripathi: @ankurankan I was a bit confused on this, I thought may be someone may want this under some scenarios. I didn't really put much of a thought in this though, I will change it if you think so.

ankurankan: @anaviltripathi I don't think we should be allowing this in any case.

Issue #590, opened 2016-01-19
New test for Factors


khalibartan:
I was thinking of rewriting the test for Factor class.
- [ ] I'll separate the tests for Factor, TabularCPD and JointProbabilityDistribution classes and will try to add more tests for Factors and JointProbabilityDistribution (not for CPD because @ankurankan is looking forward to remove some redundant properties from it).
- [ ] I'll test cases where variables can be a list hashable objects like tuples etc.
- [ ] I hope to write tests that which checks the boundary conditions for the methods.
- [ ] Coverage of the current skipping test for reduce method.

@ankurankan Should I do the following things or not.
Other people can also add in there views, related to tests.

mvnnn:
@khalibartan  i am new contributor. i like to work on this. should i do ?

khalibartan:
@mvnnn This is not an easy issue in my view, actually I wanted to work on this one, I haven't discuss things with @ankurankan .

ankurankan:
@khalibartan You can go ahead with this. We need to clean up a lot of tests. Would be a good start.

khalibartan:
@ankurankan Will send a PR ASAP.

khalibartan:
Would someone like to work on it. It looks like I won't be able to give necessary time because of my mid-term examinations. I would be great if someone could finish it up. Otherwise I'll work on it after my mid-terms are over.

ankurankan:
Hi Developers,

I am a newbie here but would like to give this a shot. Can someone please
elaborate a bit on the issue?

Thanks

khalibartan:
What we are trying to do is have new test for our Factor class objects. The current tests are not testing things to the boundary conditions. Also test don't test the factor class , where scope can be a list of immutable objects.

abhilashrj:
@anaviltripathi Are u working on this right now?

ankurankan:
@abhi95 I don't think anyone is working on this right now. So, you can go ahead with this if you like.

anaviltripathi:
@abhi95 No, I am not working on this. You can go ahead with, and make sure to include some asymmetric cases whenever you write tests. Currently many tests use symmetric factors/CPDs.

abhilashrj:
@anaviltripathi by asymmetric factors do you mean variables with different cardinality

anaviltripathi:
@abhi95 Yes, and different values too at different places in the table. Some of the current tests have the same set of numbers for all the places.

abhilashrj:
[...]

PR #656, opened 2016-03-04
Tests for Factor where variables are hashable #590.

abhilashrj:
@ankurankan I have added tests where variables of Factor are tuples. Please review and suggest changes if any. What other hashable types can variables of Factor take?

### DIFF
--- a/pgmpy/tests/test_factors/test_Factor.py
+++ b/pgmpy/tests/test_factors/test_Factor.py
@@ -56,41 +56,42 @@ class TestFactorMethods(unittest.TestCase):
                            card3, np.arange(np.prod(card3), dtype=np.float))
         tup1 = ('x1', 'x2')
         tup2 = ('x2', 'x3')
-        tup3 = ('x3', 'x4')
-        self.phi4 = Factor([tup1, tup2, tup3],[2, 3, 4],np.random.uniform(3, 10, size=24))
-        self.phi5 = Factor([tup1, tup2, tup3],[2,3,4],range(24))
+        tup3 = ('x3', (1, 'x4'))
+        self.phi4 = Factor([tup1, tup2, tup3], [2, 3, 4], np.random.uniform(3, 10, size=24))
+        self.phi5 = Factor([tup1, tup2, tup3], [2, 3, 4], range(24))
 
-        card6 = [4,2,1,3,5,6]
+        card6 = [4, 2, 1, 3, 5, 6]
         self.phi6 = Factor([tup1, tup2, tup3, tup1+tup2, tup2+tup3, tup3+tup1],
-                               card6, np.arange(np.prod(card6), dtype = np.float))
+                           card6, np.arange(np.prod(card6), dtype=np.float))
 
     def test_scope(self):
         self.assertListEqual(self.phi.scope(), ['x1', 'x2', 'x3'])
         self.assertListEqual(self.phi1.scope(), ['x1', 'x2', 'x3'])
-        self.assertListEqual(self.phi4.scope(), [('x1','x2'),('x2','x3'),('x3','x4')])
+
+        self.assertListEqual(self.phi4.scope(), [('x1', 'x2'), ('x2', 'x3'), ('x3', (1, 'x4'))])
 
     def test_assignment(self):
         self.assertListEqual(self.phi.assignment([0]), [[('x1', 0), ('x2', 0), ('x3', 0)]])
         self.assertListEqual(self.phi.assignment([4, 5, 6]), [[('x1', 1), ('x2', 0), ('x3', 0)],
-                                                              [('x1', 1), ('x2', 0), ('x3', 1)],
-                                                              [('x1', 1), ('x2', 1), ('x3', 0)]])
+                             [('x1', 1), ('x2', 0), ('x3', 1)],
+                             [('x1', 1), ('x2', 1), ('x3', 0)]])
 
         self.assertListEqual(self.phi1.assignment(np.array([4, 5, 6])),
                              [[('x1', 0), ('x2', 2), ('x3', 0)],
-                              [('x1', 0), ('x2', 2), ('x3', 1)],
-                              [('x1', 1), ('x2', 0), ('x3', 0)]])
-
-        self.assertListEqual(self.phi4.assignment(np.array([11,12,23])),
-                              [[(('x1','x2'), 0), (('x2','x3'), 2), (('x3','x4'), 3)],
-                               [(('x1','x2'), 1), (('x2','x3'), 0), (('x3','x4'), 0)],
-                               [(('x1','x2'), 1), (('x2','x3'), 2), (('x3','x4'), 3)]])
+                             [('x1', 0), ('x2', 2), ('x3', 1)],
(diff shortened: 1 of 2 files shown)


ankurankan:
@khalibartan Can you please review this ?

pgmpy/tests/test_factors/test_Factor.py
```diff
@@ -124,21 +161,39 @@ def test_normalize(self):
...
         values = [('A', 0), ('D', 0), ('F', 0), ('H', 1)]
         phi3_reduced = self.phi3.reduce(values, inplace=False)
         # Previously a sorting error caused these to be different
         np_test.assert_array_equal(phi3_reduced.values.shape, phi3_reduced.cardinality)
 
+        values = [(('x1','x2'), 0), (('x3','x4','x1','x2'), 0)]
+        phi6_reduced = self.phi6.reduce(values, inplace = False)
+        np_test.assert_array_equal(phi6_reduced.values.shape, phi6_reduced.cardinality)
+
     @unittest.skip
```
khalibartan: @abhi95 Fix this skipping test.

pgmpy/tests/test_factors/test_Factor.py
```diff
@@ -124,21 +161,39 @@ def test_normalize(self):
...
     def test_reduce_shape(self):
         values = [('A', 0), ('D', 0), ('F', 0), ('H', 1)]
         phi3_reduced = self.phi3.reduce(values, inplace=False)
         # Previously a sorting error caused these to be different
         np_test.assert_array_equal(phi3_reduced.values.shape, phi3_reduced.cardinality)
 
+        values = [(('x1','x2'), 0), (('x3','x4','x1','x2'), 0)]
+        phi6_reduced = self.phi6.reduce(values, inplace = False)
+        np_test.assert_array_equal(phi6_reduced.values.shape, phi6_reduced.cardinality)
+
```
khalibartan: @abhi95 Try reducing parameters with `inplace = True`

pgmpy/tests/test_factors/test_Factor.py
```diff
@@ -54,10 +54,20 @@ def setUp(self):
         card3 = [3, 3, 3, 2, 2, 2, 2, 2, 2]
         self.phi3 = Factor(['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I'],
                            card3, np.arange(np.prod(card3), dtype=np.float))
+        tup1 = ('x1', 'x2')
+        tup2 = ('x2', 'x3')
+        tup3 = ('x3', 'x4')
```
khalibartan: @abhi95 If you could replace this tuple by something `('x3', (1, 'x4') )`, since we don't restrict variables to have similar types. Try even the cases like where a variable is a tuple and other one a frozenset.

pgmpy/tests/test_factors/test_Factor.py
```diff
@@ -99,6 +118,17 @@ def test_marginalize(self):
         self.phi1.marginalize(['x3'])
         np_test.assert_array_equal(self.phi1.values, np.array(66))
 
+    def test_marginalize_hashable(self):
```
khalibartan: Rename this function with test_marginalize1 or test_marginalize2 something like that.

pgmpy/tests/test_factors/test_Factor.py
```diff
@@ -146,6 +201,11 @@ def test_complete_reduce(self):
         np_test.assert_array_equal(self.phi1.cardinality, np.array([]))
         np_test.assert_array_equal(self.phi1.variables, OrderedDict())
 
+        self.phi5.reduce([ (('x1', 'x2'),1), (('x1', 'x2'),0), (('x1', 'x2'),3) ])
```
khalibartan: Follow PEP8.

khalibartan:
@abhi95 We use PEP8 coding style , please follow that https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md#code-guidelines .
Also you haven't implemented tests for divide and marginalize.Implement them and also add test for checking the `__hash__` method works fine with these kind of parameters.

abhilashrj:
@khalibartan Thanks for the input. I will make necessary changes.

pgmpy/tests/test_factors/test_Factor.py
ankurankan: @khalibartan @abhi95 Don't use `1` or `2` in function names. Come up with something that describes what the test is for.

ankurankan:
@khalibartan @abhi95 For PEP8 we allow 120 char line instead of 79.

pgmpy/tests/test_factors/test_Factor.py
khalibartan: @abhi95  Okay then, use `test_marginalize_scope_hashable` or something like that.

khalibartan:
@ankurankan I have mentioned that in guide to which link points.
Anyways I wanted to ask that we should work with 79 character limit not 120. While working with split screen (for my case its gvim), I find it really difficult to work. I suggest that we should try to follow 79.

yashu-seth:
@ankurankan @khalibartan I feel the same. Is there a particular reason for keeping the 120 char limit?

abhilashrj:
@ankurankan @khalibartan I have made the necessary changes.
I had to skip tests for hash function because of the following error.

```
  File "/home/abhilash/FOSS/pgmpy/pgmpy/tests/test_factors/test_Factor.py", line 442, in test_hash
    self.assertEqual(hash(phi3), hash(phi4))
  File "/home/abhilash/FOSS/pgmpy/pgmpy/factors/Factor.py", line 783, in __hash__
    variables = sorted(self.variables)
TypeError: unorderable types: tuple() < str()
```

We cannot use sorted function when variables are of different types. A solution to this would be to write a function which would decide prioritize the order of datatypes.

yashu-seth:
@abhi95 This seems to be a bug in the `hash` method. Can you fix this? If you want you can also open a separate issue for this.

abhilashrj:
@yashu-seth I will open an issue for this and look into it.

ankurankan:
@abhi95 You can test it with a custom hashable object. Like write a simple class. Add a `__hash__` method in it and test if everything works.

abhilashrj:
@ankurankan I will add tests with custom hashable objects. 
Please have a look at #661. Is it okay if I commit to this PR, the fix for #661.

ankurankan:
@abhi95 #661 is will need some fix in Factor's `__hash__` method. So create a separate PR for that.

abhilashrj:
@ankurankan @khalibartan I have added test cases with custom hashable objects. Please check.

khalibartan:
@abhi95 Tests are falling. Can you please check?

ankurankan:
@abhi95 Looks like a python 2 and 3 compatibility issue. We use six: https://pythonhosted.org/six/ in pgmpy to make it compatible to both 2 and 3. So, look into it in case if tests are failing because of this.

abhilashrj:
@ankurankan I think if u merge https://github.com/pgmpy/pgmpy/pull/665 it will work. I saw the failing test cases output it is because of that bug.
@khalibartan Thanks for reviewing.

PR #662, opened 2016-03-08
Fixes bug in repetitive calling of GibbsSampling.sample

khalibartan:
@ankurankan Actually bug was there because when the first time call was being made

``` python
        if start_state is None and self.state is None:
            self.state = self.random_state()
        else:
            self.set_start_state(start_state)
```

both `start_state` and `self.state` were None but next time only `start_state` was None, and which was being set as `self.state`


khalibartan:
@ankurankan I was just about fix the corresponding line in generate sample. Could you fix that?

PR #663, opened 2016-03-09
Adding fixes for the DBN initialze_intial_state method

aelnouby:
This is a proposed fix for the issue in #452

The problem as i understood it, was that nodes with no intra-connections aren't replicated across time slices, this was fixed in the `add_edge` method

Also, the new replica if in the 0th slice, it won't have any parents, so for its cpd i used marginalization over 1th slice parents.

This an example that used to fail previously:

``` python
from pgmpy.models import DynamicBayesianNetwork as dbn
from pgmpy.factors import TabularCPD
import numpy as np
model = dbn()
model.add_edges_from([(('a',0),('b',1)),(('c',0),('b',1))])
values = np.split(np.array([0.2,0.3,0.4,0.1]),2)
cpd = TabularCPD(('b',1), 2, [[0.5,0.5,0.5,0.5],[0.5,0.5,0.5,0.5]], [('c',0),('a',0)],[2,2])
model.add_cpds(cpd)
model.initialize_initial_state()
```

### DIFF
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -191,6 +191,8 @@ class DynamicBayesianNetwork(DirectedGraph):
 
         if start[1] == end[1]:
             super(DynamicBayesianNetwork, self).add_edge((start[0], 1 - start[1]), (end[0], 1 - end[1]))
+        else:
+            super(DynamicBayesianNetwork,self).add_node((end[0], 1 - end[1]))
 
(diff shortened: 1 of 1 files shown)


khalibartan:
@aelnouby Also fix the error mentioned in #557 .

aelnouby:
@khalibartan I think this PR already fixes this issue as well, the issue was that there are nodes that weren't replicated, so when you explicitly added nodes, they were added. 

Right now, nodes are supposedly replicated correctly.

PR #578, opened 2016-01-15
Fixed bug in hash method for factor class

khalibartan:
The earlier method took order into account .see #574 
Ping @ankurankan , @yashu-seth I don't like this implementation much. Can you suggest any other ?

Issue #574, opened 2016-01-15
Bug in hash method for factor class


khalibartan:
The current implementation takes order into account in the case where it shouldn't

```
In [10]: phi1 = Factor(['x2','x1'], [2,2], [1,3,2,4])

In [11]: phi = Factor(['x1','x2'], [2,2], [1,2,3,4] )

In [12]: hash(phi)
Out[12]: -8224978590549700758

In [13]: hash(phi1)
Out[13]: 851726010497848334

In [14]: phi1 == phi
Out[14]: True

In [18]: print(phi)
╒══════╤══════╤══════════════╕
│ x1   │ x2   │   phi(x1,x2) │
╞══════╪══════╪══════════════╡
│ x1_0 │ x2_0 │       1.0000 │
├──────┼──────┼──────────────┤
│ x1_0 │ x2_1 │       2.0000 │
├──────┼──────┼──────────────┤
│ x1_1 │ x2_0 │       3.0000 │
├──────┼──────┼──────────────┤
│ x1_1 │ x2_1 │       4.0000 │
╘══════╧══════╧══════════════╛

In [19]: print(phi1)
╒══════╤══════╤══════════════╕
│ x2   │ x1   │   phi(x2,x1) │
╞══════╪══════╪══════════════╡
│ x2_0 │ x1_0 │       1.0000 │
├──────┼──────┼──────────────┤
│ x2_0 │ x1_1 │       3.0000 │
├──────┼──────┼──────────────┤
│ x2_1 │ x1_0 │       2.0000 │
├──────┼──────┼──────────────┤
│ x2_1 │ x1_1 │       4.0000 │
╘══════╧══════╧══════════════╛

```

As per my understanding it should have been the same.

khalibartan:
@ankurankan @yashu-seth Am I wrong here ?

yashu-seth:
@khalibartan I also think that it should not take order into account.

ankurankan:
@khalibartan You are right. Both should give the same hash value. @khalibartan @yashu-seth Would any of you like to fix this ?

khalibartan:
@ankurankan I can think of only one implementation and that happens to be quite slow. I'll try to replicate what ` _str ` does and then optimally apply the hash function. But since `__hash__` method has to be used frequently I want its implementation to be robust and fast too. This implementation will be as good as python hash methods in terms of robustness but will be slow. Do you have any other implementation in mind .

yashu-seth:
[...]

PR #665, opened 2016-03-09
Fixes bug in hash method of Factor

abhilashrj:
@ankurankan @khalibartan Please review and suggest changes.


pgmpy/factors/Factor.py
```diff
@@ -780,7 +780,9 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
     def __hash__(self):
-        variables = sorted(self.variables)
+        var_with_hashes = [(var, hash(var)) for var in self.variables]
```
khalibartan: @abhi95 You could have used map instead. Also, why you used `(var, hash(var))` not `(hash(var), var)`

khalibartan:
@abhi95 Add the corresponding tests.

abhilashrj:
@khalibartan I made the necessary changes and used maps.

pgmpy/factors/Factor.py
```diff
@@ -780,7 +780,9 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
     def __hash__(self):
-        variables = sorted(self.variables)
+        var_with_hashes = list(map((lambda var: (hash(var), var)), self.variables))
```
ankurankan: @abhi95 Any specific reason for using `(hash(var), var)` ? I mean using a tuple here.

ankurankan: @abhi95 No need to explicitly convert it to `list`. `sorted` can handle `map` objects.

abhilashrj: @ankurankan (hash(var), var) is required because we also need the variable ordering to generate the hash of Factor object.
I thought map can just be used to operate on all list values similarly. 
My bad, I could have used the map object itself to sorted as it takes any iterable.

ankurankan: @abhi95 Can't we just replace the variables with their hash values ?

khalibartan: @ankurankan I don't think we can. If you see how we are finding the "hash value of factor" by swapping the axes according to the sorted order list of variables. Otherwise we have to change the code, but I cannot think of any more efficient way for [this](https://github.com/pgmpy/pgmpy/pull/665/files#diff-6b19001780e97d8d06329b912a2c0e03L785).

abhilashrj: @ankurankan Even I can't think of a better method here.

abhilashrj:
@ankurankan I used map object in sorted function. Suggest changes if any.

pgmpy/factors/Factor.py
```diff
@@ -780,7 +780,9 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
     def __hash__(self):
-        variables = sorted(self.variables)
+        var_with_hashes = map((lambda var: (hash(var), var)), self.variables)
+        var_with_hashes = sorted(var_with_hashes, key=lambda element: element[0])
```
khalibartan: There is no need of key now.

abhilashrj: Oh it sorts based on first element by default. Thanks for showing.

pgmpy/factors/Factor.py
```diff
@@ -780,7 +780,9 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
     def __hash__(self):
-        variables = sorted(self.variables)
+        var_with_hashes = map((lambda var: (hash(var), var)), self.variables)
+        var_with_hashes = sorted(var_with_hashes)
+        variables = list(map(lambda var_with_hash: var_with_hash[1], var_with_hashes))
         phi = self.copy()
```
ankurankan, line 785: @abhi95 @khalibartan 

``` python
variable_hashes = [hash(variable) for variable in self.variables]
sorted_var_hashes = sorted(variable_hashes)
for axis in range(phi.values.ndim):
    exchange_index = variables_hashes.index(sorted_var_hashes[axis])
    variable_hashes[axis], variable_hashes[exchange_index] = (variable_hashes[exchange_index], variable_hashes[axis])
```

Why shouldn't something like this work ?

khalibartan, line 785: @abhi95 @ankurankan Yeah it should work fine!!

abhilashrj, line 785: @ankurankan @khalibartan I agree. I will commit once more.

pgmpy/factors/Factor.py
```diff
@@ -780,10 +780,12 @@ def __ne__(self, other):
         return not self.__eq__(other)
 
     def __hash__(self):
-        variables = sorted(self.variables)
+        variable_hashes = [hash(variable) for variable in self.variables]
+        sorted_var_hashes = list(sorted(variable_hashes))
```
ankurankan: @abhi95 No need to use `list`. `sorted` returns a list

abhilashrj:
@ankurankan I made the necessary changes.

pgmpy/factors/Factor.py
```diff
@@ -780,10 +780,12 @@ def __ne__(self, other):
...
         phi = self.copy()
         for axis in range(phi.values.ndim):
-            exchange_index = phi.variables.index(variables[axis])
+            exchange_index = variable_hashes.index(sorted_var_hashes[axis])
+            variable_hashes[axis], variable_hashes[exchange_index] = (variable_hashes[exchange_index], variable_hashes[axis])
             phi.variables[axis], phi.variables[exchange_index] = (phi.variables[exchange_index],
```
ankurankan: This is not going to work. Please test it. Also you will have to make changes in the tests.

abhilashrj: @ankurankan I will make necessary changes in the test. What is not going to work can you please explain?

ankurankan: @abhi95 Because we need to use the hash values of the objects for computing the final hash. Right now str (in line 794) would return different value for two different objects having the same attributes.

abhilashrj: @ankurankan Aren't we sorting to avoid that. I mean we are deciding a specific order for variables and making necessary changes to cardinality and values so that the hash is same if variables are rearranged. Did u mean something else here?

khalibartan: @abhi95 You haven't sorted phi.variables. You either need to sort them or don't use them at all (use there hash values instead).  See the earlier bug that was fixed, by me. You can  refer to #574 and try the test case mentioned at there. I believe it fail.

khalibartan: @ankurankan Test are passing , first even I thought they should fail, because of the test case that were written by me in #578 but later I noticed that line 789 hasn't been changes so we still are swapping the things accordingly in variables too, but this shouldn't have been done. @abhi95 in line 794 use hash values instead of variables and remove the line 789. Also please squash the commits. Here is a guide http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

abhilashrj: @khalibartan Do you want me to replace 
`hash(str(phi.variables) + str(phi.values) + str(phi.cardinality))`
with
`hash(sorted_var_hashes+ str(phi.values) + str(phi.cardinality))` ?
I will clean the commit history also.

khalibartan: @abhi95 sorted_var_hashes is an list type object you cannot concatenate it with strings so make changes accordingly. Also I want you to remove line 789 as it is not relevant.

abhilashrj:
@ankurankan @khalibartan Finally the code will look like this.

```
def __hash__(self):
    variable_hashes = [hash(variable) for variable in self.variables]
    sorted_var_hashes = sorted(variable_hashes)
    phi = self.copy()
    for axis in range(phi.values.ndim):
        exchange_index = variable_hashes.index(sorted_var_hashes[axis])
        variable_hashes[axis], variable_hashes[exchange_index] = (variable_hashes[exchange_index], variable_hashes[axis])
        phi.cardinality[axis], phi.cardinality[exchange_index] = (phi.cardinality[exchange_index],
                                                                  phi.cardinality[axis])
        phi.values = phi.values.swapaxes(axis, exchange_index)
    return hash(str(sorted_var_hashes) + str(phi.values) + str(phi.cardinality))
```

khalibartan:
@abhi95 No the join operation will not be collision free. Consider if we have got a factor with two variables with hash value 4 and 9, and another one with hash value 49. In the first case if you concatenate it, we will get collision with the second case. You just simply convert the list to string directly no need to map the things. `str(variable_hashes)` will do the job.

abhilashrj:
@ankurankan Is this fine. Can I commit?

Issue #573, opened 2016-01-15
ValueError: Values array must be of size:64


OKotevska:
I'm having this error, any idea how it can be solved:
data = np.random.randint(low=0, high=4, size=(40, 4))

 raise ValueError("Values array must be of size: {size}".format(size=np.product(cardinality)))
ValueError: Values array must be of size: 64

ankurankan:
@CloudDemo Can you share the whole code  and the complete traceback of the error ?

khalibartan:
@CloudDemo Share the exact steps so that we can reproduce it. And moreover you may like to look at it https://guides.github.com/features/issues/

OKotevska:
This is the program:

import numpy as np
import pandas as pd
from pgmpy.models import BayesianModel

data = np.random.randint(low=0, high=4, size=(40, 4))
data = pd.DataFrame(data, columns=['cost', 'quality', 'location', 'no_of_people'])

train = data[:32] 
test = data[32:].drop('no_of_people', axis=1)

restaurant_model = BayesianModel(
[('location', 'cost'),
('quality', 'cost'),
('location', 'no_of_people'),
('cost', 'no_of_people')])
restaurant_model.fit(train)

---

Traceback (most recent call last):

  File "<ipython-input-39-652112fbc3bb>", line 1, in <module>
    runfile('/home/onk3/UbuntuShared/pgm/pgm-4.py', wdir='/home/onk3/pgmpy')

  File "/home/onk3/anaconda3/lib/python3.5/site-packages/spyderlib/widgets/externalshell/sitecustomize.py", line 699, in runfile
    execfile(filename, namespace)

  File "/home/onk3/anaconda3/lib/python3.5/site-packages/spyderlib/widgets/externalshell/sitecustomize.py", line 88, in execfile
    exec(compile(open(filename, 'rb').read(), filename, 'exec'), namespace)

  File "/home/onk3/UbuntuShared/pgm/pgm-4.py", line 40, in <module>
    restaurant_model.fit(train)

  File "/home/onk3/pgmpy/pgmpy/models/BayesianModel.py", line 557, in fit
    cpds_list = estimator.get_parameters()

  File "/home/onk3/pgmpy/pgmpy/estimators/MLE.py", line 75, in get_parameters
    evidence_card=parent_card.astype('int'))

  File "/home/onk3/pgmpy/pgmpy/factors/CPD.py", line 139, in **init**
[...]

Issue #352, opened 2015-03-28
Bug in BayesianModel's fit method


ankurankan:
``` python3
from sklearn.datasets import load_iris

iris = load_iris()

data = pd.DataFrame(iris.data.astype('int'), columns=['x1', 'x2', 'x3', 'x4'])

data['y'] = iris.target

data = data.reindex(np.random.randint(0, 150, 150))

model = BayesianModel([('x1', 'y'), ('x2', 'y'), ('x3', 'y'), ('x4', 'y')])

model.fit(data)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-107-d9a8c16dbfe9> in <module>()
----> 1 model.fit(data)

/home/ankur/pgmpy/pgmpy/models/BayesianModel.py in fit(self, data)
    541             if not parents:
    542                 state_counts = data.ix[:, node].value_counts()
--> 543                 cpd = TabularCPD(node, node_card[node],
    544                                  state_counts.values[:, np.newaxis])
    545                 cpd.normalize()

/usr/local/lib/python3.4/dist-packages/pandas/core/indexing.py in __getitem__(self, key)
   1198     def __getitem__(self, key):
   1199         if type(key) is tuple:
-> 1200             return self._getitem_tuple(key)
   1201         else:
   1202             return self._getitem_axis(key, axis=0)

/usr/local/lib/python3.4/dist-packages/pandas/core/indexing.py in _getitem_tuple(self, tup)
   1416                 continue
   1417 
-> 1418             retval = getattr(retval, self.name)._getitem_axis(key, axis=axis)
   1419 
   1420             # if the dim was reduced, then pass a lower-dim the next time

/usr/local/lib/python3.4/dist-packages/pandas/core/indexing.py in _getitem_axis(self, key, axis)
   1470                 self._is_valid_integer(key, axis)
   1471 
-> 1472             return self._get_loc(key, axis=axis)
   1473 
   1474     def _convert_to_indexer(self, obj, axis=0, is_setter=False):

/usr/local/lib/python3.4/dist-packages/pandas/core/indexing.py in _get_loc(self, key, axis)
     89 
     90     def _get_loc(self, key, axis=0):
---> 91         return self.obj._ixs(key, axis=axis)
     92 
     93     def _slice(self, obj, axis=0, typ=None):

/usr/local/lib/python3.4/dist-packages/pandas/core/frame.py in _ixs(self, i, axis)
   1726             """
   1727 
[...]

PR #667, opened 2016-03-14
MLE: Fix missing CPD columns

chrisittner:
Fix for a bug in `MaximumLikelihoodEstimator.get_parameters()`, reported in #352, #573 and #579. See my [comment](https://github.com/pgmpy/pgmpy/issues/579#issuecomment-196003936) for explanation.

Additionally changed some docstrings, they weren't saying the truth :).

### DIFF
--- a/pgmpy/estimators/MLE.py
+++ b/pgmpy/estimators/MLE.py
@@ -2,6 +2,7 @@ from pgmpy.estimators import BaseEstimator
 from pgmpy.factors import TabularCPD
 from pgmpy.models import BayesianModel
 import numpy as np
+import pandas as pd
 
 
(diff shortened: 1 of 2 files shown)


chrisittner:
plus basic test for MLE

yashu-seth:
@chrisittner There are other cases for which more tests can be added. I just opened an issue #668 . If you want, please have a look at it.

chrisittner:
@yashu-seth good! I also noticed tests were missing, but here I just included a test for this specific bug. I won't work on #668 this or next week, but might do it if it is still around later on.

yashu-seth:
@chrisittner Cool.

Issue #637, opened 2016-02-18
Update README.md


khalibartan:
Since we now have pgmpy at PyPI , https://github.com/pgmpy/pgmpy/blob/dev/README.md#download , https://github.com/pgmpy/pgmpy/blob/dev/README.md#installation these sections needs updated. Also [ Dependencies ](https://github.com/pgmpy/pgmpy/blob/dev/README.md#dependencies) section doesn't show pyparsing as our dependency . We need to update that too.

PR #674, opened 2016-03-22
Added install instructions

souravsingh:
This patch adds the link to PyPI to Installation and Download sections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ pgmpy has following non optional dependencies:
 - Numpy 1.9.2 
 - Cython 0.21 
 - Pandas 0.15.1 
+- pyparsing 2.0.3
 
 Download
 =========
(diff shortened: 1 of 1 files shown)


souravsingh:
Fixes #637

README.md
```diff
@@ -23,6 +23,7 @@ pgmpy has following non optional dependencies:
 - Numpy 1.9.2 
 - Cython 0.21
```
khalibartan: remove cython also. It is no longer a dependency .

README.md
```diff
@@ -31,7 +32,9 @@ You can either clone the git repo with:
 ```
 git clone https://github.com/pgmpy/pgmpy
 ```
-or download a zip from: https://github.com/pgmpy/pgmpy/archive/dev.zip
+or download a zip from: [https://github.com/pgmpy/pgmpy/archive/dev.zip](https://github.com/pgmpy/pgmpy/archive/dev.zip)
+
+Alternatively, one can download a .tar.gz file from the PyPI archive [https://pypi.python.org/pypi/pgmpy](https://pypi.python.org/pypi/pgmpy)
```
ankurankan, line 37: @souravsingh Mention that it would be the last release version and not the latest code.

README.md
```diff
@@ -56,6 +59,10 @@ Then install using:
 sudo python3 setup.py install
 ```
 
+Alternatively, To install pgmpy, Obtain it from PyPI by using pip
```
ankurankan: @souravsingh This seems better to me:  `Alternatively you can install pgmpy using pip`

ankurankan:
@souravsingh After making the changes, squash all your commits in a single commit.

khalibartan:
Refer #845. And since PyPi version is old and buggy we don't want users to install that.

Issue #534, opened 2015-12-11
Logical Error during Inference


Erotemic:
I've come across what looks like a serious issue. Exact Inference seems to be producing incorrect results for map_query. I've provide a small test case that reproduces this error in both python 2 and python 3. 

**For context**: 
In this section I'm going to describe the context of the problem I'm working on. This will hopefully provide an intuition about the model that is causing the error. 

I am applying Bayesian networks to photographic zebra identification. I have 3 zebras pictures: a, b, and c. I want to know the "name" of each zebra (who each zebra is). For the purposes of this test I'm given an oracle that tells me whether each pair of pictures  (ab, bc, ac) depicts either the same individual or shows different individuals. 

**Network Setup**
I've encoded this knowledge in a Bayesian network with two types of nodes: name nodes and matching nodes. 
- There are 3 name nodes, each representing a different zebra picture. Each of these nodes has a uniform probability distribution over 3 possible name labels. 
- There are also 3 matching nodes for each pair of name nodes. These nodes denote if each pair of pictures matches. The parents of these matching nodes are two name nodes. In this test case two pictures match if and only if they have the same name. 

**The Bug**
In this test case I'm given that a and b are different, a and c are different and b and c are the same. 
More concisely this is what the oracle tells me about each pair of images: ab=different, ac=different, bc=same

Therefore the MAP state assignment of (Na, Nb, and Nc) must be one where Nb and Nc have the same state and Na has a different state from the previous two. There will be an ambiguity between which name is which as a name is simply a label. I would expect the only possible assignments of Na, Nb, and Nc to be: (1, 0, 0), (1, 2, 2), (0, 1, 1), (0, 2, 2), (2, 0, 0), and (2, 1, 1).

However, the MAP probability given by VariableElimination and VariableElimination is Na, Nb, and Nc are assigned to same name state. ie. (0, 0, 0) 

Furthermore, if I manually compute the joint probability distribution by multipling the factors and then reduce over the states where Mab=0, Mac=0, and Mbc=1, and then take the state with the maximum value it gives me a valid assignment of names: (Na=1, Nb=0, Nc=0). Unless I'm incorrect about what map_query is supposed to do, it should return the same value as this brute force method. 

**Code to Reproduce Bug**

``` python
import numpy as np
import pgmpy
from itertools import product
import pgmpy.inference
import pgmpy.factors
import pgmpy.models
import pgmpy.extern

print('Testing map_query bug')

[...]

PR #488, opened 2015-10-09
Optimize variable elimination

ankurankan:
Changes:
1. Adds elimination order 
2. Adds barren nodes 
3. Adds reduce over evidence
4. Changes the expected type of the evidence argument.
5. Remove node method in bayesian network

PR #678, opened 2016-04-12
Fix joint distribution error during Exact Inference

MadDataScience:
This may be the same bug identified by @Erotemic in issue #534

I noticed that, when given multiple variables, _variable_elimination computes
them each independently, marginalizing over the others. This is a problem when
the variables are not independent. I added a flag, marginal_independent,
which, if set to False, will return the joint factor for all variables queried.
It defaults to True in order to maintain backwards compatibility.

The default (old) behavior returns a dictionary where the keys are the variables
and the values are the factors. Since the new version only returns a single
joint probability distribution, it seems unnecessary to return it nested in a
dictionary. As a result, the new version simply returns the factor directly.
I realize this is a no-no as it means the method returns two different data
types based on which flag is set. But it carries the benefit that the query can
now be nested directly in anything that expects a Factor. I hope that this
improved ease of use justifies the breaking of convention, and that we can
eventually drop the old way entirely.

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -12,7 +12,7 @@ from pgmpy.models import JunctionTree
 
 
 class VariableElimination(Inference):
-    def _variable_elimination(self, variables, operation, evidence=None, elimination_order=None):
+    def _variable_elimination(self, variables, operation, evidence=None, elimination_order=None, marginal_independent=True):
         """
         Implementation of a generalized variable elimination.
 
@@ -28,6 +28,11 @@ class VariableElimination(Inference):
         elimination_order: list, array-like
             list of variables representing the order in which they
             are to be eliminated. If None order is computed automatically.
+        marginal_independent: bool,
+            a flag saying whether or not to assume marginal independence on
+            multiple variables. Defaults to True to maintain backwards
+            compatibility, however, should never be true when inferring multiple
+            variables (can always marginalize them later)
         """
         # Dealing with the case when variables is not provided.
         if not variables:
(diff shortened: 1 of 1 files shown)


khalibartan:
@MadDataScience 

> I noticed that, when given multiple variables, _variable_elimination computes
> them each independently, marginalizing over the others. This is a problem when
> the variables are not independent.

Yeah, this is indeed a problem.

Returning two different data types can be really a mess, and user have to be really careful for this. I'll not agree on returning two different data types.

> But it carries the benefit that the query can
> now be nested directly in anything that expects a Factor. I hope that this
> improved ease of use justifies the breaking of convention, and that we can
> eventually drop the old way entirely.

I don't completely agree at here with the advantage. Well we can have an additional parameter for this, I believe that will provide user a more flexibility. Right now if user doesn't want all the marginal factors but a complete factor , he needs to iterate over and than compute what we have already computed, so it is a waste of calculation and time. But if user wants to have marginal factors than current implementation is better.
@ankurankan What are your views?

ankurankan:
@MadDataScience @khalibartan Yes, you guys are right. I think we should by default just return the joint factor on all the variables in the query list. And I don't think we need the extra parameter. Also @MadDataScience you will need to make similar change in `BeliefPropagation`.

MadDataScience:
Okay, I think I did what you asked. I tested it with a BeliefPropagation query with two variables and it returned as expected. I have _not_ tested MAP queries. I left it with the default behavior as before because I didn't want to start mucking around with nosetests and whatever else would blow up if suddenly the default behavior changed.

chebee7i:
It seems like a dev decision is required for this PR to move forward. The implementation here fixes the issue, but I don't like that tries to preserve the existing behavior. In my opinion, when you query with multiple variables, it should always return the joint distribution of those variables not their marginals.

What do you guys want to do? And what else needs to be done?

chebee7i:
Any thoughts on this? I would like to help, but I don't know codebase and ramifications well enough to know what should be done.

ankurankan:
@chebee7i I think this depends on #488. If you would like to work on #488 please go ahead. I think you will need to fix tests mostly.

Issue #677, opened 2016-04-11
fit(data) exemple throws TypeError: ('Values: Expected type int or type float, got ', dtype('int64'))


anapinkg:
From the pgmpy API reference, I tried the example under `fit(data)`:

``` python
import numpy as np
import pandas as pd`
from pgmpy.models import BayesianModel
values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)), columns=['A', 'B', 'C', 'D', 'E'])
model = BayesianModel([('A', 'B'), ('C', 'B'), ('C', 'D'), ('B', 'E')])
model.fit(values)
model.get_cpds()
```

and I got the error

``` bash
> TypeError: ('Values: Expected type int or type float, got ', dtype('int64'))

**Traceback:**
  File "<ipython-input-11-bb819455e71e>", line 1, in <module>
    runfile('/home/ana/.spyder2-py3/temp.py', wdir='/home/ana/.spyder2-py3')

  File "/home/ana/anaconda2/envs/pgmpy-env/lib/python3.4/site-packages/spyderlib/widgets/externalshell/sitecustomize.py", line 699, in runfile
    execfile(filename, namespace)

  File "/home/ana/anaconda2/envs/pgmpy-env/lib/python3.4/site-packages/spyderlib/widgets/externalshell/sitecustomize.py", line 88, in execfile
    exec(compile(open(filename, 'rb').read(), filename, 'exec'), namespace)

  File "/home/ana/.spyder2-py3/temp.py", line 15, in <module>
    model.fit(values)

  File "/home/ana/anaconda2/envs/pgmpy-env/lib/python3.4/site-packages/pgmpy-0.1.0-py3.4.egg/pgmpy/models/BayesianModel.py", line 569, in fit
    cpds_list = estimator.get_parameters()

  File "/home/ana/anaconda2/envs/pgmpy-env/lib/python3.4/site-packages/pgmpy-0.1.0-py3.4.egg/pgmpy/estimators/MLE.py", line 65, in get_parameters
    state_counts.values[:, np.newaxis])

  File "/home/ana/anaconda2/envs/pgmpy-env/lib/python3.4/site-packages/pgmpy-0.1.0-py3.4.egg/pgmpy/factors/CPD.py", line 139, in __init__
    super(TabularCPD, self).__init__(variables, cardinality, values.flatten('C'))

[...]

PR #682, opened 2016-05-02
Fix for np.int64 bug in MaximumLikelihoodEstimation

yashu-seth:
This bug occurs because numpy does not take numpy.int64 to be an instance
of int. 
For more details refer -
https://github.com/numpy/numpy/issues/2951

Fixes issue #677

### DIFF
--- a/pgmpy/factors/Factor.py
+++ b/pgmpy/factors/Factor.py
@@ -96,10 +96,7 @@ class Factor(object):
         if isinstance(variables, six.string_types):
             raise TypeError("Variables: Expected type list or array like, got string")
 
-        values = np.array(values)
-
-        if values.dtype != int and values.dtype != float:
-            raise TypeError("Values: Expected type int or type float, got ", values.dtype)
+        values = np.array(values, dtype=float)
 
         if len(cardinality) != len(variables):
             raise ValueError("Number of elements in cardinality must be equal to number of variables")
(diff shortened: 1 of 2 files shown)


yashu-seth:
@ankurankan @khalibartan

khalibartan:
@yashu-seth remove the files generated during nosetests.

yashu-seth:
@khalibartan Oh sorry. I forgot to remove them.

ankurankan:
@yashu-seth @khalibartan I don't like this. There should be some cleaner and readable solution for this.

ankurankan:
@yashu-seth Can you try checking for integers using `np.integers` ?

yashu-seth:
@ankurankan `np.integer` also does not work. 
An another solution that can work is -

``` python
values.dtype.kind not in np.typecodes['AllInteger']
```

I found this [here](http://stackoverflow.com/a/22472662/5658017). None of the other solutions on the thread work for `int64`.

yashu-seth:
Even this solution does not seem clean. Maybe we can keep it as it is with a comment in the code describing the problem and we can have a separate issue for it. So that it can be dealt later with newer numpy versions.

yashu-seth:
@ankurankan What should be done about this?

yashu-seth:
@ankurankan I think this should be given attention as soon as possible. A lot of windows users might be facing problems. I was playing with the titanic dataset when I got the same error again. Hence, ended up using this branch.

yashu-seth:
@khalibartan @chrisittner Can you please suggest a better solution for this?

chrisittner:
Hm. So in the `Factor.__init__`-method `values` is converted to `np.array`. If the (implicitly and apparently platform-dependent-ly) chosen `dtype` is not `np.int` or `np.float` but some other numeric `dtype` an error is thrown. Why can't we just cast `values` to an `np.float`-array at this point, that seems an appropriate `dtype` for the Factor/CPD values anyways?

yashu-seth:
@chrisittner Yes that seems the best option, we should change the values datatype to float itself It would save us with the type checks too.

yashu-seth:
@ankurankan Please have a look.

PR #686, opened 2016-05-08
Adds base class for continuous node representation

yashu-seth:
This PR deals with the basic continuous node representation feature. It will comprise a base class for Continuous node representation along with various methods to discretize the continuous variables into discrete factors.
This involves the first three weeks of my GSoC project.


yashu-seth:
@ankurankan This is ready for a review. I will then proceed to the tests. Please ignore the PEP8 fixes for now. I'll fix them soon.

yashu-seth:
@ankurankan @khalibartan @chrisittner I cannot figure out why appveyor and travis are failing. There are no such errors on my local environment.

ankurankan:
@yashu-seth: Appveyor will fail. I haven't configured it yet. Will check
for travis.

pgmpy/factors/ContinuousNode.py
```diff
@@ -0,0 +1,230 @@
...
+    the objects.
+    This supports an extra method to discretize the continuous distribution
+    into a discrete factor using various methods.
+    """
+    def __init__(self, pdf, lb=None, ub=None):
+        """
+        Parameters
+        ----------
+        pdf : function
+        The user defined probability density function.
```
ankurankan: @yashu-seth Check out the formatting in other docstrings. Add a few extra spaces on the description line.

yashu-seth: @ankurankan Okay. I will do it.

pgmpy/factors/ContinuousNode.py
```diff
@@ -0,0 +1,230 @@
...
+        super(ContinuousNode, self).__init__(momtype=0, a=lb, b=ub)
+
+    def _pdf(self, *args):
+        """
+        Defines the probability density function of the given
+        continuous variable.
+        """
+        return self.pdf(*args)
+
+    def _rounding_method(self, frm, to, step):
```
ankurankan: @yashu-seth I think all the discretization methods should be a separate class. Each of them should be inheriting a base abstract class. And then we can directly pass that class to the `discretize` method. This will allow the user to add a discretization method very easily. It will also give more flexibility and clarity for writing a new method. What do you think?

yashu-seth: @ankurankan Yes, that would be a great way to do this.

So, here is what I think can be done.

``` python
class BaseDiscretizer():
    def __init__(self, frm, to, step):
         #code
         pass

class RoundingDiscretizer(BaseDiscretizer):
    def get_discrete_values(self):
         # code for discretiztion algorithm
         return factor

class CustomDiscretizer(BaseDiscretizer):
    def __int__(self, node, frm, to, step, function, *args, **kwargs):
        #code
        super(CustomDiscretize, self).__init__(node, frm, to, step)

    def get_discrete_values(self):
         def get_discrete_values(self):
         # code for discretiztion algorithm
         return factor
```

And the Continuous Node class can be like - 

``` python
class ContinuousNode(self, pdf, lb=None, ub=None):
    ...

    def discretize(self, method=None, **kwargs):
         #code
         pass
```

This is somewhat similar to the way the BaseEstimator class has been implemented.

yashu-seth: @ankurankan I think we can have all the continuous part of the library in a separate directory "continuous" with sub-directories  "discretizers" and "represenataion"  etc. or may be "factor" instead of represenatation.

ankurankan: @yashu-seth Yeah, the API looks good. I feel like this class should be called  `ContinuousFactor` ? Because we would be associating the instances of this class to the node just like we associate instances of `Factor` to nodes.

yashu-seth:
@ankurankan I have added the tests as well. This is ready for a review. 
I cannot figure out why travis ci is failing.

yashu-seth:
@ankurankan I think this explains the travis-ci failure - https://github.com/travis-ci/travis-ci/issues/4816
But I could not figure out exactly what is to be done. Please have a look at this.

khalibartan:
@yashu-seth I was thinking that the models, should have `get_value` method, which takes a vector as input ( the values of parameter ) and returns the probability at that setting.

yashu-seth:
@khalibartan That can be done but will make even more stronger assumptions as there will be no continuous range for discretization. Moreover, I think that, to incorporate these continuous factors into pgmpy models discretizing it over continuous interval would be the best fit. I think you can have a look at the R package [discretize {actuar}](http://www.inside-r.org/packages/cran/actuar/docs/discretize). It has also implemented these methods in a similar fashion.

khalibartan:
@yashu-seth  Okay.

ankurankan:
@yashu-seth The failing travis seems to be a conda issue. I trying to fix it. Will let you know.

ankurankan:
@yashu-seth: Finally figured it out. Just add `conda install libgfortran` before the build statement (last line) in `continuous_integration/install.sh`.

yashu-seth:
@ankurankan Thanks, finally Travis-ci does not fail.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,45 @@
...
+    frm, to: float
+        the range over which the function will be discretized.
+    step: float
+        the discretization step (or span, or lag)
+
+    """
+
+    def __init__(self, factor, frm, to, step):
+        if not isinstance(factor, ContinuousFactor):
+            raise ValueError("{} is not a valid ContinuousFactor object."
```
ankurankan: @yashu-seth Put a name inside `{}`

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,45 @@
...
+        the continuous factor representing the distribution
+        to be discretized.
+    frm, to: float
+        the range over which the function will be discretized.
+    step: float
+        the discretization step (or span, or lag)
+
+    """
+
+    def __init__(self, factor, frm, to, step):
```
ankurankan: @yashu-seth I don't think `step` is a required parameter for every discretization function. Think of some method that samples only from high probability regions, the step size will vary for each sample. Having a parameter for no of samples (which would be the states) makes more sense to me. What do you think ?

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Returns a list of strings representing the values about
+        which the discretization method calculates the probabilty
+        masses.
+        Default value is the points -
+        [frm, frm+step, frm+2*step, ......... , to-step]
+        unless the method is overridden by a subclass.
+        """
+        x = np.arange(self.frm, self.to, self.step)
+        labels = list('x='+str(i) for i in x)
```
ankurankan: @yashu-seth Use `.format` instead of `+`.

pgmpy/continuous/discretize/rounding.py
```diff
@@ -0,0 +1,47 @@
...
+
+import numpy as np
+
+from pgmpy.factors import ContinuousFactor
+from pgmpy.continuous.discretize import BaseDiscretizer
+
+
+class RoundingDiscretizer(BaseDiscretizer):
+    """
+    Class for discretization using the rounding method.
```
ankurankan: @yashu-seth This docstring is not very clear. Can you reword it?

pgmpy/continuous/discretize/rounding.py
```diff
@@ -0,0 +1,47 @@
...
+    >>> std_normal.discretize(RoundingDiscretizer, frm=-3, to=3, step=0.5)
+    [0.001629865203424451, 0.009244709419989363, 0.027834684208773178,
+     0.065590616803038182, 0.120977578710013, 0.17466632194020804,
+     0.19741265136584729, 0.17466632194020937, 0.12097757871001302,
+     0.065590616803036905, 0.027834684208772664, 0.0092447094199902269]
+    """
+
+    def get_discrete_values(self):
+        # for x=[frm]
+        discrete_values = [self.factor.cdf(self.frm+self.step/2)
```
ankurankan: @yashu-seth Add spaces before and after `+`.

pgmpy/continuous/discretize/base.py
yashu-seth: @ankurankan Should we rename step to no_of_samples so that it can be used accordingly in each discretization algorithm?

pgmpy/continuous/discretize/rounding.py
ankurankan: @yashu-seth Can't you also include this inside the loop rather that doing this separately ?

pgmpy/continuous/discretize/base.py
ankurankan: @yashu-seth You will also have to make some minor changes. In case of uniform sampling `no_of_samples = (to - frm) / step`

pgmpy/continuous/discretize/rounding.py
yashu-seth: @ankurankan The formula is different for x=[frm] and the other points so I kept it separate for clarity.

pgmpy/continuous/discretize/base.py
yashu-seth: @ankurankan But that would change the way the method is being used. Can't we add an another parameter for step and not use no_of_samples for uniform methods?

yashu-seth: @ankurankan Also I was trying to keep the API similar to R {actuar} discretize methods. Should I just add step as a separate parameter for these methods?

ankurankan: @yashu-seth Change the names `frm` and `to`. Maybe use `low` and `high` as in numpy's API.

ankurankan: @yashu-seth I think keeping `no_of_samples` makes more sense. We would like to tell how many states we want in our discretized factor. If we ask for step size the user will have to calculate it from the number of states.

yashu-seth: @ankurankan Should I do this for the `BaseDiscretizer` or the other classes as well?

pgmpy/continuous/discretize/rounding.py
ankurankan: @yashu-seth Wouldn't looping over `range(self.frm + self.step, self.to, self.step)` give the same result ?

pgmpy/continuous/discretize/base.py
ankurankan: @yashu-seth Everywhere.

pgmpy/continuous/discretize/rounding.py
yashu-seth: @ankurankan I am rewording it to - 
This class uses the rounding method for discretizing the given continuous distribution.
Is it okay?

pgmpy/continuous/discretize/base.py
yashu-seth: @ankurankan Yes, now I think keeping the no_of_samples would be more sensible. I will change it.

pgmpy/continuous/discretize/rounding.py
yashu-seth: @ankurankan For x=[frm], 
the algorithm does not include lower half it only takes cdf from, frm+step/2 to frm
but for other points it takes cdf from x+step/2 to x-step/2.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,45 @@
...
+    def get_labels(self):
+        """
+        Returns a list of strings representing the values about
+        which the discretization method calculates the probabilty
+        masses.
+        Default value is the points -
+        [frm, frm+step, frm+2*step, ......... , to-step]
+        unless the method is overridden by a subclass.
+        """
+        x = np.arange(self.frm, self.to, self.step)
```
ankurankan: @yashu-seth I think all the statements in this method can be combined in a single statement. And if you are not combining don't use names like `x` for variables.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,45 @@
...
+                             .format(factor))
+
+        self.factor = factor
+        self.frm = frm
+        self.to = to
+        self.step = step
+
+    def get_labels(self):
+        """
+        Returns a list of strings representing the values about
```
ankurankan: @yashu-seth Need to add the docstring in the proper format (with examples).

pgmpy/continuous/discretize/rounding.py
ankurankan: @yashu-seth I meant the whole description of Rounding Method is not very clear.

pgmpy/continuous/discretize/unbiased.py
```diff
@@ -0,0 +1,97 @@
...
+from pgmpy.factors import ContinuousFactor
+from pgmpy.continuous.discretize import BaseDiscretizer
+
+
+class UnbiasedDiscretizer(BaseDiscretizer):
+    """
+    Class for discretization using the unbiased method.
+
+    The unbiased method for discretization is the matching of the
+    first moment method. It invovles calculating the first order
```
ankurankan: @yashu-seth Typo in `involves`.

pgmpy/continuous/discretize/rounding.py
yashu-seth: @ankurankan You mean the way I have explained the formulas?

yashu-seth:
@ankurankan Is the way I have described the formulas for the discretization classes in the doc string not clear?

yashu-seth:
@ankurankan I have replaced the step parameter with the cardinality parameter and made the other suggested changes. I have also edited docs for `RoundingDiscretizer` and `UnbiasedDiscretizer` to make it more clear. Please have a look.

ankurankan:
@yashu-seth I would like to suggest a few changes in the API to make it more consistent with the rest of pgmpy.
1. `ContinuousFactor` should accept `variables` as a parameter. In pgmpy, we have tried to keep factor/CPD separate from the model structure and for that factor needs to have the information of the variables on which it is defined.
2. I would suggest to have a directory structure like this:

```
├── factors
│   ├── discrete
│   │     ├── Factor.py
│   │     ├── CPD.py
│   │     ├── other
│   ├── continuous
│   │     ├── Discretize.py
│   │     ├── ContinuousFactor.py
│   │     ├── other
```
1. I don't think we need to have separate files for all the discretize methods. They can be put in the same file.
2. You need to add `get_discrete_values` as an abstract method in the base class because it needs to be defined if the user inherits the base class.
3. Does the individual discretize classes need access to the `ContinuousFactor` object ? We should keep it separate from factor since discretize is a property/function of distribution. So, we should associate discretize with just the pdf.

yashu-seth:
@ankurankan The `ContinuousFactor` class can only represent uni-variate distributions. So it can have a variable attribute. The reason I have not added it is because we cannot use these factors directly into models. After discretiztion we will have to insert the discrete values into a `TabularCPD` or `Factor`. Just adding a variable name every time does not make much sense to me for this class.
Can you please have a look at my blog [post](http://yashu-seth.github.io/blog/2016/05/29/ContinuousFactorDiscretize), I have described how we can use the discrete values to form TabularCPDs. If you have some suggestions or feel otherwise please let me know.

yashu-seth:
@ankurankan I will change the directory and file structure as suggested. I will also add the `get_discrete_values` method as an abstract method in the base class.

yashu-seth:
@ankurankan The individual discretize methods, can use other methods provided by the `scipy.stats.rv_continuous` class. Like, the `RoundingDiscretizer` uses the cdf of the distribution. There can be other methods that require nth order moments in their algorithm too. Just letting the discretizer classes get access to the pdf method would require re-initializing the `ContinuousFactor` object internally to get access to other methods like cdf, moments etc. I think if we have a class that can provide such functionalities, why not use it directly?

yashu-seth:
@ankurankan I think the way I have created this class does not make it a factor representation truly. It is more of a helper class for providing the discretizing functionality. A true factor class would be a class that can represent relations between continuous child - continuous parents and/or discrete child - continuous parents and/or continuous child - discrete parents. 
Like we have `TabularCPD` and `Factor` to represent discrete child - discrete parents relationship and we will have `LinearGaussianCPD` and `CanonicalFactor` for continuous child - continuous parents.

khalibartan:
@ankurankan @yashu-seth For file structure I propose we divide the complete directory structure of pgmpy into discrete and continuous, something like this

```
├── pgmpy
│   ├── discrete
│   │     ├── factors
│   │     │      ├── Factor.py
│   │     │      ├── CPD.py
│   │     ├── models
│   │     ├── inference
│   ├── continuous
│   │     ├── factors
│   │     ├── models
│   │     ├── inference
```

rather than dividing the current individual directory structure into continuous and discrete.

pgmpy/tests/test_continuous/test_discretizers.py
```diff
@@ -0,0 +1,153 @@
...
+import numpy.testing as np_test
+from scipy.stats import gamma, expon
+
+from pgmpy.factors import ContinuousFactor
+from pgmpy.continuous.discretize import BaseDiscretizer
+from pgmpy.continuous.discretize import RoundingDiscretizer
+from pgmpy.continuous.discretize import UnbiasedDiscretizer
+
+
+# class TestBaseDiscretizer(unittest.TestCase):
```
khalibartan: @yashu-seth Why is this section commented ?

yashu-seth: @khalibartan I need to either change them or remove them. The work is still
in progress as there are few things that need to be changed in this PR.

ankurankan:
@yashu-seth Why are you making the `ContinuousFactor` to work only on single variables. No matter how the parent  child relation is (as you have mentioned in the blog post), it's pdf can be represented by a single function. So, this is my idea of `ContinuousFactor`:

``` python
ContinuousFactor
    def __init__(var, pdf):
        "Initialize the values"
    def discretize(method):
        "Return maybe a Factor instance"
    def sum/divide/product(other factor):
        "returns a function (pdf)"
    def marginalize(var):
        "integrate over one axis and return the function (or replace self.pdf if inplace=True)"
    def reduce(var):
        "Return a new function representing the new pdf"
    def assignment(x1, [x2, x3,. ..]):
        "Return an single float value"
```

And then the special cases of say LinearGaussianCPD should inherit this ContinuousFactor class and in the special cases we should be able to compute these methods more efficiently. So, we can overload them with the efficient implementation.

Are we on the same page or you are thinking of some other implementation?

yashu-seth:
@ankurankan I also think that we should have base classes. But I think we should have a slightly different implementation road map. Here is what I think - 

We should have base classes but, instead of one base class we should have two
 **`ContinuousFactor`** and a **`HybridFactor`.**

Reasons for this choice,
- I think pure(continuous - continuous) relationships are different from hybrid(discrete - continuous) relationships. For example, in a discrete parent - continuous child relationship we will have different pdf functions for each state of the discrete parent. I don't understand how we can use a single pdf function to represent such relationships. (It would be great if you can help me on this).
- The `HybridFactor` class would be subclassed by classes like `CanonicalTable` which would itself be using classes like the `Factor` class (the one we already have) and `CanonicalFactor` (the one that I think should be subclassed by `ContinuousFactor`) . (Although I am not 100% sure about its implementation, but it will be something having Factor tables with CanonicalFactor objects as the table contents). Keeping `CanonicalFactor` and  `CanonicalTable` derived by the same base class does not make much sense to me. 

I believe by keeping the two families separate we can generalize things in a more efficient way.

Now, if we do have two separate base classes, there could be two cases,
- We will be discretizing the continuous distributions to obtain  `Factor` objects and then applying inference, like the way I discussed on my blog post.
- We will be directly using these classes to perform inference tasks. For example, we will be using the `JointGaussianDistribution` class to perform Hamilton sampling etc. and the `Canonical Factor`  or the `CanonicalTable` class to perform the sum product inference tasks.

Now our base classes should handle both these cases and they will. These classes will behave exactly the way you suggested in your previous comment and so will their subsequent derivatives.

Earlier I was planning to make this class behave as a helper class for the dicretiztion functionality and have a separate class for the base `ContinuousFactor` . Hence, I was keeping this class restricted to one variable. But now I think the `ContinuousFactor` class can also behave as a helper class in various places. Hence, I will implement it the way you suggested.

As for the other parts of the implementation road map, please let me know if you think differently.

yashu-seth:
@ankurankan There is one problem, scipy.stats.rv_continuous does not allow subclassing multivariate distributions. Hence,  we will either have to keep the base ContinuousFactor class and this class separate as I earlier suggested or employ some other way to do away with this problem. I would need help on this.

ankurankan:
@yashu-seth I think you can represent both the cases you have mentioned using the same class. So, let's say we have a discrete (A) -- continuous (B) relationship. So we can simply have a pdf function (in this case pdf of B is `e**(-b)` if `a = 0` and `e**(-2b) if`a=1`) like this:

``` python
pdf =  lambda b, a: if state_of_a == 0: return e**(-b) elif state_of_b == 1: return e**(-2b)
```

And I agree with the rest of the implementation you have mentioned in the comment. We will be doing direct inference for the special cases in which we can otherwise we will use discretization. What do you think ?

yashu-seth:
@ankurankan Okay. Then we will have only one base class.
What about the problem that `scipy.stats.rv_continuous` does not allow subclassing multivariate distributions?

yashu-seth:
@ankurankan I cannot proceed without getting this resolved, it would be great if you look into this as soon as possible.

yashu-seth:
@ankurankan The problem is that since `scipy.stats.rv_continuous` works only with univariate distributions, we cannot use multivarite pdfs in its derived class. Hence, I would not be able to do both - allow multivarite pdfs and make this class derived from `scipy.stats.rv_continuous`. Would need your help on this.

ankurankan:
@yashu-seth Have a look at this: https://github.com/scipy/scipy/blob/master/scipy/stats/_multivariate.py#L174. Since the generic class doesn't have much functionality, we can implement our own.

yashu-seth:
@ankurankan I wanted to ensure that the `ContinuousFactor` class can have both multivariate and univariate pdfs. For univariate cases it should be able to access all the functionalities of the class `scipy.stats.rv_continuous` and for mulitivariate cases just allow such pdfs to be defined. 
We can implement our own generic class for multivariate distribution, but what is the best way to keep both the functionality in the same class?
If we simply go on defining multivariate pdfs on a `scipy.stats.rv_continuous` subclass, it will simply produce an error for other methods like for `cdf` and `moment` which I guess is okay. I just wanted to confirm what would be the best way to do this?

ankurankan:
@yashu-seth You can write a base class for multivariate random variable and
then in the `ContinuousFactor` class you can inherit both `rv_continuous`
and the other multivariate base class. And in the **init** method you can
control which parent class to use for representation based on the number of
variables given.

yashu-seth:
@ankurankan Okay. I was also thinking that multiple inheritance would be a feasible option. I will do the same.

yashu-seth:
@ankurankan Is there a particular way to restrict method resolution to only
one parent class. I looked up but could not find a feasible solution.

ankurankan:
@yashu-seth I don't think so. But can you explain your use case a bit ?

yashu-seth:
@ankurankan Like in the ContinuousFactor.**init** when we have more than
one variable, we would would want this class to be derived by the
multivariate base class and when single variable then by the rv_continuous
class. What would be the best way to implement this switch?

ankurankan:
@yashu-seth You can simply use `Parent1.__init__(..)` or `Parent2.__init__(...)` based on which one you want to call. 
Also since you are working with multiple inheritance I would suggest you to read Guido's blog post on MRO.

yashu-seth:
@ankurankan Actually that is not the problem. I should have been more
clear.  I want to make sure that when we are using a multivariate
distribution we get absolutely no access to the cdf method provided by the
rv_continuous class. I posted a question about this.
http://stackoverflow.com/questions/37677097/how-to-to-restrict-method-resolution-to-only-one-parent-in-python-multiple-inher

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,72 @@
...
+
+from six import with_metaclass
+from abc import ABCMeta, abstractmethod
+
+import numpy as np
+
+from pgmpy.factors import ContinuousFactor
+
+
+class BaseDiscretizer(with_metaclass(ABCMeta)):
```
ankurankan: @yashu-seth Why have you inherited from `with_metaclass(ABCMeta)`?

pgmpy/continuous/discretize/rounding.py
```diff
@@ -0,0 +1,50 @@
...
+
+    The probability mass in x=[low+step, low+2*step, ......... , high-step] is,
+    cdf(x+step/2)-cdf(x-step/2)
+
+    where, cdf is the cumulative density function of the distribution
+    and step = (high-low)/cardinality.
+
+    Examples
+    --------
+    >>> from pgmpy.factors import ContinuousFactor
```
ankurankan: @yashu-seth Import `RoundingDiscretizer`

ankurankan:
@yashu-seth @khalibartan I don't like this file structure. I don't see the point in dividing the whole project into `continuous` and `discrete`. Since the model structure part and inference part is going to be common in both the types. So, I would suggest you to go with the structure that I had suggested. Also I don't think we should put the discretize methods in separate files.

pgmpy/continuous/discretize/rounding.py
ankurankan: @yashu-seth `import numpy as np`

ankurankan: @yashu-seth `import numpy as np`

ankurankan:
@yashu-seth Also check for PEP8.

``` python
(py3)ankur@ankur:~/pgmpy (pr/686 $%)$ pep8 pgmpy/continuous/* --max-line-length=120
pgmpy/continuous/discretize/rounding.py:42:28: W503 line break before binary operator
pgmpy/continuous/discretize/rounding.py:48:36: W503 line break before binary operator
pgmpy/continuous/discretize/unbiased.py:60:28: W503 line break before binary operator
pgmpy/continuous/discretize/unbiased.py:95:9: E731 do not assign a lambda expression, use a def
pgmpy/continuous/factors/ContinuousFactor.py:10:1: W293 blank line contains whitespace
pgmpy/continuous/factors/ContinuousFactor.py:15:1: W293 blank line contains whitespace
pgmpy/continuous/factors/ContinuousFactor.py:20:1: W293 blank line contains whitespace
```

yashu-seth:
@ankurankan I still cant figure out how to ensure that rv_continuous
methods are restricted only to univariate distributions, in case multiple
inheritance is being used with ContinuousFactor class.

yashu-seth:
@ankurankan Help required.

khalibartan:
@ankurankan Okay.

khalibartan:
@yashu-seth I think this might help you with your doubt http://python-history.blogspot.in/2010/06/method-resolution-order.html.

ankurankan:
@yashu-seth What you can do is define two separate classes for single variables and multiple variables. And don't inherit them in `ContinuousFactor`. Rather just create an object of either class (based on the number of variables given) and assign it to some variable. And you can then simply use that variable to call the methods of the base classes.

ankurankan:
@yashu-seth I completely forgot that we will just be using the multivariate case in `ContinuousFactor`. Since a factor is used to represent how 2 or more variables interact with each other. So, I don't think we even need to write the code for the single variable case. What do you think ?

yashu-seth:
@ankurankan  Then shouldn't a separate class be created for `ContinuousFactor` that deals with mulitvariate distributions? And leave this existing class as `ContinuousNode` because we would need scipy.stats.rv_continuous in various discretization algorithms anyways?

ankurankan:
@yashu-seth Yeah you can do that. But I don't think we will need the
`rv_continuous` class later. But let's just leave that for now.

yashu-seth:
@ankurankan Okay. Then I will keep this PR for `ContinuousNode` and the various discretization algorithms. I will create base continuous factor class in a separate PR.

I think `ContinuousNode` objects would be useful in HybridFactors as well. I would love to discuss the future of rv_continuos whenever you get time.

pgmpy/continuous/discretize/base.py
yashu-seth: @ankurankan I have used the abc module to implement the abstract method `get_discrete_values`. It requires using a metaclass `abc.ABCMeta`. Since metaclasses are used differently in python 2 and 3, I used `six.with_metaclass`. An another way to do this would be using a decorator `@add_metaclass(ABCMeta)` on the `BaseDiscretizer` class.
If there is a better way to do this, please let me know. I will change it.

yashu-seth:
@khalibartan I went through that post in the very beginning. It is useful but does not solve the problem. Anyways we have resolved the issues for now. Thanks.

yashu-seth:
@ankurankan I think we should have a separate folder for the discretize.py. And we should import the discretization classes as `from pgmpy.discretize import RoundingDiscretizer`. Should I put discretize.py into pgmpy/discretize/discretize.py ?

yashu-seth:
@ankurankan I have made the changes and this is ready for a review. But the travis ci is still failing because of that libgfortran error, even though I have added "conda install libgfortran".

ankurankan:
@yashu-seth Have been trying to fix the issue for the past 2 hours. Can't figure this out.
And the other strange thing is that it's not failing on other branches.

yashu-seth:
@ankurankan I also looked this up for quite some time but could not find the reason.

ankurankan:
@yashu-seth Can you try to create a patch of this branch and apply on some new branch and then check? Also remove the `conda install libgfortran`.

yashu-seth:
@ankurankan Okay. I will try this asap.

ankurankan:
@yashu-seth Fixed finally: https://travis-ci.org/ankurankan/pgmpy. There's some issue with this branch.

yashu-seth:
@ankurankan Should I create a new branc with a patch of this now?

ankurankan:
@yashu-seth Yes you can do that and group your changes in multiple commits to keep things clear.

yashu-seth:
@ankurankan Did you figure out the reason behind this as well? Because the same problem is happening with PR #693

ankurankan:
@yashu-seth No idea. I will try to find the commit after which this started happening. That might give us some idea.

yashu-seth:
@ankurankan I am also trying git biscet. I will let you know if I find anything. It seems conda and scipy together have problems.

yashu-seth:
@ankurankan This was working fine by adding `conda install libgfortran` . The error reoccurred after I added this commit - [moved all discretization classes in one file - discretize.py](https://github.com/pgmpy/pgmpy/pull/686/commits/d4cda32609a0ee0d3b51cd070b8216d3dbaa32ff).
I added two more commits after this, all three together.

yashu-seth:
@ankurankan Travis-Ci shows no errors after I removed the commit -  "moved all discretization classes in one file - discretize.py" . This commit broke things for some reason. But I still can't figure out what makes it break in PR #693

khalibartan:
Wired indeed, as test were passing for python3.3 . Though reason for broken test is missing libgfortran.so , but it should fail for python3.3 also if such is the case.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,72 @@
...
+    Parameters
+    ----------
+    factor: ContinuousFactor
+        the continuous factor representing the distribution
+        to be discretized.
+    low, high: float
+        the range over which the function will be discretized.
+    cardinality: int
+        the number of states required in the discretized output.
+
```
ankurankan: @yashu-seth Add examples here.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,72 @@
...
+class BaseDiscretizer(with_metaclass(ABCMeta)):
+    """
+    Base class for the discretizer classes in pgmpy. The discretizer
+    classes are used to discretize a continuous random variable
+    distribution into discrete probability masses.
+
+    Parameters
+    ----------
+    factor: ContinuousFactor
```
ankurankan: @yashu-seth Add one blank line after each parameter.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,72 @@
...
+        to be discretized.
+    low, high: float
+        the range over which the function will be discretized.
+    cardinality: int
+        the number of states required in the discretized output.
+
+    """
+    def __init__(self, factor, low, high, cardinality):
+        if not isinstance(factor, ContinuousFactor):
+            raise ValueError("{factor} is not a valid ContinuousFactor object."
```
ankurankan: @yashu-seth Raise the error in this format: `factor expected type: ContinuousFactor. Got: ...`. We have used this format in pgmpy. Just search for it.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,72 @@
...
+        >>> node = ContinuousFactor(norm(0).pdf)
+        >>> base = BaseDiscretizer(node, -5, 5, 20)
+        >>> base.get_labels()
+        ['x=-5.0', 'x=-4.5', 'x=-4.0', 'x=-3.5', 'x=-3.0', 'x=-2.5',
+         'x=-2.0', 'x=-1.5', 'x=-1.0', 'x=-0.5', 'x=0.0', 'x=0.5',
+         'x=1.0', 'x=1.5', 'x=2.0', 'x=2.5', 'x=3.0', 'x=3.5', 'x=4.0', 'x=4.5']
+
+        """
+        step = (self.high - self.low) / self.cardinality
+        labels = list('x={i}'.format(i=str(i))
```
ankurankan: @yashu-seth 

```
%timeit [i for i in range(10000)]
1000 loops, best of 3: 348 µs per loop

%timeit list(i for i in range(10000))
1000 loops, best of 3: 517 µs per loop
```

And I don't see any other benefits of the way you have written this.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,72 @@
...
+        self.low = low
+        self.high = high
+        self.cardinality = cardinality
+
+    @abstractmethod
+    def get_discrete_values(self):
+        """
+        This method implements the algorithm to discretize the given
+        continuous distribution.
+        It must be implemented by all the subclasses of BaseDiscretizer.
```
ankurankan: @yashu-seth Try to add extra blank line after paragraphs. Makes things much more cleaner.

pgmpy/continuous/discretize/base.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        pass
+
+    def get_labels(self):
+        """
+        Returns a list of strings representing the values about
+        which the discretization method calculates the probabilty
+        masses.
+        Default value is the points -
```
ankurankan: @yashu-seth Same here.

pgmpy/continuous/discretize/rounding.py
ankurankan: @yashu-seth You can also write this as:

```
For the rounding method, the probability mass is:
cdf(x+step/2) - cdf(x)                for x = low
cdf(x+step/2) - cdf(x-step/2)         for low < x < high
```

This is a more standard form for writing equations like this.

pgmpy/continuous/discretize/rounding.py
```diff
@@ -0,0 +1,50 @@
...
+     0.19741265136584729, 0.17466632194020937, 0.12097757871001302,
+     0.065590616803036905, 0.027834684208772664, 0.0092447094199902269]
+    """
+
+    def get_discrete_values(self):
+        step = (self.high - self.low) / self.cardinality
+
+        # for x=[low]
+        discrete_values = [self.factor.cdf(self.low + step/2)
+                           - self.factor.cdf(self.low)]
```
ankurankan: @yashu-seth This raises a PEP8 error. Since we allow 120 character line in pgmpy, you can put the whole thing in a single line. And in my opinion that's more readable also.

pgmpy/continuous/discretize/rounding.py
```diff
@@ -0,0 +1,50 @@
...
+
+        # for x=[low+step, low+2*step, ........., high-step]
+        x = np.arange(self.low + step, self.high, step)
+        for i in x:
+            discrete_values.append(self.factor.cdf(i + step/2)
+                                   - self.factor.cdf(i - step/2))
```
ankurankan: @yashu-seth Doing an `extend` instead of `append` will decrease the run time by 50%.

```
a = []

%timeit for i in range(1000): a.append(i)
The slowest run took 6.39 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 72.9 µs per loop

a = []

%timeit a.extend([i for i in range(1000)])
10000 loops, best of 3: 34.7 µs per loop
```

pgmpy/continuous/discretize/unbiased.py
```diff
@@ -0,0 +1,102 @@
...
+    For this method,
+
+    The probability mass in x=[low] is,
+    (E(x) - E(x + step))/step + 1 - cdf(x)
+
+    The probability mass in x=[low+step, low+2*step, ........., high-step],
+    (2 * E(x) - E(x - step) - E(x + step))/step
+
+    for x=[high],
+    (E(x) - E(x - step))/step - 1 + cdf(x)
```
ankurankan: @yashu-seth Same here. Use the standard form of representation.

pgmpy/continuous/discretize/unbiased.py
```diff
@@ -0,0 +1,102 @@
...
+
+    where, E(x) is the first limiting moment of the distribution
+    about the point x, cdf is the cumulative density function
+    and step = (high-low)/cardinality.
+
+    For details, refer Klugman, S. A., Panjer, H. H. and Willmot,
```
ankurankan: @yashu-seth Add a separate section for references (like parameters, examples). Search for reference in pgmpy.

pgmpy/continuous/discretize/unbiased.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # for x=[low]
+        discrete_values = [(lev(self.low) - lev(self.low + step)) / step
+                           + 1 - self.factor.cdf(self.low)]
+
+        # for x=[low+step, low+2*step, ........., high-step]
+        x = np.linspace(self.low + step, self.high - step, self.cardinality-2)
+        for i in x:
+            discrete_values.append((2 * lev(i) - lev(i - step) -
+                                    lev(i + step)) / step)
```
ankurankan: @yashu-seth Same here. Use extend instead of append.

ankurankan:
@yashu-seth Please check why this is happening:

``` python
In [29]: ContinuousNode(norm(0, 1).pdf)
Out[29]: <pgmpy.factors.continuous.ContinuousNode.ContinuousNode at 0x7f55bdd09da0>

In [30]: gaus = _

In [31]: gaus.discretize(RoundingDiscretizer, low=-3, high=3, cardinality=10)
Out[31]: 
[0.0021170757714105521,
 0.01439744675977396,
 0.048942780706043479,
 0.1172529240779015,
 0.19802845246428794,
 0.23582284437790496,
 0.19802845246428846,
 0.11725292407790089,
 0.048942780706041522,
 0.014397446759775723,
 0.0029835496606569478]

In [32]: len(_)
Out[32]: 11
```

It's returning 11 values instead of 10.

ankurankan:
@yashu-seth Also I tried plotting the discretized values with respect to the actual distribution:
figure_1
Blue line is actual distribution: norm(0, 1)
I have used UnbiasedDiscretizer. Green dots are discretized using cardinality=10
Blue dots are discretized using cardinality=20
Yellow dots are discretized using cardinality=30

I was expecting the discretized values to get closer to the actual distribution when I increase the value of cardinality. Isn't that the expected behavior ?

yashu-seth:
@ankurankan The blue line is the probability density function while the discretized outputs are the estimated probability masses for the points in a particular range and they always sum to 1. I don't think they should approach the density function as we increase the cardinality with fixed range.

yashu-seth:
@ankurankan I have made the suggested changes and fixed the cardinality bug in RoundingDiscretizer.

ankurankan:
@yashu-seth Looks good. But the travis is still failing. Can you create a new PR for this ?

ankurankan:
@yashu-seth Have merged this. There was some issue with scipy version and conda. So, have updated all the requirement versions to latest.

yashu-seth:
@ankurankan Okay. That is great.

PR #693, opened 2016-05-30
ContinousFactor and Joint Gaussian Representation

yashu-seth:
This PR deals with 
- the creation of a base class `ContinuousFactor` for multivariate representations.
- the creation of the class `JointGaussainDistribution` - a model to represent the gaussain random variables.


pgmpy/models/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,229 @@
...
+        >>> copy_dis.covariance_matrix
+        matrix([[ 4,  2, -2],
+                [ 2,  5, -5],
+                [-2, -5,  8]])
+        >>> copy_dis.precision_matrix
+        matrix([[ 0.3125    , -0.125     ,  0.        ],
+                [-0.125     ,  0.58333333,  0.33333333],
+                [ 0.        ,  0.33333333,  0.33333333]])
+        """
+        copy_distribution = JointGaussianDistribution(self.variables, self.mean_vector,
```
khalibartan: @yashu-seth Numpy array and matrix objects are mutable. Copy method wouldn't actually create a copy.

pgmpy/models/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,229 @@
...
+
+        self.mean_vector = np.matrix(np.reshape(mean_vector, (n, 1)))
+
+        self.covariance_matrix = np.matrix(covariance_matrix)
+
+        if self.covariance_matrix.shape != (n, n):
+            raise ValueError("Each dimension of the covariance matrix must\
+                                 be equal to the number of variables.")
+
+        self._precision_matrix = None
```
khalibartan: @yashu-seth why `_precision_matrix`, not `precision_matrix` ?

pgmpy/models/continuous/JointGaussianDistribution.py
yashu-seth: @khalibartan I was not aware of this. Although I checked the copy method was working fine which is strange. Anyways I will definitely look into it. Thanks for letting me know.

khalibartan:
@yashu-seth I (as a user) would like to have more flexibility by giving inputs as an list or any array type object in mean_vector. A more better way will be convert all the array type structure user inputs into the vector form we require. Something like this

```
a = np.array([1, 2, 3])
b = np.array([[1, 2, 3]])
c = np.array([[1], [2], [3]])
d = [1, 2, 3]
# all a, b, c, d can be given as parameter without a problem
```

pgmpy/models/continuous/JointGaussianDistribution.py
khalibartan: Strange it shouldn't have.

```
In [1]: import numpy as np

In [2]: a = np.array([1,2,3])

In [3]: a
Out[3]: array([1, 2, 3])

In [4]: b = a

In [5]: b[1] = 4

In [6]: a
Out[6]: array([1, 4, 3])

In [7]: b
Out[7]: array([1, 4, 3])
```

though numpy provides copy method which each object so you can pass copy of each object as parameter.

yashu-seth:
@khalibartan It works exactly as you want for a, c and d. Permitting b is not a problem but I think too much flexibility in the input would only increase the number of type checks .

``` python
>>> from pgmpy.models import JointGaussianDistribution as JGD
>>> import numpy as np
>>> a = np.array([1, 2, 3])
>>> b = np.array([[1, 2, 3]])
>>> c = np.array([[1], [2], [3]])
>>> d = [1, 2, 3]
>>> dis = JGD(['x1', 'x2', 'x3'], a,
                        np.array([[4, 2, -2], [2, 5, -5], [-2, -5, 8]]))
>>> dis.mean_vector
matrix([[1],
        [2],
        [3]])

>>> dis = JGD(['x1', 'x2', 'x3'], c,
                        np.array([[4, 2, -2], [2, 5, -5], [-2, -5, 8]]))
>>> dis.mean_vector
matrix([[1],
        [2],
        [3]])

>>> dis = JGD(['x1', 'x2', 'x3'], d,
                        np.array([[4, 2, -2], [2, 5, -5], [-2, -5, 8]]))
>>> dis.mean_vector
matrix([[1],
        [2],
        [3]])
```

pgmpy/models/continuous/JointGaussianDistribution.py
yashu-seth: @khalibartan It works fine because in the **init** method the np.matrix command creates a copy.

``` python
>>> a = np.matrix([[1,2,3,4],[0,0,0,0]])
a
matrix([[1, 2, 3, 4],
        [0, 0, 0, 0]])
>>> b = a
>>> a[1,1] = 99999
>>> a
matrix([[    1,     2,     3,     4],
        [    0, 99999,     0,     0]])
>>> b
matrix([[    1,     2,     3,     4],
        [    0, 99999,     0,     0]])

>>> a = np.matrix([[1,2,3,4],[0,0,0,0]])
>>> b = np.matrix(a)
>>> a[1,1] = 99999
>>> a
matrix([[    1,     2,     3,     4],
        [    0, 99999,     0,     0]])
>>> b
matrix([[1, 2, 3, 4],
       [0, 0, 0, 0]])
```

khalibartan: Indeed, it didn't look at it.

khalibartan:
@yashu-seth Is there a specific reason for choosing matrix over arrays ? numpy arrays are more flexible than matrix. It took me a hard time to debug certain sections of my code, because I thought arrays and matrix work in a similar fashion.

yashu-seth:
@khalibartan Yes, matrix multiplication is required for methods like `to_CanonicalFactor`. Have a look at its docs.

yashu-seth:
@khalibartan Can you work with numpy matrices without converting them into numpy arrays every time? If not we will have to decide what would be the best way.

khalibartan:
@yashu-seth We can obtain the product of numpy arrays too using `np.dot()`. Maybe this link can be of some help http://stackoverflow.com/questions/4151128/what-are-the-differences-between-numpy-arrays-and-matrices-which-one-should-i-u
And I couldn't properly understand this question here

> Can you work with numpy matrices without converting them into numpy arrays every time? If not we will have to decide what would be the best way.

Though from my understanding the answer might be all numpy matrices are convertible to numpy arrays but converse is not true.
@ankurankan what are your views about this?

yashu-seth:
@khalibartan By the question I meant, that it would be unfeasible to convert numpy matrices to numpy arrays and vice versa. So it would be better we find a common representation that can work in all the cases. You will be working with 2d numpy arrays and they are always convertible to numpy matrices although as I said we would want to avoid converting them from one form to another. Anyways, let me go through that link you sent, I'll let you know if I find anything that can be done in a feasible manner.

khalibartan:
@yashu-seth I'm currently using numpy.matrices because I saw in your proposal that you will be using them. But I want to have a discussion over the matter.

yashu-seth:
@khalibartan After going through that post I also feel that arrays should be used, anyways I think we should ask @ankurankan before deciding what is best.

khalibartan:
@yashu-seth Yes ! as I said arrays provide more flexibility than matrices.  We can really have discussion in this week meeting over the matter.

yashu-seth:
@khalibartan I think we should wait.Let @ankurankan look into this. I would suggest not to use matrices for now just because that is how I planned it in the proposal.

yashu-seth:
@khalibartan Cool.

khalibartan:
@yashu-seth I have finished up the Hamiltonian class, but things won't be a big change as I have to just replace products with `numpy.dot`. But things can get messy in NUTS.

yashu-seth:
@khalibartan Same with me.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,252 @@
...
+               [-3],
+               [4]]))
+        >>> dis.covariance
+        matrix([[4, 2, -2],
+               [2, 5, -5],
+               [-2, -5, 8]])
+        """
+        self.variables = variables
+        # dimension of the mean vector and covariance matrix
+        n = len(self.variables)
```
ankurankan: @yashu-seth Don't use variable names like `n`.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,252 @@
...
+        if len(mean) != n:
+            raise ValueError("Length of mean_vector must be equal to the\
+                                 number of variables.")
+
+        self.mean = np.asarray(np.reshape(mean, (n, 1)), dtype=float)
+
+        self.covariance = np.asarray(covariance, dtype=float)
+
+        if self.covariance.shape != (n, n):
+            raise ValueError("Each dimension of the covariance matrix must\
```
ankurankan: @yashu-seth Or something like: `"Should be a square matrix of order no of variables. Expected dimension {n} x {n}. Got: {shape}".format( .... )`

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,252 @@
...
+import six
+import numpy as np
+from scipy.stats import multivariate_normal
+
+
+class JointGaussianDistribution(object):
+    """
+    In its most common representation, a multivariate Gaussian distribution
+    over X1...........Xn is characterized by an n-dimensional mean vector μ,
+    and a symmetric nXn covariance matrix Σ.
```
ankurankan: @yashu-seth Don't use a capital X. Replace with `n x n`.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,252 @@
...
+    This is the base class for its representation.
+    """
+    def __init__(self, variables, mean, covariance):
+        """
+        Parameters
+        ----------
+        variables: iterable of any hashable python object
+            The variables for which the distribution is defined.
+        mean: nX1, array like
```
ankurankan: @yashu-seth Same here use `n x 1`.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,252 @@
...
+        array([[ 0.3125    , -0.125     ,  0.        ],
+                [-0.125     ,  0.58333333,  0.33333333],
+                [ 0.        ,  0.33333333,  0.33333333]])
+        """
+        copy_distribution = JointGaussianDistribution(self.variables.copy(), self.mean.copy(),
+                                                      self.covariance.copy())
+        copy_distribution._precision_matrix = self._precision_matrix
+        return copy_distribution
+
+    def to_CanonicalFactor(self):
```
ankurankan: @yashu-seth Replace with `to_canonical_factor`.

yashu-seth:
@ankurankan I have added the base continuous factor class. Please have a look.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,109 @@
...
+    Base class for factors representing various multivariate
+    representations.
+    """
+    def __init__(self, variables, pdf, mean=None, covariance=None, lb=None, ub=None):
+        """
+        Parameters
+        ----------
+        variables: list or array-like
+            The variables for wich the distribution is defined.
+        pdf: function
```
ankurankan, line 21: @yashu-seth One blank line after each parameter.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,109 @@
+import six
+
+
+class ContinuousFactor(object):
```
ankurankan, line 9: @yashu-seth Need to add methods like marginalize, reduce, product etc. in this.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    Base class for factors representing various multivariate
+    representations.
+    """
+    def __init__(self, variables, pdf, mean=None, covariance=None, lb=None, ub=None):
```
ankurankan: @yashu-seth Why are you accepting mean, covariance, lb and ub in this. Those are parameters specific to JointGaussian

yashu-seth: @ankurankan `lb` and `ub` define the bounds for the pdf for each dimension. There may be cases when the pdf cannot be defined for -inf to +inf. This would be required in methods like `marginalize`.
I added mean and covariance because I think there will be some form of mean and covariance representation in most of the subclasses. Do you feel otherwise?

ankurankan: @yashu-seth 

> lb and ub define the bounds for the pdf for each dimension. There may be cases when the pdf cannot be defined for -inf to +inf. This would be required in methods like marginalize.

If the pdf isn't defined somewhere it should be handled by the pdf function not `ContinuousFactor`.

> I added mean and covariance because I think there will be some form of mean and covariance representation in most of the subclasses

I don't think we should have mean and covariance as arguments. We will be needing these in special cases only and can accept those parameters in that (like in JointGaussian).

yashu-seth: @ankurankan Okay, then I will remove these parameters.

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth, line 9: @ankurankan I understand the implementation for marginalize. How should reduce, product, add etc. be defined?

ankurankan, line 9: @yashu-seth You will have to do something like this:

``` python
def reduce(state, inplace=True):
    def reduced_pdf(a, b, d):
        return self.pdf(a, b, c=state, d)
    self.pdf = reduced_pdf

def product(factor2, inplace=False):
     def new_pdf(a, b, c, d):
         return self.pdf(a, b, c) * factor2.pdf(b, c, d)
     return ContinuousFactor(variables=self.variables \union factor2.variables, pdf=new_pdf)
```

yashu-seth:
@ankurankan We need to maintain the same order of self.variables and the input parameters of the pdf function. To maintain order in self.variables, methods like marginalize and reduce would take O (n^2) with respect to the number of variables. Should I look for ways to do this at a better performance? Or       O (n^2) would be fine for these methods?

ankurankan:
@yashu-seth I think there should be more efficient ways. But since the number of variables won't be huge, there won't be much difference in runtime. So, just check once if you can find simpler/faster implementation.

yashu-seth:
@ankurankan Okay.

ankurankan:
@yashu-seth Always avoid merging some other branch. Makes the git history difficult to read (if needed). Rather use rebase to replay these changes on the latest dev.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,187 @@
...
+        ...     return (np.power(x, 1)*np.power(y, 2))/beta(x, y)
+        >>> from pgmpy.factors import ContinuousFactor
+        >>> drichlet_factor = ContinuousFactor(['x', 'y'], drichlet_pdf)
+        >>> drichlet_factor.scope()
+        ['x', 'y']
+        >>> drichlet_factor.assignemnt(5,6)
+        226800.0
+        """
+        if isinstance(variables, six.string_types):
+            raise TypeError("Variables: Expected type list or array like, got string")
```
ankurankan: @yashu-seth Wouldn't checking if it's array type better ? And then you can throw an error that `expected type list or array, got ...`

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,187 @@
...
+        ['x', 'y', 'z']
+        >>> phi.pdf(1, 2, 3)
+        24.0
+        >>> phi.reduce([('y', 2)])
+        >>> phi.variables
+        ['x', 'z']
+        >>> phi.pdf(1, 3)
+        24.0
+        """
+        if isinstance(values, six.string_types):
```
ankurankan: @yashu-seth Same here.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,187 @@
...
+            raise TypeError("values: Expected type list or array-like, got type str")
+
+        phi = self if inplace else self.copy()
+
+        var_to_remove = [var for var,value in values]
```
ankurankan: Space after `,`.

yashu-seth:
@ankurankan Scipy does not provide functionality for integration over more that three variables. I am having trouble implementing the `ContinuousFactor.marginalize` method when the number of variables to be marginalized is more than 3. Should I look for some other integration methods or only allow marginalization for at max 3 variables and maybe let users redefine the pdf if they want marginalization for more than 3 variables.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,330 @@
...
+            return distribution
+
+    def reduce(self, values, inplace=True):
+        """
+        Reduces the distribution to the context of the given variable values.
+
+        The formula for the obtained conditional distribution is given by -
+
+        For, N(X_j | X_i = x_i) ~ MN(mu_j.i, sig_j.i)
+
```
khalibartan, line 166: @yashu-seth I think this link would be helpful in writing documentation with maths http://www.sphinx-doc.org/en/stable/ext/math.html

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,330 @@
...
+        >>> from pgmpy.factors import JointGaussianDistribution as JGD
+        >>> dis = JGD(['x1', 'x2', 'x3'], np.array([[1], [-3], [4]]),
+        ...             np.array([[4, 2, -2], [2, 5, -5], [-2, -5, 8]]))
+        >>> dis.variables
+        ['x1', 'x2', 'x3']
+        >>> dis.mean
+        array([[ 1],
+               [-3],
+               [4]]))
+        >>> dis.covariance
```
khalibartan: @yashu-seth covariance is changed to array.

khalibartan:
I don't know whether this should be changed or not but if we keep mean this shape:

```
>>>model.mean
array([1, 2])
```

rather than 

```
>>>model.mean
array([[1],
       [2]])
```

we will get different output when we do the product (consider covariance as np.eye(2) )

```
>>> # For first case 
>>> np.dot(model.mean, np.dot(model.covariance, model.mean))
5.0
>>># For the second case
>>> np.dot(model.mean.T, np.dot(model.covariance, model.mean))
array([[5.]])
```

For the earlier case I don't have to take care of the fact that output is an array, but for later I need to convert it to float (etc), to perform operations like raising it to a power of some constant or something like that which we cannot do in cases of arrays. Though its my opinion.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,330 @@
...
+        matrix([[4, 2, -2],
```
khalibartan: @yashu-seth array instead of matrix.

yashu-seth:
@khalibartan  I kept the mean vector as a column matrix (here array) on purpose. Because it is how the mean vector is defined, and all the formulas require it to be a column matrix. 

In the first case I don't think np.dot(model.covariance, model.mean) should be defined because model.covariance is order 2 x 2 and model.mean is order 1 x 2. 
If I do this, I get - 

``` python
>>> mean = np.matrix([1,2])
>>> mean
matrix([[1, 2]])
>>> cov = np.matrix([[1,3],[0,2]])
>>> cov*mean
  File "<pyshell#52>", line 1, in <module>
    cov*mean
  File "C:\Python34\lib\site-packages\numpy\matrixlib\defmatrix.py", line 341, in __mul__
    return N.dot(self, asmatrix(other))
ValueError: shapes (2,2) and (1,2) not aligned: 2 (dim 1) != 1 (dim 0)
>>> 
```

We are using numpy arrays but I think all multiplications should follow matrix multiplication rules for numpy arrays as well. I don't know why np.dot(model.covariance, model.mean) is allowed in the first place.
Besides I feel all np.dot results should always be np arrays and not floats.

And I do not understand what you mean by this - " (consider covariance as np.eye(2) ) ". Can you please explain? Edit - Okay, I understood this line.

khalibartan:
@yashu-seth Yes, they should be arrays in terms of implementation, but mathematically dot product of two vectors should be value not a matrix. There is no issue my sides either in having model.mean as column matrix, I just wanted to give another opinion. 

Also I was going through your implementation, you have used **Σ** [here](https://github.com/pgmpy/pgmpy/pull/693/files#diff-f109d8468ed85865be49612488891a6dR282) which I think is a unicode. You can change your docstring to unicode in that  case. I think currently it will raise an error in python 2.x .

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth: @ankurankan Would we be having a separate check for this in pgmpy.utils?

ankurankan: @yashu-seth No, this is just a single check. So, no need to add that in utils.

ankurankan:
@yashu-seth Please let me know when this is ready for review.

yashu-seth:
@ankurankan This is ready for a review. I wanted help on the marginalize
method. Besides that this is ready.

yashu-seth:
@ankurankan I have added the `marginalize` method. Please have a look.
I have not implemented the product method because I think simply returning the product of the two pdfs would not ensure that the new pdf would integrate to 1. Please let me know if you feel otherwise.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,335 @@
...
+            raise ValueError("Length of mean_vector must be equal to the\
+                                 number of variables.")
+
+        self.mean = np.asarray(np.reshape(mean, (no_of_var, 1)), dtype=float)
+        normal_pdf = lambda *args: multivariate_normal.pdf(args, self.mean.reshape(1, no_of_var)[0],
+                                                           covariance)
+        self.covariance = np.asarray(covariance, dtype=float)
+        self._precision_matrix = None
+
+        if self.covariance.shape != (no_of_var, no_of_var):
```
khalibartan, line 63: @yashu-seth @ankurankan I think we should also check whether covariance is a symmetric matrix or not. But I don't know any efficient way of doing that.

yashu-seth, line 63: @khalibartan @ankurankan I will look into it. I also forgot to add type
checks for array elements. Will do that too.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,259 @@
...
+        >>> def drichlet_pdf(x, y):
+        ...     return (np.power(x, 1)*np.power(y, 2))/beta(x, y)
+        >>> from pgmpy.factors import ContinuousFactor
+        >>> drichlet_factor = ContinuousFactor(['x', 'y'], drichlet_pdf)
+        >>> drichlet_factor.scope()
+        ['x', 'y']
+        >>> drichlet_factor.assignemnt(5,6)
+        226800.0
+        """
+        if not isinstance(variables, list):
```
ankurankan: @yashu-seth I don't think the condition you have written matches with the error message that you have written. You will need to check if `variable` is iterable and is not a string. Maybe write a function for this in utils because we will be using it multiple times.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,259 @@
...
+
+            return pdf(*reduced_args, **reduced_kwargs)
+
+        phi.variables = var_to_keep
+        phi.pdf = reduced_pdf
+
+        if not inplace:
+            return phi
+
+    def marginalize(self, variables, inplace=True):
```
ankurankan, line 205: @yashu-seth You also need to add the `product` and `divide` methods.

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth: @ankurankan Wouldn't just checking if its not a string suffice?

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth, line 205: @ankurankan For `product` and `divide` , I think simply dividing and multiplying the pdfs would not ensure that the new pdf always integrated to 1 for all the variables. Don't we need proper definitions for it so that all operations ensure that new pdf is valid?

pgmpy/factors/continuous/ContinuousFactor.py
ankurankan: @yashu-seth We just want to accept something that is iterable like set, dict, list, tuple, np.array. User can pass anything, so you need to check if it's iterable or not. And because strings are iterables and we don't want to accept that we need to check for that also.

yashu-seth: @ankurankan I wanted only lists to be allowed in the methods. Providing too much flexibility just increases the type checks and we convert them to lists anyways in the implementation. What do you think?

ankurankan: @yashu-seth I don't think allowing iterables would complicate anything. You just need to check if it has the `__iter__` attribute and all of them behave in the same way. So, it's just one check and no extra case handling. What do you think?

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth, line 205: @ankurankan Like whenever we define such methods we need to ensure that the factors remain in the same parametric family. `Marginalize` and `reduce` does that. But if we define `product` and `divide` this way (i.e. by mutplying or dividing the pdf) we cannot always ensure this property. Without this we will not be able to use them in inference. What do you think?

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth: @ankurankan But during the implementation we will anyways convert it into lists. I do not understand why for assigning variables users would want to give a dict, set or np.array but not straightaway lists or tuples?

pgmpy/factors/continuous/ContinuousFactor.py
ankurankan, line 205: @yashu-seth I don't think that should be a problem. We can always normalize the pdf. That reminds me that you need to add `normalize` method also.

yashu-seth, line 205: @ankurankan The problem is not just that to perform inference we need to ensure that all intermediate factors formed during the task must belong to the same parametric family. I don't think we can ensure that `product` and `divide` would make this happen. That is why we separate classes (like the CanonicalFactor) that ensure this property for special cases. I don't think we can fix this by normalizing it. Besides how should the normalize method be implemented? I went through the initial part of the chapter Inference in Hybrid Networks but could not find a suitable way to do this. What should be done?

ankurankan, line 205: @yashu-seth You will need to keep the factor in the same parametric family only in special cases. For example you can apply variable elimination on any pdf. You will just be integrating and eliminating variables.

For normalizing just divide the pdf with it's integral over the whole space.

yashu-seth:
@ankurankan That is great then. I will add these method too. Can you tell me other sources that deals with variable elimination on continuous random variables in detail , the principles and techniques book mostly deals with special cases. 
I meant I went through the chapter inference in hybrid networks for this parametric property maintainence and it mentioned that same parametric family property must be maintained. That is what got me thinking in the first place.

yashu-seth:
@ankurankan If we will be performing inference directly on ContinuousFactor then please have a look at the `marginalize` method. The way I have currently implemented it does not make it feasible (would take a lot of time) for repetitive marginalization.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,337 @@
...
+        >>> dis.pdf([0,0,0])
+        0.0014805631279234139
+        """
+        no_of_var = len(variables)
+
+        if len(mean) != no_of_var:
+            raise ValueError("Length of mean_vector must be equal to the\
+                                 number of variables.")
+
+        self.mean = np.asarray(np.reshape(mean, (no_of_var, 1)), dtype=float)
```
khalibartan, line 57: @yashu-seth Is there a particular reason for using `np.asarray()` rather than `np.array()`? `np.array()` creates a new object, thus avoids mutability issues, but same cannot be said about `np.assary()`.

pgmpy/factors/continuous/ContinuousFactor.py
khalibartan: @yashu-seth We have allowed all array like structures in all places, (though allowing dictionary do raise some doubts in my mind). Sets, tuples and np.array should be allowed. For eg user can use set in case of large number of variables to keep variable name distinct. I my views we cannot really presume things like this. Also you should check for distinct variable names to avoid ambiguous results later which can be hard to debug.

yashu-seth: @khalibartan Putting in checks for every single ambiguity does not make much sense to me. To an extent I think we should follow garbage in garbage out.
As far as the variables being list is concerned, yes it needs to be changed. But I think it should only be allowed to be lists or tuples or sets. What do you think?

pgmpy/factors/continuous/JointGaussianDistribution.py
yashu-seth, line 57: @khalibartan np.array always makes a copy of the object, while np.asarray only makes a copy when necessary. If mean and covariance are already np.arrays we can avoid making an unnecessary copy. What mutability issues can we have here?

khalibartan, line 57: @yashu-seth Yes I know that np.assary create a copy when necessary. Agreed at here, no need of unnecessary copy.

pgmpy/factors/continuous/ContinuousFactor.py
khalibartan: @yashu-seth @ankurankan 

> To an extent I think we should follow garbage in garbage out.

My thoughts differ at here. In my perspective we should raise error at here. 

> But I think it should only be allowed to be lists or tuples or sets. What do you think? 

Yes I also agree there is no need to allow every iterable object especially dictionary object. But I'm a bit confused to what things we should allow and what not. But allowing all kind of iterable objects doesn't make much sense to me.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,393 @@
+from __future__ import division
+
+import six
```
ankurankan: @yashu-seth `six` is not a dependency. Import it from `pgmpy.extern`.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,345 @@
...
+    and a symmetric n x n covariance matrix Σ.
+    This is the base class for its representation.
+    """
+    def __init__(self, variables, mean, covariance):
+        """
+        Parameters
+        ----------
+        variables: iterable of any hashable python object
+            The variables for which the distribution is defined.
+        mean: n x 1, array like
```
ankurankan, line 26: @yashu-seth Blank lines between parameters.

pgmpy/factors/continuous/JointGaussianDistribution.py
```diff
@@ -0,0 +1,345 @@
...
+    def marginalize(self, variables, inplace=True):
+        """
+        Modifies the distribution with marginalized values.
+
+        Parameters
+        ----------
+
+        variables: iterator
+                List of variables over which to marginalize.
+        inplace: boolean
```
ankurankan, line 101: @yashu-seth Same here.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,393 @@
...
+
+        phi = self if inplace else self.copy()
+
+        var_to_remove = [var for var, value in values]
+        var_to_keep = [var for var in self.variables if var not in var_to_remove]
+
+        reduced_var_index = [(self.variables.index(var), value) for var, value in values]
+        pdf = self.pdf
+
+        def reduced_pdf(*args, **kwargs):
```
ankurankan, line 184: @yashu-seth Look into `types.FunctionType`. Should be cleaner than this.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,393 @@
...
+        pdf = self.pdf
+
+        pdf_mod = integrate.nquad(pdf, [[-np.inf, np.inf] for var in self.variables])[0]
+
+        phi.pdf = lambda *args: pdf(*args) / pdf_mod
+
+        if not inplace:
+            return phi
+
+    def operate(self, other, operation, inplace=True):
```
ankurankan: @yashu-seth Add product and divide methods. Rename this to `_operate` to make it an internal method.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -0,0 +1,393 @@
...
+        >>> from pgmpy.factors import ContinuousFactor
+        >>> from scipy.stats import multivariate_normal
+        >>> normal_pdf = lambda x: multivariate_normal(x, [0, 0], [[1, 0], [0, 1]])
+        >>> phi = ContinuousFactor(['x1', 'x2'], normal_pdf)
+        >>> phi.scope()
+        ['x1', 'x2']
+        """
+        return self.variables
+
+    def assignment(self, *args):
```
ankurankan, line 67: @yashu-seth Also handle the case of **kwargs. The user might pass named arguments.

yashu-seth, line 67: @ankurankan I purposely restricted this to only _args. 
Allowing *_kwargs would make the various modified pdfs (in `marginalize`, `_operate` etc.) even more complicated. They would require added checks for args and kwargs. I tried to implement this with `reduce` but it only made it complicated and I was thinking of restricting reduce method also to *args only. What should be done?

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth, line 184: @ankurankan Are you talking about the reduced_pdf ?

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth, line 67: @ankurankan And the current implementation of these methods is very poor. Should we leave them with just the *args , until we can find a better way to implement these methods?

ankurankan, line 67: @yashu-seth I don't get how it's making it more complicated. Can you elaborate please ?

pgmpy/factors/continuous/ContinuousFactor.py
ankurankan, line 184: @yashu-seth Yes. With `types.FunctionType` you will be able to dynamically define functions in a cleaner way.

pgmpy/factors/continuous/ContinuousFactor.py
yashu-seth, line 67: @ankurankan If we are allowing kwargs in `self.pdf` then, the pdf obtained after operations like `reduce`,  `marginalize`, `_operate` should also allow kwargs parameters.

Take for example the `reduced_pdf` in `reduce``, it checks if there are args or there are kwargs parameters and then if there are both. Then it maps the arguments accordingly.  I think this would increase the overhead and complication on the already slow implemented methods.

PR #700, opened 2016-06-15
Fix travis continuous node

yashu-seth:


ankurankan:
@yashu-seth I think you have cherry-picked the commits. Right ? Don't do this it will fail again. Do this:

``` bash
git checkout support_for_continuous_nodes
git format-patch dev --stdout > continuous.patch
git checkout dev
git checkout -b feature/new_continuous
git apply continuous.patch
```

You will then have all the changes in your current branch. After that just do git add and git commit. Just try to group  things over a few commits.

yashu-seth:
@ankurankan How can we make a patch for the last 5 commits only? This would preserve most of the commit history. Since tavis started failing only after the 5th last commit, this can be done.

yashu-seth:
@ankurankan I think I found it. Looking into it now.

ankurankan:
@yashu-seth This should work:

```
git format-patch HEAD~5..HEAD --stdout > patch.patch
```

yashu-seth:
@ankurankan This is still failing :-(  Besides we cannot resolve it like this in every PR. PR #693 has the same problems. Can't we ignore this for now? Tests are passing in one environment at least.

ankurankan:
@yashu-seth No, ignoring this would just make things more difficult for us in future. Just create a patch and add it to a new branch as I have mentioned here: https://github.com/pgmpy/pgmpy/pull/700#issuecomment-226101140

yashu-seth:
@ankurankan Is it necessary to create a new branch everytime? Because I did the same as you suggested for this branch and just to check made only commit to see if travis passes or not. But it fails again.

ankurankan:
@yashu-seth I don't know. It worked for me on a new branch.

yashu-seth:
@ankurankan Okay. Then let me try this on a new branch.

PR #701, opened 2016-06-15
Fix travis for Continuous Node PR

yashu-seth:


yashu-seth:
@ankurankan No relief here too. What should be done?

pgmpy/factors/__init__.py
```diff
@@ -2,6 +2,7 @@
 from .FactorSet import FactorSet, factorset_product, factorset_divide
 from .CPD import TabularCPD
 from .JointProbabilityDistribution import JointProbabilityDistribution
+from .continuous.ContinuousNode import ContinuousNode
```
khalibartan, line 5: @yashu-seth I think this will fail in python2, try running nosetests for this in python2 in your local system.

yashu-seth, line 5: @khalibartan The tests are passing. Can you please check this too if you have doubts? I don't think there would be a problem.

khalibartan, line 5: @yashu-seth Okay.

PR #706, opened 2016-06-27
No U Turn Sampler

khalibartan:
This PR Deals with implementing following two samplers:
- NUTS
- NUTS with dual averaging

### DIFF
--- a/pgmpy/inference/continuous/sampling.py
+++ b/pgmpy/inference/continuous/sampling.py
@@ -396,7 +396,7 @@ class HamiltonianMCda(HamiltonianMC):
     def __init__(self, model, grad_log_pdf=None, simulate_dynamics=LeapFrog, delta=0.65):
 
         if not isinstance(delta, float) or delta > 1.0 or delta < 0.0:
-            raise AttributeError(
+            raise ValueError(
                 "delta should be a floating value in between 0 and 1")
 
         self.delta = delta
@@ -633,7 +633,7 @@ class NoUTurnSampler(HamiltonianMCda):
      (1.7139810571103862, 2.809135711303245, 5.690811523613858), ...,
      (-0.7742669710786649, 2.092867703984895, 6.139480724333439),
      (1.3916152816323692, 1.394952482021687, 3.446906546649354),
-     (-0.2726336476939125, 2.6230854954595357, 2.923948403903159)], 
+     (-0.2726336476939125, 2.6230854954595357, 2.923948403903159)],
               dtype=[('x', '<f8'), ('y', '<f8'), ('z', '<f8')])
 
     References
@@ -654,8 +654,8 @@ class NoUTurnSampler(HamiltonianMCda):
         Initalizes root node of the tree, i.e depth = 0
         """
 
-        position_bar, momentum_bar, grad_bar = self.simulate_dynamics(self.model, position, momentum, stepsize,
-                                                                      self.grad_log_pdf).get_proposed_values()
+        position_bar, momentum_bar, _ = self.simulate_dynamics(self.model, position, momentum, stepsize,
+                                                               self.grad_log_pdf).get_proposed_values()
 
         _, logp_bar = self.model.get_gradient_log_pdf(position_bar, self.grad_log_pdf)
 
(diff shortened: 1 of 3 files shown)


khalibartan:
@ankurankan The `_build_tree` method for NUTS and NUTS with dual averaging is slightly different for now I have wrote different code for both. Is that okay, or should I try to accommodate things in a single method with if else etc.

pgmpy/inference/continuous/sampling.py
```diff
@@ -591,3 +592,231 @@ def generate_sample(self, initial_pos, num_adapt, num_samples, trajectory_length
...
+        samples[0] = tuple(initial_pos)
+        position_m = initial_pos
+
+        for i in range(1, num_samples):
+            # Genrating sample
+            position_m = self._sample(position_m, stepsize)
+            samples[i] = position_m
+
+        if HAS_PANDAS == True:
+            return pd.DataFrame.from_records(samples)
```
yashu-seth, line 836: @khalibartan There could be case when the user would want a np.recarry output despite having pandas installed. May be we can have an optional parameter to ensure this. What do you think?

khalibartan, line 836: @yashu-seth Earlier in HMC I have done the same, but @ankurankan told me to return pandas.DataFrame if available (i.e current way of returning ).

yashu-seth, line 836: @khalibartan Okay. I have not gone through all the methods. Can you tell me if this output would be used in some other pgmpy method?

khalibartan:
@ankurankan Please take a look at things. I haven't added tests for `_build_tree` method. Should I add tests for it? I'll make some changes in it after PR about joint gaussian distribution is merged.

pgmpy/inference/continuous/sampling.py
```diff
@@ -49,12 +51,12 @@ class HamiltonianMC(object):
     >>> samples = sampler.sample(initial_pos=np.array([1, 1]), num_samples = 10000,
     ...                          trajectory_length=2, stepsize=0.4)
     >>> samples
-    array([(5e-324, 5e-324), (-2.2348941964735225, 4.43066330647519),
-           (-2.316454719617516, 7.430291195678112), ...,
-           (-1.1443831048872348, 3.573135519428842),
-           (-0.2325915892988598, 4.155961788010201),
-           (-0.7582492446601238, 3.5416519297297056)], 
-          dtype=[('x', '<f8'), ('y', '<f8')])
+    rec.array([(1.0, 1.0), (-3.1861687131079086, 3.7940994520145654),
```
ankurankan, line 54: @khalibartan Format this properly.

pgmpy/inference/continuous/sampling.py
```diff
@@ -591,3 +593,629 @@ def generate_sample(self, initial_pos, num_adapt, num_samples, trajectory_length
...
+    >>> from pgmpy.models import JointGaussianDistribution as JGD
+    >>> import numpy as np
+    >>> mean = np.array([1, 2, 3])
+    >>> covariance = np.array([[4, 0.1, 0.2], [0.1, 1, 0.3], [0.2, 0.3, 8]])
+    >>> model = JGD(['x', 'y', 'z'], mean, covariance)
+    >>> sampler = NUTS(model=model, grad_log_pdf=None, simulate_dynamics=LeapFrog)
+    >>> samples = sampler.sample(initial_pos=np.array([0.1, 0.9, 0.3]), num_samples=20000,stepsize=0.4)
+    >>> samples
+    rec.array([(0.1, 0.9, 0.3),
+     (-0.27303886844752756, 0.5028580705249155, 0.2895768065049909),
```
ankurankan, line 632: @khalibartan Format this properly.

ankurankan:
@khalibartan This PR looks good. I am merging this right now. And please create a new PR for the remaining tests and other changes.

pgmpy/inference/continuous/sampling.py
khalibartan, line 632: @ankurankan This is the actual output representation. I have copied it.

Issue #712, opened 2016-07-27
XBIFReader incompatible with SAMIAM export (and spec?)


finnhacks42:
The CPDs come in transposed - such that they are no longer valid, for example:

Z      |  Z_0  | Z_1 |           Z      |  Z_0  | Z_1 | 
X_0  |  .8    |   .4   |     =>  X_0  |  .8    |   .2   |
X_1  |  .2    |   .6   |           X_1  |  .4    |   .6   |

According to the spec at http://www.cs.cmu.edu/~fgcozman/Research/InterchangeFormat/ the tables should be read:

"taking the GIVEN variables first and then
the FOR variables (if all variables were binary, we would say binary
counting with least significant digit in the right)."

PR #713, opened 2016-08-01
Fix/xml bif transposed cpd

finnhacks42:
Fixes issue #712. This was a bug not caught by existing tests as they didn't have coverage of get_model(). I've added a test and updated code and test code for XMLBIFReader and XMLBIFWriter.

### DIFF
--- a/pgmpy/readwrite/XMLBIF.py
+++ b/pgmpy/readwrite/XMLBIF.py
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
 
 import string
-import re
 from io import BytesIO
+from pgmpy.extern.six.moves import map
 
 # TODO input and output state
 
@@ -321,16 +321,15 @@ class XMLBIFWriter(object):
         return outcome_tag
 
     def _make_valid_state(self, state):
-        """Transform the input statename into a valid state in XMLBIF. 
+        """Transform the input statename into a valid state in XMLBIF.
         XMLBIF states must start with a letter an only contain letters,
         numbers and underscores.
         """
         s = str(state)
-        s_allowed = re.sub("[^a-zA-Z0-9]", "_", s)
-
-        if s_allowed[0] not in string.ascii_letters:
-            s_allowed = "state" + s_allowed
-        return s_allowed
+        s_fixed = "".join(map(lambda x: x if x.isalnum() else "_",s))
+        if s_fixed[0] not in string.ascii_letters:
+            s_fixed = "state" + s_fixed
+        return s_fixed
 
     def get_properties(self):
         """
--- a/pgmpy/readwrite/XMLBIF.py
+++ b/pgmpy/readwrite/XMLBIF.py
@@ -2,6 +2,7 @@
 
 
 from io import BytesIO
+import pyparsing as pp
 
 
 # TODO input and output state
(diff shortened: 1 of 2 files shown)


pgmpy/readwrite/XMLBIF.py
```diff
@@ -303,10 +316,22 @@ def get_states(self):
...
                 outcome_tag[var].append(state_tag)
         return outcome_tag
 
+    def _make_valid_state(self, state):
+        """Transform the input statename into a valid state in XMLBIF. 
+        XMLBIF states must start with a letter an only contain letters,
+        numbers and underscores.
+        """
+        s = str(state)
+        s_allowed = re.sub("[^a-zA-Z0-9]", "_", s)
```
khalibartan: @finnhacks42 Don't use regular expressions. We use pyparsing instead of regular expressions.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -1,5 +1,12 @@
 #!/usr/bin/env python
 
+import string
+import re
+from io import BytesIO
+
+# TODO input and output state
```
khalibartan, line 8: @finnhacks42 Can you add in the support for input and output state, our CPD representation supports input and output state.

pgmpy/readwrite/XMLBIF.py
```diff
@@ -233,20 +242,24 @@ def __str__(self):
         """
         if self.prettyprint:
             self.indent(self.xml)
-        return etree.tostring(self.xml, encoding=self.encoding)
+        f = BytesIO()
+        et = etree.ElementTree(self.xml)
+        et.write(f, encoding='utf-8', xml_declaration=True)
+        return f.getvalue().decode('utf-8')  # your XML file, encoded as UTF-8
+        # return etree.tostring(self.xml, encoding=self.encoding)
```
khalibartan: @finnhacks42 This needs a explanation. What was the issue with previous behaviour ? Also why we are always returning XML file with UTF-8 encoding if we have encoding parameter.

pgmpy/readwrite/XMLBIF.py
finnhacks42, line 8: Happy to try and do that but may take a while as I don't understand how the state works in the CPD representation. Maybe better in a separate pull request? Are there any examples of creating a CPD with state (couldn't find any in existing readers)

khalibartan, line 8: @finnhacks42 Look for state name decorator file in utils. State name decorator adds in an extra layer that take care of mapping of variables.

pgmpy/readwrite/XMLBIF.py
finnhacks42: previous behaviour was returning bytes from **str** (meaning calling print on the writer resulted in an error). It was also not outputting a header - which is required for some other programs to be able to read ie JavaBayes.

I'll fix to use encoding parameter

pgmpy/readwrite/XMLBIF.py
khalibartan: @ankurankan What  are your views ? Should we use pyparsing or regular expression at here. This expression is quite simple to understand.

khalibartan:
@ankurankan +1 for merge after mentioned changes are made.

ankurankan:
@finnhacks42 Thanks for the patch. :)

Issue #716, opened 2016-08-02
TableCPD type checking too strict


finnhacks42:
TableCPD init method requires variable_card to be an int. This makes it incompatible with model.get_cardinality(.), which returns a numpy.int64

The fix is trivial, replace

`if not isinstance(variable_card, int):`

with 
`import numbers`
`if not isinstance(variable_card,numbers.Integral)`

should I create a pull request?

yashu-seth:
@finnhacks42 model.get_cardinality calls self.get_cpds which looks for the cardinality attribute in the CPD object. I don't understand why it produces a numpy.int64 type. 
It would be great if you can share a reproducible example. It would make things clearer.

finnhacks42:
The following prints class 'numpy.int64' and then throws a TypeError 

```
model = BayesianModel([('X','Y')])
cpd_x = TabularCPD(variable='X', variable_card=2, values=[[0.6, 0.4]])
cpd_y = TabularCPD(variable='Y', variable_card=2,
                   values=[[0.95, 0.2],
                           [0.05, 0.8]],
                   evidence=['X'],
                   evidence_card=[2])
model.add_cpds(cpd_x,cpd_y)
model.check_model()

x_card = model.get_cardinality('X')

print(type(x_card))
cpd_new = TabularCPD(variable='X', variable_card=x_card, values=[[0.6, 0.4]])
```

In general should we be doing instance checks explicitly against native python int? A numpy int works just as well. It would seem more pythonic to accept either.

khalibartan:
@finnhacks42 Yes, indeed we should be allowing all type integers, native python and numpy integer types. You can send a pull request for that.

ankurankan:
Fixed in #717

PR #717, opened 2016-08-05
Fix/typechecking variable elimination table cpd

### LINKED ISSUE #715: Inconsistent behaviour of query in VariableElimination when passing a single variable
If query is passed a single variable rather than a list, it works if the variable name is a single character and fails with an obtuse error otherwise. 

The documentation does say to pass a list but there are many examples, ie in the notebooks where a single variable is passed. 

Can we either handle passing single variables (ie check and insert them into a list) or check and throw an informative error?

Simple code to reproduce an error below. It occurs because the set subtraction (on line 57 of ExactInference) happens to work if the single variable is one character long because set('x')==set(['x'])

`
from pgmpy.factors import TabularCPD
model = BayesianModel([('Z', 'Xx')])
cpd_z = TabularCPD(variable='Z', variable_card=2, values=[[0.7, 0.3]])
cpd_x = TabularCPD(variable='Xx', variable_card=2, 
                   values=[[0.6, 0.2],
                           [0.4, 0.8]],evidence=['Z'],evidence_card=[2])

model.add_cpds(cpd_z, cpd_x)
model.check_model()

infer = VariableElimination(model)
infer.query('Xx')['Xx']`

finnhacks42:
Fixes #715 and #716 

Changes VariableElimination query to check variable_card against numbers.Integral instead of int and to check variables is not a string

Changes TableCPD init to expect evidence and evidence card are passed as list-like things not string to be consistent.

### DIFF
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -4,6 +4,7 @@ from __future__ import division
 
 from itertools import product
 from warnings import warn
+import numbers
 
 import numpy as np
 
@@ -107,7 +108,7 @@ class TabularCPD(Factor):
 
         variables = [variable]
 
-        if not isinstance(variable_card, int):
+        if not isinstance(variable_card, numbers.Integral):
             raise TypeError("Event cardinality must be an integer")
         self.variable_card = variable_card
 
--- a/pgmpy/factors/CPD.py
+++ b/pgmpy/factors/CPD.py
@@ -86,7 +86,7 @@ class TabularCPD(Factor):
     values: 2d array, 2d list or 2d tuple
         values of the cpd table
 
-    evidence: string, array-like
+    evidence: array-like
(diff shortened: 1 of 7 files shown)


pgmpy/factors/CPD.py
```diff
@@ -107,34 +108,23 @@ def __init__(self, variable, variable_card, values,
...
         if evidence_card is not None:
-            if not isinstance(evidence_card, (list, tuple)):
-                if isinstance(evidence_card, np.ndarray):
-                    evidence_card = evidence_card.tolist()
-                elif isinstance(evidence_card, (int, float)):
-                    evidence_card = [evidence_card]
-                else:
-                    raise TypeError("Must be a list, tuple or array of variable names ",
-                                    "and variable name can be any hashable python object")
+            if isinstance(evidence_card,numbers.Real):
```
ankurankan: @finnhacks42 PEP 8

PR #719, opened 2016-08-18
Removes pandas dependency from sampling in discrete models

khalibartan:
Ping @ankurankan .

### DIFF
--- a/pgmpy/base/__init__.py
+++ b/pgmpy/base/__init__.py
@@ -1,6 +1,6 @@
 from .UndirectedGraph import UndirectedGraph
 from .DirectedGraph import DirectedGraph
-from .OptionalDependency import *
+from .OptionalDependency import HAS_PANDAS
 
 __all__ = ['UndirectedGraph',
            'DirectedGraph',
--- a/pgmpy/base/__init__.py
+++ b/pgmpy/base/__init__.py
@@ -1,5 +1,7 @@
 from .UndirectedGraph import UndirectedGraph
 from .DirectedGraph import DirectedGraph
+from .OptionalDependency import *
 
 __all__ = ['UndirectedGraph',
-           'DirectedGraph']
+           'DirectedGraph',
+           'HAS_PANDAS']
--- a/pgmpy/sampling/NUTS.py
+++ b/pgmpy/sampling/NUTS.py
@@ -4,13 +4,9 @@ from __future__ import division
 
 import numpy as np
 
-from pgmpy.sampling import HamiltonianMCDA
-from pgmpy.sampling import LeapFrog
+from pgmpy.sampling import HamiltonianMCDA, LeapFrog, _return_samples
 from pgmpy.utils import _check_1d_array_object, _check_length_equal
-from pgmpy.base import HAS_PANDAS
 
-if HAS_PANDAS:
-    import pandas
 
 class NoUTurnSampler(HamiltonianMCDA):
     """
@@ -212,7 +208,6 @@ class NoUTurnSampler(HamiltonianMCDA):
 
         Examples
         ---------
-        >>> # If environment has a installation of pandas
         >>> from pgmpy.sampling import NoUTurnSampler as NUTS, GradLogPDFGaussian, LeapFrog
         >>> from pgmpy.factors.continuous import JointGaussianDistribution as JGD
         >>> import numpy as np
@@ -220,7 +215,8 @@ class NoUTurnSampler(HamiltonianMCDA):
         >>> covariance = np.array([[6, 0.7, 0.2], [0.7, 3, 0.9], [0.2, 0.9, 1]])
         >>> model = JGD(['x', 'y', 'z'], mean, covariance)
         >>> sampler = NUTS(model=model, grad_log_pdf=GradLogPDFGaussian, simulate_dynamics=LeapFrog)
-        >>> samples = sampler.sample(initial_pos=np.array([1, 1, 1]), num_samples=10, stepsize=0.4, return_type='dataframe')
+        >>> samples = sampler.sample(initial_pos=np.array([1, 1, 1]), num_samples=10,
+        ...                          stepsize=0.4, return_type='dataframe')
         >>> samples
                   x         y         z
         0  1.000000  1.000000  1.000000
@@ -251,14 +247,7 @@ class NoUTurnSampler(HamiltonianMCDA):
             position_m = self._sample(position_m, stepsize)
             samples[i] = position_m
 
-        if return_type.lower() == "dataframe":
-            if HAS_PANDAS:
-                return pandas.DataFrame.from_records(samples)
-            else:
-                warn("Pandas installation not found. Returning numpy.recarray object")
-                return samples
-        else:
-            return samples
+        return _return_samples(return_type, samples)
 
     def generate_sample(self, initial_pos, num_samples, stepsize=None):
         """
@@ -354,7 +343,8 @@ class NoUTurnSamplerDA(NoUTurnSampler):
     >>> covariance = np.array([[-2, 7, 2], [7, 14, 4], [2, 4, -1]])
     >>> model = JGD(['x', 'v', 't'], mean, covariance)
     >>> sampler = NUTSda(model=model, grad_log_pdf=GradLogPDFGaussian)
-    >>> samples = sampler.sample(initial_pos=np.array([0, 0, 0]), num_adapt=10, num_samples=10, stepsize=0.25, return_type='recarray')
+    >>> samples = sampler.sample(initial_pos=np.array([0, 0, 0]), num_adapt=10, num_samples=10,
+    ...                          stepsize=0.25, return_type='recarray')
     >>> samples
     rec.array([(0.0, 0.0, 0.0),
      (0.06100992691638076, -0.17118088764170125, 0.14048470935160887),
@@ -508,7 +498,6 @@ class NoUTurnSamplerDA(NoUTurnSampler):
 
         Examples
         ---------
-        >>> # If environment has a installation of pandas
         >>> from pgmpy.sampling import NoUTurnSamplerDA as NUTSda, GradLogPDFGaussian, LeapFrog
         >>> from pgmpy.factors.continuous import JointGaussianDistribution as JGD
         >>> import numpy as np
@@ -516,7 +505,8 @@ class NoUTurnSamplerDA(NoUTurnSampler):
         >>> covariance = np.array([[16, -3], [-3, 13]])
         >>> model = JGD(['x', 'y'], mean, covariance)
         >>> sampler = NUTSda(model=model, grad_log_pdf=GradLogPDFGaussian, simulate_dynamics=LeapFrog)
-        >>> samples = sampler.sample(initial_pos=np.array([12, -4]), num_adapt=10, num_samples=10, stepsize=0.1, return_type='dataframe')
+        >>> samples = sampler.sample(initial_pos=np.array([12, -4]), num_adapt=10, num_samples=10,
+        ...                          stepsize=0.1, return_type='dataframe')
         >>> samples
                    x          y
         0  12.000000  -4.000000
@@ -560,14 +550,7 @@ class NoUTurnSamplerDA(NoUTurnSampler):
             else:
                 stepsize = stepsize_bar
 
-        if return_type.lower() == "dataframe":
-            if HAS_PANDAS:
-                return pandas.DataFrame.from_records(samples)
-            else:
-                warn("Pandas installation not found. Returning numpy.recarray object")
-                return samples
-        else:
-            return samples
+        return _return_samples(return_type, samples)
 
     def generate_sample(self, initial_pos, num_adapt, num_samples, stepsize=None):
         """
--- a/pgmpy/sampling/NUTS.py
+++ b/pgmpy/sampling/NUTS.py
@@ -3,16 +3,14 @@
 from __future__ import division
 
 import numpy as np
-try:
-    from pandas import DataFrame
-    HAS_PANDAS = True
-except ImportError:
-    HAS_PANDAS = False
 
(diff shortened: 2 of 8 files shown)


pgmpy/inference/Sampling.py
```diff
@@ -132,17 +142,32 @@ def rejection_sample(self, evidence=None, size=1):
...
+
             prob = max(len(_sampled) / _size, 0.01)
-            sampled = sampled.append(_sampled)
-        sampled.reset_index(inplace=True, drop=True)
-        return sampled[:size]
+
+            if HAS_PANDAS is True:
+                _sampled = _sampled.to_records(False)
+
+            for j in range(len(_sampled)):
```
ankurankan: @khalibartan Isn't there a better way of doing this ?

khalibartan: @ankurankan I don't think there is. `numpy.arrays` don't have any append method.

ankurankan: @khalibartan Wouldn't something like this: http://stackoverflow.com/questions/16246643/adding-records-to-a-numpy-record-array work ?

khalibartan: @ankurankan Yes we can do that. I have made the changes.

pgmpy/inference/Sampling.py
```diff
@@ -132,17 +142,30 @@ def rejection_sample(self, evidence=None, size=1):
         """
         if evidence is None:
             return self.forward_sample(size)
-        sampled = DataFrame(columns=self.topological_order)
+        types = [(var_name, 'int') for var_name in self.topological_order]
+        sampled = np.zeros(0, dtype=types).view(np.recarray)
         prob = 1
-        while len(sampled) < size:
-            _size = int(((size - len(sampled)) / prob) * 1.5)
+        i = 0
```
ankurankan: @khalibartan Don't use `i` as a variable name.

pgmpy/inference/Sampling.py
```diff
@@ -132,17 +142,30 @@ def rejection_sample(self, evidence=None, size=1):
...
             for evid in evidence:
-                _sampled = _sampled[_sampled.ix[:, evid[0]] == evid[1]]
+                _sampled = _sampled[_sampled[evid[0]] == evid[1]]
+
             prob = max(len(_sampled) / _size, 0.01)
-            sampled = sampled.append(_sampled)
-        sampled.reset_index(inplace=True, drop=True)
-        return sampled[:size]
+
+            if HAS_PANDAS is True:
```
ankurankan: @khalibartan @yashu-seth Can we add an extra parameter to these methods for the kind of output that we want. So, the default value would be `None` and in that case it should work as it is now (checking if user has pandas or not and then work accordingly). But it the parameter is specified for either `recarray` or `dataframe` it will try to return that object type. I don't think this will add too much overhead and will help us in dealing with situations like this. For example later on when we add inference using generated samples, we will need to know what data type are these methods returning. What do you guys think ?

yashu-seth: @ankurankan Yes, I think that would be a better option. That would allow us to get the particular data type (pandas or recarry) according to the requirement.

khalibartan: @ankurankan @yashu-seth I did the same earlier in continuous sampling :P . I think we should use string only. Not setting default value to `None` instead to `recarray`. This gives a sure behaviour. Setting the parameter to 'None' and then returning two different types based on installation is kind of ambiguous. In case of only string as  a parameter user is sure what he is doing but in the later case (one with None) he must be careful if he forgets to set parameter. And I cannot see any advantages of setting it to `None`.

ankurankan: @khalibartan If we just use strings and let's say we keep the default
string to pandas and the user doesn't have pandas installed. In that case
what should we do? Should we throw an error or fall back to recarray? And
if we are doing the falling back thing its the same as having None. And I
don't think throwing an error is a good way of handling this. Your opinion ?

khalibartan: @ankurankan Here is what I'm thinking. We should set default value to recarry arrays. Now if user give 'dataframe' as an argument and he doesn't have a installation of pandas then it will throw error that pandas is not installed. Falling back isn't a good option here. This can also happen in case of default value being set to None. Or the best option IMO will be to just return recarray object. Since recarry arrays can be transformed into DataFrame easily (in a single line) this should not be onerous task for user. This way things will be uniform and will not depend on environment installations or user end carelessness.

ankurankan: @khalibartan But in our last discussion (gitter) we had decided to accept and return pandas dataframes whenever available. Because most of the people work with pandas and this will just add an extra step for them.

khalibartan: @ankurankan We can do that , I just suggested an alternative option. 

But for returning two different data types I wouldn't like to go for setting default value to `None` and then returning data types based on env installations.

ankurankan: @khalibartan But isn't that the best option we have?

khalibartan: @ankurankan I don't agree. If some want to have samples as dataframe then he must explicitly specify it, I don't see any problem in specifying a parameter. I am not able to understand why second option is better.

yashu-seth: @khalibartan We don't want to add an extra step for users already having pandas installed. Specifying a string parameter for getting dataframes would add one. We should just ensure that pandas is not a compulsary dependency. Since pandas are so ubiquitous not returning pandas (if installed) does not sound a very good option to me.

khalibartan: @yashu-seth I find this behaviour to be ambiguous. Lets say I have got two envs one contains pandas and other just specific to pgmpy. Both envs has pgmpy installed. Now if I create a script in first env which has pandas and didn't specify the parameter for return type, later I switched the env and tried to run my script it will raise an error that will be difficult for user understand, like indexing of this type not supported etc. But with the case I'm suggesting it will raise error pandas not found. And also there is no extra step in specifying a parameter.

yashu-seth: @khalibartan How about instead of raising an error or directly falling back to recarrys we issue a warning then fall back to numpy recarrys?

khalibartan: @yashu-seth I think you didn't catch me, if we return recarray then later it will give error something like indexing not supported etc.

> Now if I create a script in first env which has pandas and didn't specify the parameter for return type, later I switched the env (which doesn't have pandas) and tried to run my script it will raise an error that will be difficult for user understand, like indexing of this type not supported etc because now data type will be recarray instead of dataframe which user designed script for.

ankurankan: @khalibartan @yashu-seth I agree with @khalibartan the code should be shareable and should work on different machines. So, in that case maybe we can keep the default value of the parameter on pandas and if we can't find pandas we will throw a warning and fall back onto recarray. This way the user will know what actually happened. 

I don't want the user to explicitly specify the parameter because firstly it adds one extra step (thus making it more complicated for the user; he will have to read the documentation to know what parameters are allowed). Also this is not a very important parameter, so I really don't want to bother the user with specifying it.

khalibartan:
@ankurankan Review the changes .

pgmpy/sampling/Sampling.py
```diff
@@ -1,15 +1,21 @@
...
+import itertools
+from warnings import warn
 
 import networkx as nx
 import numpy as np
-from pandas import DataFrame
-from pgmpy.extern.six.moves import map, range
+try:
+    from pandas import DataFrame
+    HAS_PANDAS = True
```
ankurankan: @khalibartan Rather than checking this in every file, we should move this to maybe `pgmpy/__init__.py` or some other separate file (along with other such dependency checks). And then we should import that variable in these files and use it. This will be much cleaner and move all the dependencies checking code in one place.

ankurankan (changes requested):

pgmpy/sampling/Sampling.py
ankurankan: @khalibartan Any updates on this ?

khalibartan:
@ankurankan I'm bit busy with my end sems I'll finish this on 3rd of
December

pgmpy/sampling/NUTS.py
```diff
@@ -251,10 +253,14 @@ def sample(self, initial_pos, num_samples, stepsize=None):
             position_m = self._sample(position_m, stepsize)
             samples[i] = position_m
 
-        if HAS_PANDAS is True:
-            return pd.DataFrame.from_records(samples)
-
-        return samples
+        if return_type.lower() == "dataframe":
+            if HAS_PANDAS is True:
```
ankurankan: @khalibartan Also PEP8. Shouldn't compare boolean values. Just simply use `if HAS_PANDAS`.

pgmpy/sampling/Sampling.py
```diff
@@ -131,19 +155,36 @@ def rejection_sample(self, evidence=None, size=1):
...
-                _sampled = _sampled[_sampled.ix[:, evid[0]] == evid[1]]
+                _sampled = _sampled[_sampled[evid[0]] == evid[1]]
+
             prob = max(len(_sampled) / _size, 0.01)
-            sampled = sampled.append(_sampled)
-        sampled.reset_index(inplace=True, drop=True)
-        return sampled[:size]
 
-    def likelihood_weighted_sample(self, evidence=None, size=1):
+            if HAS_PANDAS is True:
```
ankurankan: @khalibartan Also won't need to do this if you get `_sampled` as `recarray` in line 164.

pgmpy/sampling/Sampling.py
```diff
@@ -156,10 +197,13 @@ def likelihood_weighted_sample(self, evidence=None, size=1):
             None if no evidence
         size: int
             size of sample to be generated
+        return_type: string
```
ankurankan: @khalibartan Add the options for the string in the first line. Something like `return_type: string (dataframe | recarray)`.

ankurankan (changes requested):

ankurankan:
@khalibartan Any updates on this?

khalibartan:
@ankurankan I forgot about it.. I will try to finish it by tomorrow.

khalibartan:
@ankurankan See the changes.

khalibartan:
@ankurankan Can you review the changes?

ankurankan (approved):

ankurankan:
@yashu-seth Can you please review this ?

yashu-seth:
@ankurankan Will look into this by tomorrow.

pgmpy/base/__init__.py
```diff
@@ -1,5 +1,7 @@
 from .UndirectedGraph import UndirectedGraph
 from .DirectedGraph import DirectedGraph
+from .OptionalDependency import *
```
yashu-seth: @khalibartan This can be just from .OptionalDependency import HAS_PANDAS

pgmpy/sampling/NUTS.py
```diff
@@ -251,10 +251,14 @@ def sample(self, initial_pos, num_samples, stepsize=None):
...
 
-        if HAS_PANDAS is True:
-            return pd.DataFrame.from_records(samples)
-
-        return samples
+        if return_type.lower() == "dataframe":
+            if HAS_PANDAS:
+                return pandas.DataFrame.from_records(samples)
+            else:
+                warn("Pandas installation not found. Returning numpy.recarray object")
```
yashu-seth: @khalibartan warnings module has not been imported. This should give an error.

pgmpy/sampling/NUTS.py
```diff
@@ -3,16 +3,14 @@
...
-except ImportError:
-    HAS_PANDAS = False
 
 from pgmpy.sampling import HamiltonianMCDA
 from pgmpy.sampling import LeapFrog
 from pgmpy.utils import _check_1d_array_object, _check_length_equal
+from pgmpy.base import HAS_PANDAS
 
+if HAS_PANDAS:
+    import pandas
```
yashu-seth, line 10: @khalibartan PEP8. Expected 2 blank lines. Found one.

pgmpy/sampling/Sampling.py
```diff
@@ -1,18 +1,21 @@
...
 from pgmpy.inference import Inference
 from pgmpy.models import BayesianModel, MarkovChain, MarkovModel
 from pgmpy.utils.mathext import sample_discrete
+from pgmpy.extern.six.moves import map, range
+from pgmpy.base import HAS_PANDAS
 
 State = namedtuple('State', ['var', 'state'])
 
+if HAS_PANDAS:
+    import pandas
```
yashu-seth, line 17: @khalibartan PEP8. Same here.

pgmpy/sampling/Sampling.py
```diff
@@ -67,19 +75,29 @@ def forward_sample(self, size=1):
...
-                evidence = sampled.ix[:, evidence].values
-                weights = list(map(lambda t: cached_values[tuple(t)], evidence))
+                evidence = np.vstack([sampled[i] for i in evidence])
+                weights = list(map(lambda t: cached_values[tuple(t)], evidence.T))
             else:
                 weights = cpd.values
             sampled[node] = sample_discrete(states, weights, size)
-        return sampled
+
+        if return_type.lower() == "dataframe":
```
yashu-seth: @khalibartan I am not too sure about this. But this block has been repeated in every method. Can we somehow devise a way to not make it redundant?

yashu-seth (changes requested):

pgmpy/sampling/Sampling.py
khalibartan: @yashu-seth Cool. I'll create a function to do this.

khalibartan:
@yashu-seth @ankurankan Please see the changes. I have made the necessary changes.

PR #720, opened 2016-08-19
Constraint-based structure estimation

chrisittner:
# Basic constraint-based structure estimation

This PR adds `estimators/ConstraintBasedEstimator.py`. The class implements the PC algorithm to construct a partially directed acyclic graph (PDAG), either from a given set of independencies or by using independence tests on a discrete data set. The PDAG can then be completed to obtain a `BayesianModel`.
### `test_conditional_independence(self, X, Y, Zs=[])`

Chi-square conditional independence test (PGM book,  18.2.2.3, page 789)
### `build_skeleton(nodes, independencies)`

Build undirected graph from independencies/1st part of PC algorithm (PGM book, 3.4.2.1, page 85, like Algorithm 3.3)
### `skeleton_to_pdag(skel, seperating_sets)`

Orients compelled edges of skeleton/2nd part of PC ([Neapolitan, Learning Bayseian Networks](http://www.cs.technion.ac.il/~dang/books/Learning%20Bayesian%20Networks%28Neapolitan,%20Richard%29.pdf), Section 10.1.2, page 550, Algorithm 10.2)
### `pdag_to_dag(pdag)`

Method to (faithfully) orient the remaining edges to obtain `BayesianModel` (Implemented as described [here](http://www.jmlr.org/papers/volume2/chickering02a/chickering02a.pdf) on page 454 last paragraph (in text)).

Finally three methods that combine the above parts for convenient access:
### `estimate(self, significance_level=0.01)`

 -> returns `BayesianModel` estimate for data
### `estimate_skeleton(self, significance_level=0.01)`

 -> returns `UndirectedGraph` estimate for data
### `estimate_from_independencies(nodes, independencies)`

 -> static, takes set of independencies and estimates `BayesianModel`.
## Examples:

``` python
import pandas as pd
import numpy as np
from pgmpy.base import DirectedGraph
from pgmpy.estimators import ConstraintBasedEstimator
from pgmpy.independencies import Independencies

data = pd.DataFrame(np.random.randint(0, 5, size=(2500, 3)), columns=list('XYZ'))
data['sum'] = data.sum(axis=1)
print(data)

# estimate BN structue:
c = ConstraintBasedEstimator(data)
model = c.estimate()
print("Resulting network: ", model.edges())
```

Output:

```
      X  Y  Z  sum
0     1  3  4    8
1     3  3  0    6
2     4  4  1    9
...  .. .. ..  ...
2497  0  4  2    6
2498  0  3  1    4
2499  2  1  3    6

[...]

PR #721, opened 2016-08-20
MMHC Structure Estimator

chrisittner:
The MMHC algorithm for learning BN structure. MMHC consists in a constraint-based algorithm for BN skeleton learning (MMPC) and then uses score-based local search (hill climbing with tabu list) to orient the edges of the skeleton.

Implementation according to Algorithm 1, 2, and 3 in the [paper](http://www.dsl-lab.org/supplements/mmhc_paper/paper_online.pdf). I will still add some of the perfomance optimizations from Section 6 of the paper.

-> This needs to be merged after PR #720, because it uses the `test_conditional_independence`-method from there. I'll rebase once #720 is merged. Until then tests will fail.
## Reference

Tsamardinos et al., The max-min hill-climbing Bayesian network structure learning algorithm (2005),
http://www.dsl-lab.org/supplements/mmhc_paper/paper_online.pdf

### DIFF
--- a/pgmpy/estimators/MmhcEstimator.py
+++ b/pgmpy/estimators/MmhcEstimator.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python
+from pgmpy.utils.mathext import powerset
+from pgmpy.base import UndirectedGraph
+from pgmpy.models import BayesianModel
+from pgmpy.estimators import StructureEstimator, HillClimbSearch, BdeuScore
+from pgmpy.independencies import Independencies, IndependenceAssertion
+
+
+class MmhcEstimator(StructureEstimator):
+    def __init__(self, data, **kwargs):
+        """
+        Implements the MMHC hybrid structure estimation procedure for
+        learning BayesianModels from discrete data.
+
+        Parameters
+        ----------
+        data: pandas DataFrame object
+            datafame object where each column represents one variable.
+            (If some values in the data are missing the data cells should be set to `numpy.NaN`.
+            Note that pandas converts each column containing `numpy.NaN`s to dtype `float`.)
+
+        state_names: dict (optional)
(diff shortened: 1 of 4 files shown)


pgmpy/tests/test_estimators/test_MmhcEstimator.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def test_estimate(self):
+        self.assertSetEqual(set(self.est1.estimate().edges()),
+                            set([('X', 'sum'), ('Y', 'sum'), ('Z', 'sum')]))
+        self.assertSetEqual(set(self.est1.estimate(significance_level=0.001).edges()),
+                            set([('X', 'sum'), ('Y', 'sum'), ('Z', 'sum')]))
+
+    def tearDown(self):
+        del self.data1
+        del self.est1
```
yashu-seth, line 26: @chrisittner I think tests for `mmpc` should also be added.

pgmpy/estimators/MmhcEstimator.py
```diff
@@ -0,0 +1,205 @@
...
+        >>> skel.edges()
+        [('X', 'Z'), ('Y', 'Z')]
+        >>> # X, Y dependent, but conditionally independent given Z:
+        >>> sep_sets
+        {('X', 'Y'): ('Z',)}
+        """
+
+        nodes = self.state_names.keys()
+
+        def is_independent(X, Y, Zs):
```
yashu-seth, line 143: @chrisittner Can't this also be moved to the base class like the `test_conditional_independence` method?

chrisittner: Yep. Will fix.

khalibartan:
@chrisittner When will you finish up this work ?

PR #731, opened 2016-09-01
Refactor dir structure

### LINKED ISSUE #729: Directory structure change
- [x] Create a new `discrete` dir in factors and move `CPD.py`, `Factor.py` and `JointProbabilityDistribution.py` in it. The idea is to separate things related to discrete and continuous into individual directories and keep the common things (like `FactorSet`) in the current factor dir.
- [x] Move `discretize.py` into `factors/continuous` dir
- [x] Remove `ContinuousNode.py`. I don't see the point of having it.
- [x] Create a new directory at top level for sampling methods since they really don't belong to the inference part.
- [ ] Two new directories in `inference` directory namely `exact_inference` and `approximate_inference`. (Not required right now)
- [x] Rename `Factor` to `DiscreteFactor`
- [x] Common name for `NoUTurnSamplerDA` and `HamiltonianMCda`.  
- [x] Change the tests directory structure according to the above changes.  
  @khalibartan @yashu-seth Opinion ?

ankurankan:
Closes #729


ankurankan:
@yashu-seth @khalibartan @chrisittner Reviews please.

khalibartan:
+1 . Just I wanted to ask when we are going to support inference algorithms for continuous cases like variable elimination for gaussian networks will we create a new class it or handle it as a different case in current variable elimination class ? If we are going to to create a new class then we will need to change directory structure of inference algorithms also.

ankurankan:
@khalibartan I will start working on that (along with other features to complete) after I finish the documentation. And I will separate the directory then.

PR #741, opened 2016-09-09
Adds BaseFactor class

ankurankan:


khalibartan:
@ankurankan I think BaseFactor class should be a abstract class and it should have abstract methods like scope, copy, reduce, etc. The current BaseFactor class doesn't make much sense to me.

ankurankan:
@khalibartan Yeah you are right. I will make the changes.

PR #745, opened 2016-09-14
Enable inference in single-variable DBNs and add test case.

jamieforth:
Previously creating a DBNInference instance from a network consisting of only a single variable (i.e. a DBN representation of a Markov Chain) failed due to assuming edges alone were sufficient to construct new network objects. This patch introduces a number of conditions to check the number of nodes in a network before proceeding. A test case for a simply Markov Chain is also included.

### DIFF
--- a/pgmpy/base/DirectedGraph.py
+++ b/pgmpy/base/DirectedGraph.py
@@ -184,8 +184,11 @@ class DirectedGraph(nx.DiGraph):
         [('intel', 'grade'), ('intel', 'diff'), ('grade', 'diff')]
         """
         moral_graph = UndirectedGraph(self.to_undirected().edges())
-
-        for node in self.nodes():
-            moral_graph.add_edges_from(itertools.combinations(self.get_parents(node), 2))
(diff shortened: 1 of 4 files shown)


ankurankan:
@jamieforth I don't know if you have noticed but the `nodes` method of the DBN is an overloaded method so it doesn't return the original nodes. The original nodes in the DBN would be (basically it saves nodes from the first 2 slices): 

``` python3
In [17]: super(DynamicBayesianNetwork, dbn).nodes()
Out[17]: [('Weather', 1), ('Weather', 0)]
```

Do you think there would be a better possible solution using this ?

yashu-seth:
@jamieforth What is the status of this PR?

jamieforth:
@yashu-seth I did briefly look into an alternative solution using the super class nodes() method but an elegant solution wasn't immediately apparent to me, and as the patch solved my immediate problem I didn't dig much further. I'll try and take another look over the next couple of weeks and update the PR accordingly.

PR #747, opened 2016-09-15
Doc fix reflecting change in directory structure.

sudk1896:
The example in `README.md` uses the old directory structure. This PR corrects it.
`from pgmpy.factors import TabularCPD`

to

`from pgmpy.factors.discrete import TabularCPD`

Also, the installation instructions, use python3 to install. In my opinion, this is a bit confusing. Most distros ship with `python2.7.x` as default. we should instead do
`sudo python setup.py install`
and explicitly tell the user to do `python3` if they want to install it for python3.
@ankurankan Review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,12 @@ $ conda install --file requirements.txt  # or requirements-dev.txt
 Then install using:
 
 ```bash
+sudo python setup.py install
+```
+
+To install it on python3:
+
+```bash
 sudo python3 setup.py install
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,7 +122,7 @@ Here is a small snippet of pgmpy API
 ```
 python3
```
khalibartan, line 117: Change this line too.

ankurankan: @khalibartan @sudk1896 I think it's just in the wrong place. Line 116 should be ```python.

ankurankan: @khalibartan @sudk1896 I think it's just in the wrong place. Line 116 should be ```python.

ankurankan (changes requested):

PR #758, opened 2016-10-02
Refactor base

ankurankan:
Changes:
1. Removes `**kwargs` argument from `add_node`, `add_edge`, `add_nodes_from` and `add_edges_from` methods.
2. Adds `weight` argument to above methods.
3. Improves docstrings and doctests
4. Rename `check_clique` method to `is_clique`.
5. Changes in test according to the above changes.


ankurankan:
@khalibartan @yashu-seth Reviews ?

khalibartan:
@ankurankan what additional configuration parameters was kwargs allowing?

ankurankan:
@khalibartan It was just according to the networkx API in which you can store any parameter for the node or the edge. But we were just using it to store weight in pgmpy. We can still store extra values by accessing the underlying dictionary but removing **kwargs just makes the API more clear.

khalibartan:
@ankurankan Seems fine to me.

PR #777, opened 2016-12-08
Adds a few methods that will be used for modified Variable Elimination

ankurankan:


ankurankan:
@khalibartan @yashu-seth Reviews please ?

pgmpy/models/BayesianModel.py
```diff
@@ -118,6 +118,94 @@ def add_edge(self, u, v, **kwargs):
...
+        ...                        ('A', 'D'), ('D', 'C')])
+        >>> values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 4)),
+        ...                       columns=['A', 'B', 'C', 'D'])
+        >>> model.fit(values)
+        >>> model.get_cpds()
+        [<TabularCPD representing P(A:2) at 0x7f28248e2438>,
+         <TabularCPD representing P(B:2 | A:2) at 0x7f28248e23c8>,
+         <TabularCPD representing P(C:2 | B:2, D:2) at 0x7f28248e2748>,
+         <TabularCPD representing P(D:2 | A:2) at 0x7f28248e26a0>]
+        >>> model.remove_node(['A', 'B'])
```
khalibartan: @ankurankan This should be `model.remove_nodes_from(['A', 'B'])`. Also it returns None. So we cannot have an output upon execution of method. Call model.get_cpds() to display the list of cpds.

pgmpy/models/BayesianModel.py
```diff
@@ -761,3 +849,44 @@ def is_imap(self, JPD):
...
+         <TabularCPD representing P(C:2 | B:2) at 0x7f2824944240>]
+        """
+        model_copy = BayesianModel()
+        model_copy.add_nodes_from(self.nodes())
+        model_copy.add_edges_from(self.edges())
+        if self.cpds:
+            cpd_copy = []
+            for cpd in self.cpds:
+                cpd_copy.append(cpd.copy())
+            model_copy.add_cpds(*cpd_copy)
```
khalibartan: @ankurankan I think instead of this
```
        if self.cpds:
            cpd_copy = []
            for cpd in self.cpds:
                cpd_copy.append(cpd.copy())
            model_copy.add_cpds(*cpd_copy)
```
We can do this
```
        if self.cpds:
            for cpd in self.cpds:
                model_copy.add_cpds(cpd.copy())
```
There is no need to store copy of cpds in a list and add_cpds method will have same execution time as it is iterating over the list.

khalibartan:
@ankurankan I was thinking from now on we don't add address based outputs in doc-string. What say ?

pgmpy/models/BayesianModel.py
```diff
@@ -118,6 +118,94 @@ def add_edge(self, u, v, **kwargs):
...
+         <TabularCPD representing P(C:2 | B:2, D:2) at 0x7f28248e2748>,
+         <TabularCPD representing P(D:2) at 0x7f28248e26a0>]
+        """
+        affected_nodes = [v for u, v in self.edges() if u == node]
+
+        for affected_node in affected_nodes:
+            node_cpd = self.get_cpds(node=affected_node)
+            if node_cpd:
+                node_cpd.marginalize([node], inplace=True)
+                # self.remove_cpds(node_cpd)
```
yashu-seth: @ankurankan I think these comments can be removed.

pgmpy/base/DirectedGraph.py
```diff
@@ -288,3 +288,30 @@ def moralize(self):
...
+        return [node for node, out_degree in self.out_degree_iter() if
+                out_degree == 0]
+
+    def get_roots(self):
+        """
+        Returns a list of roots of the graph.
+
+        Examples
+        --------
+        >>> from pgmpy.base  import DirectedGraph
```
yashu-seth: @ankurankan Extra space after ```pgmpy.base```

ankurankan:
@khalibartan We are not really adding address based outputs. It's just the repr of the CPD and it also shows what the CPD is defined on. I don't feel there is a problem in that.

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -173,6 +173,24 @@ def test_is_iequivalent(self):
         G3 = BayesianModel([('W', 'V'), ('W', 'X'), ('Y', 'X'), ('Z', 'Y')])
         self.assertFalse(G3.is_iequivalent(G2))
 
+    def test_copy(self):
+        model_copy = self.G1.copy()
+        self.assertEqual(sorted(self.G1.nodes()), sorted(model_copy.nodes()))
+        self.assertEqual(sorted(self.G1.edges()), sorted(model_copy.edges()))
+        self.assertNotEqual(id(self.G1.get_cpds('diff')),
+                            id(model_copy.get_cpds('diff')))
```
yashu-seth, line 181: @ankurankan I think more tests for the copy method can be added. Tests to ensure that changes in original do not get reflected in the copy and vice versa.

khalibartan:
@ankurankan Yes I know that it cpds representation.There is no problem unless we might want to add doctests. Also can you squash 8 commits into 4 (first 3 commits and one about test).

ankurankan:
@khalibartan Yes, doctests will fail in these cases. Do you have any alternative way in mind to show CPDs of a network?

khalibartan:
@ankurankan I was also having doubts in that, because we need addresses in repr and we cannot print the cpds. I'll tell you whenever I come up with something for this. You can raise an issue for that.

yashu-seth (changes requested):

yashu-seth (approved):

PR #781, opened 2016-12-12
ContinuousFactor for accepting scipy.stats distributions

kris-singh:


ankurankan:
@kris-singh Tests are failing. Please fix them.

khalibartan:
@kris-singh Can you re-label the PR to something related to context of the problem which PR solves. It will help us in reviewing.

pgmpy/factors/continuous/ContinuousFactor.py
```diff
@@ -4,17 +4,21 @@
 
 import numpy as np
 import scipy.integrate as integrate
+from scipy.stats import multivariate_normal, matrix_normal, dirichlet
+import scipy
```
ankurankan, line 8: @kris-singh Do we need both of these import statements?

ankurankan:
@khalibartan @kris-singh @yashu-seth I am having second thoughts about this implementation. It will get too complicated and nested when we have a lot of distributions. I have created a new issue #782 for discussing a possible new architecture.

ankurankan (changes requested):

ankurankan:
@kris-singh We would like to have to code according to the new architecture that we discussed in #782. So closing this.

Issue #780, opened 2016-12-11
Reducing a joint gaussian distribution created from a LinearGaussianBayesianNetwork shows unstable behaviour.


chriskamphuis:
Reducing a joint gaussian distribution created from a linear gaussian bayesian network shows different outcomes when running the same program multiple times This [example](http://termbin.com/vzos) shows this behaviour. 
I suspect that either the joint gaussian distribution is not created the same in all runs, or there is a small error in the reduce function. I checked the math with a book I have, and I think the reduce function is working properly. 
It is harder to check if the joint distribution is created properly, because it changes the order of variables in different runs. This gives different orders of rows and columns in the covariance matrix. (Which should not be a problem)

yashu-seth:
@Chriskamphuis Can you reproduce the different outcomes you got? It would be easier to look into the bug.

ankurankan:
@yashu-seth Just try running it multiple times ( in different instances of python), it will give different results.

chriskamphuis:
@yashu-seth 
I just ran it twice and got respectively: `[[ 1.052213]]` and `[[ 0.367078]]`

khalibartan:
@Chriskamphuis I'm not able to open the link. It is blocked for Indian users by Department of Telecommunications, Government of India. Maybe you can put that up in a github gist and post the link or comment it here.

ankurankan:
@khalibartan Here you go: https://gist.github.com/ankurankan/3077a0b0cffa7c733790d55eb83028b7

khalibartan:
Yes indeed, it gives different results each time, but one thing when I ran this in ipython (in same session) it gave same output. :/

yashu-seth:
@ankurankan @khalibartan @Chriskamphuis I get the same value for ```jgd.mean``` for different instances of python 2. ( ```[[ 0.960813]]```) . While it changes for python 3.  The ```jgd.covariance``` remains same throughout. I also noticed that ```nx.topological_sort``` gives different order of variables for different instances of python 3.

ankurankan:
@yashu-seth But all the different topological sorts are correct. And the algorithm ideally shouldn't be dependent on different topological sorts.

yashu-seth:
@ankurankan This can be closed now.

PR #783, opened 2016-12-13
Fixes issue #780

yashu-seth:
#780

### DIFF
--- a/pgmpy/factors/continuous/CanonicalFactor.py
+++ b/pgmpy/factors/continuous/CanonicalFactor.py
@@ -260,9 +260,7 @@ class CanonicalFactor(ContinuousFactor):
         h_j = self.h[index_to_reduce]
 
         # The values for the reduced variables.
-        y = np.array([value for index,
-                      value in sorted([(self.variables.index(var), value) for var,
-                                      value in values])]).reshape(len(index_to_reduce), 1)
+        y = np.array([value for var, value in values]).reshape(len(index_to_reduce), 1)
 
         phi.variables = [self.variables[index] for index in index_to_keep]
         phi.K = K_i_i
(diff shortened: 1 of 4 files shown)


yashu-seth:
@ankurankan @khalibartan @Chriskamphuis I found the problem in the reduce methods. The values were being sorted according to variable indices. Hence, it was giving correct output only when the input parameter was sorted according to the variables. 
With @Chriskamphius' [example](https://gist.github.com/ankurankan/3077a0b0cffa7c733790d55eb83028b7) every new instance created a new order of variables for ```JointGaussianDistribution``` but the input parameters of the reduce method remained same. 
Reviews please.

yashu-seth:
@Chriskamphuis Do you know the correct result for your example?

chriskamphuis:
@yashu-seth 
Not for this example, I have another example that I have the solution for. I can test it with that example if you want?

yashu-seth:
@Chriskamphuis Yeah sure. Just keep in mind that the current dev branch has a slight change in ```LinearGaissianCPD``` API. You can have a look at its API doc. If you want may be we can add you examples as tests.

chriskamphuis:
@yashu-seth Okay, I can write some tests later this week if that is ok with you? Today I am a bit busy, but I think I can add 2 or 3 tests.

yashu-seth:
@Chriskamphuis Okay. Just let me know the results when you verify this with your known solution.

chriskamphuis:
@yashu-seth Hmm, I now see I only have an example for a JGD that was not derive from a linear Bayesian network. I can send you that data, I do not now if that is sufficient enough for you?

yashu-seth:
@Chriskamphuis I don't think that would be sufficient. Anyways you can share it.

ankurankan:
@yashu-seth You can compute something by hand and then use that to test. Shouldn't be too difficult I think.

yashu-seth:
@ankurankan The hand computed tests are already there. I also added few tests to ensure the problem is fixed. I thought maybe if we can get the solution for the Chriskamphuis' example it would be great. But since we don't have it, the current tests should suffice.

Issue #779, opened 2016-12-11
Query methods should be able to return probability values of specific states.


ankurankan:
Currently query methods can only return a Factor/TabularCPD instance over a variable. It should also work in the case of queries like:

```python
model.query([('var', 0)], evidence=[(something)]
```
In this case it should return the probability value of `P(var=0 | evidence)`.

raghavg7796:
I am new here. Can I help to contribute ?

yashu-seth:
@raghavg7796 Yes. Just let us know if you face any problem.

ankurankan:
@raghavg7796 This should be fairly easy. Currently we return the whole `DiscreteFactor` object on calling `query`, you just need to modify it to return a specific value if the state is provided.

raghavg7796:
Should we make a query method like value_query or something or it should be implicit in the query method only?

raghavg7796:
And if implicit how we will distinguish between [A, B, C, D] and [A, 0, B, 2]

ankurankan:
@raghavg7796 I am sorry the original comment was wrong. So, there should be two possible ways of querying.
```python
model_infer.query(['A', 'B'], evidence=evidence_values)
model_infer.query([('A', 0)], evidence=evidence_values)
```
In general the first argument would either be a 1-D array of 2-D array/list of tuples.

raghavg7796:
Should value given for a variable be its one of the state or value should be between [0, cardinality]

raghavg7796:
And I am getting another confusion that what if ('A', 0) is itself a variable. We can have tuples as variables also??

yashu-seth:
@raghavg7796 It can be both. But for the time being you don't have to worry about this. Let us have a discussion on gitter and I'll explain it to you.

ankurankan:
@yashu-seth @raghavg7796  I didn't think of the case when the variable names are tuples like in the case of Dynamic Bayesian Networks. Wouldn't there be an issue because of the new query argument ?

yashu-seth:
@ankurankan Yes I was also thinking about this. There could be issues with the state name architecture too with tuple variable names.

raghavg7796:
@ankurankan @yashu-seth I think we can use this with Bayesian Model only as inference with Bayesian Model requires string.

raghavg7796:
There can be 3 inputs:
1. infer.query(variables = [('A', ('A', 0), ('A', 2), 'B'...
[...]

PR #790, opened 2016-12-18
Modified query

raghavg7796:
I have modified query method to accept ('A', 0) and give the answer in a dictionary with key as ['A_0'].

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -127,9 +127,9 @@ class VariableElimination(Inference):
         variable_list = []
 
         for var in variables:
-            if isinstance(var, tuple):
+            if isinstance(var, tuple) and var not in self.variables:
                 variable_list.append(var[0])
-            if isinstance(var, str):
+            else:
                 variable_list.append(var)
 
         query_factors = self._variable_elimination(variable_list, 'marginalize',
@@ -138,11 +138,11 @@ class VariableElimination(Inference):
         query_dis_factor = {}
 
         for var in variables:
-            if isinstance(var, tuple):
+            if isinstance(var, tuple) and var not in self.variables:
                 query_factor = copy.copy(query_factors[var[0]])
                 query_factor.reduce([(var[0], var[1])])
-                query_dis_factor[var[0] + '=' + str(var[1])] = query_factor
-            if isinstance(var, str):
+                query_dis_factor[var[0]] = query_factor
+            else:
                 query_dis_factor[var] = query_factors[var]
 
         return query_dis_factor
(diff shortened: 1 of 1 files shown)


raghavg7796:
Please can someone explain me why these tests got failed.!

raghavg7796:
My PR on asking infer.query(['A', ('B', 0), 'C', ('A', 0)]), it will give result in a dict as 
dic['A'] = Discrete Factor, dic['C'] = Discrete Factor for A, dic['B=0'] = Discrete Factor with one value..
Should I have a better input/output format...and what in case of tuple variables??

yashu-seth:
@raghavg7796 I think the keys should be not be in the format 'B=0' but only 'B'. Only difference will be that the factor having B must be reduced  on ('B', 0).

@ankurankan I hope this isn't what you also intended.

yashu-seth:
@raghavg7796 @ankurankan The DBN inference tests are failing. We need to do something about this tuple problem.

yashu-seth:
@raghavg7796 A test related to state names is also failing. I know the reason. Complete this query modification first and then we will discuss it.

raghavg7796:
@yashu-seth @ankurankan For case of tuple handling, we can make a new query method like val_query or query_with_val

ankurankan:
@raghavg7796 @yashu-seth I am not in the favor of having a new method for this. Let's hold this PR till we have a better idea.

raghavg7796:
@ankurankan @yashu-seth My nostests are passing except for 3 where they can't find mock. Why here tests are failing? I have added support for tuple variables. Please review my PR.

yashu-seth:
@raghavg7796 The state name test is failing because it does not account for the case when the parameter to be mapped should be returned as it is. Which could be a case here because for inputs like ```['A', 'B']``` no mapping is required. You can look into the state_name file yourself to fix this. In case you have problems, just let me know.

raghavg7796:
@ankurankan @yashu-seth I fixed the errors. 1 more problem remains what if user asks 2 value_queries with 1 variable like query([('A', 0), ('A', 1)])? What to do in that case when user asks multiple queries for a single variable.?

yashu-seth:
@raghavg7796 I can't find a good solution for this. If you remove the state name decorator it won't work for the user defined states.
You asked me about the state name error so I told you about it, but I also believe this PR should be held until we find a better solution.

raghavg7796:
@ankurankan @yashu-seth I have implemented the case of user defined states. You can see the lattest commit. It handles the case of user defined states as well as internal states. Infact, everything is working fine, I just need to see what output will be given in case of multiple queries on same variable(may be a factor product).

ankurankan:
Closing this as we don't have a proper solution for this right now. Possible solutions can be discussed in #779

Issue #773, opened 2016-10-24
Decorator for state names should preserve the actual method signature


ankurankan:
http://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions

yashu-seth:
@ankurankan I looked into the thread. Using ```functools.wraps``` seems an option, but it won't preserve the functions's arguments list. Should we use this for now and keep the issue open?

ankurankan:
@yashu-seth Yes, we should wait for a better solution. There's no point in changing the current code if it doesn't get better.

raghavg7796:
@ankurankan @yashu-seth How about this?
http://stackoverflow.com/questions/147816/preserving-signatures-of-decorated-functions

yashu-seth:
@raghavg7796 We looked into this but could not find anything feasible. Do you have any ideas?

ankurankan:
@raghavg7796 @yashu-seth Have you guys looked into wrapt: https://pypi.python.org/pypi/wrapt ? They mention something about preserving signatures here: http://wrapt.readthedocs.io/en/stable/decorators.html. But I haven't checked it out in detail.

raghavg7796:
@ankurankan @yashu-seth I checked it out and it is giving perfect results. Should I make a PR adding the changes?

ankurankan:
@raghavg7796 Yeah sure.

raghavg7796:
@ankurankan @yashu-seth Although using wrapt gives proper documentation, it is failing nosetests because wrapt has different method of defining decorators. I am trying to look into this but still is getting problems. Please someone look.

yashu-seth:
@raghavg7796 Do you have a PR for this?

raghavg7796:
@yashu-seth I have a PR but it is giving like 290 errors, i.e. wherever it is being called.

yashu-seth:
@raghavg7796 I have not used the wrapt module before. Just make sure you have studied python decorators a bit.

raghavg7796:
@yashu-seth I studied the wrapt module and used it but it have some changed implementation like it doesn't allows *args, **kwargs in its wrapper but allows them as args and kwargs which are not editable then etc. So it is creating problems but I will still try to use it in our code.

ankurankan:
@raghavg7796 I find Graham Dumpleton's blog series on decorators quite useful: http://blog.dscpl.com.au/search/label/decorators. Maybe have a look if you have any doubts.

[...]

PR #793, opened 2016-12-22
Added wrapt.decorator to preserve documentation

raghavg7796:
Solving #773

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: python
 virtualenv:
     system_site_packages: false
 env:
+    - DISTRIB="conda" PYTHON_VERSION="3.5" COVERAGE="true"
     - DISTRIB="conda" PYTHON_VERSION="3.4" COVERAGE="true"
-    - DISTRIB="conda" PYTHON_VERSION="3.3" COVERAGE="true"
     - DISTRIB="conda" PYTHON_VERSION="2.7" COVERAGE="true"
 
 before_install:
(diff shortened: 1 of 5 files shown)


raghavg7796:
@ankurankan Tests are failing as they can't import wrapt. What should be done?

yashu-seth:
@raghavg7796 I am not too sure if we would want to have an extra dependency for this. That is why we passed on the idea of using decorator module in the first place.
@ankurankan What do you think?

raghavg7796:
I was too avoiding and tried to solved it using other methods like functools but I think we have to use dependencies for python2. Moreover @ankurankan suggested wrapt, so I think he will be OK with this.

raghavg7796:
@yashu-seth @ankurankan Then we can just use the code of decorator in wrapt. Its a few lines of code(70 lines) and can be added in another module in utils.???

ankurankan:
@yashu-seth @raghavg7796 I am fine adding a dependency for this. @raghavg7796 For travis to install wrapt you will need to add it to `requirements.txt` and `requirements-dev.txt`

raghavg7796:
@ankurankan I added it to requirements and requirements-dev ... still not working

yashu-seth:
@raghavg7796 Can you please also look into the decorator module and compare it with wrapt for our use case here.

raghavg7796:
@yashu-seth I studied decorator module and founded it difficult to be used in case of class based decorators like ours. While wrapt is properly structured for bath class and function decorators.

Issue #791, opened 2016-12-21
Issue with Markov Chain Sampling Method


kris-singh:
from pgmpy.models.MarkovChain import MarkovChain as MC
from pgmpy.factors.discrete import State
model = MC(['intel', 'diff'], [3, 2])
intel_tm = {0: {0: 0.2, 1: 0.4, 2:0.4}, 1: {0: 0, 1: 0.5, 2: 0.5}, 2: {2: 1}}
model.add_transition_model('intel', intel_tm)
diff_tm = {0: {0: 0.5, 1: 0.5}, 1: {0: 0.25, 1:0.75}}
model.add_transition_model('diff', diff_tm)
model.prob_from_sample([State('diff', 0)])
this could should produce an output like
array([ 0.27, 0.4 , 0.18, 0.23, ..., 0.29])
but it produces an ouptut like [0, 0, 0, ....0]

kris-singh:
I will reimplement the sample method for markovchain network. I will create a PR soon

raghavg7796:
@kris-singh Running it first time, it gives [0, 0, 0, .....0] and second time it gives [0.99, 1, 1, .......1]. I think it somewhat depends on start_state which we have not initialized and is set to random. This is just a possible guess but I don't know what's the actual reason.

raghavg7796:
@ankurankan Please confirm.

kris-singh:
@raghavg7796  the mehod is not correctly implemented you can look at #789 for a more clear explanation. @ankurankan I have made the changes and will create a PR today

kris-singh:
@ankurankan 
 For the stationarity condition i have implemented it as 
```python
def test_prob_from_sample(self):`
        model = MC(['a'], [4])
        transition_mat = np.array([[.95, .05, 0, 0], 
                                    [0, 0.9, 0.09, 0.01], 
                                    [0, 0.05, 0.9, 0.05], 
                                    [0.8, 0, 0.05, 0.15]])
        model.add_transition_model('a', transition_mat)
        S, U = eig(transition_mat.T)
        stationary = np.array(U[:, np.where(np.abs(S - 1.) < 1e-8)[0][0]].flat)
        stationary = stationary / np.sum(stationary)
        probabilites = []
        for k in range(0, transition_mat.shape[0]):
            probabilites.append(round(model.prob_from_sample([State('a', k)],window_size=10000),4))
        print probabilites,stationary
        for i in range(0, transition_mat.shape[0]):
[...]

Issue #789, opened 2016-12-17
Remove mock from tests


ankurankan:
Currently some tests in `MarkovChain` use mock. Remove it as it's not very readable.

kris-singh:
@ankurankan i read about mock test. So my question is for something like the start_state_test it is in some cases intialized to random state. I think this things can only be checked by using a mock object.
So how would we check something like that

ankurankan:
@kris-singh I don't think we can remove it then. I would suggest you to add some tests for the long term behavior of the `MarkovChain`.

kris-singh:
@ankurankan Can you explain the long term behaviour ?

ankurankan:
@kris-singh Read about the convergence of Markov Chains to stationary distributions.

kris-singh:
@ankurankan yes i know about it. What sort of test did you had in mind. Is is something like  the transition matrix had to ergodic or something else

ankurankan:
@kris-singh No, just run the chain for a while and see if it converges to the stationary distribution.

kris-singh:
@ankurankan 
prob_from_sample method returns a list of zeros which is strange when the expected output given in the docs is array([ 0.27,  0.4 ,  0.18,  0.23, ..., 0.29])

```python
from pgmpy.models.MarkovChain import MarkovChain as MC
from pgmpy.factors.discrete import State
model = MC(['intel', 'diff'], [3, 2])
intel_tm = {0: {0: 0.2, 1: 0.4, 2:0.4}, 1: {0: 0, 1: 0.5, 2: 0.5}, 2: {2: 1}}
model.add_transition_model('intel', intel_tm)
diff_tm = {0: {0: 0.5, 1: 0.5}, 1: {0: 0.25, 1:0.75}}
model.add_transition_model('diff', diff_tm)
model.prob_from_sample([State('diff', 0)])
```

ankurankan:
@kris-singh Can you try debugging and find what the problem is ?

kris-singh:
@ankurankan I think i figured it out. I will try to explain.If you look at the sample funciton in markovchain.
We pass a parameter size=k which indicates we should do a random walk k from the all the var and all the states but we do not do this k times we generate sample_discrete only var*cardinalities times which can be much less than say 1000. Hence we need to change this. I think this should be added as seprate issue.
`for var in self.transition_models.keys():`
` ....for st in self.transition_models[var]:`
[...]

PR #796, opened 2016-12-22
Fixes Sampling Markov Chain PR#791 #789

kris-singh:
This PR solves solves 2 Issues #791 Sampling problem and #789 Stationarity Test for Distribution. In Stationarity Test we check against steady state values obtained from the Eigen value and check the steady state distribution obtained from the samples that we generate these values have to be within a user defined parameter called tolerance to constitute a Passing of the stationarity test.

### DIFF
--- a/pgmpy/models/MarkovChain.py
+++ b/pgmpy/models/MarkovChain.py
@@ -5,7 +5,6 @@ from warnings import warn
 import numpy as np
 from pandas import DataFrame
 from scipy.linalg import eig
-import math
 
 from pgmpy.factors.discrete import State
 from pgmpy.utils import sample_discrete
@@ -284,7 +283,7 @@ class MarkovChain(object):
             for st in self.transition_models[var]:
                 var_states[var][st] = list(self.transition_models[var][st].keys())
                 var_values[var][st] = list(self.transition_models[var][st].values())
-                samples[var][st] = sample_discrete(var_states[var][st], var_values[var][st], size = size)
+                samples[var][st] = sample_discrete(var_states[var][st], var_values[var][st], size=size)
 
         for i in range(size - 1):
             for j, (var, st) in enumerate(self.state):
@@ -369,22 +368,23 @@ class MarkovChain(object):
                 self.state[j] = State(var, next_st)
             yield self.state[:]
 
-    def check_stationarity(self, tolerance=0.2, sample=None):
+    def is_stationarity(self, tolerance=0.2, sample=None):
         """
-        Checks if the given markov model is stationary and checks the steady state
+        Checks if the given markov chain is stationary and checks the steady state
         probablity values for the state are consistent.
-
 
         Parameters:
         -----------
         tolerance: float
-            reprsents the diff between actual steady state value and the computed value
+            represents the diff between actual steady state value and the computed value
         sample: [State(i,j)]
             represents the list of state which the markov chain has sampled
 
         Return Type:
         ------------
-        None
+        Boolean
+        True, if the markov chain converges to steady state distribution within the tolerance
+        False, if the markov chain does not converge to steady state distribution within tolerance
 
         Examples:
         ---------
@@ -396,20 +396,29 @@ class MarkovChain(object):
         >>> model.add_transition_model('intel', intel_tm)
         >>> diff_tm = {0: {0: 0.5, 1: 0.5}, 1: {0: 0.25, 1:0.75}}
         >>> model.add_transition_model('diff', diff_tm)
-        >>> model.check_stationarity()
+        >>> model.is_stationarity()
+        True
         """
         keys = self.transition_models.keys()
+        return_val = True
         for k in keys:
-            transition_mat = DataFrame(self.transition_models[k]).as_matrix()
+            # convert dict to numpy matrix
+            transition_mat = np.array([np.array(list(self.transition_models[k][i].values()))
+                                       for i in self.transition_models[k].keys()], dtype=np.float)
             S, U = eig(transition_mat.T)
             stationary = np.array(U[:, np.where(np.abs(S - 1.) < 1e-8)[0][0]].flat)
             stationary = (stationary / np.sum(stationary)).real
+
             probabilites = []
-            window_size = 10000 if sample == None else len(sample) 
+            window_size = 10000 if sample is None else len(sample)
             for i in range(0, transition_mat.shape[0]):
                 probabilites.extend(self.prob_from_sample([State(k, i)], window_size=window_size))
-            if any(np.abs(i) > tolerance for i in np.subtract(probabilites,stationary)):
-                raise ValueError('Model is not stationary')
+            if any(np.abs(i) > tolerance for i in np.subtract(probabilites, stationary)):
+                return_val = return_val and False
+            else:
+                return_val = return_val and True
+
+        return return_val
 
     def random_state(self):
         """
--- a/pgmpy/models/MarkovChain.py
+++ b/pgmpy/models/MarkovChain.py
@@ -4,6 +4,8 @@ from warnings import warn
 
 import numpy as np
 from pandas import DataFrame
+from scipy.linalg import eig
+import math
 
 from pgmpy.factors.discrete import State
 from pgmpy.utils import sample_discrete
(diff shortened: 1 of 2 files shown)


kris-singh:
i will also solve issue #789 in this request. As soon as we get a clear way of finding the stationarity condition is satisfied.

kris-singh:
@ankurankan can you review this and also for the stationarity condition i am thinking of implementing that as rabbe's test for convergence of the series

kris-singh:
@ankurankan @yashu-seth @khalibartan  this is ready for review..

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +369,48 @@ def generate_sample(self, start_state=None, size=1):
                 self.state[j] = State(var, next_st)
             yield self.state[:]
 
+    def check_stationarity(self, tolerance=0.2, sample=None):
```
khalibartan: @kris-singh @ankurankan IMO instead of raising a `ValueError` when markov chain is not stationary, it should return a Boolean value. Or is there any particular reason for this design?

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +369,48 @@ def generate_sample(self, start_state=None, size=1):
...
+        """
+        Checks if the given markov model is stationary and checks the steady state
+        probablity values for the state are consistent.
+
+
```
khalibartan: @kris-singh Remove extra blank line.

khalibartan (changes requested):

kris-singh:
@khalibartan your comments are updated. @ankurankan can you review this as i think this is high priority it solves the sampling error #791.

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +369,49 @@ def generate_sample(self, start_state=None, size=1):
...
+        >>> from pgmpy.models.MarkovChain import MarkovChain
+        >>> from pgmpy.factors.discrete import State
+        >>> model = MarkovChain()
+        >>> model.add_variables_from(['intel', 'diff'], [3, 2])
+        >>> intel_tm = {0: {0: 0.2, 1: 0.4, 2:0.4}, 1: {0: 0, 1: 0.5, 2: 0.5}, 2: {0: 0.3, 1: 0.3, 2: 0.4}}
+        >>> model.add_transition_model('intel', intel_tm)
+        >>> diff_tm = {0: {0: 0.5, 1: 0.5}, 1: {0: 0.25, 1:0.75}}
+        >>> model.add_transition_model('diff', diff_tm)
+        >>> model.check_stationarity()
+        """
```
yashu-seth, line 401: @kris-singh Add the output for ```model.check_stationarity```

pgmpy/models/MarkovChain.py
```diff
@@ -4,6 +4,8 @@
 
 import numpy as np
 from pandas import DataFrame
+from scipy.linalg import eig
+import math
```
yashu-seth: @kris-singh PEP8 
http://stackoverflow.com/questions/22722976/import-order-coding-standard

yashu-seth (changes requested):

yashu-seth (approved):

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +368,50 @@ def generate_sample(self, start_state=None, size=1):
...
+        Parameters:
+        -----------
+        tolerance: float
+            reprsents the diff between actual steady state value and the computed value
+        sample: [State(i,j)]
+            represents the list of state which the markov chain has sampled
+
+        Return Type:
+        ------------
+        None
```
khalibartan: @kris-singh Return type is Boolean not none.

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +368,50 @@ def generate_sample(self, start_state=None, size=1):
...
+        True
+        """
+        keys = self.transition_models.keys()
+        for k in keys:
+            transition_mat = DataFrame(self.transition_models[k]).as_matrix()
+            S, U = eig(transition_mat.T)
+            stationary = np.array(U[:, np.where(np.abs(S - 1.) < 1e-8)[0][0]].flat)
+            stationary = (stationary / np.sum(stationary)).real
+            probabilites = []
+            window_size = 10000 if sample == None else len(sample)
```
khalibartan: @kris-singh Why have you chosen 10000 as a default length?

kris-singh: This the default length previously used i didn't change it

khalibartan: @kris-singh 10000 is really a large number for burn-in period. For some complex chains it can be time consuming. Also we haven't mentioned anything about this in docstring. @ankurankan What do you suggest here ?

I was thinking about using a probabilistic approach for choosing this window size.

khalibartan:
@kris-singh Can you change the sections which are using pandas.Dataframe objects to use recarray as default? We want to make pandas as optional dependency. For reference you can checkout my PR #719

pgmpy/models/MarkovChain.py
ankurankan: @kris-singh I think the method name should be `is_stationary` and it should return True or False.

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +368,50 @@ def generate_sample(self, start_state=None, size=1):
                 self.state[j] = State(var, next_st)
             yield self.state[:]
 
+    def check_stationarity(self, tolerance=0.2, sample=None):
+        """
+        Checks if the given markov model is stationary and checks the steady state
```
ankurankan: @kris-singh Use `Markov Chain` instead of markov model.

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +368,50 @@ def generate_sample(self, start_state=None, size=1):
...
+        Parameters:
+        -----------
+        tolerance: float
+            reprsents the diff between actual steady state value and the computed value
+        sample: [State(i,j)]
+            represents the list of state which the markov chain has sampled
+
+        Return Type:
+        ------------
+        Boolean
```
ankurankan, line 385: @kris-singh Explain when do you expect True and when do you expect False.

pgmpy/models/MarkovChain.py
ankurankan: @khalibartan I think that the burn in period really depends on the distribution and the start state. So, we can't have a general number for that. What probabilistic approach are you suggesting ?

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +368,50 @@ def generate_sample(self, start_state=None, size=1):
...
+        >>> intel_tm = {0: {0: 0.2, 1: 0.4, 2:0.4}, 1: {0: 0, 1: 0.5, 2: 0.5}, 2: {0: 0.3, 1: 0.3, 2: 0.4}}
+        >>> model.add_transition_model('intel', intel_tm)
+        >>> diff_tm = {0: {0: 0.5, 1: 0.5}, 1: {0: 0.25, 1:0.75}}
+        >>> model.add_transition_model('diff', diff_tm)
+        >>> model.check_stationarity()
+        True
+        """
+        keys = self.transition_models.keys()
+        for k in keys:
+            transition_mat = DataFrame(self.transition_models[k]).as_matrix()
```
ankurankan: @kris-singh Do you really need to convert it to a `DataFrame` ? We are trying to have pandas as an optional dependency.

pgmpy/models/MarkovChain.py
```diff
@@ -367,6 +368,50 @@ def generate_sample(self, start_state=None, size=1):
...
+            S, U = eig(transition_mat.T)
+            stationary = np.array(U[:, np.where(np.abs(S - 1.) < 1e-8)[0][0]].flat)
+            stationary = (stationary / np.sum(stationary)).real
+            probabilites = []
```
ankurankan, line 412: @kris-singh Add blank lines between different sections of code.

ankurankan (changes requested):

pgmpy/models/MarkovChain.py
khalibartan: @ankurankan Yes i know that burn-in period depends upon the start state and transition probabilities that's why I was suggesting not to use some large value. This might not sound good but i was thinking of using a random value out of a set of most likely values (like choosing a random value out of (100, 1000, 10000) or something like that.

pgmpy/models/MarkovChain.py
khalibartan: @kris-singh I think there is really no need of Dataframe here.

pgmpy/models/MarkovChain.py
ankurankan: @khalibartan Yeah I agree we should have some smaller value. Or maybe by default we should have 0 burn in period. But I don't get the point of choosing a random value.

kris-singh: Should we let the user decide the burn in period. with param like burn_in_period this can be categorical like small medium long

kris-singh:
@ankurankan i didn't understand why the check failed.

ankurankan (approved):

Issue #794, opened 2016-12-22
Method like predict that gives probabilities rather than predictions.


raghavg7796:
A method like predict which given the predict_data will give probabilities for the missing variables for user to see the actual probabilities.

raghavg7796:
I am working on this and will create a PR tomorrow.

raghavg7796:
@ankurankan This can be closed now.

PR #797, opened 2016-12-23
Added model.predict_probability

raghavg7796:
Added a new method that gives probabilities of missing variables given a predict data #794 

            B_0         B_1
        80  0.439178    0.560822
        81  0.581970    0.418030
        82  0.488275    0.511725
        83  0.581970    0.418030
        84  0.510794    0.489206
        85  0.439178    0.560822
        86  0.439178    0.560822
        87  0.417124    0.582876
        88  0.407978    0.592022
        89  0.429905    0.570095
        90  0.581970    0.418030
        91  0.407978    0.592022
        92  0.429905    0.570095
        93  0.429905    0.570095
        94  0.439178    0.560822
        95  0.407978    0.592022
        96  0.559904    0.440096
        97  0.417124    0.582876
        98  0.488275    0.511725
        99  0.407978    0.592022`

Also a new error test for predict that increases the coverage.

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -714,7 +714,7 @@ class BayesianModel(DirectedGraph):
         >>> predict_data.drop('E', axis=1, inplace=True)
         >>> y_pred = model.predict(predict_data)
         >>> y_pred
-            B
+            E
         800 0
         801 1
         802 1
@@ -735,7 +735,7 @@ class BayesianModel(DirectedGraph):
             raise ValueError("No variable missing in data. Nothing to predict")
 
         elif set(data.columns) - set(self.nodes()):
-            raise ValueError("data has variables which are not in the model")
+            raise ValueError("Data has variables which are not in the model")
 
         missing_variables = set(self.nodes()) - set(data.columns)
         pred_values = defaultdict(list)
@@ -799,7 +799,7 @@ class BayesianModel(DirectedGraph):
             raise ValueError("No variable missing in data. Nothing to predict")
 
         elif set(data.columns) - set(self.nodes()):
-            raise ValueError("data has variables which are not in the model")
+            raise ValueError("Data has variables which are not in the model")
 
         missing_variables = set(self.nodes()) - set(data.columns)
         pred_values = defaultdict(list)
@@ -809,7 +809,8 @@ class BayesianModel(DirectedGraph):
             states_dict = model_inference.query(variables=missing_variables, evidence=data_point.to_dict())
             for k, v in states_dict.items():
                 for l in range(len(v.values)):
-                    pred_values[k + '_' + str(l)].append(v.values[l])
+                    state = self.get_cpds(k).state_names[k][l]
+                    pred_values[k + '_' + str(state)].append(v.values[l])
         return pd.DataFrame(pred_values, index=data.index)
 
     def get_factorized_product(self, latex=False):
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -714,11 +714,20 @@ class BayesianModel(DirectedGraph):
         >>> predict_data.drop('E', axis=1, inplace=True)
         >>> y_pred = model.predict(predict_data)
         >>> y_pred
-        array([0, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 1, 1, 0, 1, 1, 0, 1, 0, 1, 1, 0, 1,
-               1, 0, 0, 0, 1, 0, 0, 1, 0, 1, 1, 1, 1, 0, 1, 0, 0, 1, 0, 1, 1, 0, 1,
-               1, 1, 1, 1, 1, 1, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 1, 1,
-               1, 0, 1, 1, 0, 0, 1, 0, 0, 1, 1, 0, 0, 0, 1, 0, 1, 0, 1, 1, 0, 0, 1,
-               1, 1, 1, 0, 0, 0, 1, 0])
+            B
+        800 0
+        801 1
+        802 1
+        803 1
+        804 0
+        ... ...
+        993 0
+        994 0
+        995 1
+        996 1
+        997 0
+        998 0
+        999 0
         """
         from pgmpy.inference import VariableElimination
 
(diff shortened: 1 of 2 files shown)


raghavg7796:
Why my coverage decreased, I included proper tests for the case also.?

pgmpy/models/BayesianModel.py
```diff
@@ -714,11 +714,20 @@ def predict(self, data):
         >>> predict_data.drop('E', axis=1, inplace=True)
         >>> y_pred = model.predict(predict_data)
         >>> y_pred
-        array([0, 0, 0, 1, 0, 0, 1, 0, 0, 1, 0, 1, 1, 0, 1, 1, 0, 1, 0, 1, 1, 0, 1,
-               1, 0, 0, 0, 1, 0, 0, 1, 0, 1, 1, 1, 1, 0, 1, 0, 0, 1, 0, 1, 1, 0, 1,
-               1, 1, 1, 1, 1, 1, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0, 0, 1, 0, 0, 0, 1, 1,
-               1, 0, 1, 1, 0, 0, 1, 0, 0, 1, 1, 0, 0, 0, 1, 0, 1, 0, 1, 1, 0, 0, 1,
-               1, 1, 1, 0, 0, 0, 1, 0])
+            B
```
yashu-seth: @raghavg7796 This should be E

yashu-seth:
@raghavg7796 I think predicting probabilities should not be done with a separate function. It can be done using a parameter in the ```predict``` method itself.
@ankurankan What do you think?

@raghavg7796 Besides, you need to account for the state names. Fit method automatically creates the state_names dict when data has string values. So when state_names dict is present you need the output accordingly.

pgmpy/models/BayesianModel.py
raghavg7796: Oh! Sorry. I will correct that

raghavg7796:
@yashu-seth @ankurankan I thought predict_probability will be more explicit to user that's why a separate function.

raghavg7796:
@yashu-seth Predict also gives values rather than state names. Should I create a PR modifying that also.

yashu-seth:
@raghavg7796 There are some other issues with ```BayesianModel.predict``` and ```state_names``` also. I have a PR (#800) for an already existing bug.  I will cover this problem also in that PR.

raghavg7796:
@ankurankan Should this be a separate method or added as an argument in existing predict?

ankurankan:
@raghavg7796 @khalibartan @yashu-seth I think it should a separate method. Just makes more sense to me to keep it as a separate method. And you need to fix the state name thing. @yashu-seth But this doesn't depend on #800. Right ?

raghavg7796:
@ankurankan @yashu-seth I have fixed the state name thing. Please review.

pgmpy/models/BayesianModel.py
```diff
@@ -738,6 +747,72 @@ def predict(self, data):
...
+        94  0.439178    0.560822
+        95  0.407978    0.592022
+        96  0.559904    0.440096
+        97  0.417124    0.582876
+        98  0.488275    0.511725
+        99  0.407978    0.592022
+        """
+        from pgmpy.inference import VariableElimination
+
+        if set(data.columns) == set(self.nodes()):
```
khalibartan, line 798: @ankurankan @raghavg7796 @yashu-seth Do we always want data to be a `pandas.Dataframe` object ?

yashu-seth:
@ankurankan If this will be separate function then it won't depend on #800.

pgmpy/models/BayesianModel.py
ankurankan: @khalibartan But this would be the only way if we are supporting multiple variables in the queries.

pgmpy/models/BayesianModel.py
```diff
@@ -738,6 +747,72 @@ def predict(self, data):
                 pred_values[k].append(v)
         return pd.DataFrame(pred_values, index=data.index)
 
+    def predict_probability(self, data):
```
ankurankan, line 750: @raghavg7796 I don't like this implementation since most of the code is common in this and predict method. I would suggest you to call `predict_probability` method in `predict`. Or you can also have a new method which has the common part of the code.

ankurankan (changes requested):

pgmpy/models/BayesianModel.py
raghavg7796: @ankurankan So you suggest a single method that calls two methods(predict and predict_probability) based on a parameter.  Please correct me if I am getting this wrong.

pgmpy/models/BayesianModel.py
khalibartan: @ankurankan I'm not in favour of `pandas.Dataframe` . I think we should search for alternatives a bit more.

pgmpy/models/BayesianModel.py
ankurankan: @raghavg7796 No. Since most of the code is common in both `predict` and `predict_probability`, I was suggesting something like:

```python
def _predict():
     # common code

def predict():
     # call _predict and do something with results

def predict_probability():
     # call _predict and do something with the values
```
or 
```python
def predict_probabilities():
     # compute the probabilities

def predict():
     # call predict_probabilities and use that to give the final predictions
```

pgmpy/models/BayesianModel.py
ankurankan: @khalibartan But returning a pandas DataFrame makes sense to me. And in most of the cases the user would be passing in dataframe and therefore will be expecting a dataframe to be returned.

pgmpy/models/BayesianModel.py
raghavg7796: @ankurankan The two methods use different methods of Inference (query and map_query). I can use maximize in query to calculate predict but that would be useless as we already have map_query for that. Therefore the common part would just consist of initializing Inference.

pgmpy/models/BayesianModel.py
ankurankan: @khalibartan I totally missed that this will add a dependency on pandas. @raghavg7796 Can you make changes such that it only returns a DataFrame only when pandas is available otherwise returns a recarray ?

pgmpy/models/BayesianModel.py
ankurankan: @raghavg7796 Okay. Then this seems fine.

raghavg7796:
@ankurankan I have made a common _predict. Please review.

pgmpy/models/BayesianModel.py
raghavg7796: @ankurankan Please can you explain in detail what you want.

pgmpy/models/BayesianModel.py
```diff
@@ -690,6 +690,32 @@ def fit(self, data, estimator_type=None, state_names=[], complete_samples_only=T
...
         self.add_cpds(*cpds_list)
 
+    def _predict(self, data, mode):
+        from pgmpy.inference import VariableElimination
+
+        if set(data.columns) == set(self.nodes()):
+            raise ValueError("No variable missing in data. Nothing to predict")
+
+        elif set(data.columns) - set(self.nodes()):
+            raise ValueError("data has variables which are not in the model")
```
ankurankan: @raghavg7796 The d in `data` should be in caps.

pgmpy/models/BayesianModel.py
```diff
@@ -690,6 +690,32 @@ def fit(self, data, estimator_type=None, state_names=[], complete_samples_only=T
...
+
+        missing_variables = set(self.nodes()) - set(data.columns)
+        pred_values = defaultdict(list)
+
+        model_inference = VariableElimination(self)
+        for index, data_point in data.iterrows():
+            if mode == 'predict':
```
ankurankan: @raghavg7796 There's no point in doing this if we are using an if-else here. I mentioned in my earlier comment that your implementation was okay. Can you revert this ?

ankurankan (changes requested):

pgmpy/models/BayesianModel.py
raghavg7796: @ankurankan Sorry! I made that commit before reading your comment. Yes! I will revert that.

raghavg7796:
@ankurankan Reagrding pandas dataframe, please can you explain in detail what you want. I got that you want return type to be same as input type?

ankurankan:
@raghavg7796 If pandas is installed on the user's system then it should return a pandas dataframe otherwise it should return a `np.recarray`.

raghavg7796:
@ankurankan I can modify return like that. But we are accepting data as pandas Dataframe and using many methods of pandas Dataframe. Therefore User will always have to install pandas for that.?

ankurankan:
@raghavg7796 Sorry, my bad. I think we won't be able to remove pandas in this case.

ankurankan (approved):

Issue #698, opened 2016-06-13
Bug in BayesianModel.predict


yashu-seth:
``` python
wd = "C:/work/R/"
import os
os.chdir(wd)

import pandas
import numpy as np

##data_link - "https://www.kaggle.com/c/titanic/download/train.csv"
data = pandas.read_csv("kaggle.csv")
data2 = data[["Survived", "Sex", "Pclass"]]

intrain = np.random.rand(len(data2)) < 0.8

dtrain = data2[intrain]
dtest = data2[~intrain]

##print(len(dtrain), len(dtest))

from pgmpy.models import BayesianModel
titanic = BayesianModel()
titanic.add_edges_from([("Sex", "Survived"), ("Pclass", "Survived")])
titanic.fit(dtrain)
for cpd in titanic.get_cpds():
    print(cpd)

╒═════════════╤══════════╕
│ Sex(female) │ 0.340909 │
├─────────────┼──────────┤
│ Sex(male)   │ 0.659091 │
╘═════════════╧══════════╛
╒═════════════╤══════════════════════╤════════════════════╤═════════════════════╤═════════════════════╤═════════════════════╤═════════════════════╕
│ Pclass      │ Pclass(1)            │ Pclass(1)          │ Pclass(2)           │ Pclass(2)           │ Pclass(3)           │ Pclass(3)           │
├─────────────┼──────────────────────┼────────────────────┼─────────────────────┼─────────────────────┼─────────────────────┼─────────────────────┤
│ Sex         │ Sex(female)          │ Sex(male)          │ Sex(female)         │ Sex(male)           │ Sex(female)         │ Sex(male)           │
├─────────────┼──────────────────────┼────────────────────┼─────────────────────┼─────────────────────┼─────────────────────┼─────────────────────┤
│ Survived(0) │ 0.029411764705882353 │ 0.6372549019607843 │ 0.08333333333333333 │ 0.8452380952380952  │ 0.49107142857142855 │ 0.8597122302158273  │
├─────────────┼──────────────────────┼────────────────────┼─────────────────────┼─────────────────────┼─────────────────────┼─────────────────────┤
│ Survived(1) │ 0.9705882352941176   │ 0.3627450980392157 │ 0.9166666666666666  │ 0.15476190476190477 │ 0.5089285714285714  │ 0.14028776978417265 │
╘═════════════╧══════════════════════╧════════════════════╧═════════════════════╧═════════════════════╧═════════════════════╧═════════════════════╛
[...]

PR #800, opened 2016-12-23
fixes BayesianModel.predict bug from issue #698

yashu-seth:
Refer #698 for details.
This is not ready for a merge. Work is still in progress.

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -731,7 +731,8 @@ class BayesianModel(DirectedGraph):
         missing_variables = set(self.nodes()) - set(data.columns)
         pred_values = defaultdict(list)
 
-        model_inference = VariableElimination(self)
+        # Send state_names dict from one of the estimated CPDs to the inference class.
+        model_inference = VariableElimination(self, state_names=self.get_cpds()[0].state_names)
         for index, data_point in data.iterrows():
             states_dict = model_inference.map_query(variables=missing_variables, evidence=data_point.to_dict())
             for k, v in states_dict.items():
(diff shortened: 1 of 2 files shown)


yashu-seth:
This is not ready for a merge. Work is still in progress.

ankurankan:
@yashu-seth Please let me know when this is ready for review.

yashu-seth:
@ankurankan I have fixed the bug described in #698. 
But currently for ```fit``` and ```predict``` method it is required that the data should be in string format. Should we leave it to the users to convert the input data into string by mentioning it in the docs and raising necessary type checks? Or should we internally convert the data into string format?

yashu-seth:
@ankurankan What are your views on this?

ankurankan:
@yashu-seth I really don't want any code related to state names in the general code base. That will just make the whole codebase confusing. Isn't there any other way to fix this ?

yashu-seth:
@ankurankan The CPDs created using the fit method create state name dicts automatically. So for predict we need these state names. There is an internal call to ```Inference``` class in the predict method and therefore it need the state names dict as an input parameter. 

An another way would be to wrap the return value of the predict method. It returns a pandas data frame. This case does not exist in ```StateNameDecorator```, so it will have to be added. Since, currently pgmpy models do need a state name dict, a case has to be added in the state name decorator where it checks for a state name dict in the CPDs associated with the model as well.

But I feel the former method is better because anyway the ```predict``` method calls ```Inference.map_query``` which uses the state name functionality appropriately.

ankurankan:
@yashu-seth Do we need to update some tests for this ?

yashu-seth:
@ankurankan I have added some more tests for the ```predict``` method. This is ready for a review.

Issue #722, opened 2016-08-21
Uniform arguments for similar methods [INCOMPLETE]


ankurankan:
- [ ] `active_trail_nodes` accepts single nodes as argument whereas `local_independencies` accept both node and iterables.
- [x] `local_independencies` in `BayesianModel` returns an `Independencies` object whereas in `NaiveBayes` it returns a list.
- [ ] `estimator_type` should be replaced by `estimator` in all the `fit`/`predict` methods.
- [ ] Should `get_cpds` accept multiple arguments ?
- [x] `get_cpds` accept an argument but `get_factors` doesn't.
- [ ] `get_cardinality` in `BayesianModel` and `MarkovModel` are different. 
- [ ] Method for finding local independencies is called `get_local_independencies` in `MarkovModel` whereas it's `local_independencies` in `BayesianModel`. Also it accepts arguments in case of `BayesianModel`
- [ ] No `copy` method in `BayesianModel`.
- [ ] `get_cardinality` in `ClusterGraph` and `FactorGraph` doesn't accept any arguments.
- [ ] `get_cardinality` of `Factor` only accepts `lists` and returns `dict` which is different from it's behaviour in other classes..
- [ ] `get_cpd` in `TabularCPD` doesn't seem to be a suitable name for the method.

@khalibartan @yashu-seth If you have noticed any non uniform behavior please add here.

chrisittner:
- [ ] Similar to 4.: I used `prior_type` several times as a keyword argument in `estimators/`, but just `prior` seems nicer. Also `tabu_length`->`tabu`. I'll soon make a PR with small changes for the estimators.

raghavg7796:
I would like to work on this. And in 1.: I think add_edges_from should continue having a list, while add_nodes_from can accept *args

ankurankan:
@raghavg7796 Sure go ahead. But I don't think `add_edges_from` and `add_nodes_from` should have different arguments.

raghavg7796:
@ankurankan Should for 1. 
add_nodes_from and add_edges_from accept *argv 
or add_cpds should accept a iterable?
And changing anyone will require a lot of change in tests...How should I proceed?

raghavg7796:
@ankurankan And in 2. Should active_trail_nodes now return a dict with keys as variables in the iterable given to it and values as sets of active_trail_nodes?

raghavg7796:
[...]

PR #801, opened 2016-12-23
Returned independencies object in NaiveBayesian local_Independecy #722

raghavg7796:
In reference to #722


raghavg7796:
I think error is due to #802

yashu-seth:
@raghavg7796 The coverage in a PR is affected by the changes of that particular PR only. But there is only a slight decrease in coverage, so you don't have to worry.

PR #806, opened 2016-12-25
Modified docstring of NaiveBayes.local_independencies

raghavg7796:
Sorry! I forgot to update the docstring of the method.

### DIFF
--- a/pgmpy/models/NaiveBayes.py
+++ b/pgmpy/models/NaiveBayes.py
@@ -141,9 +141,8 @@ class NaiveBayes(BayesianModel):
 
     def local_independencies(self, variables):
         """
-        Returns a list of independencies objects containing the local independencies
-        of each of the variables. If local independencies does not exist for a variable
-        it gives a None for that variable.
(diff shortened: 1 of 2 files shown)


raghavg7796:
Edited only docstring...still showing failed test

pgmpy/models/NaiveBayes.py
```diff
@@ -141,9 +141,8 @@ def active_trail_nodes(self, start, observed=None):
 
     def local_independencies(self, variables):
         """
-        Returns a list of independencies objects containing the local independencies
-        of each of the variables. If local independencies does not exist for a variable
-        it gives a None for that variable.
+        Returns a independencies object containing the local independencies
```
ankurankan: @raghavg7796 I would suggest you to use something like this: `Returns an instance of Independencies`. Feels technically more correct.

ankurankan (changes requested):

ankurankan (approved):

Issue #809, opened 2016-12-27
Floating point error in Hill climbing algorithm prevents the procedure from stopping.


karttikeya:
Once the score gets low enough ( < e-10 ) sometimes the procedure gets stuck in a loop because of very slight change in score by an operation and again a perceived increase by reversing it or doing it again in case of flipping the edge.

ankurankan:
@karttikeya Can you please share your code? It will be easier for us to reproduce it.

karttikeya:
Sure. I was trying to construct a DAG for feature selection using IAMB algorithm. Following is the snippet: 

```python
import pandas as pd 
from pgmpy.models import BayesianModel
from pgmpy.estimators import BdeuScore
from pgmpy.estimators import HillClimbSearch
data = pd.read_csv('print.csv')
hc = HillClimbSearch(data, scoring_method=BicScore(data))      
best_model = hc.estimate()                                     
print best_model.edges()
```

I have modified the `HillClimbSearch.py` to print the operation executed and the score achieved. I have attached the corresponding output and also the input file Print.csv . 
[output.txt.txt](https://github.com/pgmpy/pgmpy/files/675281/output.txt.txt)

[Print.zip](https://github.com/pgmpy/pgmpy/files/675120/Print.zip)

karttikeya:
I have submitted a simple patch for the problem; to break from the loop in case the score becomes too small.However, the patch shows a coveralls error which is to be expected due to the typical nature of the error and code. Kindly review.

ankurankan:
Fixed in #811

PR #811, opened 2016-12-29
Update HillClimbSearch.py 

karttikeya:
This is a fix for: #809

### DIFF
--- a/pgmpy/estimators/HillClimbSearch.py
+++ b/pgmpy/estimators/HillClimbSearch.py
@@ -136,7 +136,7 @@ class HillClimbSearch(StructureEstimator):
         >>> est.estimate(max_indegree=1).edges()
         [('J', 'A'), ('B', 'J')]
         """
-
+        epsilon = 1e-10
         nodes = self.state_names.keys()
         if start is None:
             start = BayesianModel()
@@ -157,6 +157,8 @@ class HillClimbSearch(StructureEstimator):
                     best_score_delta = score_delta
 
             if best_operation is None:
+                break
+            if best_score_delta < epsilon:
                 break
             elif best_operation[0] == '+':
                 current_model.add_edge(*best_operation[1])
--- a/pgmpy/estimators/HillClimbSearch.py
+++ b/pgmpy/estimators/HillClimbSearch.py
@@ -156,7 +156,7 @@ class HillClimbSearch(StructureEstimator):
                     best_operation = operation
                     best_score_delta = score_delta
 
-            if best_operation is None or best_score_delta < epsilon :
+            if best_operation is None or best_score_delta < epsilon:
                 break
             elif best_operation[0] == '+':
                 current_model.add_edge(*best_operation[1])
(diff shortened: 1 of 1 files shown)


karttikeya:
Can somebody please explain what this particular comment means ? I am new to the community.

raghavg7796:
@karttikeya Coveralls tests for coverage, i.e., how much of your code is covered by tests. If you have edited a file and added some new feature that is not covered by tests, then it will show this error.

pgmpy/estimators/HillClimbSearch.py
```diff
@@ -158,6 +158,8 @@ def estimate(self, start=None, tabu_length=0, max_indegree=None):
 
             if best_operation is None:
                 break
+            if best_score_delta < epsilon:
```
kris-singh: Rather than doing this it's better that we have
eps = 1e-5
if prev_score_delta - score_delta < eps
.... then break

karttikeya: Do you mean prev_best_score_delta and best_score_delta ? 
Because in that case, it would cause to break prematurely in case best_score_delta is large enough but changes only a little in the next edge selection.

kris-singh: No, i mean something like x_curr - x_prev where x_curr indicates the current value of the score of the system. We will keep the values of score of the model in iteration i and iteration i-1 if the score dosen't change much then we can sure that we have reached some local maxima. This is a standard technique in ML

karttikeya: Got it. But isn't best_score_delta already the maximum positive change by which we can increase the score given the current network. So, it already is best_possible - curr ?

ankurankan: @karttikeya I think your changes are fine. @kris-singh I don't think there is a very easy way to do what you are suggesting.

pgmpy/estimators/HillClimbSearch.py
```diff
@@ -158,6 +158,8 @@ def estimate(self, start=None, tabu_length=0, max_indegree=None):
 
             if best_operation is None:
```
ankurankan: @karttikeya I think both the if conditions can be merged in a single statement. Something like:
```python
if best_operation is None or best_score_delta < epsilon:
    break
```

ankurankan (changes requested):

pgmpy/estimators/HillClimbSearch.py
karttikeya: Sure! Done.

karttikeya: I experimented with eps a bit. Seems that 1e-8 is good on a 64-bit machine too.

ankurankan (approved):

pgmpy/estimators/HillClimbSearch.py
```diff
@@ -156,7 +156,7 @@ def estimate(self, start=None, tabu_length=0, max_indegree=None):
                     best_operation = operation
                     best_score_delta = score_delta
 
-            if best_operation is None:
+            if best_operation is None or best_score_delta < epsilon :
```
ankurankan: @karttikeya PEP8. You shouldn't have a space before `:`

ankurankan (changes requested):

pgmpy/estimators/HillClimbSearch.py
karttikeya: Apologies for the carelessness!

ankurankan (approved):

PR #812, opened 2017-01-03
Active_trail_nodes accepts iterables

raghavg7796:
Solving #722 first.

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -423,7 +423,7 @@ class BayesianModel(DirectedGraph):
         Page 75 Algorithm 3.1
         """
         if observed:
-            observed_list = [observed] if isinstance(observed, str) else observed
+            observed_list = observed if isinstance(observed, (list, tuple)) else [observed]
         else:
             observed_list = []
         ancestors_list = self._get_ancestors_of(observed_list)
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -433,7 +433,7 @@ class BayesianModel(DirectedGraph):
         # down -> from child to parent
 
         active_trails = {}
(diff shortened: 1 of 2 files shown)


pgmpy/models/BayesianModel.py
```diff
@@ -410,9 +412,9 @@ def active_trail_nodes(self, start, observed=None):
         >>> student.add_nodes_from(['diff', 'intel', 'grades'])
         >>> student.add_edges_from([('diff', 'grades'), ('intel', 'grades')])
         >>> student.active_trail_nodes('diff')
```
khalibartan, line 414: @raghavg7796 I'm not in a agreement with allowing strings for single variable. @ankurankan  I think we are doing this in other places also. Was there any particular reason for this ?

raghavg7796: @khalibartan Bayesian Model specifies nodes should be strings so I was using strings for single variable.
@ankurankan But also there is another problem that add_edge method specifies its argument to be any hashable python objects. This is in contradiction with docstring of Bayesian Model.

raghavg7796:
@ankurankan I think this is ready for merge. Please review.

pgmpy/models/BayesianModel.py
ankurankan: @raghavg7796 Earlier we used to support only strings for node names but later we refactored the code to accept any hashable object. So, we might have missed changing the docstring somewhere. Can you create an issue for that ?

And I think you will need to do some fixes for that in this PR. Right ?

raghavg7796: @ankurankan OK! I will create the issue and make the required changes in this PR.

ankurankan:
@raghavg7796 Let me know when this is ready.

raghavg7796:
@ankurankan This is ready now.

PR #814, opened 2017-01-04
Solving PR#722

raghavg7796:
In reference to #722 . 
- [x] estimator_type should be replaced by estimator in all the fit/predict methods.
- [ ]  Should get_cpds accept multiple arguments ?
- [x]  get_cpds accept an argument but get_factors doesn't.
- [x]  get_cardinality in BayesianModel and MarkovModel are different.
- [ ]  Method for finding local independencies is called get_local_independencies in MarkovModel whereas it's local_independencies in BayesianModel. Also it accepts arguments in case of BayesianModel
- [ ]  No copy method in BayesianModel.
- [ ]  get_cardinality in ClusterGraph and FactorGraph doesn't accept any arguments.
- [ ]  get_cardinality of Factor only accepts lists and returns dict which is different from it's behaviour in other classes..
- [ ]  get_cpd in TabularCPD doesn't seem to be a suitable name for the method.

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -315,20 +315,14 @@ class BayesianModel(DirectedGraph):
                 cpd = self.get_cpds(cpd)
             self.cpds.remove(cpd)
 
-    def get_cardinality(self, node=None):
+    def get_cardinality(self):
         """
-        Returns the cardinality of the node. Throws an error if the cpd for the
-        queried node hasn't been added to the network.
-        If node is not specified returns a dictionary with the nodes as keys
+        Returns a dictionary with the nodes as keys
         and their respective cardinality as values.
 
         Parameters
         ----------
-        node: Any hashable python object.
-
-        Returns
-        -------
-        int: The cardinality of the node.
+        None.
 
         Examples
         --------
@@ -342,15 +336,10 @@ class BayesianModel(DirectedGraph):
         >>> student.add_cpds(cpd, cpd2)
         >>> student.get_cardinality()
         defaultdict(int, {'diff': 2, 'grade': 2})
-        >>> student.get_cardinality('grade')
-        defaultdict(int, {'grade': 2})
         """
         cardinalities = defaultdict(int)
-        if node:
-            cardinalities[node] = self.get_cpds(node).cardinality[0]
-        else:
-            for cpd in self.get_cpds():
-                cardinalities[cpd.variable] = cpd.cardinality[0]
+        for cpd in self.get_cpds():
+            cardinalities[cpd.variable] = cpd.cardinality[0]
         return cardinalities
 
     def check_model(self):
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -630,7 +630,7 @@ class BayesianModel(DirectedGraph):
         mm = self.to_markov_model()
         return mm.to_junction_tree()
 
-    def fit(self, data, estimator_type=None, state_names=[], complete_samples_only=True, **kwargs):
+    def fit(self, data, estimator=None, state_names=[], complete_samples_only=True, **kwargs):
         """
         Estimates the CPD for each variable based on a given data set.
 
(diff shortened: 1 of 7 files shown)


raghavg7796:
Please review this. I will then add more commits to this PR.

pgmpy/models/BayesianModel.py
```diff
@@ -674,16 +674,16 @@ def fit(self, data, estimator_type=None, state_names=[], complete_samples_only=T
...
-            estimator_type = MaximumLikelihoodEstimator
+        if estimator is None:
+            estimator = MaximumLikelihoodEstimator
         else:
-            if not issubclass(estimator_type, BaseEstimator):
+            if not issubclass(estimator, BaseEstimator):
                 raise TypeError("Estimator object should be a valid pgmpy estimator.")
 
-        estimator = estimator_type(self, data, state_names=state_names,
+        _estimator = estimator(self, data, state_names=state_names,
```
yashu-seth: @raghavg7796 I don't think this extra variable (```_estiamtor```) is required. You can directly get the cpds_list. If you are using the variable give it some other name.

pgmpy/models/MarkovModel.py
```diff
@@ -157,7 +158,14 @@ def get_factors(self):
         >>> student.add_factors(factor)
         >>> student.get_factors()
```
yashu-seth, line 161: @raghavg7796 Add the output for both the input cases.

yashu-seth (changes requested):
@raghavg7796 You can address other parts of the issue #722 as well.

pgmpy/models/MarkovModel.py
yashu-seth: @raghavg7796 Add another factor. This would make both the outputs different.

pgmpy/models/BayesianModel.py
```diff
@@ -315,10 +315,12 @@ def remove_cpds(self, *cpds):
                 cpd = self.get_cpds(cpd)
             self.cpds.remove(cpd)
 
-    def get_cardinality(self, node):
+    def get_cardinality(self, node=None):
```
yashu-seth: @raghavg7796 This would make this method inconsistent with ```BayesianModel.get_cardinality``` .  Any particular reason for adding the ```None``` condition?

raghavg7796: @yashu-seth The reason was to make it consistent with the MarkovModel.get_cardinality

yashu-seth: @raghavg7796 Okay. But, I am of the opinion that we should keep the same data type of the return value. Which means it should return a dictionary even if the ```node``` is specified . Similar changes should be made in ```MarkovModel.get_cardinality``` also. What do you think?

raghavg7796: @yashu-seth But then I think it will be of no use to take a variable node. Both methods will be like
>>>get_cardinality()['node']
>>>get_cardinality('node')['node']

yashu-seth: @raghavg7796 Then I think it should just return a dictionary without taking any argument.

raghavg7796: @yashu-seth OK! I will do that.

ankurankan:
@khalibartan @yashu-seth Can you guys please review this ?

yashu-seth (changes requested):
@raghavg7796 I think we agreed that ```get_cardinality``` should not accept an argument. The recent changes do not reflect that. Do you have other ideas?

yashu-seth:
@raghavg7796 Please mention the issues you have addressed from #722, using a checkbox or something else. It would be easier to review the changes.

raghavg7796:
- [ ]

raghavg7796:
- [x] estimator_type should be replaced by estimator in all the fit/predict methods.
- [ ]  Should get_cpds accept multiple arguments ?
- [x]  get_cpds accept an argument but get_factors doesn't.
- [x]  get_cardinality in BayesianModel and MarkovModel are different.
- [ ]  Method for finding local independencies is called get_local_independencies in MarkovModel whereas it's local_independencies in BayesianModel. Also it accepts arguments in case of BayesianModel
- [ ]  No copy method in BayesianModel.
- [ ]  get_cardinality in ClusterGraph and FactorGraph doesn't accept any arguments.
- [ ]  get_cardinality of Factor only accepts lists and returns dict which is different from it's behaviour in other classes..
- [ ]  get_cpd in TabularCPD doesn't seem to be a suitable name for the method.

ankurankan:
@raghavg7796 Don't put everything in a single PR. Will be really difficult to review. Split as much as you can.

raghavg7796:
@ankurankan OK! I will split it.

PR #815, opened 2017-01-04
Refactor Continuous Distribution and Continuous Factor

ankurankan:
- [x] New BaseDistribution Class
- [x] New CustomDistribution Class
- [ ] Refactor old ContinuousFactor Class

### DIFF
--- a/pgmpy/factors/distributions/CustomDistribution.py
+++ b/pgmpy/factors/distributions/CustomDistribution.py
@@ -1,0 +1,501 @@
+from pgmpy.factors.distributions import BaseDistribution
+
+import numpy as np
+from scipy import integrate
+
+
+class CustomDistribution(BaseDistribution):
+    def __init__(self, variables, distribution, *args, **kwargs):
+        """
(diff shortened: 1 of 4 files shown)


yashu-seth:
@ankurankan I think methods like ```reduce```, ```marginalize``` etc. should be in the base distribution not in the custom distribution. There can be cases where I have a different (better) implementation for a particular method but I want the other methods to be inherited.

For example, if we define a gamma distribution and we know the direct results for reducing a multivariate gamma distribution we can implement it by overriding it. But for its other methods, for example ```product``` we want the default behavior where the PDFs get multiplied, then we would want to inherit it from the base distribution.

ankurankan:
@yashu-seth Yes you are right. Also would you like to take over this from here ? I will be quite busy with my exams for the next couple of weeks.

yashu-seth:
@ankurankan Okay. I will look into it.

ankurankan:
Not relevant anymore.

PR #827, opened 2017-01-17
Fix documentation in CPD.py

rishabmarkand:


pgmpy/factors/discrete/CPD.py
```diff
@@ -149,6 +149,9 @@ def __repr__(self):
     def get_values(self):
         """
         Returns the cpd
+        
+        Examples
+        --------
         >>> from pgmpy.factors import TabularCPD
```
yashu-seth: @rishabmarkand This should be from ```pgmpy.factors.discerte import TabularCPD```. Can you please fix this too.

yashu-seth (approved):

PR #837, opened 2017-01-30
added BaseDistribution class

yashu-seth:


ankurankan:
@yashu-seth I was thinking about how the fallback option would work. So, let's say someone implements a new Distribution class and implements just the marginalize method (which has an efficient implementation and just needs to modify the parameters of the distribution). Now, let's say he calls the product method in which case our implementation would fall back to the default product implementation and now our returned distribution would not be the same class of distribution and none of the methods implemented by the user would work anymore.

So, for dealing with situations like this I think we will need to make all the methods (which might change the distribution) abstract. Or we can change the object type to `CustomDistribution` whenever we  fall back to the default implementation. What do you think ?

yashu-seth:
@ankurankan 

> So, let's say someone implements a new Distribution class and implements just the marginalize method (which has an efficient implementation and just needs to modify the parameters of the distribution). 

If parameters of the distribution is modified, shouldn't the pdf also get updated after the marginalize call? For example in the Joint Gaussian Distribution the pdf is defined according to the parameters and any change in the parameters changes the pdf as well.

I think all the distributions' pdf should be consistent with the parameters. So that if we call a method like product later that has a default implementation it should work.

But yes, we have to make the pdf attribute abstract. What do you think?

yashu-seth:
@ankurankan Can you please update me on your views regarding this?

ankurankan:
@yashu-seth I am still thinking about this. I will update you once I am sure that this should work.

ankurankan:
@yashu-seth I think we misunderstood each other.

So, in general we will have 2 types of distributions:
1) The standard ones which can be specified just by the parameters.
2) The ones which have a pdf function given by the user.

Now in the 1st case, let's say the user tries to do a product. Now if the other distribution is a conjugate of the original distribution, the final distribution will be in the same distribution class. For example Gaussian, multiplying with another gaussian will result in a gaussian. So, in these cases we will just need to store the parameters of the distributions. But in the case when the user multiplies a gaussian with some random distribution, the final distribution wouldn't be of the same class. In this case we will have to convert the Gaussian Distribution to a user defined distribution(`CustomDistribution` or `BaseDistribution` based on our implementation) and then multiply using our fall back method.

And I think it's better to keep all the default methods in the `CustomDistribution` rather than the `BaseClass` because:
1. Those methods are actually to work in the case of `CustomDistribution`.
2. It's better to convert the objects to `CustomDistribution` rather than `BaseDistribution` as it is more intuitive and technically correct.

What do you think ?

yashu-seth:
@ankurankan Okay, I also think this would be the best way to approach this. Will soon start working on this.

linzhp:
This PR could be safely closed

ankurankan:
Moved to #862

PR #839, opened 2017-02-06
FactorGraph: Added copy method

lohani2280:
Fixes https://github.com/pgmpy/pgmpy/issues/489

### DIFF
--- a/pgmpy/models/FactorGraph.py
+++ b/pgmpy/models/FactorGraph.py
@@ -391,3 +391,39 @@ class FactorGraph(UndirectedGraph):
             raise ValueError('DiscreteFactor for all the random variables not defined.')
 
         return np.sum(factor.values)
+
+    def copy(self):
+        """
+        Returns a copy of the model.
+
+        Returns
+        -------
+        FactorGraph : Copy of FactorGraph
+
+        Examples
+        --------
+        >>> from pgmpy.models import FactorGraph
+        >>> from pgmpy.factors.discrete import DiscreteFactor
+        >>> import numpy as np
+        >>> G = FactorGraph()
+        >>> G.add_nodes_from([('a', 'b'), ('b', 'c')])
+        >>> phi1 = DiscreteFactor(['a', 'b'], [2, 2], np.random.rand(4))
+        >>> phi2 = DiscreteFactor(['b', 'c'], [2, 2], np.random.rand(4))
+        >>> G.add_factors(phi1, phi2)
+        >>> G.add_nodes_from([phi1, phi2])
+        >>> G.add_edges_from([('a', phi1), ('b', phi1),
+        ...                   ('b', phi2), ('c', phi2)])
+        >>> G_copy = G.copy()
+        >>> G_copy.nodes()
+        [<Factor representing phi(b:2, c:2) at 0xb4badd4c>, 'b', 'c',
+          'a', <Factor representing phi(a:2, b:2) at 0xb4badf2c>]
+
+        """
+        clone_FactorGraph = FactorGraph(self.edges())
+        clone_FactorGraph.add_nodes_from(self.nodes())
+
+        if self.factors:
+           factors_copy = [factor.copy() for factor in self.factors]
+           clone_FactorGraph.add_factors(*factors_copy)
+
+        return clone_FactorGraph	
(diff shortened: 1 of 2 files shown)


pgmpy/models/FactorGraph.py
```diff
@@ -391,3 +391,39 @@ def get_partition_function(self):
...
+        >>> G.add_nodes_from([phi1, phi2])
+        >>> G.add_edges_from([('a', phi1), ('b', phi1),
+        ...                   ('b', phi2), ('c', phi2)])
+        >>> G_copy = G.copy()
+        >>> G_copy.nodes()
+        [<Factor representing phi(b:2, c:2) at 0xb4badd4c>, 'b', 'c',
+          'a', <Factor representing phi(a:2, b:2) at 0xb4badf2c>]
+
+        """
+        clone_FactorGraph = FactorGraph(self.edges())
```
ankurankan, line 422: @lohani2280 We use `lower_case_with_underscores` naming convention. So, please change the variable name here.

pgmpy/models/FactorGraph.py
```diff
@@ -391,3 +391,39 @@ def get_partition_function(self):
...
+        """
+        Returns a copy of the model.
+
+        Returns
+        -------
+        FactorGraph : Copy of FactorGraph
+
+        Examples
+        --------
+        >>> from pgmpy.models import FactorGraph
```
ankurankan, line 405: @lohani2280 Import order should be (PEP 8): 
1. standard library imports
2. related third party imports
3. local application/library specific imports

pgmpy/models/FactorGraph.py
```diff
@@ -391,3 +391,39 @@ def get_partition_function(self):
...
 
         return np.sum(factor.values)
+
+    def copy(self):
+        """
+        Returns a copy of the model.
+
+        Returns
+        -------
+        FactorGraph : Copy of FactorGraph
```
ankurankan, line 401: @lohani2280 `Copy of the FactorGraph instance` is a bit more clearer.

pgmpy/tests/test_models/test_FactorGraph.py
```diff
@@ -280,5 +280,25 @@ def test_check_model4(self):
         self.graph.add_edges_from([('a', phi3), ('c', phi3)])
         self.assertRaises(ValueError, self.graph.check_model)
 
+    def test_copy(self):
+        self.graph.add_nodes_from(['a', 'b', 'c'])
+        phi1 = DiscreteFactor(['a', 'b'], [2, 2], np.random.rand(4))
+        phi2 = DiscreteFactor(['b', 'c'], [2, 2], np.random.rand(4))
```
ankurankan, line 286: @lohani2280 I am seeing these 2 statements repeated elsewhere in this class. Can you please move all of them to the `setUp` method. Also make sure to put that in a separate commit.

ankurankan (changes requested):
Seems fine except for the minor issues I have put in comments.

lohani2280:
@ankurankan Thanks for your review. I will make the changes as per your comments and synchronize my PR very soon :)

pgmpy/models/FactorGraph.py
lohani2280: @ankurankan Sorry , but can you please tell me what's wrong here as i have observed many similar import statements and cannot identify what is wrong here.

pgmpy/tests/test_models/test_FactorGraph.py
lohani2280: @ankurankan  These two statements are present at many places but its not same everywhere . I mean if you see these in every method of the class , you will observe that there are places where the arguments are a bit different . Please correct me if i am wrong.

pgmpy/models/FactorGraph.py
ankurankan: @lohani2280 The imports should be in this order:
```python
import numpy as np
from pgmpy.models import FactorGraph
from pgmpy.factors.discrete import DiscreteFactor
```

pgmpy/tests/test_models/test_FactorGraph.py
ankurankan: @lohani2280 For example we have the same statements in line 255 and 256, 237 and 238, 224 and 225.

ankurankan (changes requested):

PR #840, opened 2017-02-06
Update Contribution Guide:

ankurankan:
Changes regarding strictly enforcing commit messages and rebasing branches before pull requests so that we can have a linear git history.

### DIFF
--- a/Contributing.md
+++ b/Contributing.md
@@ -1,6 +1,6 @@
 # Contributing to pgmpy
 
-Hi! Thanks for your interest in contributing to [pgmpy](pgmpy — pgmpy 0.1.0 documentation) :D . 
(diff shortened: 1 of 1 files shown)


lohani2280:
@ankurankan I think we should update our contribution guide by adding one more point regarding the commit structure. Our commit structure in the PR should be of the form:
```
short log: commit message

commit body
Closes <issue number>
```
where `short log` is the area/filename where you make the change `commit message` is the very brief description of the change made by you and any other additional details go into the `commit body`.
This will make our git history a bit more uniform and also help new contributors.

ankurankan:
@lohani2280 The commit message link has all this I think.

lohani2280:
@ankurankan oops sorry my bad . Yeah the link do contain all this.

PR #853, opened 2017-02-15
Move tabulate and six to separate folders:

ankurankan:
Just to make things more cleaner and adds license for six.

@khalibartan @yashu-seth I have been trying to move six and tabulate to separate dirs. But it's behaving funny or I am doing something silly. Can you guys please have a look and see what's wrong ?
I don't think we even need `__init__` in the subdirs.


pgmpy/extern/__init__.py
```diff
@@ -1,3 +1,5 @@
-from .tabulate import tabulate
+from .six import six
+from .tabulate.tabulate import tabulate
 
-__all__ = ['tabulate']
+__all__ = ['tabulate',
+           'six',]
```
khalibartan: @ankurankan Remove `,` before the closing braces.

pgmpy/extern/tabulate/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .tabulate import tabulate
+
+__all__ = ['tabulate',]
```
khalibartan: @ankurankan Remove `,` before the closing braces.

khalibartan:
@ankurankan You can omit the `__init__` file but you will need to change how we are exporting them to other modules. IMO there is no need for it. Even I'm not able to understand reason :P.

ankurankan:
Rather moved licenses inside the files. So closing this.

PR #883, opened 2017-05-19
Fix dirichlet typos in docstrings

sroecker:


ankurankan:
@sroecker Thanks for the PR.

PR #884, opened 2017-06-03
Add the ability to use a pre-count column when computing the state-counts

npfp:
In the case, the counts of the various states are already available in a
column of the dataframe, this feature allows to use them to compute the state-counts.

It therefore allows use of more compact dataframes.

### DIFF
--- a/pgmpy/estimators/base.py
+++ b/pgmpy/estimators/base.py
@@ -7,7 +7,7 @@ from scipy.stats import chisquare
 
 
 class BaseEstimator(object):
-    def __init__(self, data, state_names=None, complete_samples_only=True):
+    def __init__(self, data, state_names=None, complete_samples_only=True, count_column=None):
         """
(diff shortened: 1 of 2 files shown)


ankurankan:
@npfp I am not sure if adding a new parameter helps in any way. @khalibartan Opinion ?

khalibartan:
@ankurankan @npfp Even I'm not sure about any benefit of doing so. But I don't think pre-count would be available in most of the cases.

npfp:
Hi,

Thanks for your replies. Here are two answers to your remarks. 

- The benefit is that if you want to apply pgmpy to a dataframe which is too big to be loaded in memory, you could use a MapReduce job to pre-count the states and then apply pgmpy to it. 

- the default behaviour doesn't change and it is of course still compatible with dataframes which don't have the pre-count column. 

Thanks,
N.

PR #886, opened 2017-07-03
Update Monte Hall Problem.ipynb

jianguda:
==>
from pgmpy.factors.CPD import TabularCPD


khalibartan:
@ycaxgjd Which version of pgmpy are you using? The patch isn't compatible with latest version. In latest version we have moved TabularCPD to pgmpy.factors.discrete.

PR #894, opened 2017-09-12
MarkovModel.py: Added node as an optional argument in get_cardinality method

lohani2280:
As per the discussion now get_cardinality of MarkovModel accepts node as an optional argument and return the cardinality of the node. If node is not specified returns a dictionary with the given factors as keys and their respective cardinality as values.

### DIFF
--- a/pgmpy/models/MarkovModel.py
+++ b/pgmpy/models/MarkovModel.py
@@ -177,10 +177,18 @@ class MarkovModel(UndirectedGraph):
         for factor in factors:
             self.factors.remove(factor)
 
-    def get_cardinality(self):
+    def get_cardinality(self, node=None):
         """
-        Returns a dictionary with the given factors as keys and their respective
-        cardinality as values.
+        Returns the cardinality of the node. If node is not specified returns
+        a dictionary with the given variable as keys and their respective cardinality
+        as values.
+
+        Parameters
+        ----------
+        node: any hashable python object (optional)
+            The node whose cardinality we want. If node is not specified returns a
+            dictionary with the given variable as keys and their respective cardinality
+            as values.
 
         Examples
         --------
@@ -190,14 +198,22 @@ class MarkovModel(UndirectedGraph):
         >>> factor = DiscreteFactor(['Alice', 'Bob'], cardinality=[2, 2],
         ...                 values=np.random.rand(4))
         >>> student.add_factors(factor)
+        >>> student.get_cardinality(node='Alice')
+        2
         >>> student.get_cardinality()
         defaultdict(<class 'int'>, {'Bob': 2, 'Alice': 2})
         """
-        cardinalities = defaultdict(int)
-        for factor in self.factors:
-            for variable, cardinality in zip(factor.scope(), factor.cardinality):
-                cardinalities[variable] = cardinality
-        return cardinalities
+        if node:
+            for factor in self.factors:
+                for variable, cardinality in zip(factor.scope(), factor.cardinality):
+                    if node == variable:
+                       return cardinality
+        else:
+            cardinalities = defaultdict(int)
+            for factor in self.factors:
+                    for variable, cardinality in zip(factor.scope(), factor.cardinality):
+                        cardinalities[variable] = cardinality
+            return cardinalities
 
     def check_model(self):
         """
--- a/pgmpy/tests/test_models/test_MarkovModel.py
+++ b/pgmpy/tests/test_models/test_MarkovModel.py
@@ -115,6 +115,19 @@ class TestMarkovModelMethods(unittest.TestCase):
         self.graph.remove_factors(phi1, phi2, phi3)
         self.assertDictEqual(self.graph.get_cardinality(), {})
 
+    def test_get_cardinality_with_node(self):
+
+        self.graph.add_edges_from([('a', 'b'), ('b', 'c'), ('c', 'd'),
+                                   ('d', 'a')])
+
+        phi1 = DiscreteFactor(['a', 'b'], [2, 2], np.random.rand(4))
+        phi2 = DiscreteFactor(['c', 'd'], [1, 2], np.random.rand(2))
+        self.graph.add_factors(phi1, phi2)
+        self.assertEqual(self.graph.get_cardinality('a'), 2)
+        self.assertEqual(self.graph.get_cardinality('b'), 2)
+        self.assertEqual(self.graph.get_cardinality('c'), 1)
+        self.assertEqual(self.graph.get_cardinality('d'), 2)
+
     def test_check_model(self):
         self.graph.add_edges_from([('a', 'b'), ('b', 'c'), ('c', 'd'),
                                    ('d', 'a')])


pgmpy/models/MarkovModel.py
```diff
@@ -177,10 +177,15 @@ def remove_factors(self, *factors):
         for factor in factors:
             self.factors.remove(factor)
 
-    def get_cardinality(self):
+    def get_cardinality(self,node=None):
```
ankurankan: @lohani2280 PEP 8. Space after `,`

pgmpy/models/MarkovModel.py
```diff
@@ -177,10 +177,15 @@ def remove_factors(self, *factors):
...
         """
-        Returns a dictionary with the given factors as keys and their respective
-        cardinality as values.
+        Returns the cardinality of the node. If node is not specified returns
+        a dictionary with the given factors as keys and their respective cardinality
```
ankurankan: @lohani2280 Shouldn't this be `given variable as keys` instead of `given factors as keys`.

pgmpy/models/MarkovModel.py
```diff
@@ -177,10 +177,15 @@ def remove_factors(self, *factors):
...
+        as values.
+
+        Parameters
+        ----------
+        node: Any hashable python object(optional).
```
ankurankan: @lohani2280 Add a small description about node. Check the docstrings elsewhere to get an idea.

pgmpy/models/MarkovModel.py
```diff
@@ -190,14 +195,22 @@ def get_cardinality(self):
         >>> factor = DiscreteFactor(['Alice', 'Bob'], cardinality=[2, 2],
         ...                 values=np.random.rand(4))
         >>> student.add_factors(factor)
+        >>> student.get_cardinality('Alice')
```
ankurankan: @lohani2280 It's generally better to use named arguments to keep things clear. So, `student.get_cardinality(node='Alice')` would be better in my opinion.

pgmpy/models/MarkovModel.py
```diff
@@ -190,14 +195,22 @@ def get_cardinality(self):
...
         """
-        cardinalities = defaultdict(int)
-        for factor in self.factors:
-            for variable, cardinality in zip(factor.scope(), factor.cardinality):
-                cardinalities[variable] = cardinality
-        return cardinalities
+        if node:
+            for factor in self.factors:
+                for variable, cardinality in zip(factor.scope(), factor.cardinality):
+                    if node==variable:
```
ankurankan: PEP8. Space before and after `==`.

pgmpy/tests/test_models/test_MarkovModel.py
```diff
@@ -115,6 +115,25 @@ def test_get_cardinality(self):
...
+        self.graph.add_edges_from([('a', 'b'), ('b', 'c'), ('c', 'd'),
+                                   ('d', 'a')])
+
+        phi1 = DiscreteFactor(['a', 'b'], [1, 2], np.random.rand(2))
+        self.graph.add_factors(phi1)
+        self.assertEqual(self.graph.get_cardinality('a'), 1)
+        self.assertEqual(self.graph.get_cardinality('b'), 2)
+        self.graph.remove_factors(phi1)
+
+        phi1 = DiscreteFactor(['a', 'b'], [2, 2], np.random.rand(4))
```
ankurankan: Both these use cases seem the same to me. Let me know if I am wrong otherwise just keep one of these.

ankurankan (changes requested):

lohani2280:
@ankurankan I have made all the suggested changes. Please review them.

pgmpy/models/MarkovModel.py
```diff
@@ -190,14 +198,22 @@ def get_cardinality(self):
...
-            for variable, cardinality in zip(factor.scope(), factor.cardinality):
-                cardinalities[variable] = cardinality
-        return cardinalities
+        if node:
+            for factor in self.factors:
+                for variable, cardinality in zip(factor.scope(), factor.cardinality):
+                    if node == variable:
+                       return cardinality
+        else:
+            cardinalities = defaultdict(int)
```
khalibartan, line 212: @lohani2280 Is there any particular need for defaultdict?

pgmpy/models/MarkovModel.py
```diff
@@ -177,10 +177,18 @@ def remove_factors(self, *factors):
...
+        Returns the cardinality of the node. If node is not specified returns
+        a dictionary with the given variable as keys and their respective cardinality
+        as values.
+
+        Parameters
+        ----------
+        node: any hashable python object (optional)
+            The node whose cardinality we want. If node is not specified returns a
+            dictionary with the given variable as keys and their respective cardinality
+            as values.
```
khalibartan, line 191: @lohani2280 Specify return type.
```
Returns
--------
```

khalibartan:
@ankurankan @lohani2280 IMO we can always return a dict. I don't see any problem in returning a dict even if variable was specified.

pgmpy/models/MarkovModel.py
lohani2280: @khalibartan This method was implemented using `defaultdict` only ,so i didn't  thought of changing it. Now, after doing some experiments, i think a normal dict would also suffice this implementation.
Its because, `defaultdict` is similar to the usual dictionary (dict) container, but it has one difference:  that it is initialized with a function  that takes no arguments and provides the default value for a nonexistent key. Thus , a defaultdict will never raise a `KeyError` if you try to get an item with a key that is not currently in the dictionary unlike usual dict container.
Now ,here in this method we are always appending `variables` and its corresponding `cardinality` into the dict container.We are never retrieving a value using a key(either an existing or non-existing key). So ,if their is no retrieval then a normal dict would itself suffice . Please correct me if i am wrong.

pgmpy/models/MarkovModel.py
lohani2280: Yes , will do it soon :)

pgmpy/models/MarkovModel.py
lohani2280: @khalibartan @ankurankan I think one more thing is lacking in this implementation . If `node` passed while calling get_cardinality, is not present in the model then it should throw an error. Right now this is not happening.

lohani2280:
@khalibartan  Technically, always returning a `dict` is fine. But IMO if someone queries for the `cardinality` of a particular `node` then giving the precise output as the `cardinality` of that very `node` is better than giving the entire `dict` as output.

khalibartan:
@lohani2280 
>Technically, always returning a dict is fine. But IMO if someone queries for the cardinality of a particular node then giving the precise output as the cardinality of that very node is better than giving the entire dict as output.

No issues with that.

pgmpy/models/MarkovModel.py
ankurankan: @lohani2280 I agree with you. Could you please create an issue for this ?

lohani2280: @ankurankan Do you want me to create an issue regarding `defaultdict` implementation or regarding ` If node passed while calling get_cardinality, is not present in the model then it should throw an error. Right now this is not happening.`  ?

ankurankan: @lohani2280 Both.

PR #899, opened 2017-09-28
BayesianModel.py: Updated get_cardinality method and added test

lohani2280:
In reference to issue https://github.com/pgmpy/pgmpy/issues/722 and https://github.com/pgmpy/pgmpy/issues/893 . As per the discussion now get_cardinality of BayesianModel accepts node as an optional argument and return the cardinality of the node. If node is not specified returns a dictionary with the given factors as keys and their respective cardinality as values.

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -316,20 +316,49 @@ class BayesianModel(DirectedGraph):
                 cpd = self.get_cpds(cpd)
             self.cpds.remove(cpd)
 
-    def get_cardinality(self, node):
+    def get_cardinality(self, node=None):
         """
         Returns the cardinality of the node. Throws an error if the CPD for the
         queried node hasn't been added to the network.
 
         Parameters
         ----------
-        node: Any hashable python object.
+        node: Any hashable python object(optional).
+              The node whose cardinality we want. If node is not specified returns a
+              dictionary with the given variable as keys and their respective cardinality
+              as values.
 
         Returns
         -------
-        int: The cardinality of the node.
+        int or dict : If node is specified returns the cardinality of the node.
+                      If node is not specified returns a dictionary with the given
+                      variable as keys and their respective cardinality as values.
+
+        Examples
+        --------
+        >>> from pgmpy.models import BayesianModel
+        >>> from pgmpy.factors.discrete import TabularCPD
+        >>> student = BayesianModel([('diff', 'grade'), ('intel', 'grade')])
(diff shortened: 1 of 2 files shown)


pgmpy/models/BayesianModel.py
```diff
@@ -316,20 +316,45 @@ def remove_cpds(self, *cpds):
                 cpd = self.get_cpds(cpd)
             self.cpds.remove(cpd)
 
-    def get_cardinality(self, node):
+    def get_cardinality(self, node=None):
         """
         Returns the cardinality of the node. Throws an error if the CPD for the
-        queried node hasn't been added to the network.
+        queried node hasn't been added to the network.If node is not specified returns
+        a dictionary with the given variable as keys and their respective cardinality
```
ankurankan: @lohani2280 Some of the content from here can be moved to the `Returns` section. Usually try to keep this section short.

ankurankan (changes requested):

pgmpy/models/BayesianModel.py
lohani2280: @ankurankan Yeah , i will update the pr very soon,

pgmpy/models/BayesianModel.py
```diff
@@ -316,20 +316,50 @@ def remove_cpds(self, *cpds):
...
-        node: Any hashable python object.
+        node: Any hashable python object(optional).
+              The node whose cardinality we want. If node is not specified returns a
+              dictionary with the given variable as keys and their respective cardinality
+              as values.
 
         Returns
         -------
-        int: The cardinality of the node.
+        int: If node is specified returns the cardinality of the node.
```
ankurankan: @lohani2280 In my opinion having something like `int or dict: The description with the conditions` is better because have two separate `int` and `dict` points might suggest multiple return values.

ankurankan (changes requested):

ankurankan:
@lohani2280 Also when you are done with the changes, please squash all your commits into a single commit and have a look at this https://robots.thoughtbot.com/5-useful-tips-for-a-better-commit-message and http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html for writing a good commit message.

pgmpy/models/BayesianModel.py
lohani2280: @ankurankan Yes, it would be better . I will update the pr soon.

ankurankan (approved):

PR #903, opened 2017-10-01
Fixed broken docstring in inference classes:

khalibartan:
@ankurankan Example of this [method](https://github.com/pgmpy/pgmpy/pull/903/files#diff-b93aab348de4b1ee04af568a90bd6dc2R408) is giving an error can you take a look.


ankurankan:
@khalibartan Thanks for the patch. Can you please rebase the current dev on your brach ?

khalibartan:
@ankurankan Done!

PR #910, opened 2017-10-07
encoding problem fixed in XMLBIF test cases

### LINKED ISSUE #909: 'TestXMLBIFReaderMethods' and 'TestXMLBIFWriterMethodsString' test cases are failing on Python 3
a `nosetests -v` is failing with

> ERROR: test_get_edges (pgmpy.tests.test_readwrite.test_XMLBIF.TestXMLBIFReaderMethods)
> Traceback (most recent call last):
>   File "/home/spyder/Desktop/pgmpy/pgmpy/tests/test_readwrite/test_XMLBIF.py", line 130, in setUp
>     self.reader = XMLBIFReader(string=TEST_FILE)
>   File "/home/spyder/Desktop/pgmpy/pgmpy/readwrite/XMLBIF.py", line 55, in \_\_init\_\_
>     self.network = etree.fromstring(string).find('NETWORK')
>   File "src/lxml/lxml.etree.pyx", line 3228, in lxml.etree.fromstring (src/lxml/lxml.etree.c:79609)
>   File "src/lxml/parser.pxi", line 1843, in lxml.etree._parseMemoryDocument (src/lxml/lxml.etree.c:119069)
> ValueError: Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration.

and

> ERROR: test_write_xmlbif_statefull (pgmpy.tests.test_readwrite.test_XMLBIF.TestXMLBIFWriterMethodsString)
> Traceback (most recent call last):
>   File "/home/spyder/Desktop/pgmpy/pgmpy/tests/test_readwrite/test_XMLBIF.py", line 281, in setUp
>     reader = XMLBIFReader(string=TEST_FILE)
>   File "/home/spyder/Desktop/pgmpy/pgmpy/readwrite/XMLBIF.py", line 55, in \_\_init\_\_
>     self.network = etree.fromstring(string).find('NETWORK')
>   File "src/lxml/lxml.etree.pyx", line 3228, in lxml.etree.fromstring (src/lxml/lxml.etree.c:79609)
>   File "src/lxml/parser.pxi", line 1843, in lxml.etree._parseMemoryDocument (src/lxml/lxml.etree.c:119069)
> ValueError: Unicode strings with encoding declaration are not supported. Please use bytes input or XML fragments without declaration.
>

dasayan05:
Fixes #909

### DIFF
--- a/pgmpy/readwrite/XMLBIF.py
+++ b/pgmpy/readwrite/XMLBIF.py
@@ -52,7 +52,7 @@ class XMLBIFReader(object):
         if path:
             self.network = etree.ElementTree(file=path).getroot().find('NETWORK')
         elif string:
-            self.network = etree.fromstring(string).find('NETWORK')
+            self.network = etree.fromstring(string.encode('utf-8')).find('NETWORK')
         else:
             raise ValueError("Must specify either path or string")
         self.network_name = self.network.find('NAME').text
--- a/pgmpy/tests/test_readwrite/test_XMLBIF.py
+++ b/pgmpy/tests/test_readwrite/test_XMLBIF.py
@@ -20,7 +20,7 @@ except ImportError:
         except ImportError:
             warnings.warn("Failed to import ElementTree from any known place")
 
-TEST_FILE = """<?xml version="1.0" encoding="US-ASCII"?>
+TEST_FILE = """<?xml version="1.0"?>
 
 
 <!--


ankurankan:
@dasayan05 Thanks for the PR. But I don't think we need to fix the tests because we expect a string in https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/readwrite/XMLBIF.py#L34. So, it would be better to add a condition in the `__init__` method to handle this situation.

dasayan05:
As we always expect an `str` in `XMLBIFReader(string)`, changing

`self.network = etree.fromstring(string).find('NETWORK')`
to
`self.network = etree.fromstring(string.encode('utf-8')).find('NETWORK')`
would work absolutely **fine**.

But it turns out the root cause of this issue ( [https://bugs.launchpad.net/lxml/+bug/613302](https://bugs.launchpad.net/lxml/+bug/613302) ) is the fact that `lxml.etree.fromstring` expects an `unicode` string to not have an encoding declared but in the test cases, we indeed have encoding declaration
https://github.com/pgmpy/pgmpy/blob/3959d60562fb07121441bfd36b2cd536882325cb/pgmpy/tests/test_readwrite/test_XMLBIF.py#L23

So, what to do here ? need help @ankurankan

ankurankan:
@dasayan05 I think we should make the changes in the `__init__` method to encode the string in `utf-8` and remove the explicit encoding in the test string. But make sure to test the code without `lxml` also, since it is an optional dependency.

dasayan05:
Now all tests are passing

- on python 2
  - with lxml
  - without lxml
- on python 3
  - with lxml
  - without lxml

ankurankan:
@dasayan05 I just tried running the tests with just encoding the string and without removing the explicit encoding in the test string and the tests don't fail. Any idea why this might be happening ? Also the link for the bug in your last comment was broken. Can you check please ?

dasayan05:
**Yes you are right. Both of my commits individually fixes the tests** But

1. Reason for my first commit ( `.encode('utf-8')` ):
By converting `str` to `bytes`, it just **suppresses** the issue, doesn't really **solve** it. I just did it to handle passing any strings _with encoding declaration_ in future. You may not have this commit but then you have to make sure nobody passes an XML string with encoding declaration ever again.

2. Reason for my second commit (removing encoding declaration):
This actually **solves** the issue by following what `lxml.etree.fromstring` wants. You may not have this commit but then you are not following `lxml`'s guidelines.

ankurankan:
@dasayan05 Got it. I just checked out the issue on the lxml issue tracker also. Thanks for the PR.

PR #918, opened 2017-10-23
[WIP] Support for Continuous variables

ankurankan:
- [x] Base classes for continuous variables
- [x] Support for Gaussian variables [WIP]
- [ ] Change Factor and CPD classes to work with continuous variables
- [ ] Inference algorithms working with continuous variables 

### PS: Merge using `--no-ff` option. I would like to have all the changes in a single commit in the history.


ankurankan:
@khalibartan @yashu-seth Reviews please. More changes coming, but it will be a lot to review in a single go.

khalibartan:
@ankurankan My end sem exams are going on, so I don't I will be able to get enough time to review the PR I'll review some parts on Thursday.

khalibartan:
@ankurankan Is there any particular reason for supporting creating distributions with string directly?

pgmpy/factors/continuous/LinearGaussianCPD.py
```diff
@@ -29,106 +31,56 @@ class LinearGaussianCPD(BaseFactor):
     p(Y |x) = N(β0 + β.T * x ; σ2)
```
khalibartan, line 31: @ankurankan Can you modify the formula to use μ instead? Feels more relatable with mean.

pgmpy/factors/continuous/LinearGaussianCPD.py
```diff
@@ -29,106 +31,56 @@ class LinearGaussianCPD(BaseFactor):
...
-        evidence: iterable of any hashabale python objects
-            An iterable of the parents of the variable. None if there are no parents.
+        dist: Instance of pgmpy.factors.distributions.GaussianDistribution (optional)
+            The GaussianDistribution which is to be represented by the LinearGaussianCPD
+            instance. The GaussianDistribution should have only one variable.
+            Not required if all other parameters are given.
 
         Examples
         --------
+        For representing the following CPD:
```
khalibartan, line 65: @ankurankan I think we should put this in comment. Allows to directly copy paste the doc in ipython console.

pgmpy/factors/continuous/LinearGaussianCPD.py
```diff
@@ -29,106 +31,56 @@ class LinearGaussianCPD(BaseFactor):
...
-        Examples
-        --------
         >>> from pgmpy.factors.continuous import LinearGaussianCPD
-        >>> cpd = LinearGaussianCPD('Y',  [0.2, -2, 3, 7], 9.6, ['X1', 'X2', 'X3'])
-        >>> copy_cpd = cpd.copy()
-        >>> copy_cpd.variable
-        'Y'
-        >>> copy_cpd.evidence
-        ['X1', 'X2', 'X3']
+        >>> cpd = LinearGaussianCPD(variable='Y', mean=[2, 3, -4, 1], cov=1)
```
khalibartan, line 69: @ankurankan Needs both examples with dist and without it.

pgmpy/factors/distributions/GaussianDistribution.py
```diff
@@ -1,581 +1,675 @@
...
+        if evidence is not None and len(evidence):
+            self.evidence = np.atleast_1d(evidence)
+        else:
+            self.evidence = np.array([])
+
+        # Reshaping because mean can be either 1-D or 2-D
+        self.mean = np.atleast_2d(mean).reshape(self.variables.size, self.evidence.size + 1)
+
+        # Both variables and evidence should be either a str or 1-D array.
+        if not self.variables.ndim == 1:
```
khalibartan: @ankurankan Can you put an empty line before each `elif block`?

khalibartan (changes requested):

khalibartan:
@ankurankan I have reviewed changes except for GaussianDistribution and below. It will take me some time for that.There are a lot of changes in Gaussian Distribution. Also, have you tested the Dynamic Bayesian Network with continuous distributions?

pgmpy/factors/continuous/distributions/CustomDistribution.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 from scipy import integrate
 
-from pgmpy.factors.distributions import BaseDistribution
+from pgmpy.factors import BaseDistribution
```
harishkashyap, line 6: @ankurankan, minor: shouldn't the dirichlet be x1^alpha-1, x2^beta-1?

harishkashyap, line 6: I would think that each distribution, such as dirichlet or beta could be added into the repo itself instead of custom distribution class as they would be useful to all given that they are conjugate priors that are always used. thoughts @ankurankan ?

yashu-seth:
@ankurankan The reason I had written a separate class with canonical factor implementation was because they have direct results for sum product operations which would heavily be used in the Inference methods. It was intended to be an internal representation to facilitate Inference on Gaussian networks because conversion of Gaussian distribution to canonical factors and vice versa has direct methods. What do you think?

harishkashyap:
@yashu-seth, makes sense. perhaps a different name then?

harishkashyap (commented):
@ankurankan This is amazing work. I just have a question about discretizing the continuous variables. what is the intention here? We might have to block any hybrid models where discrete variables could be used along with other discrete CPDs as per Daphne's statement "Basic
hybrid network
continuity assumptions that hold in almost all domains imply certain relationships that hold between probabilities associated with “nearby” discretized values of a continuous variable. However, such constraints are very hard to capture in a discrete distribution, where there is no notion that two values of the variable are “close” to each other."

PR #928, opened 2017-11-04
Fix appveyor

ankurankan:
PS: Never hated windows users more


khalibartan (approved):
Merge +1. Got another reason to hate windows :joy:

PR #932, opened 2017-11-05
Removed the inconsistent behavior of query method in Variable Elimination and fixed small typo errors

### LINKED ISSUE #683: Inconsistent behavior of query method in Variable Elimination
``` python
import numpy as np
import pandas as pd
 values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)),
                      columns=['A', 'B', 'C', 'D', 'E'])
model = BayesianModel([('A', 'B'), ('C', 'B'), ('C', 'D'), ('B', 'E')])
model.fit(values)
inference = VariableElimination(model)

inference.query(variables=None)
Out[5]: 
{<Factor representing phi(D:2, C:2) at 0x7f2c3a07abe0>,
 <Factor representing phi(B:2, A:2, C:2) at 0x7f2c3a07ae48>,
 <Factor representing phi(A:2) at 0x7f2c3a07af98>,
 <Factor representing phi(C:2) at 0x7f2c3a07ab00>,
 <Factor representing phi(E:2, B:2) at 0x7f2c3a07af60>}

inference.query(variables=['A'])
Out[6]: {'A': <Factor representing phi(A:2) at 0x7f2c58718400>}
```

Returns a set when `variables` is `None`. Otherwise returns a `dict`.

lohani2280:

#### Changes
- Removed the inconsistent behavior of query method in Variable Elimination.Now the query method 
  returns an empty dict for no variables specified
- FactorGraph.py: Fixed a small typo error in docstring of get_cardinality method.
- DynamicBayesianNetwork.py: Fixed small typo error in the docstring.

💔Thank you!

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -40,10 +40,7 @@ class VariableElimination(Inference):
 
         # Dealing with the case when variables is not provided.
         if not variables:
-            all_factors = []
-            for factor_li in self.factors.values():
-                all_factors.extend(factor_li)
-            return set(all_factors)
+            return {}
 
         eliminated_variables = set()
         working_factors = {node: {factor for factor in self.factors[node]}
--- a/pgmpy/tests/test_inference/test_ExactInference.py
+++ b/pgmpy/tests/test_inference/test_ExactInference.py
@@ -93,7 +93,7 @@ class TestVariableElimination(unittest.TestCase):
                                           np.array([0.772727, 0.227273]))
 
     def test_max_marginal(self):
-        np_test.assert_almost_equal(self.bayesian_inference.max_marginal(), 0.1659, decimal=4)
+        np_test.assert_almost_equal(self.bayesian_inference.max_marginal(['A','R','J','Q','G','L']), 0.0516, decimal=4)
 
     def test_max_marginal_var(self):
         np_test.assert_almost_equal(self.bayesian_inference.max_marginal(['G']), 0.5714, decimal=4)
@@ -107,9 +107,8 @@ class TestVariableElimination(unittest.TestCase):
                                     0.3260, decimal=4)
 
     def test_map_query(self):
-        map_query = self.bayesian_inference.map_query()
-        self.assertDictEqual(map_query, {'A': 1, 'R': 1, 'J': 1, 'Q': 1, 'G': 0,
-                                         'L': 0})
+        map_query = self.bayesian_inference.map_query(['J','Q'])
+        self.assertDictEqual(map_query, {'Q': 1, 'J': 1})
 
     def test_map_query_with_evidence(self):
         map_query = self.bayesian_inference.map_query(['A', 'R', 'L'],
@@ -217,7 +216,7 @@ class TestVariableEliminationMarkov(unittest.TestCase):
                                           np.array([0.772727, 0.227273]))
 
     def test_max_marginal(self):
-        np_test.assert_almost_equal(self.markov_inference.max_marginal(), 0.1659, decimal=4)
+        np_test.assert_almost_equal(self.markov_inference.max_marginal(['A','R','J','Q','G','L']), 0.0516, decimal=4)
 
     def test_max_marginal_var(self):
         np_test.assert_almost_equal(self.markov_inference.max_marginal(['G']), 0.5714, decimal=4)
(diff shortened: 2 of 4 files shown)


pgmpy/inference/ExactInference.py
```diff
@@ -40,10 +40,7 @@ def _variable_elimination(self, variables, operation, evidence=None, elimination
 
         # Dealing with the case when variables is not provided.
         if not variables:
-            all_factors = []
-            for factor_li in self.factors.values():
-                all_factors.extend(factor_li)
-            return set(all_factors)
+            return {}
```
ankurankan, line 43: @lohani2280 I can't understand this change. Elaborate please ? And I think the tests are failing because of this.

ankurankan (dismissed):

pgmpy/inference/ExactInference.py
ankurankan: Sorry, my bad. Hadn't read the whole issue.

ankurankan (changes requested):
Tests should be easy to fix. Just pass in all the variables instead of `None` to the `query` or `map_query` methods of the failing tests.

lohani2280:
@ankurankan Edited the PR as suggested. Please review it.

pgmpy/tests/test_inference/test_ExactInference.py
```diff
@@ -93,7 +93,7 @@ def test_query_multiple_times(self):
                                           np.array([0.772727, 0.227273]))
 
     def test_max_marginal(self):
-        np_test.assert_almost_equal(self.bayesian_inference.max_marginal(), 0.1659, decimal=4)
+        np_test.assert_almost_equal(self.bayesian_inference.max_marginal(['A','R','J','Q','G','L']), 0.0516, decimal=4)
```
ankurankan, line 96: @lohani2280 Any idea why these values are changing ? And should they ?

lohani2280: @ankurankan With  `_variable_elimination` method we are implementing  variable elimination in a generalised way.Many other methods are using this method internally. So, any change here will affect other methods too. 
In our issue we are only concerned with inconsistency with the `query` method.We just want an empty dict to be returned if no variables is specified.I think implementing it locally in the `query` method would be the best idea. I would synchronise the PR soon. Please suggest your views.

ankurankan: @lohani2280 I don't think implementing it locally would be a good idea because we have other methods which use `_variable_elimination` and we would like to have the behavior consistent across all these methods. Plus it will add boilerplate code. Could you just check once if the values returned now are correct or the previous values were correct ?

lohani2280: @ankurankan The values are actually not changing. If we run the query `bayesian_inference.max_marginal(['A','J','R','Q','L','G'])` , either on this PR or without this, value is same, i.e, 0.0516. Its just that if we run the query `bayesian_inference.max_marginal()`  on this PR it throws an error because this method call for `_variable_elimination` for which we have just applied, that it should  return an empty dict on passing an empty list of variables.

ankurankan: @lohani2280 Okay. But isn't calling `max_marginal()` in the previous version and calling `max_marginal(['A', 'J', 'R', 'Q', 'L', 'G'])` in this PR equivalent ? I mean that in the previous version, didn't passing no argument meant computing max_marginal over all the variables in the network ? Because if that's the case the value for your new query should still be `0.1659`. Right ?

lohani2280: @ankurankan Right. Let me look a bit more deeply and then i'll get back again.

lohani2280: @ankurankan Calling `max_marginal()` in the previous version meant computing max_marginal over all the variables in the network. Also, calling  `max_marginal(['A', 'J', 'R', 'Q', 'L', 'G'])` in the previous version meant computing max_marginal over all the variables in the network. 
In the previous version if we run the query  `bayesian_inference.max_marginal()` , we get `0.1659`. However for `bayesian_inference.max_marginal(['A', 'J', 'R', 'Q', 'L', 'G'])` we get `0.0516`. Ideally, both the query should have same answer.Thus,there is an ambiguity. IMO  value of `max_marginal()` is correct and `max_marginal(['A', 'J', 'R', 'Q', 'L', 'G'])` isn't.
I think i found the bug -- 
When we query `max_marginal(['A', 'J', 'R', 'Q', 'L', 'G'])`, ` _variable_elimination` should have returned https://github.com/pgmpy/pgmpy/blob/fe2ca2c4ea4d7e1f4a785b9a03a01146826974cf/pgmpy/inference/ExactInference.py#L90
instead of 
https://github.com/pgmpy/pgmpy/blob/fe2ca2c4ea4d7e1f4a785b9a03a01146826974cf/pgmpy/inference/ExactInference.py#L98.
Please suggest your opinion.

PR #934, opened 2017-11-06
Fix map_query method of VariableElimination and BeliefPropagation

### LINKED ISSUE #863: Incorrect MAP inference result for VariableElimination and BeliefPropagation on 2-site Markov Model
I have a simple code where the inference on a 2-site Markov Net seems to be incorrect:
There are two binary sites a and b with one factor:
```
mm = MarkovModel([("a", "b")])
factor_ab = DiscreteFactor(variables=["a", "b"], cardinality=[2, 2], values=[0, 3, 2, 2])

bp = BeliefPropagation(model=mm)
bp.calibrate()
ve = VariableElimination(model=mm)
```
If I ask for the most likely value of a or b, it should yield "1" in both cases (in absence of evidence), but neither VE nor BP give the correct answer:
```
>>> ve.map_query(variables=["a"])
{'a': 0}
>>> ve.map_query(variables=["b"])
{'b': 1}
```
And BP gives the same result. However, both correctly predict the joined most probable assignment:
```
>>> ve.map_query(variables=["a", "b"])
{"a": 0, "b": 1}
```

ankurankan:
Because of change in behavior of maximize at some time, needed to switch the passed argument
to marginalize. Plus adds code to BeliefPropagation's map_query to return states rather than
Factor instances.

Fixes the case when no argument is passed

#### Changes
1. Change passed argument to `marginalize` from `maximize`.
2. Change `BeliefPropagation`'s `map_query` to return state instead of `DiscreteFactor`.
3. Fix for the case when no argument is passed to `map_query` of `BeliefPropagation`.

💔Thank you!

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -171,6 +171,9 @@ class VariableElimination(Inference):
         """
         Computes the MAP Query over the variables given the evidence.
 
+        Note: When multiple variables are passed, it returns the map_query for each
+        of them individually.
+
         Parameters
         ----------
         variables: list
@@ -195,8 +198,8 @@ class VariableElimination(Inference):
         >>> inference = VariableElimination(model)
         >>> phi_query = inference.map_query(['A', 'B'])
         """
-        elimination_variables = set(self.variables) - set(evidence.keys()) if evidence else set()
-        final_distribution = self._variable_elimination(elimination_variables, 'maximize',
+        # TODO:Check the note in docstring. Change that behavior to return the joint MAP
+        final_distribution = self._variable_elimination(variables, 'marginalize',
                                                         evidence=evidence,
                                                         elimination_order=elimination_order)
         # To handle the case when no argument is passed then
@@ -661,6 +664,9 @@ class BeliefPropagation(Inference):
         """
         MAP Query method using belief propagation.
 
+        Note: When multiple variables are passed, it returns the map_query for each
+        of them individually.
+
         Parameters
         ----------
         variables: list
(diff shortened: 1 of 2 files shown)


ankurankan:
Please open #863 after merging this.

khalibartan:
@ankurankan  I'm not able to understand this
>Because of change in behavior of maximize at some time

pgmpy/inference/ExactInference.py
```diff
@@ -171,6 +171,10 @@ def map_query(self, variables=None, evidence=None, elimination_order=None):
         """
         Computes the MAP Query over the variables given the evidence.
 
+        TODO: Change this behavior to return the joint MAP
```
khalibartan: @ankurankan Please put TODO in inline comments rather than docstring.

khalibartan (changes requested):

ankurankan:
Sorry, I meant at some point in time.
> @ankurankan I'm not able to understand this
> 
> Because of change in behavior of maximize at some time

pgmpy/inference/ExactInference.py
ankurankan: @khalibartan But why ? This seems to be the best place to put this todo.

khalibartan: @ankurankan Usual standard is to use #TODO (even IDE support identification of `# TODO:`) rather than putting TODO in docstring. I think this is because TODO are for devs not for users.

ankurankan: @khalibartan Done.

khalibartan (approved):

khalibartan:
@ankurankan  I just noticed that our doc is bit inconsistent. I mean ceratin docstring contains note about return type with
```
Returns
-------
```
certain methods don't.

ankurankan:
@khalibartan I try to keep it as consistent as possible. Maybe this is because these methods were written way back. And I am not improving these right now because I will be completely refactoring this part for #918 and will fix all the minor issues with it.

khalibartan:
@ankurankan Okay. I'm merging this if there are no changes to be made in this PR.

ankurankan:
Thanks 👍

PR #937, opened 2017-11-07
Fixes issue with disconnected node when reading models from files

### LINKED ISSUE #936: Reading a XML BIF with isolated nodes
The XMLBIFReader class doesn't allow to return the model from a BIF XML file where there are isolated nodes. 

I can create the XMLBIFReader:
`bif = pgmpy.readwrite.XMLBIF.XMLBIFReader('diabetes.xml')`

but when trying to return the model:
`bif_model = bif.get_model()`

It returns an exception because is trying to load the CPD of the isolated node, but this node is not included in the BayesianModel:
 `ValueError: ('CPD defined on variable not in the model', <TabularCPD representing P(skin:1) at 0x7f669d905c10>)` arises.

Reviewing the  `XMLBIFReader.get_model()`, we can see that the Bayesian model is created from the edges: 

`model = BayesianModel(self.get_edges())`

As an isolated node has no edge by definition, I think that the BayesianModel can't know that an isolated node exists when adding the cpds to the BayesianModel.

I provide the following XML BIF as a MWE. It is a XML BIF learnt using WEKA BayesNet class.
[diabetes.zip](https://github.com/pgmpy/pgmpy/files/1449591/diabetes.zip)

I think that even when an isolated node is not the most useful thing to do, it is a nuisance that we can't load the rest of the network.

Execution environment.
pgmpy: 0.1.3
Python: 2.7
OS: Ubuntu 16.04

ankurankan:

#### Changes
1. All the `get_model` methods now use `add_nodes_from` to add nodes to the network instead of directly using `add_edges_from`.
2. Minor change in tests for `ProbModelXML` since explicitly adding nodes add a new parameter `weight` for each node. And since it doesn't affect rest of the codebase in any way, I added this parameter in tests.

💔Thank you!


pgmpy/readwrite/BIF.py
```diff
@@ -284,9 +284,10 @@ def get_model(self):
         <pgmpy.models.BayesianModel.BayesianModel object at 0x7f20af154320>
         """
         try:
-            model = BayesianModel(self.variable_edges)
-            model.name = self.network_name
+            model = BayesianModel()
             model.add_nodes_from(self.variable_names)
+            model.add_edges_from(self.variable_edges)
+            model.name = self.network_name
```
khalibartan, line 290: This and earlier one are same.

khalibartan (dismissed):

khalibartan (dismissed):
@ankurankan Are you going to add tests for this? I believe we should.

khalibartan (approved):

PR #943, opened 2017-11-09
Changes made in sampling methods to accomodate state_name

### LINKED ISSUE #942: Sampling methods in discrete case don't use state_names in provided in CPD
Currently, sampling methods don't use state names if provided in CPD which makes it difficult to understand the output.

khalibartan:
created a different method to handle state name to support extensibility needed for continuous models.

#### Changes
- Added changes to accommodate `state_name` in `Sampling.py` methods

💔Thank you!

### DIFF
--- a/pgmpy/sampling/Sampling.py
+++ b/pgmpy/sampling/Sampling.py
@@ -9,7 +9,7 @@ from pgmpy.inference import Inference
 from pgmpy.models import BayesianModel, MarkovChain, MarkovModel
 from pgmpy.utils.mathext import sample_discrete
 from pgmpy.extern.six.moves import map, range
-from pgmpy.sampling import _return_samples
+from pgmpy.sampling import _return_samples, _map_to_state_name
 
 
 State = namedtuple('State', ['var', 'state'])
@@ -29,6 +29,7 @@ class BayesianModelSampling(Inference):
     --------------
     forward_sample(size)
     """
+
     def __init__(self, model):
         if not isinstance(model, BayesianModel):
             raise TypeError("Model expected type: BayesianModel, got type: ", type(model))
@@ -87,6 +88,7 @@ class BayesianModelSampling(Inference):
                 weights = cpd.values
             sampled[node] = sample_discrete(states, weights, size)
 
+        sampled = _map_to_state_name(self.model, sampled)
         return _return_samples(return_type, sampled)
 
     def pre_compute_reduce(self, variable):
@@ -158,6 +160,7 @@ class BayesianModelSampling(Inference):
 
             i += len(_sampled)
 
+        sampled = _map_to_state_name(self.model, sampled)
         return _return_samples(return_type, sampled)
 
     def likelihood_weighted_sample(self, evidence=None, size=1, return_type="dataframe"):
@@ -230,6 +233,7 @@ class BayesianModelSampling(Inference):
                 else:
                     sampled[node] = sample_discrete(states, cpd.values, size)
 
+        sampled = _map_to_state_name(self.model, sampled)
         return _return_samples(return_type, sampled)
 
 
@@ -268,7 +272,9 @@ class GibbsSampling(MarkovChain):
     1      0    0
     2      1    1
     """
+
     def __init__(self, model=None):
+        self.model = model
         super(GibbsSampling, self).__init__()
         if isinstance(model, BayesianModel):
             self._get_kernel_from_bayesian_model(model)
@@ -388,6 +394,7 @@ class GibbsSampling(MarkovChain):
                 self.state[j] = State(var, next_st)
             sampled[i + 1] = np.array([st for var, st in self.state])
 
+        sampled = _map_to_state_name(self.model, sampled)
         return _return_samples(return_type, sampled)
 
     def generate_sample(self, start_state=None, size=1):
--- a/pgmpy/sampling/base.py
+++ b/pgmpy/sampling/base.py
@@ -3,6 +3,7 @@ from warnings import warn
 import numpy as np
 
 from pgmpy import HAS_PANDAS
+from pgmpy.models import BayesianModel
 from pgmpy.utils import _check_1d_array_object, _check_length_equal
 
 if HAS_PANDAS:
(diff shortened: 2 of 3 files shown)


pgmpy/sampling/base.py
```diff
@@ -418,3 +419,35 @@ def _return_samples(return_type, samples):
             return samples
     else:
         return samples
+
+
+def _map_to_state_name(model, samples):
+    """
+        A utility function to map samples to corresponding state name
+    """
+    # Assuming all cpd in model either have state name or not
```
ankurankan, line 428: @khalibartan In my opinion this is not a good assumption. I think it would be better to check and throw an error message accordingly. It would be much easier for the user to debug the issue as well.

pgmpy/sampling/base.py
```diff
@@ -418,3 +419,35 @@ def _return_samples(return_type, samples):
...
+            types.append(('_weight', 'float'))
+        else:
+            types = [(var_name, object) for var_name in samples.dtype.names]
+            index = len(samples.dtype)
+
+        named_samples = np.zeros(samples.size, dtype=types).view(np.recarray)
+
+        for node in samples.dtype.names[:index]:
+            cpd = getattr(model, method)(node)
+            named_samples[node] = np.array(list(map(lambda x: cpd.state_names[node][x], samples[node])))
```
ankurankan, line 446: @khalibartan Can use `np.fromiter`. That should be faster.

ankurankan: Or a faster approach would be to create an array with state names and then just index over it. For example:
```python
states = np.array(['state_1', 'state_2'])
samples = np.array([0, 0, 1, 1, 0, 1, 0])
state_samples = states[samples]
```

pgmpy/sampling/base.py
```diff
@@ -418,3 +419,35 @@ def _return_samples(return_type, samples):
             return samples
     else:
         return samples
+
+
+def _map_to_state_name(model, samples):
+    """
+        A utility function to map samples to corresponding state name
```
ankurankan, line 426: Could you add an example here as well ? It's difficult to understand what exactly it's supposed to do by looking at the code.

pgmpy/sampling/Sampling.py
```diff
@@ -87,6 +88,7 @@ def forward_sample(self, size=1, return_type='dataframe'):
                 weights = cpd.values
             sampled[node] = sample_discrete(states, weights, size)
 
+        sampled = _map_to_state_name(self.model, sampled)
```
ankurankan, line 91: I do think we should mention this in the docstring that it returns state names if defined.

pgmpy/sampling/Sampling.py
```diff
@@ -158,6 +160,7 @@ def rejection_sample(self, evidence=None, size=1, return_type="dataframe"):
 
             i += len(_sampled)
 
+        sampled = _map_to_state_name(self.model, sampled)
```
ankurankan, line 163: I would prefer `_map_to_state_name` to be from `return_samples`. In my opinion it belongs there, since `_return_samples` is supposed to postprocess the samples. Also, with that we will not need to call this method everywhere.

ankurankan (changes requested):

pgmpy/sampling/Sampling.py
khalibartan: @ankurankan Okay. But I think users will assume that it will return state name because he would want that. But no harm in writing.

pgmpy/sampling/Sampling.py
khalibartan: @ankurankan No. return_samples is also being used by hmc and nuts. So it will make code messy IMO.

pgmpy/sampling/base.py
khalibartan: @ankurankan It is an internal method and is not supposed to be used by the user. That's why I didn't add it in the first place

pgmpy/sampling/base.py
khalibartan: @ankurankan Are you suggesting that I should check if all have a state_name or not ? If certain cpd have then should I convert them to state_name and let other be numerics or something else.

pgmpy/sampling/base.py
khalibartan: @ankurankan I tried using np.fromiter but it gave the error regarding variable size while assigning to the array. Lemme check for the second suggestion though.

pgmpy/sampling/Sampling.py
ankurankan: @khalibartan Yeah, true. But it's always better to be explicit.

pgmpy/sampling/Sampling.py
ankurankan: @khalibartan Yeah, but `return_samples` is supposed to be like a post processor, so any cases regarding that should be handled inside it. And if implementing that messes up the code too much, we should have a different solution to that, maybe implementing 2 separate post processors for different cases. But calling 2 post processing methods doesn't make much sense.

pgmpy/sampling/base.py
ankurankan: But it would be good to have an example for our own reference later.

pgmpy/sampling/base.py
ankurankan: @khalibartan Yes, and if all of them doesn't name state names defined, maybe show a warning and fall back to just returning numerical values. Having a mixture of state names and numeric states is a good idea but only if it's not computationally too expensive. We wouldn't want to reduce the performance of the general case for dealing with a specific case.

pgmpy/sampling/Sampling.py
khalibartan: @ankurankan I'm in favour of keeping these two separately. I don't see any problem in this and it keeps the code clean.

khalibartan: @ankurankan What should be done here?  Both functions have different work.Two different post processors for the different case will lead to redundancy. Merging them into one can be done if we want to.

ankurankan: @khalibartan I do think that we should merge these two.

PR #958, opened 2017-12-17
Adds Markov Blanket to Bayesian Models

harishkashyap:
#### Changes
* Added Markov Blanket to Bayesian Models

### DIFF
--- a/pgmpy/base/DirectedGraph.py
+++ b/pgmpy/base/DirectedGraph.py
@@ -315,3 +315,25 @@ class DirectedGraph(nx.DiGraph):
         ['A', 'E']
         """
         return [node for node, in_degree in self.in_degree().items() if in_degree == 0]
+    
+    def get_children(self, node):
+        """
+        Returns a list of children of node.
+
+        Throws an error if the node is not present in the graph.
+
+        Parameters
+        ----------
+        node: string, int or any hashable python object.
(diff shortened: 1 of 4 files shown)


harishkashyap:
@ankurankan , I thought of adding get_markov_model but I saw that markov_model has the blanket implemented otherwise (http://pgmpy.org/_modules/pgmpy/models/MarkovModel.html) and hence had it for uniformity. Anycase, have made that change. The tests are also updated.

ankurankan:
@harishkashyap We would like to have only 2 commits for this. One with the children method and it's tests and the second for the markov blanket and it's tests. We want to keep each commit for a specific feature so that it's easier to rollback in future if required. Also, we don't want any merge commits. So, I would suggest you to start with a new branch off the pgmpy:dev, and add your changes and do a couple of commits on that and create a new PR from the new branch.

harishkashyap:
Makes sense. Would rebasing —fondly with two commits make sense? First one being the children method and second commit being the blanket? That way we can only a single PR and you can always roll back to a feature?

harishkashyap:
Sorry autocorrect. I meant —ff-only and I can put up a single PR with two commits in order by rebasing

ankurankan:
@harishkashyap Thanks. Closing this since duplicate of #959

Issue #961, opened 2017-12-20
Variable naming inconsistency?


theUltraMarine19:
### Subject of the issue
In line 54, 55 of pgmpy/pgmpy/models/LinearGaussianBayesianNetwork.py, cpd.variables has been checked for, whereas, line 66 of  pgmpy/pgmpy/factors/continuous/LinearGaussianCPD.py has self.variable specified and not variables. Hence, I think that this error is being raised. Highly obliged if you look into it.

### Your environment
* pgmpy version : 0.1.6
* Python version : 3.4
* Operating System : Ubuntu 14.04

### Steps to reproduce
from pgmpy.models import *
model = LinearGaussianBayesianNetwork([('hist_dist', 'dist_relevance'),('curr_dist', 'dist_relevance'),('monthly_exp','offer_interest'),('cat_merchant_affinity','offer_interest'),
					  ('offer_interest','offer_hit_effect'),('dist_relevance','offer_hit_effect')])

model.add_nodes_from(['hist_dist', 'curr_dist', 'dist_relevance','monthly_exp','cat_merchant_affinity','offer_interest','offer_hit_effect'])

from pgmpy.factors.continuous import LinearGaussianCPD

hist_dist_cpd = LinearGaussianCPD('hist_dist', [200], 20)
curr_dist_cpd = LinearGaussianCPD('curr_dist', [400], 50)
dist_rel_cpd  = LinearGaussianCPD('dist_relevance', [0, 0.5, 0.5], 0.1, ['hist_dist', 'curr_dist'])

monthly_exp_cpd = LinearGaussianCPD('monthly_exp', [45000], 5000)
cat_merchant_cpd = LinearGaussianCPD('cat_merchant_affinity', [0.6], 0.2)
offer_interest_cpd = LinearGaussianCPD('offer_interest', [0, 0.7, 0.3], 0.1, ['monthly_exp', 'cat_merchant_affinity'])

offer_hit_effect_cpd = LinearGaussianCPD('offer_hit_effect', [0.1, 0.3, 0.6], 0.2, ['dist_relevance', 'offer_interest'])
model.add_cpds(hist_dist_cpd, curr_dist_cpd, dist_rel_cpd, monthly_exp_cpd, cat_merchant_cpd, offer_interest_cpd, offer_hit_effect_cpd)

### Expected behaviour
No error

### Actual behaviour
File "linearGaussian.py", line 25, in <module>
[...]

PR #962, opened 2017-12-20
Fixed class variable for LinearGaussianCPDs

theUltraMarine19:
### Fixes # 
Fixes Issue #961 
#### Changes
- Replaced variables with self.variables in __init__ method of LinearGaussianCPD.py 

Thank you!

### DIFF
--- a/pgmpy/factors/continuous/LinearGaussianCPD.py
+++ b/pgmpy/factors/continuous/LinearGaussianCPD.py
@@ -75,8 +75,8 @@ class LinearGaussianCPD(BaseFactor):
         self.evidence = evidence
         self.beta_vector = np.asarray(beta[1:])
 
-        variables = [variable] + evidence
-        super(LinearGaussianCPD, self).__init__(variables, pdf='gaussian',
+        self.variables = [variable] + evidence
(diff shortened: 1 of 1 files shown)


ankurankan:
@Arijit-hydrated Thanks for the PR. Could you please also add a test case which would fail on the current version ? And after adding that, please squash both the commits into a single commit using `git rebase`. And please do try to follow: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html for commit messages.

theUltraMarine19:
@ankurankan Do you want me to add the test case which fails on the current version in the Fixes section above, where I referenced the relevant issue?

Thanks :)

ankurankan:
@Arijit-hydrated No, we have unittests for the whole codebase (look into pgmpy/tests dir). So, I would like to have a tests added there for this specific issue. Let me know if something's not clear.

theUltraMarine19:
@ankurankan This is the pull request for the issue #961 which is on high priority now. I noticed this was never merged. Can you elaborate what you want added in the unittests directory?

ankurankan:
@theUltraMarine19 Just a simple unit test which fails with the current code and gets fixed after merging the PR.

PR #967, opened 2018-01-07
Fix for #897 

### LINKED ISSUE #897: BayesianModel __eq__ method
I noticed other classes such as DiscreteFactor have equality but BayesianModel does not. Perhaps something as simple as the following will do?
```
def __eq__(self, other):
    if isinstance(other, self.__class__):
        return self.__dict__ == other.__dict__
    else:
        return False
```

Vaibhavdixit02:

- [ ] Check your code additions will fail neither code linting checks nor unit test.

#### Changes
- Added __eq__ method to BayesianModel class
- Added test for the same

💔Thank you!

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -1030,3 +1030,25 @@ class BayesianModel(DirectedGraph):
         blanket_nodes = set(blanket_nodes)
         blanket_nodes.remove(node)
         return list(blanket_nodes)
+    
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            node_cpd_dict_self = {}
+            for i in self.cpds:
+                node_cpd_dict_self[i.variable] = i
+            for i in self.nodes():
+                if i not in node_cpd_dict_self:
+                    node_cpd_dict_self[i] = ''
+            node_cpd_dict_other = {}
+            for i in other.cpds:
+                node_cpd_dict_other[i.variable] = i
+            for i in other.nodes():
+                if i not in node_cpd_dict_other:
+                    node_cpd_dict_other[i] = ''
+            if self.edges() == other.edges() and node_cpd_dict_self == node_cpd_dict_other:
+                return True
+            else:
+                return False
+        else:
+            return False
+            
--- a/pgmpy/tests/test_models/test_BayesianModel.py
+++ b/pgmpy/tests/test_models/test_BayesianModel.py
@@ -226,6 +226,22 @@ class TestBayesianModelMethods(unittest.TestCase):
         self.assertRaises(ValueError, self.G1.get_cpds, 'diff')
         self.assertRaises(ValueError, self.G1.get_cpds, 'grade')
 
+    def test__eq__(self):
+        other_equal = self.G.copy()
+        cpd_a = TabularCPD('a', 2, [[0.2], [0.8]])
+        cpd_b = TabularCPD('b', 2, [[0.3, 0.7], [0.7, 0.3]],
+                       evidence=['a'],
+                       evidence_card=[2])
+        cpd_c = TabularCPD('c', 2, [[0.3, 0.7], [0.7, 0.3]],
+                       evidence=['b'],
+                       evidence_card=[2])
+        self.G.add_cpds(cpd_c, cpd_a, cpd_b)
+        other_equal.add_cpds(cpd_b, cpd_a, cpd_c)
+        other_unequal.add_cpds(cpd_c, cpd_a)
+        assertTrue(self.G == other_equal)
+        assertFalse(self.G == other_unequal)
+        assertFalse(self.G == self.G1)
+
     def tearDown(self):
         del self.G
         del self.G1


Vaibhavdixit02:
@ankurankan

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -230,6 +230,9 @@ def tearDown(self):
         del self.G
         del self.G1
 
+    def test__eq__(self):
+        model_copy = self.G1.copy()
+        self.assertTrue(G1.__eq__(model_copy))
```
khalibartan: @Vaibhavdixit02 Test are failing for this line `self.assertTrue(G1.__eq__(model_copy))`, please fix it to self.assertTrue(self.G1.__eq__(model_copy))`also add test cases when they are not equal.

khalibartan (changes requested):

pgmpy/tests/test_models/test_BayesianModel.py
Vaibhavdixit02: Fixed and added.

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -230,6 +230,10 @@ def tearDown(self):
         del self.G
         del self.G1
 
+    def test__eq__(self):
```
lohani2280: @Vaibhavdixit02  Tests are failing because you have defined `test__eq__` method after `tearDown` method.Thus, `G1` is destroyed

Vaibhavdixit02: I didn't pay attention, sorry! Fixed now.

lohani2280:
@Vaibhavdixit02  I would suggest to please amend a commit using `git commit --amend` rather than adding a new commit everytime for updating your PR. Also, i think adding `__eq__` method for `BayesianModel` and its tests should go only in one commit. Right now you have 4 commit which i don't think is necessary at all.

Vaibhavdixit02:
@lohani2280 I wasn't familiar with `git commit --amend` thank you for making me aware of it, I will definitely keep it in my mind!

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -226,6 +226,11 @@ def test_remove_nodes_from(self):
         self.assertRaises(ValueError, self.G1.get_cpds, 'diff')
         self.assertRaises(ValueError, self.G1.get_cpds, 'grade')
 
+    def test__eq__(self):
+        model_copy = self.G1.copy()
+        self.assertTrue(self.G1.__eq__(model_copy))
+        self.assertFalse(self.G1.__eq__(self.G))
```
lohani2280: @Vaibhavdixit02 To use the operator `==`  with custom objects we are defining the method `__eq__`. So why to write tests using  `self.assertFalse(self.G1.__eq__(self.G))`?. IMO a better way would be to use       
`self.assertFalse(self.G1 == self.G)` .

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -226,6 +226,11 @@ def test_remove_nodes_from(self):
         self.assertRaises(ValueError, self.G1.get_cpds, 'diff')
         self.assertRaises(ValueError, self.G1.get_cpds, 'grade')
 
+    def test__eq__(self):
+        model_copy = self.G1.copy()
+        self.assertTrue(self.G1.__eq__(model_copy))
```
lohani2280: @Vaibhavdixit02 Tests are failing here. `self.G1.__eq__(model_copy)` is returning false.

Vaibhavdixit02:
From my local testing I found that `__dict__` of 2 `BayesianModels` are not equal and that is what was causing the tests to fail, also `copy` also creates a different version of the dictionary. Is this behavior in line with the idea behind implementation of the `copy` method? Also I couldn't understand why there is difference in some values for the same keys on initializing 2 `BayesianModels` in exactly the same way.

lohani2280:
Yes __dict__ of `G1` and `model_copy = G1.copy()` is not same and that's why implementing __eq__  in the following way 
```
def __eq__(self, other):
    if isinstance(other, self.__class__):
        return self.__dict__ ==  other.__dict__
    else:
        return False
```
is returning false when `self.G1.__eq__(model_copy)`  is called.
Let's wait for @ankurankan @khalibartan review.

pgmpy/models/BayesianModel.py
```diff
@@ -1030,3 +1030,19 @@ def get_markov_blanket(self, node):
         blanket_nodes = set(blanket_nodes)
         blanket_nodes.remove(node)
         return list(blanket_nodes)
+    
+    def __eq__(self, other):
+        ignore_keys = ['root_graph','edges','out_edges']
+        d1_filtered = dict((k, v) for k,v in list(self.__dict__.items()) if k not in ignore_keys)
+        d2_filtered = dict((k, v) for k,v in list.(other.__dict__.items()) if k not in ignore_keys)
```
lohani2280: typo error

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -226,6 +226,14 @@ def test_remove_nodes_from(self):
         self.assertRaises(ValueError, self.G1.get_cpds, 'diff')
         self.assertRaises(ValueError, self.G1.get_cpds, 'grade')
 
+    def test__eq__(self):
+        G4 = BayesianModel([('a', 'd'), ('b', 'd'),
+                                ('d', 'e'), ('b', 'c')])
+        G5 = BayesianModel([('a', 'd'), ('b', 'd'),
+                                ('d', 'e'), ('b', 'c')])
+        self.assertTrue(G4 == G5)
```
lohani2280: Please add cpd to the `G4` and `G5` network and then test using `self.assertTrue(G4 == G5)`

Vaibhavdixit02: Could you please elaborate the reason for this?

lohani2280: @Vaibhavdixit02  Currently `G4` and `G5` are just directed graphs and not PGM because they do not have factors associated with it. The main aim of PGM is that it allows us to represent joint probability distribution in an efficient way. A Bayesian Network consist of a directed graph and a Condition Probability Distribution associated with each random variable.Currently, `G4` and `G5` are just directed graphs and their is no CPD association. Using `self.assertTrue(G4 == G5)`, we want to check whether `G4` Bayesian Model and `G5` Bayesian Model are equal or not.

Vaibhavdixit02: That does seem correct but I think that `G4` and `G5` are just used as instantiations of `BayesianModel` objects here and their being PGM isn't really necessary? I mean the test case is just for comparing 2 `BayesianModel` object's equality? If it is necessary I will surely add your suggestion but I just want to be clear regarding its requirement. :)

Vaibhavdixit02:
Pinging @khalibartan @ankurankan

ankurankan:
@Vaibhavdixit02 Sorry for the late reply. We don't really need to compare everything in the `__dict__` of the objects. It's unnecessary and will add computational overhead. For two Bayesian Networks to be equal we want it's network structure to be equal and to have the same CPDs. So, we just need to make sure about these.

But do keep in mind having same CPDs doesn't mean having the same object instances, so two CPDs are equal even if they are different instances but have same values. 

Also @lohani2280 is correct about test using networks having CPDs because we want the tests to cover every possible case.

Hope this helps and let me know if you have any doubts.

Vaibhavdixit02:
@ankurankan @lohani2280 made the required changes.

pgmpy/models/BayesianModel.py
```diff
@@ -1030,3 +1030,16 @@ def get_markov_blanket(self, node):
         blanket_nodes = set(blanket_nodes)
         blanket_nodes.remove(node)
         return list(blanket_nodes)
+    
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            network_struct1 = (self.nodes(),self.edges())
```
ankurankan: @Vaibhavdixit02 Please follow PEP8 (https://www.python.org/dev/peps/pep-0008/). Needs a space after the comma.

pgmpy/models/BayesianModel.py
```diff
@@ -1030,3 +1030,16 @@ def get_markov_blanket(self, node):
...
+            network_struct2 = (other.nodes(),other.edges())
+            cpd1 = self.cpds
+            cpd2 = other.cpds
+            if network_struct1 == network_struct2 and cpd1 == cpd2:
```
ankurankan: As far as I remember `self.cpds` would be a list. So `cpd1 == cpd2` would fail if the same CPDs are added to the network in a different order.

pgmpy/models/BayesianModel.py
```diff
@@ -1030,3 +1030,16 @@ def get_markov_blanket(self, node):
         blanket_nodes = set(blanket_nodes)
         blanket_nodes.remove(node)
         return list(blanket_nodes)
+    
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            network_struct1 = (self.nodes(),self.edges())
+            network_struct2 = (other.nodes(),other.edges())
+            cpd1 = self.cpds
+            cpd2 = other.cpds
```
ankurankan: Never use names with `cpd1` and `cpd2`. Come up with better variable names which explains what they are.

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -629,4 +644,4 @@ def test_get_values_raises_error(self):
         self.assertRaises(ValueError, self.graph.get_cpds, 'sat')
 
     def tearDown(self):
-        del self.graph
+        del self.graph
```
ankurankan: Add a new line at the end of the file.

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -226,6 +226,21 @@ def test_remove_nodes_from(self):
...
+        cpd_b = TabularCPD('b', 2, [[0.3, 0.7], [0.7, 0.3]],
+                       evidence=['a'],
+                       evidence_card=[2])
+        cpd_c = TabularCPD('c', 2, [[0.3, 0.7], [0.7, 0.3]],
+                       evidence=['b'],
+                       evidence_card=[2])
+        G4.add_cpds(cpd_a, cpd_b, cpd_c)
+        G5 = G4.copy()
+        self.assertTrue(G4 == G5)
+        self.assertFalse(self.G1 == self.G)
```
ankurankan: The tests need to cover more possible situations. For example different order of CPDs in the network, check on the networks with same structure but different CPDs etc.

ankurankan (changes requested):
I have put some comments in the code. Please have a look. Also, after you are done making those changes, I would like all the commits for this PR to be squashed in a single commit. Have a look at `git rebase`.

ankurankan (changes requested):
Also, I think the implementation could be a bit better. For example, I don't think we need to introduce new variables like `cpd1` and `cpd2`. They could have been easily compared directly in the `if` statement with something like `self.cpds == other.cpds`.

pgmpy/models/BayesianModel.py
Vaibhavdixit02: @ankurankan I am stuck at this part, do you have some suggestion on how to resolve this part? I tried to do something along the lines of converting the cpd objects to string (I understand that this doesn't make much sense) and then sorting it, the other alternative I came up with is checking every `cpd` of `self` to see if it is in the `other.cpds` list but this would be very in-efficient. I will appreciate some help in this part.

Vaibhavdixit02:
@ankurankan I have done all the changes, the cpds' dictionary comparison is a quite informal method at the moment, maybe you could recommend some more standard way to handle this problem?

pgmpy/models/BayesianModel.py
```diff
@@ -1030,3 +1030,16 @@ def get_markov_blanket(self, node):
         blanket_nodes = set(blanket_nodes)
         blanket_nodes.remove(node)
         return list(blanket_nodes)
+    
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
+            network_struct1 = (self.nodes(), self.edges())
+            network_struct2 = (other.nodes(), other.edges())
```
ankurankan: @Vaibhavdixit02 I don't think this would work. Since networkx implements the `node` and `edge` properties as dict, you can't be sure of the order in which it's keys (which is what `nodes()` and `edges()` method returns) are returned. Better to compare the properties directly. Also, there's no need to create new tuples here. You can just compare the things in the `if` statement below. Something like:
```python
if self.node == other.node and self.edge == other.edge and other_conditions
```

pgmpy/models/BayesianModel.py
```diff
@@ -1030,3 +1030,16 @@ def get_markov_blanket(self, node):
...
+            cpd_self = [str(i) for i in self.cpds]
+            cpd_other = [str(i) for i in other.cpds]
```
ankurankan: I don't understand what you are trying to do here. Could you please explain a bit ?

ankurankan: A good way of comparing the CPDs could be to create a dict of the form `{node: node_cpd}` for both self and other and then compare these two. Since CPD already has an implementation of `__eq__`, the comparison should work fine.

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -226,6 +226,31 @@ def test_remove_nodes_from(self):
...
+                                ('d', 'e'), ('b', 'c')])
+        cpd_a = TabularCPD('a', 2, [[0.2], [0.8]])
+        cpd_b = TabularCPD('b', 2, [[0.3, 0.7], [0.7, 0.3]],
+                       evidence=['a'],
+                       evidence_card=[2])
+        cpd_c = TabularCPD('c', 2, [[0.3, 0.7], [0.7, 0.3]],
+                       evidence=['b'],
+                       evidence_card=[2])
+        G4.add_cpds(cpd_a, cpd_b, cpd_c)
+        G5 = G4.copy()
```
ankurankan: Rather than creating so many new networks you can simply use the networks created in the `setUp` method. For example you can simply use `self.G` and it's copy instead of `G4` and `G5`. Also you can modify those by adding or remove cpds/nodes/edges.  And since unittests runs the `setUp` method before running each test you don't need to worry about modifying the objects in setUp.

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -630,3 +655,4 @@ def test_get_values_raises_error(self):
 
     def tearDown(self):
         del self.graph
+
```
ankurankan: PEP8. Space in an empty line.

ankurankan (changes requested):
I have added some comment for improving the code. Other than that as I had mentioned in the last review could you please squeeze all you commits into a single commit. For a tutorial on how to do that google for `squash commits using git rebase`

PR #1002, opened 2018-03-30
Change in MarkovModel

VSSILPA:

#### Changes
Changed factor_node to factor in line number 295 and added the corresponding test cases.

💔Thank you!

### DIFF
--- a/pgmpy/models/MarkovModel.py
+++ b/pgmpy/models/MarkovModel.py
@@ -291,8 +291,7 @@ class MarkovModel(UndirectedGraph):
         factor_graph.add_nodes_from(self.nodes())
         for factor in self.factors:
             scope = factor.scope()
-            factor_node = 'phi_' + '_'.join(scope)
-            factor_graph.add_edges_from(itertools.product(scope, [factor_node]))
+            factor_graph.add_edges_from(itertools.product(scope, [factor]))
             factor_graph.add_factors(factor)
 
         return factor_graph
--- a/pgmpy/tests/help_functions.py
+++ b/pgmpy/tests/help_functions.py
@@ -5,3 +5,9 @@ def recursive_sorted(li):
     for i in range(len(li)):
         li[i] = sorted(li[i])
     return sorted(li)
+
+
+def recursive_frozenset(edges):
+    for i in range(len(edges)):
+        edges[i] = frozenset(edges[i])
+    return edges
--- a/pgmpy/tests/test_models/test_MarkovModel.py
+++ b/pgmpy/tests/test_models/test_MarkovModel.py
@@ -213,15 +213,12 @@ class TestMarkovModelMethods(unittest.TestCase):
         phi2 = DiscreteFactor(['Bob', 'Charles'], [2, 2], np.random.rand(4))
         self.graph.add_edges_from([('Alice', 'Bob'), ('Bob', 'Charles')])
         self.graph.add_factors(phi1, phi2)
-
         factor_graph = self.graph.to_factor_graph()
         self.assertIsInstance(factor_graph, FactorGraph)
-        self.assertListEqual(sorted(factor_graph.nodes()),
-                             ['Alice', 'Bob', 'Charles', 'phi_Alice_Bob',
-                              'phi_Bob_Charles'])
-        self.assertListEqual(hf.recursive_sorted(factor_graph.edges()),
-                             [['Alice', 'phi_Alice_Bob'], ['Bob', 'phi_Alice_Bob'],
-                              ['Bob', 'phi_Bob_Charles'], ['Charles', 'phi_Bob_Charles']])
+        self.assertSetEqual(set(factor_graph.nodes()), set(['Alice', 'Bob', 'Charles', phi1, phi2]))
+        self.assertSetEqual(set(hf.recursive_frozenset(list(factor_graph.edges()))),
+                            set(hf.recursive_frozenset([('Alice', phi1), ('Bob', phi1),
+                                                        ('Bob', phi2), ('Charles', phi2)])))
         self.assertListEqual(factor_graph.get_factors(), [phi1, phi2])
 
     def test_factor_graph_raises_error(self):


khalibartan:
@ankurankan I'm not in favour of the current factor graph tests. Tests are using both factor or `str phi` based on convenience in different places.

khalibartan:
@VSSILPA Tests will fail because str and factor are different data types and we don't have an ordering operation implemented in factors.

pgmpy/tests/test_models/test_MarkovModel.py
```diff
@@ -216,12 +216,12 @@ def test_factor_graph(self):
 
         factor_graph = self.graph.to_factor_graph()
         self.assertIsInstance(factor_graph, FactorGraph)
-        self.assertListEqual(sorted(factor_graph.nodes()),
-                             ['Alice', 'Bob', 'Charles', 'phi_Alice_Bob',
-                              'phi_Bob_Charles'])
-        self.assertListEqual(hf.recursive_sorted(factor_graph.edges()),
-                             [['Alice', 'phi_Alice_Bob'], ['Bob', 'phi_Alice_Bob'],
-                              ['Bob', 'phi_Bob_Charles'], ['Charles', 'phi_Bob_Charles']])
+        self.assertListEqual(factor_graph.nodes(),
```
khalibartan: @VSSILPA  You can convert both lists to python set. Order of elements might change in list, so test might fail or pass.

VSSILPA:
Yes, the tests are failing. In some places, factor nodes are taken as string while in the other cases, it's of dicrete factor class.

pgmpy/tests/test_models/test_MarkovModel.py
```diff
@@ -216,12 +216,12 @@ def test_factor_graph(self):
...
+                             ['Alice', 'Bob', 'Charles', phi1,
+                              phi2])
+        self.assertListEqual(factor_graph.edges(),
+                             [['Alice', phi1], ['Bob', phi1],
```
khalibartan: @VSSILPA  same goes for this. Convert both of them to set. Also, note that the edge will be a tuple and not list

khalibartan (changes requested):

pgmpy/tests/test_models/test_MarkovModel.py
VSSILPA: Will recursive_sorted() function work in case of elements with different data types? Is it fine to just write factor_graph.edges() instead of recursive_sorted(factor_graph.edges())?

pgmpy/tests/test_models/test_MarkovModel.py
```diff
@@ -216,12 +216,12 @@ def test_factor_graph(self):
...
-                             ['Alice', 'Bob', 'Charles', 'phi_Alice_Bob',
-                              'phi_Bob_Charles'])
-        self.assertListEqual(hf.recursive_sorted(factor_graph.edges()),
-                             [['Alice', 'phi_Alice_Bob'], ['Bob', 'phi_Alice_Bob'],
-                              ['Bob', 'phi_Bob_Charles'], ['Charles', 'phi_Bob_Charles']])
+        self.assertSetEqual(set(factor_graph.nodes()),
+                            set(['Alice', 'Bob', 'Charles', phi1,
+                              phi2]))
+        self.assertSetEqual(set(factor_graph.edges()),
+                            set([('Alice', phi1), ('Bob', phi1),
```
khalibartan: @VSSILPA Correct order would be:
`set([('Bob', phi1), ('Bob', phi2), ('Charles', phi2), (phi1, 'Alice')])`

You can either change it to above or you can convert the tuples to frozenset in both cases. I'd like to go with latter one, add this function in help_function.py in test directory.

khalibartan (changes requested):

pgmpy/tests/test_models/test_MarkovModel.py
VSSILPA: I've converted tuples to frozenset. Is this fine?

VSSILPA: It's working for all test cases now.

pgmpy/tests/test_models/test_MarkovModel.py
```diff
@@ -215,13 +215,16 @@ def test_factor_graph(self):
         self.graph.add_factors(phi1, phi2)
 
         factor_graph = self.graph.to_factor_graph()
+
```
khalibartan: @VSSILPA Remove these empty lines. Also, can you check for pep8 errors in your code? You can use
`pycodestyle --max-line-length=120 <filename>` for this check.

pgmpy/tests/help_functions.py
```diff
@@ -5,3 +5,8 @@ def recursive_sorted(li):
     for i in range(len(li)):
         li[i] = sorted(li[i])
     return sorted(li)
+
+def convert_to_frozenset(edges) :
```
khalibartan: @VSSILPA I think we can have a better name for this function. Also IMO it should return a frozenset of frozenset.
So function name could be recursive_frozenset, if you can think of some name, please use that.

khalibartan (changes requested):
Please squash your commits.

pgmpy/models/MarkovModel.py
```diff
@@ -291,8 +291,8 @@ def to_factor_graph(self):
         factor_graph.add_nodes_from(self.nodes())
         for factor in self.factors:
             scope = factor.scope()
-            factor_node = 'phi_' + '_'.join(scope)
-            factor_graph.add_edges_from(itertools.product(scope, [factor_node]))
+#            factor_node = 'phi_' + '_'.join(scope)
```
khalibartan: @VSSILPA Remove this commented line.

khalibartan (changes requested):

ankurankan:
@VSSILPA I am not sure about this change. Using a factor object as a node might break the consistency in API for inference or learning. And that will make us add extra conditions for handling it all the higher level APIs.

khalibartan:
@ankurankan I think we should discuss upon the design of FactorGraph class. Currently, in FactorGraph class factor nodes are actual factors and that too isn't being tested. There is high inconsistency in implementation. We can discuss the design and then finalize how we want to proceed. i do too agree with the fact that adding factor as node will cause difficulty.

PR #1019, opened 2018-07-18
variable name clash in states array creation

fabriziov:
In _get_kernel_from_markov_model method of the class GibbsSampling (Sampling module), the variable name "var" used to create the array "states" clashes with the variable name "var" used in the outer for loop.

#### Changes
- Changed variable name "var" into "first_var" for the variable used to create the "states" array in _get_kernel_from_markov_model of the class GibbsSampling (Sampling module).

Thank you!

### DIFF
--- a/pgmpy/sampling/Sampling.py
+++ b/pgmpy/sampling/Sampling.py
@@ -331,7 +331,7 @@ class GibbsSampling(MarkovChain):
             factor = factors_dict[var]
             scope = set(factor.scope())
             for tup in itertools.product(*[range(card) for card in other_cards]):
-                states = [State(var, s) for var, s in zip(other_vars, tup) if var in scope]
+                states = [State(first_var, s) for first_var, s in zip(other_vars, tup) if first_var in scope]
                 reduced_factor = factor.reduce(states, inplace=False)
                 kernel[tup] = reduced_factor.values / sum(reduced_factor.values)
             self.transition_models[var] = kernel


pgmpy/sampling/Sampling.py
```diff
@@ -331,7 +331,7 @@ def _get_kernel_from_markov_model(self, model):
             factor = factors_dict[var]
             scope = set(factor.scope())
             for tup in itertools.product(*[range(card) for card in other_cards]):
-                states = [State(var, s) for var, s in zip(other_vars, tup) if var in scope]
+                states = [State(first_var, s) for first_var, s in zip(other_vars, tup) if first_var in scope]
```
khalibartan, line 334: @fabriziov Can you change this name to `v`? `first_var` doesn't seem a good name.

khalibartan (changes requested):

khalibartan:
@fabriziov Yes indeed, in Python3 this would have worked fine because of loop variable scope but in Python2.x this would be wrong.

@ankurankan This should have been caught by our unittest. We need better unittests.

@fabriziov If you don't mind can you add new tests and remove the old one which uses mock. We don't intend to use mock for testing anymore. This would be a big task and might take some time. New tests should use sample results to do simple inference task and see that values are close to results obtained by exact inference (closeness can be defined using a small threshold). You can see tests for NUTS and HMC for getting the idea.

PR #1050, opened 2019-01-04
Fix broken probability estimation for pandas >= 0.21.0 (issue #1035)

### LINKED ISSUE #1035: Probability estimation broken for pandas > 0.19.2
The problem had already been reported here, but not solved unfortunately: #1005 

With `pgmpy==0.1.6` the ML estimation of a BN works if `pandas==0.19.2` is installed, but not if the latest `pandas==0.23.4` is present.

Example code:
```python3
import pandas as pd
import numpy as np

from pgmpy.estimators import MaximumLikelihoodEstimator, BayesianEstimator

df = pd.DataFrame(np.array([
    ["A", "U"],
    ["A", "V"],
    ["B", "U"],
    ["B", "U"],
    ["A", "U"],
    ["B", "V"],
    ["A", "V"],
    ["A", "V"]
]), columns = ["Dim1", "Dim2"])

model = BayesianModel([("Dim1", "Dim2")])
model.fit(df, estimator = MaximumLikelihoodEstimator)

for cpd in model.get_cpds():
    print(cpd)
```

Output `pandas==0.19.2`:
image

Output `pandas==0.23.4`:
image

marcotama:
#### Changes
- The added code checks the type of index of the DataFrame `state_count_data` and, if that is not a MultiIndex, it makes its a MultiIndex so that it plays nice with the call to `reindex()` a few lines below.

Thank you!

### DIFF
--- a/pgmpy/estimators/base.py
+++ b/pgmpy/estimators/base.py
@@ -98,6 +98,11 @@ class BaseEstimator(object):
         C
         c1  1   1   0   0
         c2  0   0   1   0
+        >>> estimator.state_counts('C', parents=['A'])
+        A    a1   a2
+        C
+        c1  2.0  0.0
+        c2  0.0  1.0
         """
 
         # default for how to deal with missing data can be set in class constructor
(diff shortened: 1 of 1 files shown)


ankurankan:
@marcotama Thanks for the PR. The changes look good for a temporary fix but this MultiIndex thing needs to be fixed at other places as well: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/base.py#L212. If you would like to fix that also, that would be great. Otherwise I will make those changes and merge this. Also, it would be great to add tests which fail right now but is fixed by this PR.

marcotama:
Hi @ankurankan , I used the same fix for the method you specified. I also added a test for the first function, while there is no need for a test for the second function since the existing test should raise an exception. Please, confirm that this is the case.

ankurankan:
@marcotama Thanks a lot for the fix. This was a long-standing bug.

marcotama:
Happy to help! When are you planning to release the version with the fix?

ankurankan:
@marcotama Just did a release.

PR #1059, opened 2019-02-04
BIFReader fix on reordering of states

### LINKED ISSUE #1053: Bug in BIFReader function
### Subject of the issue
There is a bug in the BIFReader function. I think while reading the CPD only the order  in which the states are specified is taken into account , not the actual values of the state.

e.g. variable Pollution {low high} **high - 1 low - 0**
variable Smoker {True False} **True - 0, False-1**

probability ( Cancer | Pollution, Smoker ) {
  (low, True) 0.03, 0.97; 
  (high, True) 0.05, 0.95; **should be 1,0 instead of 0,1**
  (low, False) 0.001, 0.999;**should be 0,1 instead of 1,0**
  (high, False) 0.02, 0.98;
}

probability ( Cancer | Pollution, Smoker ) {
  (low, True) 0.03, 0.97; 
 (low, False) 0.001, 0.999;
  (high, True) 0.05, 0.95;
  (high, False) 0.02, 0.98;
}
The two CPDs are equivalent but read differently by the function.
### Your environment
* pgmpy version 0.1.7
* Python version 3.5
* Operating System Unbuntu16.04

### Steps to reproduce
Attached are two benchmark files cancer1.bif and cancer2.bif with the same data but the order of specifying the states of CPD is changed for one of the tables. Inference on both models gives different results even though the model is the same. 
Sample run file is also attached.

### Expected behaviour
The result of the query should be same in both cases since models are equivalent.

### Actual behaviour
Results with model file1:
```
+----------+---------------+
| Cancer   |   phi(Cancer) |
+==========+===============+
| Cancer_0 |        0.0116 |
+----------+---------------+
| Cancer_1 |        0.9884 |
+----------+---------------+
Results with model file2:
+----------+---------------+
| Cancer   |   phi(Cancer) |
+==========+===============+
| Cancer_0 |        0.0410 |
+----------+---------------+
| Cancer_1 |        0.9590 |
+----------+---------------+
```
[test.zip](https://github.com/pgmpy/pgmpy/files/2775737/test.zip)

khalibartan:
Fixes #1053

### DIFF
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -95,7 +95,7 @@ class BIFReader(object):
         # 1.00 or 1 or 1.00. 0.00 or 9.8e-5 etc
         num_expr = Word(nums + '-' + '+' + 'e' + 'E' + '.') + Suppress(Optional(","))
         probability_expr = Suppress('probability') + Suppress('(') + OneOrMore(word_expr) + Suppress(')')
-        optional_expr = Suppress('(') + Suppress(OneOrMore(word_expr2)) + Suppress(')')
+        optional_expr = Suppress('(') + OneOrMore(word_expr2) + Suppress(')')
         probab_attributes = optional_expr | Suppress('table')
         cpd_expr = probab_attributes + OneOrMore(num_expr)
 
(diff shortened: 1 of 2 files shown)


ankurankan (approved):

PR #1085, opened 2019-04-03
[WIP] Adds support for Causal Inference

mrklees:

#### Changes
- Adds CausalInference class with initial support for identifying deconfounding sets per the backdoor criterion.  Definitely need more tests.
- I added a Jupyter Notebook with some examples of using the CausalInference class and finding backdoor paths. 

As a note, I saw the piece in the contributing guide about translating TODO comments to issues.  I will definitely do this once the code is much closer to complete. 

Thank you!


mrklees:
Hi @ankurankan, so I'm thinking about adding a method to automatically create and run an estimator, like say a simple linear model, given the estimand that the class is producing. Do you have any preference regarding using the model in statsmodels or scikit-learn?  I notice that so far you all have been implementing estimators yourselves, but I would rather not in this case.

ankurankan:
@mrklees Yes, ideally we should provide an estimation method and I would prefer statsmodels over sklearn because of the more informative fit summary. I think we should have a method which should accept the estimator type (OLS, GLS etc) as a parameter and `kwargs` to accept the parameters which can be passed to these estimators. What do you think?

But I am not sure if we will also need to provide methods for non-linear estimation. I don't know much about non-linear cases (and if they are even used) and will have to refer to literature on it. But linear estimation is a good start and can figure out non-linear cases later.

mrklees:
Sounds good @ankurankan.  I have been playing with [a more practical example in this Colab Notebook](https://colab.research.google.com/drive/14cB6EIaGBJUvDWHS2UH-lupa0BPeDHoB), and created a simple implementation based on bootstrapping the LinearRegression model to get a confidence interval on the ATE. It is a lot of extra code though just for something that statsmodels gives you for free.

The other thought I had, after taking a look at your SEM code, was that I could write it in pytorch.  That would help limit the dependencies, and give us great performance on the bootstrapping that you might do to get a confidence interval.  If I were to go this route though, it would be helpful if you could push your pytorch helpfer functions to dev.  I could write my own, but I do love to steal cleanly written code :)

ankurankan:
@mrklees First of all, your notebook looks great. :)
I can definitely push the pytorch function to dev but I am not sure where are you using it. I thought you were planning to use statsmodels to compute the estimates.
I don't really mind an extra dependency (unless it doesn't help in reducing much code) because that would keep our codebase small, so easier to maintain. Also, we already have some features planned (like some independence tests) which would depend on statsmodels.

mrklees:
Ok, I went ahead and wrote the estimate_ate method with statsmodels. So far that seems to be working pretty well for backdoor adjustment sets. 

I should also comment on the structure.  So I ended up shifting the structure a bit after [having this short exchange](https://github.com/ijmbarr/causalgraphicalmodels/issues/1). Essentially, most of the code I have written doesn't actually occupy the inference step.  Instead, it really seems to more naturally exist as an extension of the DAG class for identifying the models independence assumptions and structure, and then being able to identify an estimand from there.  So I created a class to represent this step, called the CausalGraph class.

The inference step, where we estimate the average treatment effect, is now the main thing that lives in CausalInerence.

ankurankan:
@mrklees I don't exactly agree with the current structure. The way I view it is that Bayesian Networks has two interpretations, firstly as a representation for joint distribution (what is implemented right now) and secondly as a causal model. So, I don't think we should have two separate models for this and 
 separating these might be confusing for users. My idea for keeping all the causal methods in the `CausalInference` class was to keep things separate between these two interpretations. And properties that were common in both the interpretations would have stayed either in `DAG` or `BayesianModel`. Also, we would need these same methods to work in the case of SEMs as well. So, we could have a common `CausalInference` class for both these models.

But I do agree that some of these methods do go more naturally with the DAG class. So, I think there are two ways to do this: either add the methods to the DAG class (since both `SEM` and `BayesianModel` inherit them) or implement everything in the `CausalInference` class. I am not sure which of these is a better solution. What do you think?

mrklees:
hmmm, yeah this is a bit messy.  I certainly think the do operator makes a lot of sense in the DAG class. The ones that really give me pause are the identification methods.  I don't like them in an inference class, like I said, because they don't need any data so kind of defy the definition of inference.  But I also don't really like putting them in DAG, because they're only methods that make sense if you're thinking about a DAG as representing causal assumptions. But finally, I don't really like the idea of directly coding them into BayesianModel and SEM, because it's just duplicative. 

Maybe there's a middle point here, because I also hear your argument about ease-of-use for users, and I don't want to make the API needlessly complicated just because I'm a little pedantic. One thought is that we could leave the code for identification in its current home (CausalGraph) or perhaps somewhere that's less visible to users like a CausalMethods file, and then specifically import and wrap it within the BayesianModel and SEM classes. This way we don't have to write and maintain it twice for the two models, and it's easy for the user to call these methods from whichever model they're using to represent their assumptions. What do you think?

ankurankan:
@mrklees Actually I am leaning towards having everything in CausalInference class. The causal  methods only come into play when we want to interpret the model as a causal model but if we implement those methods on a lower lever (say CausalGraph and BayesianModel inherits it), this might be a bit confusing for users who don't really know about causal inference but just want to use the normal Bayesian Network. If we implement it in a `CausalInference` class, this would kind of provide a context switch based on the interpretation of the model. So, if they want to use the causal interpretation, the model can just be passed to the `CausalInference` class and all the causal related methods become available else for users who only care about working with distributions, they can just use the normal inference without caring about the causal methods.

Also, since these methods are used only when doing causal estimates, putting them in Causal Inference isn't the worst place. I do agree that it's not the ideal situation but in my opinion having it in Causal Inference make more sense. What do you think?

mrklees:
Lol, but why wouldn't everyone want causal methods all the time?!?! Alright, I'll make the updates. 

Besides that, I think the only thing missing is an ATE method for front door adjustment.  I could definitely write more tests, but do you anticipate anything else being required for an initial release?

ankurankan:
@mrklees Haha. Well, will need to do a survey to find an answer to that.

Everything looks good overall except for a few minor things:
1. Ideally I would like to replace `assert` with more informative error messages. And add as much relevant information as possible in the error message so that it's easier for the user to debug.
2. I would suggest putting `_variable_or_iterable_to_set` and `_powerset` methods into `pgmpy/utils`.
3. For the estimate method, I was thinking of taking the kind of regression (OLS, GLS, WLS etc) as a parameter as well.

I still haven't gone through the code in detail. But will try to do that soon and will add inline comments for that. And thanks for this PR :)

pgmpy/base/DAG.py
```diff
@@ -671,3 +671,22 @@ def _get_ancestors_of(self, obs_nodes_list):
...
+        Applies the do operator to the graph and returns a new DAG with the transformed graph.
+
+        Defined by Pearl in Causality on p. 70, the do-operator, do(X = x) has the effect of removing all edges from
+        the parents of X and setting X to the given value x.
+
+        Parameters
+        ----------
+        node : string
+            The name of the node to apply the do-operator to.
+        """
```
ankurankan, line 707: Need to add Return and Examples section.

ankurankan: Can put the `Defined by Pearl in Causality on p.70` in `References` section. Have a look at numpydoc guide: https://numpydoc.readthedocs.io/en/latest/format.html

pgmpy/base/DAG.py
```diff
@@ -671,3 +671,22 @@ def _get_ancestors_of(self, obs_nodes_list):
...
+        assert node in self.nodes()
+        edges = [(a, b) for a, b in self.edges() if b != node]
+        dag_do_x = DAG(edges)
+        # Make sure disconnected nodes aren't lost
+        [dag_do_x.add_node(n) for n in self.nodes() if n not in dag_do_x.nodes()]
```
ankurankan: Not really efficient. Maybe make a copy and remove edges using `predecessors`? Will be more readable as well.

pgmpy/inference/CausalInference.py
```diff
@@ -0,0 +1,291 @@
...
+     * Identifies adjustment variables
+     * Backdoor Adjustment
+     * Front Door Adjustment
+     * Instrumental Variable Adjustment
+
+    Parameters
+    ----------
+    model : CausalGraph
+        The model that we'll perform inference over.
+    set_nodes : list[node:str] or None
```
ankurankan: Please add a blank line above.

pgmpy/inference/CausalInference.py
```diff
@@ -0,0 +1,291 @@
...
+    References
+    ----------
+    'Causality: Models, Reasoning, and Inference' - Judea Pearl (2000)
+
+    Many thanks to @ijmbarr for their implementation of Causal Graphical models available. It served as an invaluable
+    reference. Available on GitHub: https://github.com/ijmbarr/causalgraphicalmodels
+    """
+    def __init__(self, model, latent_vars=None, set_nodes=None):
+        if not isinstance(model, BayesianModel):
+            raise NotImplementedError("Bayesian Parameter Estimation is only implemented for BayesianModel")
```
ankurankan: Wrong error message I guess?

pgmpy/utils/sets.py
```diff
@@ -0,0 +1,38 @@
...
+    ---------
+    x : None, str or Iterable[str]
+    """
+    if x is None:
+        return frozenset([])
+
+    if isinstance(x, str):
+        return frozenset([x])
+
+    if isinstance(x, set):
```
ankurankan: I think this condition can be removed and the `return frozenset(x)` should take care of this case.

pgmpy/utils/sets.py
```diff
@@ -0,0 +1,38 @@
...
+
+def _variable_or_iterable_to_set(x):
+    """
+    Convert variable, set, or iterable x to a frozenset.
+
+    If x is None, returns the empty set.
+
+    Parameters
+    ---------
+    x : None, str or Iterable[str]
```
ankurankan, line 13: Needs a `Returns` section

pgmpy/inference/CausalInference.py
```diff
@@ -0,0 +1,291 @@
...
+        Parameters
+        ----------
+        X: str
+            Intervention Variable
+        Y: str
+            Target Variable
+        Z: str or set[str]
+            Adjustment variables
+        """
+        observed = [X]+list(Z) if Z else [X]
```
ankurankan: Maybe keep the default value of `Z=[]`, then this statement can be simplified.

pgmpy/inference/CausalInference.py
```diff
@@ -0,0 +1,291 @@
...
+        Test whether Z is a valid backdoor adjustment set for estimating the causal impact of X on Y.
+
+        Parameters
+        ----------
+        X: str
+            Intervention Variable
+        Y: str
+            Target Variable
+        Z: str or set[str]
+            Adjustment variables
```
ankurankan, line 77: Returns and Examples section.

pgmpy/inference/CausalInference.py
```diff
@@ -0,0 +1,291 @@
...
+        Z = _variable_or_iterable_to_set(Z)
+
+        # 0. Get all directed paths from X to Y.  Don't check further if there aren't any.
+        directed_paths = list(nx.all_simple_paths(self.dag, X, Y))
+
+        if directed_paths == []:
+            return False
+
+        # 1. Z intercepts all directed paths from X to Y
+        unblocked_directed_paths = [
```
ankurankan, line 197: Would really suggest writing a for loop in these cases. Makes the code much more readable.

ankurankan (changes requested):

mrklees:
Thanks for the feedback! I've gone through and implemented the requested changes.  At this point I would be comfortable merging in this PR. 

There will be a few issues that I will open open as a follow up to this as well including:
  * estimating the average treatment effect using frontdoor adjustment
  * implementing identification using the ID and IDC algorithms
  * integrating these methods with SEM (once that code has been merged in).
  * Robustness checks.  The folks over at [DoWhy](https://github.com/Microsoft/dowhy) have implemented some interesting automated robustness checks to help verify if your graph is consistent with your data.  This is probably a good idea. 

I should be able to work on some of these over the next few months.  

In related news, I'm pretty excited about your SEM PR.  I think we could add some really nice counterfactual support to causal inference once that is in place. I also suspect that the current average treatment effect method should also be reimplemented with your SEM estimator once that is available in Dev. That'll be a little bit longer term of course, but I will definitely work on implementing that stuff as well.

ankurankan (approved):

ankurankan:
This looks good and looking forward to your contributions. Thanks for this PR :)

PR #1119, opened 2019-06-10
Simple changes to handle missing values in BayesianModel predict methods

### LINKED ISSUE #1118: Handle missing data in BayesianModel predict methods
### Subject of the issue
One of the advantages of the Bayesian network model is that it can handle missing data easily, but the predict methods in pgmpy fail entirely when a row in the input data frame contains `None` or `nan`. I have a couple simple changes that support missing data in these predict methods, I'll open a PR with this as the issue. I imagine there are better ways to do this, and I'm happy to take feedback and do this right.

robbymeals:
There may be a better/faster way to do this, but supporting missing variables easily in these predict methods will definitely make my own life easier.

#### Changes
- Checks if input state name is either None or nan before converting to state index, returns original value if it is instead of failing from KeyError
- Drops NA from series in map_query calls over iterrows

Thank you!

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -583,7 +583,7 @@ class BayesianModel(DAG):
         model_inference = VariableElimination(self)
         for index, data_point in data.iterrows():
             states_dict = model_inference.map_query(
-                variables=missing_variables, evidence=data_point.to_dict()
+                variables=missing_variables, evidence=data_point.dropna().to_dict()
             )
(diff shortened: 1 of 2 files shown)


ankurankan:
@robbymeals Thanks for the PR. I don't know much about different ways of dealing with missing values in BN and will have to do some literature reading for that.

For your changes could you add a warning message before dropping any rows in the data so that the user is aware of it? Also, could you please add some tests for the changes?

PR #1143, opened 2019-07-14
changes to handle state_name and state_no error when state_name is not integer #1141

### LINKED ISSUE #1141: state_name and state_no in reduce()
### Subject of the issue
The value argument in reduce() function for TabularCPD and DiscreteFactor. It looks like right now the argument should be (var_name, var_state_name), but previously it's (var_name, var_state_no). However, in pre_compute_reduce() of Sampling.py, this for loop
```python
for state_combination in itertools.product(
            *[range(self.cardinality[var]) for var in variable_evid]
        ):
            states = list(zip(variable_evid, state_combination))
            cached_values[state_combination] = variable_cpd.reduce(
                states, inplace=False
            ).values
```
seems to use (var_name, var_state_no) as argument. 

### Your environment
* newest dev branch
* 3.6.8
* Ubuntu 18.04.2 LTS

### Steps to reproduce
I have a BayesianModel that has variable_state_name as str (not the same with variable_state_no). If I call likelihood_weighted_sample() on the model, which will call pre_compute_reduce(), it gives me key error, because reduce() calls self.get_state_no() like self.get_state_no(var_name, 0)

silverdarknight:
using BIFReader.get_model (asia / ins ...) and inference.forward_sample will cause this error

`for state_combination in itertools.product(
            *[range(self.cardinality[var]) for var in variable_evid]
        ):`
maybe use state_no rather than state_name woule be better?

#### Changes
- fix forward_sample by getting state_name from self.state_names[var][state_no]

### DIFF
--- a/pgmpy/utils/state_name.py
+++ b/pgmpy/utils/state_name.py
@@ -64,7 +64,7 @@ class StateNameMixin:
         Given `var` and `state_name` return the state number.
         """
         if self.state_names:
-            return self.name_to_no[var][state_name]
+            return self.name_to_no[var][self.state_names[var][state_name]]
         else:
             return state_name
 


ankurankan:
@silverdarknight Thanks for the PR. I am sorry, but I am not able to understand why the change is required/fixes the issue. Could you please explain it a bit or better if you could write a test for it?

PR #1149, opened 2019-08-28
Fix Travis CI and Appveyor builds

pchaigno:
Blake requires Python 3.6+, so this commit drops the tests for Python 3.5. I haven't updated the `__version__` variable since the sources themselves are untouched; should I?

- [x] Make sure that the `__version__` variable has been updated.

#### Changes
- Remove tests for Python 3.5 under Travis CI and Appveyor.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@
         virtualenv:
             system_site_packages: false
         env:
-            - DISTRIB="conda" PYTHON_VERSION="3.5" COVERAGE="true"
             - DISTRIB="conda" PYTHON_VERSION="3.6" COVERAGE="true"
             - DISTRIB="conda" PYTHON_VERSION="3.7" COVERAGE="true"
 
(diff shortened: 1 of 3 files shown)


ankurankan:
@pchaigno Sorry for the delay. And thanks for the PR. Everything looks good, but I would suggest updating the `__version__` variable in this case as well (and squash the commits). It helps the users to report the exact version they are using, so if it's not updated future commits will start mismatching with the `__version__` variable.

pchaigno:
@ankurankan Done. Wouldn't it be possible to generate a version number from the commit id during packaging though?

ankurankan:
@pchaigno Thanks. The problem is that since we don't have a scheduled/frequent releases, most of the users install pgmpy directly from the source right now. In this case, the packaging solution wouldn't work but in future we would like to do something like that.

pchaigno:
If they install from source, they likely have a git commit id, no?

ankurankan:
Yeah, but it's easier to just run `pgmpy.__version__`

PR #1203, opened 2019-12-22
Fix "Inference in Bayesian Networks" Jupyter Notebook Example

Erotemic:
I went to go run the "Inference in Bayesian Networks", example and I noticed that (1) it was broken due to bad syntax in the first cell and (2) it was incompatible with the current pgmpy version. 

I simply fixed these problems. The new notebook should now work correctly.


ankurankan:
@Erotemic Thanks a lot for the PR.

PR #1212, opened 2020-01-11
Fix for Tabular CPD copy() function

echoyi:

- [ ] Make sure that the `__version__` variable has been updated.

#### Changes
- In TabularCpd copy(), use state_names.copy() instead of state_names for the copied CPD since state names is a dictionary, so that when modifying the copied CPD by doing new_CPD.state_names.update(new_state_names), it would not affect the original one.

### DIFF
--- a/pgmpy/factors/discrete/CPD.py
+++ b/pgmpy/factors/discrete/CPD.py
@@ -258,7 +258,7 @@ class TabularCPD(DiscreteFactor):
             self.get_values(),
             evidence,
             evidence_card,
-            state_names=self.state_names,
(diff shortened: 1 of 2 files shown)


ankurankan:
@echoyi Thanks for the PR. Could you also possibly add a test case that would fail with the current codebase but get fixed after this patch?

echoyi:
> @echoyi Thanks for the PR. Could you also possibly add a test case that would fail with the current codebase but get fixed after this patch?

working on it

ankurankan (approved):
Looks good. Thanks for the PR :)

PR #1219, opened 2020-01-14
Add show_progress kwarg to other inference methods

Erotemic:
Previously only the `VariableElimination.query` method had the `show_progress` argument, but other query / map_query exact inference methods would show progress and there was no option to disable it. This PR simply adds the `show_progress` kwarg to the appropriate methods.


ankurankan (changes requested):
@Erotemic travis failing because of black formatting

ankurankan:
@Erotemic Thanks for the PR.

PR #1249, opened 2020-04-09
Alarm Bayesian Network

yash-analytics:
- Created an example of Bayesian Belief Network using pgmpy and did some simple queries on the network
- The network represented is originally mentioned in [Bayesian Artificial Intelligence](https://bayesian-intelligence.com/publications/bai/book/BAI_Chapter2.pdf)


ankurankan:
@yashml Thanks for the PR.

PR #1259, opened 2020-04-30
Feature/parallel hill climb search

ssuyash28:

#### Changes
- Used joblib for calculating scores for '+', '-' and 'flip' operations parallelly.
- Added a function '_operation_execution' for enabling parallel processing.
- Added a parameter n_jobs (int)(maximum number of concurrently running jobs) with default value -1.

### DIFF
--- a/pgmpy/estimators/HillClimbSearch.py
+++ b/pgmpy/estimators/HillClimbSearch.py
@@ -1,11 +1,11 @@
 #!/usr/bin/env python
-from itertools import permutations,chain
+from itertools import permutations, chain
 
 import networkx as nx
 
 from pgmpy.estimators import StructureEstimator, K2Score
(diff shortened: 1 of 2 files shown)


ankurankan:
@ssuyash28 Thanks a lot for the PR. I have a couple of comments about this PR:

1. If I am understanding things correctly, making the algorithm parallel should make the algorithm non-deterministic i.e. the results could differ in different runs of the algorithm as the operation being selected could depend on the order of execution. Right? And if that is the case, I think we should add that as a warning in the documentation.
2. Did you by any chance check whether parallelization improves the runtime of the algorithm as parallelization would add extra overhead too? Or for what graph sizes do we start to see any substantial improvement. I think it would be nice to add that to the documentation as well, as a suggestion for when to use parallelization.

PR #1260, opened 2020-05-03
Add optional pseudo_count param for K2 in BaysianEstimator

loudly-soft:
Currently,  BayesianEstimator has a handy built-in K2 parameter that adds a pseudo-count of 1 to cpd.  I thought to expand the option to allow any pseudo-count value.


ankurankan:
@loudly-soft Thanks for the PR. According to my understanding K2 is a special case of Dirichlet with pseudo counts of 1. So, if a user wants to use a different pseudo count, they can specify the `prior_type` as `dirichlet` and specify the `pseudo_counts` parameter. So, I am not sure if this PR adds something else or am I missing something?

loudly-soft:
@ankurankan I suspected this PR may not be pure because of K2’s special meaning.  My model works much better with a fixed pseudo count < 1 so `K2` almost worked for me out of the box.

Setting the `pseudo_counts` parameter is certainly possible but it was a bit verbose so I hacked K2 to save time.

I guess the value it brings is similar to `K2` - time saving and enables quicker trial and error.

I’m okay if the PR is rejected.  I’ll leave it to the expert. :)

Thanks again for the wonderful tool!

ankurankan:
@loudly-soft My main problem here is that it could give an unexpected result for a new user since technically K2 is defined as having a pseudo count of 1.

But I do agree with you that the API should be able to handle the use case that you have described. Maybe we can change `dirichlet` such that if a single value is passed to it, it replicates the behavior that you have implemented right now. What do you think?

loudly-soft:
@ankurankan I take it you mean overloading `pseudo_counts` to allow a single value in addition to the existing 2D array.  If that’s the case, should the `K2` option be removed because it’s being superseded?  One can easily get K2 by setting `prior_type` to `dirichlet` and `pseudo_counts` to 1.

ankurankan:
@loudly-soft I would like to have K2 to just do a pseudo count of 1 as that's how it's defined. With `dirichlet`, we are just giving the user another way to specify pseudo counts, which also keeps it consistent with its definition. 

Yes, I would like the `pseudo_counts` to accept either a 2D array or a single value if the prior_type is dirichlet.

loudly-soft:
@ankurankan I'm baffled why my commits are failing CI, any idea?

loudly-soft:
Drats.  It appears I have polluted this PR with a commit unrelated to this PR.  It's rather late at night right now.  I'll sort his out later.  Apology.

pgmpy/estimators/BayesianEstimator.py
```diff
@@ -150,13 +153,16 @@ def estimate_cpd(
...
+                pseudo_counts = np.array(pseudo_counts)
+                if pseudo_counts.shape != cpd_shape:
+                    raise ValueError(
+                        "The shape of pseudo_counts for the node: {node} must be of shape: {shape}".format(
+                            node=node, shape=str(cpd_shape)
+                        )
                     )
-                )
+            else:
+                pseudo_counts = np.ones(cpd_shape, dtype=int) * pseudo_counts
```
ankurankan: @loudly-soft Sorry for the late reply. This line seems to be the culprit: https://travis-ci.org/github/pgmpy/pgmpy/jobs/685246434#L522. Also, you can run specific tests on your system to debug the problem using: `nosetests pgmpy/tests/test_estimators/test_BayesianEstimators.py:TestBayesianEstimator.test_get_parameters2`

loudly-soft:
@ankurankan Thanks.  I've fixed the issue, added test coverage and re-ran `nosetests`.

ankurankan:
@loudly-soft Looks good. Thanks for the PR.

PR #1264, opened 2020-05-17
Fix #1261

### LINKED ISSUE #1261: VariableElimination ignores identical factors
### Subject of the issue
When VariableElimination generates several identical factors (same variables, same values) only one is considered and the rest are discarded. This is due to the use of sets (see e.g. "final_distribution = set()" in ExactInference.py), which do not admit repeated elements. 

This can be fixed by switching to another data structure. I would have time, but which one? Python lists, a multiset using collections.Counter?

This issue (identical factors arising during variable eliimination) will rarely be a problem for learned networks --but it can be academic examples. May be related to other inference problems (issue #856 ?).

### Environment
* pgmpy version: 0.1.10
* Python version: 3.6.10
* Operating System: Ubuntu 18.04

### Steps to reproduce
    import numpy as np
    from pgmpy.models import MarkovModel
    from pgmpy.factors.discrete import DiscreteFactor

    g = MarkovModel([('A', 'B'), ('A', 'C')])
    f1 = DiscreteFactor(variables=['A', 'B'], cardinality=[2, 2], values=np.eye(2) * 2)
    f2 = DiscreteFactor(variables=['A', 'C'], cardinality=[2, 2], values=np.eye(2) * 2)
    g.add_factors(f1, f2)

    print(VariableElimination(g).query(['A']))  # <- gives incorrect result: [2, 2]

    # We now modify one of the factors by a tiny amount, so they are no longer equal
    g.factors[0].values += 1e-8
    print(VariableElimination(g).query(['A']))  # <- gives correct result: [4, 4]

rballester:
### Fixes # .
See #1261.

#### Changes
- In VariableElimination, each newly added factor in working_factors now keeps track of the variable whose elimination gave rise to the factor. This makes sure (a) no factors arising from the same variable are duplicated; (b) identical factors arising from different variables are kept separate as they should.

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -40,12 +40,12 @@ class VariableElimination(Inference):
         # Dealing with evidence. Reducing factors over it before VE is run.
         if evidence:
             for evidence_var in evidence:
-                for factor in working_factors[evidence_var]:
-                    factor_reduced = factor[0].reduce(
+                for factor, origin in working_factors[evidence_var]:
+                    factor_reduced = factor.reduce(
                         [(evidence_var, evidence[evidence_var])], inplace=False
                     )
                     for var in factor_reduced.scope():
-                        working_factors[var].remove(factor)
+                        working_factors[var].remove((factor, origin))
                         working_factors[var].add((factor_reduced, evidence_var))
                 del working_factors[evidence_var]
         return working_factors
@@ -172,9 +172,9 @@ class VariableElimination(Inference):
             # Removing all the factors containing the variables which are
             # eliminated (as all the factors should be considered only once)
             factors = [
-                factor[0]
-                for factor in working_factors[var]
-                if not set(factor[0].variables).intersection(eliminated_variables)
+                factor
+                for factor, _ in working_factors[var]
+                if not set(factor.variables).intersection(eliminated_variables)
             ]
             phi = factor_product(*factors)
             phi = getattr(phi, operation)([var], inplace=False)
@@ -186,11 +186,10 @@ class VariableElimination(Inference):
         # Step 4: Prepare variables to be returned.
         final_distribution = set()
         for node in working_factors:
-            for factor in working_factors[node]:
-                f = factor[0]
-                if not set(f.variables).intersection(eliminated_variables):
-                    final_distribution.add(factor)
-        final_distribution = [f[0] for f in final_distribution]
+            for factor, origin in working_factors[node]:
+                if not set(factor.variables).intersection(eliminated_variables):
+                    final_distribution.add((factor, origin))
+        final_distribution = [factor for factor, _ in final_distribution]
 
         if joint:
             if isinstance(self.model, BayesianModel):
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -32,20 +32,21 @@ class VariableElimination(Inference):
         -------
         dict: Modified working factors.
         """
+
         working_factors = {
-            node: {factor for factor in self.factors[node]} for node in self.factors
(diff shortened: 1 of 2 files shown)


ankurankan:
@rballester Thanks for the PR. I will try to review this ASAP.

ankurankan:
@rballester Thanks for the PR and sorry for the late reply. Could you please also add some test cases which would fail for the previous code, and works with these changes (the example that you had given in #1261 would work)?

pgmpy/inference/ExactInference.py
```diff
@@ -171,24 +172,25 @@ def _variable_elimination(
...
+                working_factors[variable].add((phi, var))
             eliminated_variables.add(var)
 
         # Step 4: Prepare variables to be returned.
         final_distribution = set()
         for node in working_factors:
-            factors = working_factors[node]
-            for factor in factors:
-                if not set(factor.variables).intersection(eliminated_variables):
+            for factor in working_factors[node]:
```
ankurankan: @rballester I would rather do something like `for factor, _ in working_factors[node]`. Would not need the next line then.

pgmpy/inference/ExactInference.py
```diff
@@ -32,20 +32,21 @@ def _get_working_factors(self, evidence):
...
+
         working_factors = {
-            node: {factor for factor in self.factors[node]} for node in self.factors
+            node: {(factor, None) for factor in self.factors[node]} for node in self.factors
         }
 
         # Dealing with evidence. Reducing factors over it before VE is run.
         if evidence:
             for evidence_var in evidence:
                 for factor in working_factors[evidence_var]:
```
ankurankan: Same here. Using `factor[0]` makes the code less readable as it is difficult to deduce why we are only using the first element and what are the other elements. Using `for factor, _ in ...` makes it explicit that there are only 2 elements and the second element is not used anywhere in the loop.

pgmpy/inference/ExactInference.py
```diff
@@ -171,24 +172,25 @@ def _variable_elimination(
             # Removing all the factors containing the variables which are
             # eliminated (as all the factors should be considered only once)
             factors = [
-                factor
+                factor[0]
                 for factor in working_factors[var]
```
ankurankan: Same here.

pgmpy/inference/ExactInference.py
rballester: In this case both elements of the tuple, not just the first one, are needed for line 192 (final_distribution.add(factor)).

We could do this if you prefer it:
```suggestion
            for factor, origin in working_factors[node]:
                if not set(factor.variables).intersection(eliminated_variables):
                    final_distribution.add((factor, origin))
```

pgmpy/inference/ExactInference.py
rballester: You are right, but I don't understand what you propose for this specific line of code (there's no factor[0] here).

pgmpy/inference/ExactInference.py
rballester: Yes, here we can switch to "factor, _". I have done it. Should I issue a new PR, along with the new test?

pgmpy/inference/ExactInference.py
ankurankan: Sorry, I meant to comment on the next line with `factor[0].reduce`. Changing this loop will avoid the indexing there.

pgmpy/inference/ExactInference.py
ankurankan: You can just commit and push to your branch and it should automatically update this PR.

pgmpy/inference/ExactInference.py
ankurankan: Yes, this should work.

ankurankan:
@rballester Looks good. Thanks a lot for the PR.

PR #1266, opened 2020-05-23
Implement tree-augmented naive bayes (TAN)

loudly-soft:
Added tree-augmented naive bayes to models and implemented unit test.


ankurankan:
@loudly-soft Thanks for the PR. I am not very aware of TAN literature and I do have some questions and comments about the best way to implement it. It would really help if you could help me in understanding it or point me to some literature.

1. I see that you have implemented a structure learning algorithm. Are there any efficient inference algorithms as well for TANs? 
2. The current code doesn't imply any network structure restrictions. But from a quick look, I think there should be some. Right?
3. Is Naive Bayes a special case of TANs? Because in that case, I would suggest merging the two models together as that will help us in reducing the codebase size. 
4. The learning algorithm will need to be moved to a separate file in the `estimators` directory.
5. The tests need to be more thorough and check for edge cases etc.

loudly-soft:
@ankurankan  Sorry, I don’t know if there are any efficient inference algorithms for TAN.  I’m mostly experimenting with it as a classifier for making predictions and learning more about BN in general.  Also, I did not know that structure learning is already supported in pgmpy.  My entry point to pgmpy was `examples` and it did’t showcase structure learning so I missed out. 

About TAN, the original paper can be found at:

https://www.cse.huji.ac.il/~nir/Papers/FrGG1.pdf

I don’t believe Naive Bayes is a special case of TAN because TAN requires a tree structure over the independent variables.  However, Naive Bayes and TAN would be special cases of [FAN](https://www.cs.ru.nl/P.Lucas/pgm02-lucas.pdf)
or [ETAN](https://www.sciencedirect.com/science/article/pii/S0888613X15000547#fg0010).

Hopefully, these help although I wonder now if the existing structure learning capability in pgmpy supersedes the needs for TAN/FAN/ETAN.

As for the question of network structure restrictions, let me continue reading my BN books and get back to you on that.  :)

loudly-soft:
@ankurankan If I understand your question correctly, restrictions mean manual constraints such as presence and absence of directed edge between two variables.  If that’s the case, the only manual restrictions are that there must be a directed edge from the class variable to the other variables.  The edges between the non-class variables are learned using the chow-liu algorithm.

As for efficient inference algorithm for TAN, perhaps by applying (cutset?) conditioning on the class variable, the TAN graph would reduce to a tree so that the prob prop algorithm can be applied efficiently.  Again, I'm not an expert at BN so take my words with a grain of salt.

loudly-soft:
@ankurankan Code has been moved to `estimators` with more tests added.  I’m not able to reproduce the pep257 error by running `pep257` locally.  I’ve tried numerous ways of formatting the docstring to no avail.  Any idea?

ankurankan:
@loudly-soft You can fix all the problems by just running `black .` from the base pgmpy directory. It will automatically format all your code, just make sure you have `black` installed. 

And thanks again for the PR but I will need some time to review this as I am not aware of the use cases and algorithms for TAN, so I will need to read up on it. Sorry for the delay.

loudly-soft:
@ankurankan No worries, take your time.  In case it helps with your investigation, R’s [bnlearn](https://rdrr.io/cran/bnlearn/man/naive.bayes.html) also supports TAN.  Perhaps there might be some relevant materials there.

Thanks for the suggestion of using `black`.  Oddly, it didn’t solve the issue.  Anyway, I’ll hold off further attempts for now.

loudly-soft:
@ankurankan btw, If I'm not mistaken, the [Chow-Liu algorithm](https://en.wikipedia.org/wiki/Chow%E2%80%93Liu_tree#:~:text=Chow%20and%20Liu%20provide%20a,information%20pair%20to%20the%20tree.&text=The%20main%20idea%20in%20the,in%20the%20pairwise%20marginal%20distribution.) is currently not supported in `pgmpy`.  I'm a bit curious why.

ankurankan:
@loudly-soft What exactly is failing? If you are referring to codacy failing, don't worry too much about it and I think if you add a space after the `class TreeAugmented...` line, it should get fixed. 

There's no specific reason why Chow-Liu hasn't been implemented yet, we were just focusing on more general algorithms. But you are welcome to implement it.

loudly-soft:
@ankurankan Yes, Codacy was failing and was driving me nuts for while but all is well now.  Thanks.

ankurankan:
Dup #1289

Issue #1278, opened 2020-06-09
Implement the Chow-Liu algorithm for structure learning


loudly-soft:
Feature request ticket for adding the [Chow-Liu algorithm](https://en.wikipedia.org/wiki/Chow%E2%80%93Liu_tree#:~:text=Chow%20and%20Liu%20provide%20a,information%20pair%20to%20the%20tree.&text=The%20main%20idea%20in%20the,in%20the%20pairwise%20marginal%20distribution.) to `estimators`.

loudly-soft:
I'll try to pick this up when I have the chance.

1021808202:
I see “from pgmpy.estimators import TreeSearch” in the Chow liu tree structure learning, but I cant import it in my lab.Can I use it now?

1021808202:
I have to get the BN which have 600+ nodes,you know,the leaning speed of hillclimb method is too slow.Could you please add the Chow liu tree structure learning to pgmpy as soon as possible,I will appeciate it a lot.

loudly-soft:
@1021808202  FYI, the Chow-Liu algorithm has been merged to `dev`.  Here's an example [notebook](https://github.com/pgmpy/pgmpy/blob/dev/examples/Structure%20Learning%20with%20Chow-Liu.ipynb) on how to use it.

ankurankan:
Fixed in #1289

PR #1289, opened 2020-06-22
Chow liu tree structure learning

loudly-soft:
Implemented Chow-liu tree structure learning (Issue #1278) and unit tests.


ankurankan:
@loudly-soft Quick question: Is everything from #1266 is included in this PR? If yes, I will close the older PR and go through this.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python
+
+import networkx as nx
+
+from pgmpy.estimators import StructureEstimator
+from sklearn.metrics import mutual_info_score
+from pgmpy.base import DAG
```
ankurankan: The imports should be in the order (with an empty line in between each category):
1. Python core imports
2. Third party package imports
3. pgmpy imports.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, data, class_node, root_node=None, **kwargs):
+        """
+        Search class for learning tree-augmented naive bayes (TAN) graph structure with a given set of variables.
+
+        Parameters
+        ----------
+        data: pandas DataFrame object
+            dataframe object where each column represents one variable.
+
+        class_node: node
```
ankurankan: `node` should be the expected data type. pgmpy can accept any variable name which is hashable, so can instead use `hashable object`.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
...
+            Specifies how to deal with missing data, if present. If set to `True` all rows
+            that contain `np.Nan` somewhere are ignored. If `False` then, for each variable,
+            every row where neither the variable nor its parents are `np.NaN` is used.
+            This sets the behavior of the `state_count`-method.
+
+        References
+        ----------
+        Friedman N, Geiger D and Goldszmidt M (1997). Bayesian network classifiers. Machine Learning 29: 131–163
+
+        """
```
ankurankan: Please also add an `Examples` section with some basic usage examples.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
...
+        >>> values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)), columns=['A', 'B', 'C', 'D', 'E'])
+        >>> class_node = 'A'
+        >>> est = TreeAugmentedNaiveBayesSearch()
+        >>> model = est.estimate()
+        >>> nx.draw_circular(model, with_labels=True, arrowsize=20, arrowstyle='fancy', alpha=0.3)
+        >>> plt.show()
+
+        """
+        if self.class_node not in self.data.columns:
+            raise ValueError("class node must exist in data")
```
ankurankan: The error message can be made a bit informative by also showing the given class node. So, something on the lines of: Given class node is: {class_node} but no columns in data exist by that name.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
...
+
+        if self.root_node is not None and self.root_node not in self.data.columns:
+            raise ValueError("root node must exist in data")
```
ankurankan: Same here.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # construct maximum spanning tree
+        G = nx.Graph()
+        df_features = self.data.loc[:, self.data.columns != self.class_node]
+        total_cols = len(df_features.columns)
```
ankurankan: If I am understanding this code block correctly, you are trying to compute the mutual info score between all possible combination of variables, creating an edge between them and using the mutual info as the edge weight. I think this can be simplified by computing an array of edge_weights and then creating the graph from that. Something like:
```
weights = pd.DataFrame(0, index=df_features.columns, columns=df_features.columns)
for u_index, u, v_index, v in combinations(enumerate(df_features.columns), 2):
   weights.iat[u_index, v_index] = mutual_info_score(df_features.iloc[:, u_index], df_features.iloc[:, v_index])
G = nx.from_pandas_adjacency(weights, create_using=nx.Graph)
```
I think the main benefit would be that this would be much faster because networkx's `add_node` and `add_edge` can be super slow. Also, I think this is more readable.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
...
+            for j in range(i + 1, total_cols):
+                to_node = df_features.columns[j]
+                G.add_node(to_node)
+                # edge weight is the MI between a pair of independent variables
+                mi = mutual_info_score(df_features.iloc[:, i], df_features.iloc[:, j])
+                G.add_edge(from_node, to_node, weight=mi)
+        T = nx.maximum_spanning_tree(G)
+
+        # create DAG by directing edges away from root node
+        if self.root_node:
```
ankurankan: I would suggest adding a line to set a variable `root_node` at the top after argument checks and use that throughout instead of testing it later. Something like:
```
root_node = df_features.columns[0] if self.root_node is None else self.root_node
```
After that you can just replace this if else block with: `D=nx.bfs_tree(T, root_node)`

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,103 @@
...
+        ----------
+        Friedman N, Geiger D and Goldszmidt M (1997). Bayesian network classifiers. Machine Learning 29: 131–163
+
+        """
+        self.class_node = class_node
+        self.root_node = root_node
+
+        super(TreeAugmentedNaiveBayesSearch, self).__init__(data, **kwargs)
+
+    def estimate(self):
```
ankurankan: It would be nice if you can add the major steps of the algorithm as comments. Something like:
```
Step 1: Argument checks
Step 2: Create a fully connected graph with edge weights as the mutual info score between them.
Step 3: Get the Maximum Spanning tree and orient the edges.
```

pgmpy/estimators/TreeSearch.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python
+
+import networkx as nx
+
+from pgmpy.estimators import StructureEstimator
+from sklearn.metrics import mutual_info_score
+from pgmpy.base import DAG
+
+
+class TreeSearch(StructureEstimator):
```
ankurankan, line 20: I think TreeSearch and NaiveBayesTreeSearch can be combined together by adding some extra parameters to the `estimate` method maybe: `return_type`. It will remove boilerplate code and also make my life easier to maintain. :D

pgmpy/tests/test_estimators/test_TreeAugmentedNaiveBayesSearch.py
```diff
@@ -0,0 +1,57 @@
+import unittest
+
+import pandas as pd
+
+from pgmpy.estimators import TreeAugmentedNaiveBayesSearch
+
+
+class TestTreeAugmentedNaiveBayesSearch(unittest.TestCase):
```
ankurankan: I think it would a good idea to add some more realistic tests. For example, generate data from a model and then try to learn it back the structure.

ankurankan (changes requested):

ankurankan:
@loudly-soft Sorry for the delay in reviewing this. I went through the code, everything looks good but I have added some comments mostly regarding the implementation. Please let me know if you need any help or something isn't clear.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
ankurankan: You can also add a nice `tqdm` progress bar and make the computation parallel at this step.

loudly-soft:
> @loudly-soft Quick question: Is everything from #1266 is included in this PR? If yes, I will close the older PR and go through this.

@ankurankan Oops, yes.

pgmpy/estimators/TreeSearch.py
loudly-soft: I agree and I think it may call for more.  Now that I’ve implemented TAN and the Chow-Liu algorithm, I’m tempted to roll naive bayes into the mix as well.  (I know, I did say before that naive bayes is not a special case of TAN but both are special cases of FAN or ETAN.)  However, in terms of coding, they are like Lego blocks and each block can be a standalone component:

- TAN = Chow-Liu + Naive Bayes

So, my questions are:

1. Do you see Naive Bayes as belonging to “estimators” like TAN and Chow-Liu even though it doesn’t do much structure learning from data.
2. What would be the name of this generic class that can do TAN, Chow-Liu (and Naive Bayes and maybe FAN or ETAN in the future)?

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
loudly-soft: I’d be interested in adding documentation to `examples` but I may be a little bit short on time in the coming weeks.  In the interest of not blocking this PR, perhaps I can create a separate ticket to keep track of that task?

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
loudly-soft: Do you mean replace the word `node` with `hashable object`?

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
loudly-soft: Sorry, I'm not familiar with multi-threading in Python.  Do you mean `joblib` and the step to parallelize is:

`weights.iat[u_index, v_index] = mutual_info_score(df_features.iloc[:, u_index], df_features.iloc[:, v_index])`

I assume Pandas is thread-safe, and/or concurrent writes wouldn't cause any race condition or deadlocks to the underlying data structures backing the dataframe.

pgmpy/tests/test_estimators/test_TreeAugmentedNaiveBayesSearch.py
loudly-soft: Since TAN is generally used for classification tasks, perhaps the assert statement should measure classification errors?

The steps are:

1. Manually build a TAN graph and CPDs
2. Generate N rows of data from graph
3. Learn the TAN graph from data
4. Use the learned graph to predict from the independent variables of the data
5. Measure classification errors

I don’t suppose you have code examples along this line already that I can bootstrap from?

(I think this might go nicely into `examples` as a demo as well.)

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
ankurankan: We have been using something like this for node: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/base/DAG.py#L277

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
ankurankan: Since this is the only way for a user to understand how to use the class, I would suggest adding some basic examples before merging this. The same examples that you are adding in the tests would work too. Or else, I can also add the examples if you don't have time for it.

pgmpy/estimators/TreeAugmentedNaiveBayesSearch.py
ankurankan: Yes, I was suggesting using `joblib`. We can run the `for` loop in parallel as there is no order requirement in this case. You can refer to: https://joblib.readthedocs.io/en/latest/parallel.html for parallelizing for loops using joblib. We will also need an argument `n_jobs=-1` in the method for the user to specify the number of threads they want to use. But this isn't a priority change, so can be done later as well.

pgmpy/tests/test_estimators/test_TreeAugmentedNaiveBayesSearch.py
ankurankan: Yes, something along those lines. But instead of testing for classification errors, I would suggest testing for smaller parts. For example, testing the learned structure or the learned parameters on a given structure. The problem with testing classification error directly (starting from a initial model/generated data) is that there are multiple steps where uncertainty can be introduced and it all can add up to give unreliable tests (or tests failing sometimes).

pgmpy/estimators/TreeSearch.py
ankurankan: We already have a model class for Naive Bayes (https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/NaiveBayes.py). So, I would suggest modifying it to be able to represent FAN models with special cases for Naive Bayes and TAN. The estimators should stay in the `estimators` dir and be completely separate from the model class, and should just return an instance of the learned model.

loudly-soft:
@ankurankan Thanks.  I'll try to return to this after this week at the earliest.

loudly-soft:
@ankurankan I’ve added to `examples` and they run fine with ipython but my Jupyter notebook is broken so the examples are currently missing output.

Also, looks like appveyor is choking up:

`
Build started
git clone -q https://github.com/pgmpy/pgmpy.git C:\projects\pgmpy
fatal: unable to access 'https://github.com/pgmpy/pgmpy.git/': Could not resolve host: github.com

Command exited with code 128
`

ankurankan:
@loudly-soft Thanks for the changes. I will have a look at it. I also restarted the appveyor build, seems to be running fine now.

loudly-soft:
Fixed my Jupyter notebook and fleshed out the code in `examples`.

FTR, the issue was [this](https://github.com/pytorch/pytorch/issues/2575).  On Ubuntu 14.04.6 LTS, downgrading scikit-learn to 0.20.3 fixed the problem.

ankurankan:
@loudly-soft Thanks a lot again for the PR. I went through it and everything looks great. I wanted to add a commit to this PR which changes some behavior and the algorithm should be substantially faster. Could you please check that `allow edits from maintainers` is checked for this PR (https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) as I am not able to push my commit? Also, if you have some time, could you please go through my changes and see if it looks good?

Here's a summary of my changes:

1. Changes name `return_type` to `estimator_type` as the function would return a DAG in both the cases.
2. Move `return_type` and `class_node` arguments to `estimate` method. The idea is that the user would create an estimator object and then try different estimators without needing to redefine the object.
3. Adds edge_weights parameter which can accept any function to compute edge weights. 
3. Cosmetic changes: 
 - Docs within 88 chars, for better rendering on the website
 - Using f-strings everywhere, they are cleaner and much faster
 - No definition of any `self.var` in methods other than __init__. Better for readability
 - Convert `chow-liu` to a static class.
 - Removed iteration while computing weights. 
 - Move algorithm steps inside the code.

ankurankan (approved):

loudly-soft:
@ankurankan Yeah sure, I'm curious about the changes you're making.  I'll have a look.

Hmm..  "Allow edits by maintainers" is already checked on this PR.

ankurankan:
@loudly-soft Hmm, don't know why it wasn't working earlier. But got through now. Please have a look if it looks good.

pgmpy/estimators/TreeSearch.py
```diff
@@ -100,87 +96,134 @@ def estimate(self):
...
+
 
-            # add edge from class node to other nodes
-            for node in df_features.columns:
-                D.add_edge(class_node, node)
+        n_jobs: int (default: -1)
+            Number of jobs to run in parallel. `-1` means use all processors.
 
-            return DAG(D)
+        edge_weights: str or function (default: mutual info)
```
loudly-soft: I'm not sure what the convention is in `pgmpy` but I wonder, for clarity, should `edge_weights` be renamed to `edge_weights_fn` or something similar to hint that a function (or function name) is expected.

pgmpy/estimators/TreeSearch.py
```diff
@@ -100,87 +96,134 @@ def estimate(self):
...
+            pbar = combinations(data.columns, 2)
 
-        vals = Parallel(n_jobs=n_jobs)(
-            delayed(calculate_mi)(u_index=u_index, v_index=v_index)
-            for (u_index, u), (v_index, v) in pbar
+        vals = Parallel(n_jobs=n_jobs, prefer="threads")(
+            delayed(edge_weights)(data.loc[:, u], data.loc[:, v]) for u, v in pbar
         )
+        weights = np.zeros((n_vars, n_vars))
+        weights[np.triu_indices(n_vars, k=1)] = vals
```
loudly-soft, line 217: More of a question than review.  About `Parallel`, does it preserve ordering? It must be, otherwise, the indices in these 2 blocks of code would not match up.

```
pbar = combinations(data.columns, 2)
vals = Parallel(n_jobs=n_jobs, prefer="threads")(
   delayed(edge_weights)(data.loc[:, u], data.loc[:, v]) for u, v in pbar
)
```

`weights[np.triu_indices(n_vars, k=1)] = vals`

pgmpy/estimators/TreeSearch.py
```diff
@@ -100,87 +96,134 @@ def estimate(self):
...
+        ...                       columns=['A', 'B', 'C', 'D', 'E'])
+        >>> est = TreeSearch(values, root_node='B')
+        >>> model = est.estimate(estimator_type='chow-liu')
+        """
+        # Step 0: Check for edge weight computation method
+        if edge_weights == "mutual_info":
+            edge_weights = mutual_info_score
+        elif not isinstance(edge_weights, callable):
+            raise ValueError(
+                "edge_weights either needs to be 'mutual_info' or a function of form fun(array, array)"
```
loudly-soft: Going by the docstring, should the error msg include the options 'adjusted_mutual_info' and 'normalized_mutual_info'?

pgmpy/estimators/TreeSearch.py
```diff
@@ -100,87 +96,134 @@ def estimate(self):
...
-            # add edge from class node to other nodes
-            for node in df_features.columns:
-                D.add_edge(class_node, node)
+        n_jobs: int (default: -1)
+            Number of jobs to run in parallel. `-1` means use all processors.
 
-            return DAG(D)
+        edge_weights: str or function (default: mutual info)
+            Method to use for computing edge weights. Options are:
+                1. 'mutual_info': Mutual Information Score.
```
loudly-soft, line 172: Should 'mutual_info' or one of them be default?

loudly-soft:
@ankurankan Thanks.  I've added my 2 cents.  Pls let me know if anything is not clear.

loudly-soft (commented):
@ankurankan Thanks. I've added my 2 cents. Pls let me know if anything is not clear.

pgmpy/estimators/TreeSearch.py
ankurankan: Fixed.

pgmpy/estimators/TreeSearch.py
ankurankan: Yes, it does preserve the order.

pgmpy/estimators/TreeSearch.py
ankurankan: Yes, have changed it to `edge_weights_fn` now.

ankurankan:
@loudly-soft Thanks for the comments and the PR.

loudly-soft:
@ankurankan And thanks for the nifty tool again!

Issue #1293, opened 2020-07-03
KeyError in BayesianModel.get_markov_blanket method


Lostefra:
### Subject of the issue
The method ```get_markov_blanket``` of ```BayesianModel``` raise a KeyError on istruction ```blanket_nodes.remove(node)``` when invoked on a leaf node of the network.

### Your environment
* pgmpy version: 0.1.10, dev
* Python version: 3.6
* Operating System: Ubuntu 18.04

### Steps to reproduce
```Python
from pgmpy.models import BayesianModel

model = BayesianModel([('Pop', 'EC'),   ('Urb', 'EC'),   ('GDP', 'EC'),
                       ('EC', 'FFEC'),  ('EC', 'REC'),   ('EC', 'EI'),
                       ('REC', 'CO2'),  ('REC', 'CH4'),  ('REC', 'N2O'),
                       ('FFEC', 'CO2'), ('FFEC', 'CH4'), ('FFEC', 'N2O')])

result = model.get_markov_blanket('CH4') # raise KeyError
```
<img src="data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAS0AAAFBCAYAAAAi1fnPAAAGB3RFWHRteGZpbGUAJTNDbXhmaWxl
JTIwaG9zdCUzRCUyMmFwcC5kaWFncmFtcy5uZXQlMjIlMjBtb2RpZmllZCUzRCUyMjIwMjAtMDct
MDNUMTklM0E0NCUzQTAwLjMyM1olMjIlMjBhZ2VudCUzRCUyMjUuMCUyMChYMTEpJTIyJTIwZXRh
ZyUzRCUyMmZDYXdnZE1YYmMwN2xtWmw1Q0JwJTIyJTIwdmVyc2lvbiUzRCUyMjEzLjMuNiUyMiUy
MHR5cGUlM0QlMjJnb29nbGUlMjIlM0UlM0NkaWFncmFtJTIwaWQlM0QlMjItUUVyV1lsRFc4eHgx
WGdNVWdTTCUyMiUyMG5hbWUlM0QlMjJuZXQlMjIlM0U3VnBiYzZJd0ZQNDFQcllEQVVRZlc3eTBN
MjExNm01WCUyQjdLVFNoUzZrVGd4Q3U2djN5aUpJRmhhcSUyQlV5M1JlSGMwaEN6dVg3emdsWTA2
[...]

PR #1294, opened 2020-07-03
bug fix: KeyError in BayesianModel.get_markov_blanket method #1293

Lostefra:
Bug fix for KeyError in BayesianModel.get_markov_blanket method pgmpy#1293

### DIFF
--- a/pgmpy/models/BayesianModel.py
+++ b/pgmpy/models/BayesianModel.py
@@ -783,5 +783,5 @@ class BayesianModel(DAG):
         for child_node in children:
             blanket_nodes.extend(self.get_parents(child_node))
         blanket_nodes = set(blanket_nodes)
-        blanket_nodes.remove(node)
+        blanket_nodes.discard(node)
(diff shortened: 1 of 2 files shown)


ankurankan:
@Lostefra Thanks for the PR. Could you please also add a test case ( in this file: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/tests/test_models/test_BayesianModel.py) that would fail with the previous codebase and get fixed with the changes? The example that you had given in the issue would also work.

Lostefra:
@ankurankan Ok, done it. I am not sure what is wrong with the checks

ankurankan:
@Lostefra It's because of code formatting and a minor error. I have pushed a commit fixing both. Everything else looks good. Thanks for the fix.

Lostefra:
@ankurankan Ok, seen. It is my first time ever for a PR and I never really worked with python unittest before, I am sure you can understand :) 

Cheers!

PR #1310, opened 2020-08-08
Fixed minor typos

loudly-soft:
I didn’t know pgmpy supports SEM as well!  I wonder what other features there are.  Maybe README.md could list all the bells and whistles...


ankurankan:
@loudly-soft Thanks for the PR. Yes, the documentation really needs a revamp.

PR #1327, opened 2020-09-27
Removes/renames default argument `bool=True` in chi_square function behavior.

### LINKED ISSUE #1326: Unable to run MMPC function in MMHC estimator
### Subject of the issue
`estimators.mmhc.mmpc()` crashes regardless of dataset used.

### Your environment
* pgmpy version 0.1.11.dev0
* Python version 3.6.9
* Operating System macOS catalina

### Steps to reproduce
Tell us how to reproduce this issue. Please provide a **minimal reproducible code** of
the issue you are facing if possible.

```
import pandas as pd
import numpy as np

import pgmpy
from pgmpy.estimators import MmhcEstimator

df = pd.DataFrame(
    np.random.randint(0, 2, size=(15000, 3)), columns=list("XYZ")
)
df['sum'] = df.sum(axis=1)
mmhc = MmhcEstimator(df)
skeleton = mmhc.mmpc()
```

### Expected behaviour
Tell us what should happen
Function should return an `UndirectedGraph` object.

### Actual behaviour
```
  File "explore.py", line 12, in <module>
    skeleton = mmhc.mmpc()
  File "/.../lib/python3.6/site-packages/pgmpy-0.1.11.dev0-py3.6.egg/pgmpy/estimators/MmhcEstimator.py", line 186, in mmpc
  File "/.../lib/python3.6/site-packages/pgmpy-0.1.11.dev0-py3.6.egg/pgmpy/estimators/MmhcEstimator.py", line 171, in max_min_heuristic
  File "/.../lib/python3.6/site-packages/pgmpy-0.1.11.dev0-py3.6.egg/pgmpy/estimators/MmhcEstimator.py", line 163, in min_assoc
  File "/.../lib/python3.6/site-packages/pgmpy-0.1.11.dev0-py3.6.egg/pgmpy/estimators/MmhcEstimator.py", line 163, in <genexpr>
  File "/.../lib/python3.6/site-packages/pgmpy-0.1.11.dev0-py3.6.egg/pgmpy/estimators/MmhcEstimator.py", line 159, in assoc
  File "/.../lib/python3.6/site-packages/pgmpy-0.1.11.dev0-py3.6.egg/pgmpy/estimators/CITests.py", line 154, in chi_square
KeyError: 'significance_level'
```

jovalie:

- Fixes #1326

- Rename 'bool=True' argument to 'use_sig=False'
- Update tests to match

The `bool=True` default argument in the `chi_square` function forced it to return a boolean flag indicating if the determined `p_value` was within a particular significance level threshold. While this is a useful feature, changing the return value of this function broke estimator classes that expected the return value of this function to be chi, dof, and p_value and did not provide the optional `significance_level` argument.

Furthermore, I don't think naming an argument after a data type is the most intuitive, so I renamed this argument to `use_sig` and changed its default value to `False`. I also updated the unit tests referencing this function regardless of if calls to `chi_square` referenced a `significance_level` or not.

### DIFF
--- a/pgmpy/estimators/MmhcEstimator.py
+++ b/pgmpy/estimators/MmhcEstimator.py
@@ -6,6 +6,7 @@ from pgmpy.estimators import StructureEstimator, HillClimbSearch, BDeuScore
 from pgmpy.independencies import Independencies, IndependenceAssertion
 from pgmpy.estimators.CITests import chi_square
 
+from tqdm import tqdm
 
 class MmhcEstimator(StructureEstimator):
     def __init__(self, data, **kwargs):
@@ -177,7 +178,7 @@ class MmhcEstimator(StructureEstimator):
 
         # Find parents and children for each node
         neighbors = dict()
-        for node in nodes:
+        for node in tqdm(nodes):
             neighbors[node] = []
 
             # Forward Phase
@@ -205,14 +206,14 @@ class MmhcEstimator(StructureEstimator):
                         break
 
         # correct for false positives
-        for node in nodes:
+        for node in tqdm(nodes):
             for neigh in neighbors[node]:
                 if node not in neighbors[neigh]:
                     neighbors[node].remove(neigh)
 
         skel = UndirectedGraph()
         skel.add_nodes_from(nodes)
-        for node in nodes:
+        for node in tqdm(nodes):
             skel.add_edges_from([(node, neigh) for neigh in neighbors[node]])
 
         return skel
--- a/pgmpy/estimators/CITests.py
+++ b/pgmpy/estimators/CITests.py
@@ -32,7 +32,7 @@ def independence_match(X, Y, Z, independencies, **kwargs):
     return IndependenceAssertion(X, Y, Z) in independencies
 
 
-def chi_square(X, Y, Z, data, boolean=True, **kwargs):
+def chi_square(X, Y, Z, data, use_sig=False, **kwargs):
     """
     Chi-square conditional independence test.
     Tests the null hypothesis that X is independent from Y given Zs.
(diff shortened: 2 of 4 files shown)


ankurankan:
@jovalie Thanks for the PR but I don't think this is the best way to solve the bug. I had used the argument name as `boolean` to keep it uniform across all the different independence test and wanted to keep some flexibility on how the decision for True/False is made. For example, in chi-squared test, the p-value could be very wrong when there are a lot of conditional variables, and in such cases, I would like to have a feature to make the decision based on rmse value or confidence intervals. So, the boolean argument just specifies whether I want a True/False value or not, and how that decision is made depends on the extra arguments being passed. 

In this case, I would just change this line: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/MmhcEstimator.py#L159 to pass the `boolean=False` argument.

pgmpy/estimators/MmhcEstimator.py
```diff
@@ -177,7 +178,7 @@ def max_min_heuristic(X, Zs):
 
         # Find parents and children for each node
         neighbors = dict()
-        for node in nodes:
+        for node in tqdm(nodes):
```
ankurankan, line 181: @jovalie Some users don't like to have progress bars, so we have tried to always have an argument `show_progress` in the function/method to specify whether to show progress or not. Also, we need to consider the global var `SHOW_PROGRESS` (https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/global_vars.py#L40). In this case, I would instead not use the progress bar because I don't think this is computationally expensive enough to take substantial time to run.

jovalie: Sorry. I added that in for my own sake but forgot to remove it when I realized there was a `SHOW_PROGRESS` global variable.

jovalie:
> @jovalie Thanks for the PR but I don't think this is the best way to solve the bug. I had used the argument name as `boolean` to keep it uniform across all the different independence test and wanted to keep some flexibility on how the decision for True/False is made. For example, in chi-squared test, the p-value could be very wrong when there are a lot of conditional variables, and in such cases, I would like to have a feature to make the decision based on rmse value or confidence intervals. So, the boolean argument just specifies whether I want a True/False value or not, and how that decision is made depends on the extra arguments being passed.
> 
> In this case, I would just change this line: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/MmhcEstimator.py#L159 to pass the `boolean=False` argument.

Fair enough. Adding a `bool=False` argument to the chi_square function was not the most intuitive solution when trying to figure out what was wrong with MMHC, but your explanation with the context of your other independence tests makes sense.

PR #1333, opened 2020-09-30
New BayesianModelProbability class to calc pmf values for BN models

mbaak:
Added new inference class BayesianModelProbability to calculate probability (pmf) values on Bayesian Models for entire data points.

For example do:

inference = BayesianModelProbability(model)
logp = inference.log_probability(X)

... to get the log probability array, where X is a pd.DataFrame or numpy 2d array. And do:

llsum = inference.score(X)

... to get the log-likelihood sum.

When using numpy arrays, one can also specify the ordering of the columns in X. E.g.

llsum = inference.score(X, ordering)

... where ordering is the ordering of column names in X, as used by the model to address the variables.

For this new functionality I have also added unit tests under tests_inference/.

And I also added a new inference class BayesianModelInference to avoid small overlap with the BayesianModelSampling class.

- [ X] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [ X] Make sure you are making a pull request against the **dev branch** (left side). Also you should start *your branch* off *our dev*.
- [ X] Check the commit's or even all commits' message styles matches our requested structure.

Changed:
- pgmpy/inference/__init__.py
- pgmpy/sampling/Sampling.py
New:
- pgmpy/inference/bn_inference.py
- pgmpy/tests/test_inference/test_bn_inference.py

### DIFF
--- a/pgmpy/inference/__init__.py
+++ b/pgmpy/inference/__init__.py
@@ -2,12 +2,15 @@ from .base import Inference
 from .ExactInference import BeliefPropagation
 from .ExactInference import VariableElimination
 from .dbn_inference import DBNInference
+from .bn_inference import BayesianModelInference, BayesianModelProbability
 from .mplp import Mplp
 
 __all__ = [
     "Inference",
     "VariableElimination",
     "DBNInference",
+    "BayesianModelInference",
+    "BayesianModelProbability",
     "BeliefPropagation",
     "BayesianModelSampling",
     "GibbsSampling",
--- a/pgmpy/inference/bn_inference.py
+++ b/pgmpy/inference/bn_inference.py
@@ -1,0 +1,146 @@
+from pgmpy.inference import Inference
+from pgmpy.models import BayesianModel
+import pandas as pd
+import numpy as np
+import networkx as nx
+import itertools
+
+
+class BayesianModelInference(Inference):
+    """Inference class specific to Bayesian Models
+    """
+    def __init__(self, model):
+        """Class to calculate probability (pmf) values specific to Bayesian Models
+
+        :param model: instance of BayesianModel model on which inference queries will be computed
+        """
+        if not isinstance(model, BayesianModel):
+            raise TypeError("Model expected type: BayesianModel, got type: ", type(model))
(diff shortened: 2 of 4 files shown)


mbaak:
Ping ping.

ankurankan:
@mbaak Sorry for the delay on this. I still haven't got a chance to look at it. But I will get to it as soon as possible (hopefully coming week).

ankurankan:
@mbaak Thanks for the PR and sorry once again for so much delay. If I am understanding things correctly the implementation computes the likelihood and log-likelihood for a given dataset and model? However, I am not aware of any standalone applications of likelihood, could you maybe point me to some resources so that I can have a better understanding of how everything fits together and also if the inference module is the best place to have it?

The code looks good to me. I just have a couple of minor comments:
1. For docstrings, we use the numpydoc style (https://numpydoc.readthedocs.io/en/latest/format.html).
2. For consistency, I would suggest changing the argument `X` to `data` as we have used `data` throughout the package.

mbaak:
Hi, thanks for reviewing. 

I've been working on advanced data modelling and synthetic data generation, in particular an idea to improve the BN modelling of categorical datasets (it improves the modelling of the non-linear correlations). Two things I need for that:
- The likelihood sum is a good way of checking how well the model describes the full dataset.
E.g. see sec 5.2 of https://arxiv.org/pdf/1907.00503.pdf
- Most importantly, I need the pmf value for any individual data point x. For example, if you want to build a classifier based on BNs for different data classes, having the pmf values for each class is essential.

B/c of this I think the inference class is a good place for this, but I'm happy move it elsewhere. Just let me know.

I'll go ahead and update the code with your suggestions.

mbaak:
Hi, I updated the code with your suggestions.

ankurankan:
@mbaak Thanks for the updates and the link. From my understanding, I think it would be best to have this class in something like a "model criticism" module where we have different methods to test a given model. I have a few ideas for some other methods for testing but until I implement them I think it's fine to keep this in the inference module, and I can move it later.

PR #1345, opened 2020-10-22
Fix for HillClimbSearch when scoring_method is StructureScore

### LINKED ISSUE #1344: Can no longer specify StructureScore in HillClimbSearch
### Subject of the Issue
With the updates to the HillClimbSearch (#1312) class in estimators, bugs were introduced with specifying the scoring method. Documentation was also not updated to reflect changes.

### Steps to reproduce
The code that sets scoring method seems to have moved down to the function call `estimate` rather than being set when you initialize the class - this has not been reflected in documentation:

    est = HillClimbSearch(data, scoring_method=BDeuScore(data))
    best_model = est.estimate()

will result in k2score being used regardless of scoring method passed.

Moving the scoring method down to the estimate function:

    est = HillClimbSearch(data)
    best_model = est.estimate(scoring_method=BDeuScore(data))

results in the following error `AttributeError: 'BDeuScore' object has no attribute 'lower'` due to [https://github.com/pgmpy/pgmpy/blob/38babb9e7efbeb9c296dbfcb4ea64eba34d75e61/pgmpy/estimators/HillClimbSearch.py#L224](url)

If you fix the above bug, then you would get the following error `TypeError: 'BDeuScore' object is not callable` due to [https://github.com/pgmpy/pgmpy/blob/38babb9e7efbeb9c296dbfcb4ea64eba34d75e61/pgmpy/estimators/HillClimbSearch.py#L234](url)

mjmt05:
Bug fix where HillClimbSearch estimator would error when passing scoring_method a Structure Score instance.

Updated the example in the function documentation.

- Fixes #1344 

- Added in a check to see if scoring method is a string before calling `scoring_method.lower()`
- If scoring_method is not a string assign score as the scoring_method provided.
- Updated example in function documentation.

### DIFF
--- a/pgmpy/estimators/HillClimbSearch.py
+++ b/pgmpy/estimators/HillClimbSearch.py
@@ -203,8 +203,8 @@ class HillClimbSearch(StructureEstimator):
         ... data = pd.DataFrame(np.random.randint(0, 5, size=(5000, 9)), columns=list('ABCDEFGHI'))
         >>> # add 10th dependent variable
         ... data['J'] = data['A'] * data['B']
-        >>> est = HillClimbSearch(data, scoring_method=BicScore(data))
-        >>> best_model = est.estimate()
+        >>> est = HillClimbSearch(data)
+        >>> best_model = est.estimate(scoring_method=BicScore(data))
         >>> sorted(best_model.nodes())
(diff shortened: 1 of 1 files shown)


ankurankan:
@mjmt05 Thanks for reporting this bug and the PR.

PR #1360, opened 2020-12-09
Fix for PC algorithm, skeleton_to_pdag

### LINKED ISSUE #1359: Some bugs in PC -> skeleton_to_pdag
### Some bugs in PC -> skeleton_to_pdag
When I debug the module `PC`, I found that `node_pairs` in `skeleton_to_pdag` is defined as a `combination` iterator. So after first iteration, it will not yield any more. Check this answer:

https://stackoverflow.com/a/25336739

In another words, steps 2, 3, and 4 in `skeleton_to_pdag` are not performed in all circumstances. In addition, all codes inside of those `for` loop (steps 2, 3, and 4) have never been tested out. Bugs in there are not reflected by the tests results.

### Your environment
* pgmpy 0.1.12
* Python 3.6.11
* Operating System Ubuntu Linux x86_64

### Steps to reproduce
Using a simple example of `PC` estimation

```
from itertools import combinations
from pgmpy.estimators import PC
from pgmpy.sampling import BayesianModelSampling
import networkx as nx
import matplotlib.pyplot as plt

if __name__ == "__main__":
    model = get_example_model('asia')
    samples = BayesianModelSampling(model).forward_sample(size=int(1e5))
    est = PC(data=samples)
    dag = est.estimate(variant='stable', return_type="dag", max_cond_vars=5, significance_level=0.01)
    nx.draw_circular(dag, with_labels=True)
    plt.show()
```

### Expected behaviour
`node_pairs` in `PC` are desired to be iterated several times.

### Actual behaviour
In the first step's loop, everything is correct.
Screenshot from 2020-12-08 09-01-02

In the second loop, `node_loop` will not yield, thus, step 2, 3, and 4 are all skipped in all cases. Some bugs inside of loop, too.
Screenshot from 2020-12-08 09-02-40

I would like to open a pull request to fix this. Thanks!

yzh211:

- Fixes #1359 

- Change `node_pairs` from iterator to list, so it can be looped multiple times.
- Change `pdag.remove()` to `pdag.remove_edge()`.
- Fix algorithm in Step 3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 pgmpy
 =====
-[![Build Status](https://travis-ci.org/pgmpy/pgmpy.svg?style=flat)](https://travis-ci.org/pgmpy/pgmpy)
+[![Build Status](https://api.travis-ci.com/pgmpy/pgmpy.svg?branch=dev)](https://travis-ci.com/pgmpy/pgmpy)
 [![Appveyor](https://ci.appveyor.com/api/projects/status/github/pgmpy/pgmpy?branch=dev)](https://www.appveyor.com/)
 [![codecov](https://codecov.io/gh/pgmpy/pgmpy/branch/dev/graph/badge.svg)](https://codecov.io/gh/pgmpy/pgmpy)
 [![Codacy Badge](https://api.codacy.com/project/badge/Grade/78a8256c90654c6892627f6d8bbcea14)](https://www.codacy.com/gh/pgmpy/pgmpy?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=pgmpy/pgmpy&amp;utm_campaign=Badge_Grade)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,7 +7,7 @@ environment:
 
 install:
   - "%PYTHON%\\python.exe -m pip install pip -U"
-  - "%PYTHON%\\python.exe -m pip install torch==1.4.0+cpu -f https://download.pytorch.org/whl/torch_stable.html"
+  - "%PYTHON%\\python.exe -m pip install torch==1.7.0+cpu -f https://download.pytorch.org/whl/torch_stable.html"
   - "%PYTHON%\\python.exe -m pip install -r requirements.txt -U"
   - "%PYTHON%\\python.exe -m pip install -e ."
 
--- a/pgmpy/estimators/PC.py
+++ b/pgmpy/estimators/PC.py
@@ -448,8 +448,8 @@ class PC(StructureEstimator):
             # 3) for each X-Y with a directed path from X to Y, orient edges to X->Y
             for pair in node_pairs:
                 X, Y = pair 
-                if pdag.has_edge(Y, X) and pdag.has_edge(X, Y) and
-                 len(list(nx.all_simple_paths(pdag, X, Y))) > 1:
+                if (pdag.has_edge(Y, X) and pdag.has_edge(X, Y) and 
+                    len(list(nx.all_simple_paths(pdag, X, Y))) > 1):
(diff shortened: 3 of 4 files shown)


ankurankan:
@yzh211 Thanks a lot for the PR. Could you please also add some tests covering this fix?

ankurankan:
@yzh211 Also, travis is complaining about the formatting: https://travis-ci.org/github/pgmpy/pgmpy/jobs/748475706. Running black (https://github.com/psf/black) on the file would fix it.

yzh211:
@ankurankan Hi, thanks for the information! My bad for missing the tests! More bugs are fixed in the `skeleton_to_pdag`:

1. In step 2, it should specify that no direct path from X to Y. 
2. I replaced `combimations` to `permutations`. Because in some steps, it is directional sensitive for node pair (X, Y). Using combinations cannot cover all cases (X->Y and Y->X), thus will yield wrong results.
3. Tests were added for steps 2, 3, and 4.

ankurankan:
@yzh211 Thanks for adding the tests and travis is also passing now. But I have a couple of questions about the test cases. 

1. In the first test case (https://github.com/pgmpy/pgmpy/pull/1360/files#diff-f54dc3d67c8bb46a60471463019dc1fc1e7e80f0068fbfd53ec8e8db45923304R169), the first separating set ( `frozenset({"A", "B"}): ("D",)`) doesn't seem to comply with the oriented model because if we condition the pdag on `D`, it would open the collider at `C` making `A` and `B` d-connected, which contradicts the separating set. 

2. In the second test case (https://github.com/pgmpy/pgmpy/pull/1360/files#diff-f54dc3d67c8bb46a60471463019dc1fc1e7e80f0068fbfd53ec8e8db45923304R180), I am seeing a similar issue. The separating set (`frozenset({"C", "D"}): ("B",)`) is getting violated as conditioning on `B` opens the path ( C <- A -> B <- D) making `C` and `D` d-connected, which again violates the separating set.  

The third test case seems to comply with the separating sets though. Am I not understanding something correctly here?

yzh211:
@ankurankan Thanks for the questions!

1. I think you are correct on this one. I have made the corresponding change, please check if it is valid now. 

2. For the second case, there is still a block `D->B->C` as conditioning on `B`. Based on my understanding, d-separation is valid if *any*  of the three rules holds. https://en.wikipedia.org/wiki/Bayesian_network

In fact, the second test case encompasses steps 2, 3 in the algorithm, so the first testing case maybe not necessary.

Please let me know your idea?

ankurankan:
@yzh211 Thanks for checking again. But I think there is still some bug in the code because in the first case (with the separating sets that you had specified before) I wouldn't expect it to return the model that it did before. For the second case, yes you are right that `D->B->C` is blocked when conditioned on `B` but it also opens the other path `C<-A->B<-D`. And for two variables to be d-separated, all the paths between them need to be blocked.

Also, I haven't looked at your code yet, so I will also give it a try and see if I can find where the bug is.

yzh211:
> @yzh211 Thanks for checking again. But I think there is still some bug in the code because in the first case (with the separating sets that you had specified before) I wouldn't expect it to return the model that it did before. For the second case, yes you are right that `D->B->C` is blocked when conditioned on `B` but it also opens the other path `C<-A->B<-D`. And for two variables to be d-separated, all the paths between them need to be blocked.
> 
> Also, I haven't looked at your code yet, so I will also give it a try and see if I can find where the bug is.

Thanks!

ankurankan:
@yzh211 The problem seemed to be in the test case, I have fixed it now. I also tested the new changes with the alarm model, and it increases its accuracy. Thanks for the PR.

PR #1375, opened 2021-01-29
Added a class that inherits from BDeuScore so that you can use the Ma…

### LINKED ISSUE #1337: Add MU+BDs for Bayesian network structure learning
Add in the ability to use the Marginal Uniform and Bayesian Dirichlet sparse prior from http://proceedings.mlr.press/v52/scutari16.pdf for structure learning in Bayesian networks. 

I think this will be fairly straightforward by adding in a class that inherits from the BDeuScore class in `pgmpy.estimators.StructureScore`.

mjmt05:
Marginal Uniform and Bayesian Dirichlet sparse prior for structure learning per Issue #1337.

- Fixes #1337 

- Added a class BDsScore (Bayesian Dirichlet Sparse Score) in Structure Score which inherits off of BDeuScore
- Added a test for the new class


ankurankan:
@mjmt05 Sorry for so much delay on this. I went through the PR but if I am understanding things correctly, there seems to be a problem with it right now. So, if you look at the `HillClimbSearch` class for structure learning, it uses the `local_score` method from the scoring classes to compute the change in score for any modification that it does in the network. But with the current implementation of BDsScore, the structure learning will get local scores same as BDeUScore as the `local_score` method doesn't use the `structure_prior`. I think there are multiple ways to fix it. The easiest one would be to implement `local_score` for BDsScore as well which computes the change in structure_prior as well. What do you think?

mjmt05:
Oh my bad, for some reason I thought it was calling the score method from the base StructureScore class. That sounds good if somewhat repetitive. In looking at this I also just noticed that we'll have to add in bdsscore to the list of supported_methods in HillClimbSearch.py

ankurankan:
@mjmt05 From my understanding the code shouldn't be repetitive. We could call the `BDeUScore.local_score` from `BDsScore.local_score` and then manipulate the returned score to also include the structure prior?

mjmt05:
Oh I see what you're saying, yep makes sense!

mjmt05:
Sorry for taking so long to get to this. In order to compute the local structure prior the function needs to know about the set of all possible parents for the node as the probability of an edge is different than the probability of the absence of an edge. Alternatively, if it knows about the proposed change to the dag then the prior ratio for the proposed change is trivial - I implemented the latter.

If you would prefer StructureScore class to not be aware of the proposed changes to the DAG then an alternative is to pass the model into the local_score function so that it can determine the number of possible parents??

ankurankan:
@mjmt05 I think the current implementation looks good. Thanks a lot for the PR.

Issue #1387, opened 2021-03-11
Bug: structure-learned Bayesian Network samples incorrect frequencies


mbaak:
### Subject of the issue
When one learns the structure of a Bayesian Network from data and fits the model parameters, then
the sampled frequencies of data points generated using forward_sampling are incorrect. The observed 
frequencies deviate significantly from the expected frequencies of a model. I see this with BNs constructed 
using HillClimbSearch and TreeSearch (and it may hold for more search algorithms). One can verify this 
behaviour with the new BayesianModelProbability class.
I do *not* see this discrepancy when forward sampling from a manually built ("theoretically known") 
Bayesian Network, where the sampled frequencies do agree with the expectations.

### Your environment
* pgmpy commit: 85d900cbaf25c141191973e2175ef92ebe408c23
* Python version: 3.8.5
* Operating System: MacOS

### Expected & Actual behaviour
See descriptions above.

### Steps to reproduce 

import numpy as np
import pandas as pd

import pgmpy
from pgmpy.models import BayesianModel
from pgmpy.utils import get_example_model
from pgmpy.sampling import BayesianModelSampling
from pgmpy.inference import BayesianModelProbability

# we want to compare *generated* frequencies with *true* probabilities from a model

# 1. starting off from an manually constructed model (alarm)

model = get_example_model('alarm')
samples = BayesianModelSampling(model).forward_sample(size=int(5e5))

# let's look at highest probabilities only
bmprob = BayesianModelProbability(model)
logp = bmprob.log_probability(samples)
p = np.exp(logp)
sortp = np.unique(p)[::-1]

# convert to numbers so np.unique works
df = pd.DataFrame(index = samples.index)
for c in samples.columns:
    codes, uniques = pd.factorize(samples[c])
    df[c] = codes

u, c = np.unique(df.values, return_counts=True, axis=0)
idc = np.argsort(c)[::-1]

# observed frequencies of model
sortc = c[idc] / 5e5
print(sortc[:10])

# expected frequencies of model
print(sortp[:10])

# ^^^ as one can see, in this case the sampled frequencies match well with the probabilities from the model

## 2. now recontruct a model from a dataset
# (we downsample to increase speed of HillClimbSearch)

sample = df.sample(10000)

from pgmpy.estimators import HillClimbSearch, ExhaustiveSearch, PC, TreeSearch

if True:
    est = TreeSearch(sample, root_node=df.columns[0])
[...]

PR #1388, opened 2021-03-13
Bug fix: fix sampling bias bug caused by reusing random seed

mbaak:
Fix sampling bias bug from incorrectly reuing the seed for every node.

When passing to sample_discrete the same seed it is recycled for each node,
leading to (large!) sampling bias.
Now set the seed at the start of the sampling function, and do not pass
to sample_discrete() for each node.

https://github.com/pgmpy/pgmpy/issues/1387

Move numpy.random.seed(seed) to the start of each sampling function, don't pass the 
seed to sample_discrete().
- 
-
-

### DIFF
--- a/pgmpy/sampling/Sampling.py
+++ b/pgmpy/sampling/Sampling.py
@@ -74,6 +74,9 @@ class BayesianModelSampling(BayesianModelInference):
         else:
             pbar = self.topological_order
 
+        if seed is not None:
+            np.random.seed(seed)
+
         for node in pbar:
             if show_progress and SHOW_PROGRESS:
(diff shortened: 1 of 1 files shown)


ankurankan:
@mbaak Thanks a lot for this fix. Seems like a really hard one to track down.

mbaak:
@ankurankan It took me a while, yes. :)

I would not have found (& fixed) this without the new log_probability() function for BayesianNetwork objects, so I guess that shows its value. Btw, you were thinking of the right place for that. My suggestion would be to move it to the BayesianNetwork class.

PR #1470, opened 2021-10-05
Add model.states property to BayesianNetwork class

### LINKED ISSUE #1466: Adding a dictionary of node:states to models
I often find myself needing to iterate over the states of a node.
For this, it would be useful to have a dictionary that maps the names of the nodes of a model to 

I have resorted to the workaround of initializing this myself when loading a model.

```
model = get_example_model('asia')

def get_states(model, node):
  phi = model.get_cpds(node)
  n_states = phi.get_cardinality([node])[node]
  states = [phi.get_state_names(node, i) for i in range(n_states)]
  return states

model.states = {node : get_states(model, node) for node in model.nodes}
model.states
```

1. Is this the easiest way to get access to the states?
2. Can we add a `@property` method called `states` to the base `BayesianModel` class reproducing this functionality? I'd be happy to do the PR myself if this is agreed to be a good approach.

Jsevillamol:
#1466

- Fixes #1466 

- Added a property method `states` to the BayesianNetwork class

### DIFF
--- a/pgmpy/models/BayesianNetwork.py
+++ b/pgmpy/models/BayesianNetwork.py
@@ -371,6 +371,22 @@ class BayesianNetwork(DAG):
                 cardinalities[cpd.variable] = cpd.cardinality[0]
             return cardinalities
 
+    @property
+    def states(self):
+        """
+        Returns a dictionary mapping each node to its list of possible states.
+
+        Returns
+        -------
+        state_dict: dict
+            Dictionary of nodes to possible states
+        """
+        state_names_list = [cpd.state_names for cpd in self.cpds]
+        state_dict = {
+            node: states for d in state_names_list for node, states in d.items()
+        }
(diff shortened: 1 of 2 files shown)


ankurankan:
@Jsevillamol Thanks for the PR. Black seems to be complaining about your changes. If you run `black .` in pgmpy home dir and commit the changes, it should fix it. Also, I would suggest a couple of other things:
1. The docstring needs to be formatted according to the numpydoc style, otherwise, our documentation generator would fail. You can look at the other method docstrings for examples.
2. Could you maybe add a test for the new method?

Jsevillamol:
Thank you for helping me through this @ankurankan !

I have amended the commit as you suggested.

ankurankan:
@Jsevillamol Thanks again for the PR. Everything looks good.

PR #1471, opened 2021-10-05
Functional Loops over Time

### LINKED ISSUE #1340: DynamicBayesianNetwork limitations
### Subject of the issue
From what I've understood, in `pgmpy` it's possible to define `DynamicBayesianNetwork`'s edges only in the first time slice (as in #714); so, to define a DBN as the following:

immagine

I would need to add the edges in the first time slice, like this:

immagine

However, by doing so I would create a loop, and thus `pgmpy` throws an error.

In my opinion, forcing to define edges in the first time slice is very limiting: functional loops over time are [a desired property](https://synergy.st-andrews.ac.uk/vannesmithlab/research/bayesian-networks/) of DBN, but in this way we lose such property.

FelipeGiro:

- Fixes #1340

- substitute the creation of the Bayesian model creation to a build-in `DynamicBayesianNetwork `function `get_constant_bn`.
- As `get_constant_bn `already have the CPDs, removed `add_cpds`

### NOTE:
Code is not fully functional. The line https://github.com/pgmpy/pgmpy/blob/c7201208f739b7757077e90404f4a2432e5cc679/pgmpy/inference/dbn_inference.py#L70 add the edges in a different way, resulting in the following nodes:

`NodeView(('A_0', 'A_1', 'B_1', 'C_0', 'B_0', 'C_1', <DynamicNode(A, 0) at 0x22298476f08>, <DynamicNode(B, 0) at 0x2229847f0c8>, <DynamicNode(C, 0) at 0x2229847f108>))`

As you can see, the three last nodes were already added. I don't know the reason they are represented differently.

The question is: should we change how the node is represented/compared or should we come back to `base.py` and think another solution other than `get_constant_bn `

### DIFF
--- a/pgmpy/inference/base.py
+++ b/pgmpy/inference/base.py
@@ -101,8 +101,7 @@ class Inference(object):
                     self.factors[var].append(factor)
 
         elif isinstance(self.model, DynamicBayesianNetwork):
-            self.start_bayesian_model = BayesianNetwork(self.model.get_intra_edges(0))
-            self.start_bayesian_model.add_cpds(*self.model.get_cpds(time_slice=0))
+            self.start_bayesian_model = self.model.get_constant_bn()
             cpd_inter = [
                 self.model.get_cpds(node) for node in self.model.get_interface_nodes(1)
             ]


ankurankan:
@FelipeGiro Sorry for the late reply. I am not sure what is the best way here. I remember that I had tried to change the `get_constant_bn` method to return a `DynamicNode` object for the nodes, but that caused a lot of trouble in integrating it with parameter learning code. And I couldn't find a nice way (without a lot of changes) to make it work with parameter learning and hence resorted to changing the node names with `_`. I am not sure if inference would work fine with the DynamicNode objects, but if it does, I would say we can add an additional argument to `get_constant_bn` to return either  `_` nodes or DynamicNode. Else, we can completely switch to the `_` node names.

PR #1528, opened 2022-05-11
Speed up the __hash__ method of DiscreteFactor

### LINKED ISSUE #1525: It seems spend too much time in VariableElimination._get_working_factors method when use bayesian newtork to predict
I build a `BayesianNetwork` and let i fit dataset to learn parameters.
And then I use it to `predict` and find that it is too slow because of spending too much time in `VariableElimination._get_working_factors()` method to build the `working_factors` variable (because there is a node have many edges from others in my network):

```python
working_factors = {
    node: {(factor, None) for factor in self.factors[node]}
    for node in self.factors
}
```

It seems take too much time in build a `set` about tuples of `DiscreteFactor` objects. Are there any methods to improve? Thank you.

reeingal:
The `__hash__()` method of `DiscreteFactor` use `str(phi.values)` to
construct the hash value.

The `phi.values` is a `numpy.ndarray` object, and `str(phi.values)`
will spend too much time when the shape of `numpy.ndarray` is huge.

I find that `hash(phi.values.data.tobytes())` is much faster,
so I suggest to use `str(hash(phi.values.data.tobytes()))` to construct
the hash value, which will speed up the execute time like `predict()` of
`BayesianNetwork` for large networks.

- Fixes #1525

- Modify the `__hash__()` method of `DiscreteFactor` to speed up execute time.

### DIFF
--- a/pgmpy/factors/discrete/DiscreteFactor.py
+++ b/pgmpy/factors/discrete/DiscreteFactor.py
@@ -1026,7 +1026,7 @@ class DiscreteFactor(BaseFactor, StateNameMixin):
             phi.values = phi.values.swapaxes(axis, exchange_index)
         return hash(
             str(sorted_var_hashes)
-            + str(hash(phi.values.data.tobytes()))
-            + str(phi.cardinality)
+            + str(hash(phi.values.tobytes()))
+            + str(hash(phi.cardinality.tobytes()))
             + str(state_names_hash)
         )
--- a/pgmpy/factors/discrete/DiscreteFactor.py
+++ b/pgmpy/factors/discrete/DiscreteFactor.py
@@ -1026,7 +1026,7 @@ class DiscreteFactor(BaseFactor, StateNameMixin):
             phi.values = phi.values.swapaxes(axis, exchange_index)
         return hash(
             str(sorted_var_hashes)
-            + str(phi.values)
+            + str(hash(phi.values.data.tobytes()))
             + str(phi.cardinality)
             + str(state_names_hash)
         )


pgmpy/factors/discrete/DiscreteFactor.py
```diff
@@ -1026,7 +1026,7 @@ def __hash__(self):
             phi.values = phi.values.swapaxes(axis, exchange_index)
         return hash(
             str(sorted_var_hashes)
-            + str(phi.values)
+            + str(hash(phi.values.data.tobytes()))
```
ankurankan: @reeingal `phi.values.tobytes()` seems to be slightly faster than `phi.values.data.tobytes()`.
```python
In [20]: %timeit data.data.tobytes()
22.4 µs ± 38.7 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [21]: %timeit data.tobytes()
21.4 µs ± 32.9 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

In [22]: data.tobytes() == data.data.tobytes()
Out[22]: True
```

pgmpy/factors/discrete/DiscreteFactor.py
```diff
@@ -1026,7 +1026,7 @@ def __hash__(self):
...
             + str(phi.cardinality)
```
ankurankan: Can also convert `phi.cardinality` to bytes.

ankurankan (changes requested):

pgmpy/factors/discrete/DiscreteFactor.py
reeingal: I am a new Github user, although signed up early. I modify the code according to your suggestion, and commit again, is this right?

ankurankan: Yes, this is good.

ankurankan (approved):

PR #1529, opened 2022-05-11
Speed up the learning time of EM for large dataset

### LINKED ISSUE #1526: The _compute_weights() method of ExpectationMaximization executes too long in large dataset
Hello, I use the `ExpectationMaximization` estimator to learn parameters of my `BayesianNetwork`. But it spend too much time on  `ExpectationMaximization._compute_weights()` method, because the method traverses all unique data in dataset:

```python
# pgmpy/estimators/EM.py ExpectationMaximization _compute_weights()
for i in range(data_unique.shape[0]):
    v = list(product(*[range(card) for card in latent_card.values()]))
    latent_combinations = np.array(v, dtype=int)
    df = data_unique.iloc[[i] * latent_combinations.shape[0]].reset_index(
        drop=True
    )
    for index, latent_var in enumerate(latent_card.keys()):
        df[latent_var] = latent_combinations[:, index]

    weights = df.apply(lambda t: self._get_likelihood(dict(t)), axis=1)
    df["_weight"] = (weights / weights.sum()) * n_counts[
        tuple(data_unique.iloc[i])
    ]
    cache.append(df)
```

I am wondering whether it can change to use `joblib.Parallel` to improve the speed. But I am not sure whether the `for` loop would use the result calculated by last loop. I am not familiar with the algorithm. Thank you.

reeingal:
The `_compute_weights()` method of `ExpectationMaximization` will
traverse all unique data in the dataset use `for` loop.

if the count of unique data is large, the `_compute_weights()` method
will take too much time in `for` loop.

I make a try to use `joblib.Parallel` for paralleling, and find that it
saves 70% to 80% time on my dataset with 9994 unique records and
10 columns. The number of latent variables in my test are 3.

- Fixes #1526

- change the `for` loop in  `_compute_weights()` method of `ExpectationMaximization` to parallel running expected to improve executing time.

### DIFF
--- a/pgmpy/estimators/EM.py
+++ b/pgmpy/estimators/EM.py
@@ -100,7 +100,7 @@ class ExpectationMaximization(ParameterEstimator):
 
         return pd.concat(cache, copy=False)
 
-    def _compute_weights(self, latent_card):
+    def _compute_weights(self, n_jobs, latent_card):
         """
         For each data point, creates extra data points for each possible combination
         of states of latent variables and assigns weights to each of them.
@@ -111,7 +111,7 @@ class ExpectationMaximization(ParameterEstimator):
 
         batch_size = 1000
 
-        cache = Parallel(n_jobs=-1)(
+        cache = Parallel(n_jobs=n_jobs)(
             delayed(self._parallel_compute_weights)(
                 data_unique, latent_card, n_counts, i, batch_size
             )
@@ -159,6 +159,7 @@ class ExpectationMaximization(ParameterEstimator):
 
         n_jobs: int (default: -1)
             Number of jobs to run in parallel. Default: -1 uses all the processors.
+            Suggest to use n_jobs=1 when dataset size is less than 1000.
 
         seed: int
             The random seed to use for generating the intial values.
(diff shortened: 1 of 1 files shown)


ankurankan:
@reeingal Could you run `black .` in the pgmpy directory to fix the formatting and commit the changes? The tests are failing because of that.

pgmpy/estimators/EM.py
```diff
@@ -105,6 +100,26 @@ def _compute_weights(self, latent_card):
...
+        For each data point, creates extra data points for each possible combination
+        of states of latent variables and assigns weights to each of them.
+        """
+
+        data_unique = self.data.drop_duplicates()
+        n_counts = self.data.groupby(list(self.data.columns)).size().to_dict()
+
+        batch_size = 1000
+
+        cache = Parallel(n_jobs=-1)(
```
ankurankan: @reeingal `n_jobs` here should be taken as a method argument. The `get_parameters` method has a `n_jobs` argument which it can pass to `_compute_weights` when called.

pgmpy/estimators/EM.py
```diff
@@ -105,6 +100,26 @@ def _compute_weights(self, latent_card):
...
+    def _compute_weights(self, latent_card):
+        """
+        For each data point, creates extra data points for each possible combination
+        of states of latent variables and assigns weights to each of them.
+        """
+
+        data_unique = self.data.drop_duplicates()
+        n_counts = self.data.groupby(list(self.data.columns)).size().to_dict()
+
+        batch_size = 1000
```
ankurankan: I would suggest having `batch_size` also as an argument with the default value of 1000.

ankurankan (changes requested):
@reeingal Seems to be working fine for me on realistic datasets. I tested it on the alarm and munin1 network: https://www.bnlearn.com/bnrepository/ with up to 1e5 samples and seems to be much faster in most of the cases. Though the run time is higher when the dataset size is less than or around 1e3. I would suggest adding a note in the `get_parameters` docstring to use n_jobs=1 when dataset size is less than 1000.

PR #1547, opened 2022-07-05
Fix year in release date

naveenrajm7:


ankurankan:
@naveenrajm7 Thanks for the fix 😄

PR #1605, opened 2023-02-03
to_joint_gaussian bug fixed

JonasLange:
The to_joint_gaussian() function in pgmpy/models/LinearGaussianBayesianNetwork.py did not work properly. Try the example provided in lines 144-164 of the [LinearGaussianBayesianNetwork class](https://github.com/pgmpy/pgmpy/blob/10830bc13f2db220cbec8c3714ea352c3307e6d9/pgmpy/models/LinearGaussianBayesianNetwork.py). The output does not match the expected output (i.e. mean and covariance are wrong). The changes below should fix this bug.


ankurankan:
@JonasLange Thanks for the PR.

Issue #1633, opened 2023-05-10
default attribute in Probability block


White-On:
### The default attribute in a Probability block in a BIF file making the file unreadable 
I'm currently working on several libraries working around a Bayesian network to compare the performance. But, some library generates BIF files with a default attribute (which seems to be a valid attribute according to the reference you give here in the reference section: https://pgmpy.org/readwrite/bif.html

*Attributes
Several attributes are defined at this point: property, type, table,
default and entry attributes (the entry attribute is not associated
with any keyword). 
The attribute property can appear in all types of blocks. A property
is just a string of arbitrary text to be associated with a
block.*

So I'm trying to make the default attribute as "table".  In pgmpy/readwrite/BIF. py, a RegExp seems to be the key to make that possible **at line 282**. The original one was ```.*\\n[] *table .*\n.*``` and the first thing I've noticed is ```\\n``` may not lead to expected result and may need to be ```\n```.

Another problem, If I change the word *table* with *default*, the file is correctly read without issue. But, as I want both *table* and *default* I end up trying those two RegExp: ```.*\n[] * (table|default) .*\n.*``` and  ```.*\n[] *table .*\n.*|.*\n[] *default .*\n.*``` 

Both will result in a correct reading of the BIF file, but went I try to get access to the model, I just end up with a ValueError.

In conclusion, I was wondering if there is a way for the default attribute to be considered as equal as the table attribute.

ankurankan:
@White-On I think that should be possible. I can't think of a reason why get_model is throwing an error because of the change that you have described. Would it be possible to somehow share the changes that you have made (maybe a patch file or pushing to repo fork) so that I can have a look at the error?

White-On:
This is the piece of code I've been testing on :
![Code_NAjq2rWLHk](https://github.com/pgmpy/pgmpy/assets/74248238/5b7583eb-6370-4e01-a110-3f2e86437972)

If nothing is done to the BIF reading code, the current error is the following: 
![Code_5nkgdj5cCI](https://github.com/pgmpy/pgmpy/assets/74248238/c227da4c-a883-46d8-bf41-be87fc1b94f8)

The *Bntest.bif* is the following: 
[BnTest.zip](https://github.com/pgmpy/pgmpy/files/11452533/BnTest.zip)
It is generated by a other library called **PyAgrum** and contain *default* attribute

as I replace all those *default* attribute by *table* the result is as follow :
[...]

PR #1636, opened 2023-05-15
Adding a the default attribut and binding it to table attribut

White-On:
### Linked to the issue [issue #1633]

The *default* attribut can now be read in a BIF file in the probability block. It is consider as a *table* attribut

- changing the RegExp reading table and default
- adding default to the vocabulary 
- mixing table and default values inside the test


ankurankan:
@White-On Thanks for the PR. For the test, I would suggest adding a model with `default` as the keyword instead of `table` instead of modifying the current model. This is just to make sure that our code works in both cases. And the test could be something simple, like comparing that both `default` and `table` models are equal.

ankurankan:
@White-On Thanks. Looks good now.

PR #1740, opened 2024-03-01
Feature/Faster Belief Propagation with message parsing & recursion

### LINKED ISSUE #1739: Feature: belief propagation with message parsing on factor graphs
Hi, I'm currently extending this library with belief propagation on factor graphs, using message parsing. This complements the currently implemented belief propagation which uses variable elimination.

So far it's been 10x faster for my factor graph with a dozen nodes.

I'll push a PR when done. It might help with issues: #383, #1571 , #1020

tristantreb:
### Summary
The currently implemented BP is great but I faced limitations when increasing either the number of nodes (>18) or when increasing the number of variables being queried at once (up to 6). My model's variables have cardinalities ranging in 40-120. Under some settings, the inference was either crashing or taking >20s using a Bayes net model, and even longer when using a factor graph.

I suggest this BP algorithm which uses message-parsing and recursion. The algorithm recursively parses the graph to compute and propagate the messages towards the queried variable. Currently, it doesn't allow to return the joint density

On my model, this BP is more than 10x faster. Where using the existing BP on a Bayes net took >20s, my implementation took <1s.

Let me know what you think

- Fixes #1739 

- Adds a BeliefPropagationWithMessageParsing class
- Extends the FactorGraph class with functions used for the message parsing algorithm
- Add exhaustive testing

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,6 @@ nosetests.xml
 # Auto examples generated
 docs/auto_examples/*
 docs/examples/*
+
+# Macos files
+.DS_Store
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -1242,20 +1242,71 @@ class BeliefPropagationWithMessageParsing(Inference):
     """
     Class for performing efficient inference using Belief Propagation method on factor graphs.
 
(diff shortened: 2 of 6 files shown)


ankurankan:
@tristantreb Thanks for the PR. Looks good to me. In future, I would ideally like to merge both the BP algorithm, to have a single method that works for all model types.

PR #1745, opened 2024-03-21
Add torch tests to MarginalEstimator and MirrorDescentEstimator

jakee417:

- Adds tests for the `torch.Tensor` backend.
- Adds a compat_fn for torch/np exp.
- Small typos
- Adds documentation for `FactorDict.from_dict` and `ClusterGraph.clique_beliefs`


pgmpy/estimators/MirrorDescentEstimator.py
```diff
@@ -45,7 +46,7 @@ def _calibrate(self, theta, n):
         log_z = logsumexp(mu[clique].values)
         for clique in cliques:
             mu[clique] += np.log(n) - log_z
```
ankurankan, line 48: @jakee417 This should be `mu[clique].values` ?

pgmpy/estimators/MirrorDescentEstimator.py
```diff
@@ -45,7 +46,7 @@ def _calibrate(self, theta, n):
         log_z = logsumexp(mu[clique].values)
```
ankurankan, line 46: This would need a compat_fn:
```python
def logsumexp(arr):                                                                                                                                                                                                                                                                                                                                     
     if isinstance(arr, np.ndarray):                                                                                                                                                                                                                                                                                                                     
         return special.logsumexp(arr)                                                                                                                                                                                                                                                                                                                   
     else:                                                                                                                                                                                                                                                                                                                                               
         return torch.logsumexp(arr, 0)   
```

ankurankan (changes requested):
Thanks for the PR. Have added a couple of minor comments.

pgmpy/estimators/MirrorDescentEstimator.py
jakee417: `mu[clique]` is a `DiscreteFactor` which implements addition by a `float`.

pgmpy/estimators/MirrorDescentEstimator.py
jakee417: Similar story here, `logsumexp(mu[clique].values)` reduces to a float which is compatible with `mu[clique] += np.log(n) - log_z`. `scipy.special.logsumexp` is also compatible with `torch.Tensor`:
```python
import torch
from scipy.special import logsumexp

logsumexp(torch.Tensor([10, 10, 10]))
```
When `axis=None`, this reduces to a float.

jakee417:
@ankurankan Added an explicit test for a junction tree with multiple cliques to affirm any doubts about the above comments. Also refactored the torch tests so that they can share code, a similar thing can be applied to `test_MaximumLikelihood.py`

jakee417 (commented):
@ankurankan I am not sure how this review feature works, but have addressed your previous comments

ankurankan (approved):

ankurankan:
@jakee417 Sorry for the delay. I think it looks good. If you are using any references for the implementation could you please add that in References section of the docstring (something like this: https://github.com/pgmpy/pgmpy/blob/cc2260e50712c72d71623474483bef17e15aa4a4/pgmpy/estimators/PC.py#L45)

PR #1788, opened 2024-06-19
predict method for Linear Gaussian BN

ankurankan:

- Fixes #

- 
-
-

### DIFF
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -294,10 +294,44 @@ class LinearGaussianBayesianNetwork(BayesianNetwork):
     def fit(self, data, method="mle"):
         """
         Estimates the parameters of the model using the given `data`.
+
+        Parameters
+        ----------
+        data: pd.DataFrame
+            A pandas DataFrame with the data to which to fit the model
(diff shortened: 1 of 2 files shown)


r-gheda:
Hi all! I was wondering why is pgmpy not using [canonical form representation](https://datascience.oneoffcoder.com/gaussian-network-inference.html) as it seems quite convenient for operations over Gaussian networks.
Is it planned to implement that type of representation as well in the future?

ankurankan:
@r-gheda We do have an implementation of canonical form representation and operations on them: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/factors/distributions/CanonicalDistribution.py. However, I am trying to change/reimplement the API for the Linear Gaussian BNs, hence some of these earlier implementations need to be reintegrated.

My understanding is that the covariance form was the more straightforward option for computing the posterior distributions (this PR). Do you think the canonical form is more suitable?

r-gheda:
AFAIK, canonical form has a slightly more complex marginalization and reduction methods than covariance form but has far easier factor product and division implementations.

I think this is one of the reasons why the canonical form tend to be preferred in literature when used for more complex applications such as CLG inference and Junction Tree-based methods (see Daphne Koller's book for instance).

PR #1830, opened 2024-10-12
Adding get_random() module to LinearGaussianBayesianNetwork and overriding the BayesianNetwork module.

### LINKED ISSUE #1795: Using get_random with LinearGaussianBayesianNetwork returns incorrect model object
```python
>>> from pgmpy.models import LinearGaussianBayesianNetwork
>>> random_model = LinearGaussianBayesianNetwork.get_random(n_nodes=10, edge_prob=0.3, latents=False)
>>> random_model
<pgmpy.models.BayesianNetwork.BayesianNetwork at 0x72cd227876d0>
```

Ideally, this should return an instance of `pgmpy.models.LinearGaussianBayesianNetwork` as the get_random was called for that class.

kollisaisiddartha:

- Fixes #1795 

- Added get_random() function to LinearGaussianBayesianNetwork.py
- Added unit test for get_random module in test_LinearGaussianBayesianNetwork.py file
-

### DIFF
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -432,8 +432,10 @@ class LinearGaussianBayesianNetwork(BayesianNetwork):
         raise NotImplementedError(
             "is_imap method has not been implemented for LinearGaussianBayesianNetwork."
         )
-    
-    def get_random(n_nodes=5, edge_prob=0.5, node_names=None, n_states=None, latents=False):
+
+    def get_random(
+        n_nodes=5, edge_prob=0.5, node_names=None, n_states=None, latents=False
+    ):
         """
         Returns a randomly generated Linear Gaussian Bayesian Network on `n_nodes` variables
         with edge probabiliy of `edge_prob` between variables.
@@ -461,5 +463,7 @@ class LinearGaussianBayesianNetwork(BayesianNetwork):
             If True, also creates latent variables.
 
         """
-        base_bn = BayesianNetwork.get_random(n_nodes, edge_prob, node_names, n_states, latents)
+        base_bn = BayesianNetwork.get_random(
+            n_nodes, edge_prob, node_names, n_states, latents
+        )
         return LinearGaussianBayesianNetwork(base_bn)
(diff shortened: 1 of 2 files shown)


pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
```diff
@@ -197,3 +197,13 @@ def test_get_random_cpds(self):
 
     def tearDown(self):
         del self.model, self.cpd1, self.cpd2, self.cpd3
+
+    def test_get_random(self):
```
ankurankan, line 201: @kollisaisiddartha I would suggest moving this method above the `tearDown` method. It's just a practice we have used in all the test classes.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -432,3 +432,38 @@ def is_imap(self, JPD):
...
+
+        n_states: int or dict (default: None)
+            The number of states of each variable in the form
+            {variable: no_of_states}. If a single value is provided,
+            all nodes will have the same number of states. When None
+            randomly generates the number of states.
+
+        latents: bool (default: False)
+            If True, also creates latent variables.
+
```
ankurankan, line 464: Please also add a `Returns` and `Examples` section here. Have a look at the docstring of other methods for examples.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -432,3 +432,38 @@ def is_imap(self, JPD):
...
+        """
+        base_bn = BayesianNetwork.get_random(
```
ankurankan, line 466: You would need to make a few changes here. Both Bayesian Network and LinearGaussianBayesianNetwork has two parts: 1) The model structure represented using the DAG class, 2) Parameterization (represented using `TabularCPD` for `BayesianNetwork` and `LinearGaussianCPD` for `LinearGaussianBayesianNetwork`).

When you call `BayesianNetwork.get_random`, it generates the model structure by calling the `DAG.get_random` method and then adds randomly generated `TabularCPD` instances to it. We need to do something similar for the `get_random` method here but instead of adding `TabularCPD`s, we need to add instances of `LinearGaussianCPD`. So, the main steps would be:

1. Generate the model structure using `DAG.get_random`.
2. Use the `LinearGaussianBayesianNetwork.get_random_cpds` method to generate random `LinearGaussianCPD`. This would also mean that the arguments required by `LinearGaussianBayesianNetwork.get_random` would be slightly different than `BayesianNetwork.get_random` as `LinearGaussianCPD`s require different arguments. 

I would suggest looking at the `BayesianNetwork.get_random` method to get an example of how everything should work.

ankurankan (changes requested):

ankurankan (commented):
@kollisaisiddartha Thanks for creating the PR. I have added a few comments suggesting a few changes.

PR #1831, opened 2024-10-12
Added LGBN get random module and made few test changes

kollisaisiddartha:
### Your checklist for this pull request
Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [x] Make sure you are making a pull request against the **dev branch** (left side). Also you should start *your branch* off *our dev*.
- [x] Check the commit's or even all commits' message styles matches our requested structure.

### Issue number(s) that this pull request fixes
- Fixes #1795 

### List of changes to the codebase in this pull request
- Added get_random() function to LinearGaussianBayesianNetwork.py
- Added parameters and example for the module
- Added unit test for get_random module in test_LinearGaussianBayesianNetwork.py file

PR #1833, opened 2024-10-13
Changes to get random module in LGBN

### LINKED ISSUE #1795: Using get_random with LinearGaussianBayesianNetwork returns incorrect model object

kollisaisiddartha:

- Fixes #1795 

- Adding get random module
- Adding test case for the module
-

### DIFF
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -3,7 +3,7 @@ import numpy as np
 import pandas as pd
 from sklearn.linear_model import LinearRegression
 
-from pgmpy.models import DAG
+from pgmpy.base import DAG
 from pgmpy.factors.continuous import LinearGaussianCPD
 from pgmpy.factors.distributions import GaussianDistribution
 from pgmpy.global_vars import logger
@@ -505,7 +505,7 @@ class LinearGaussianBayesianNetwork(BayesianNetwork):
         lgbn_model = LinearGaussianBayesianNetwork(dag.edges(), latents=dag.latents)
         lgbn_model.add_nodes_from(dag.nodes())
 
-        cpds = lgbn_model.get_random_cpds(loc, scale, seed)
+        cpds = lgbn_model.get_random_cpds(loc=loc, scale=scale, seed=seed)
 
         lgbn_model.add_cpds(*cpds)
         return lgbn_model
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -3,6 +3,7 @@ import numpy as np
 import pandas as pd
 from sklearn.linear_model import LinearRegression
 
+from pgmpy.models import DAG
 from pgmpy.factors.continuous import LinearGaussianCPD
 from pgmpy.factors.distributions import GaussianDistribution
 from pgmpy.global_vars import logger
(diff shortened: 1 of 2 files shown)


pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -3,6 +3,7 @@
 import pandas as pd
 from sklearn.linear_model import LinearRegression
 
+from pgmpy.models import DAG
```
ankurankan: @kollisaisiddartha I think the issue is here. The import should be `from pgmpy.base import DAG`. Trying to import `pgmpy.models` is running into a circular import.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -432,3 +433,79 @@ def is_imap(self, JPD):
...
+        if node_names is None:
+            node_names = list(range(n_nodes))
+
+        dag = DAG.get_random(
+            n_nodes=n_nodes, edge_prob=edge_prob, node_names=node_names, latents=latents
+        )
+        lgbn_model = LinearGaussianBayesianNetwork(dag.edges(), latents=dag.latents)
+        lgbn_model.add_nodes_from(dag.nodes())
+
+        cpds = lgbn_model.get_random_cpds(loc, scale, seed)
```
ankurankan: I would suggest calling functions with keyword arguments wherever possible. It makes it much easier to read, and also avoids potential errors due to the order of the arguments.

kollisaisiddartha: Made the changes but still the same error.

ankurankan:
@kollisaisiddartha No, this seems to be working fine to me. The error in the github runner now is due to the code formatting: https://github.com/pgmpy/pgmpy/actions/runs/11312865431/job/31460861469?pr=1833

pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
```diff
@@ -195,5 +195,22 @@ def test_get_random_cpds(self):
         cpds = model_lin.get_random_cpds()
         self.assertEqual(len(cpds), len(model.nodes()))
 
+    def test_get_random(self):
+        model = LinearGaussianBayesianNetwork()
+        random_model = model.get_random(
```
ankurankan: Because `get_random` is a static method, you should be able to call it directly using `LinearGaussianBayesianNetwork.get_random` instead of first initializing the object. I would suggest to directly use `LinearGaussianBayesianNetwork.get_random` instead of `model.get_random`

ankurankan (changes requested):
@kollisaisiddartha I think this is mostly ready to be merged. The only thing remaining is a few more tests as we had discussed in the other PR: https://github.com/pgmpy/pgmpy/pull/1831#discussion_r1797693320

pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
kollisaisiddartha: Should i add more models with different values of parameters and check assertion for all of them?

ankurankan: Yes, I would just check whether the arguments work as expected, for example, check if we provide `node_names`, the final model has the same node names. Some of the arguments can be more difficult to test for, for example, loc or sd as they are then used for generating random numbers, and hence would return a different value every time. For those you can specify the seed argument, so that every time the tests are run you get the same value from the random number generator.

kollisaisiddartha: Will there be any floating point difference in CPDs when generated randomly. Should I use np.allclose instead of np.array_equal. And if yes, what could be the tolerance.

pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
```diff
@@ -195,5 +195,38 @@ def test_get_random_cpds(self):
...
+        )
+        self.assertEqual(len(model3.nodes()), 5)
+        self.assertEqual(sorted(model3.nodes()), node_names)
+        self.assertEqual(len(model3.cpds), 5)
+        self.assertIsInstance(
+            model3, LinearGaussianBayesianNetwork, "Incorrect instance"
+        )
+
+        seed = 42
+        model4 = LinearGaussianBayesianNetwork.get_random(n_nodes=5, seed=seed)
```
ankurankan: Hmm, I don't think this would be testable for now. I think you remove the check for now. The DAG.get_random method does not have an argument to specify the seed right now. Because of that model4 and model5 have different structures, resulting in different CPDs. Can you create an issue for adding a seed argument to `DAG.get_random` method?

kollisaisiddartha: Sure i will create an issue for that. Should i remove checks for model4 and model5 and update the PR?

ankurankan: Yes.

ankurankan (approved):

ankurankan:
@kollisaisiddartha I think this looks good now.

PR #1836, opened 2024-10-13
Added Structural Hamming Distance metric

Nimish-4:

- Fixes #

- 
-
-

### DIFF
--- a/pgmpy/metrics/metrics.py
+++ b/pgmpy/metrics/metrics.py
@@ -427,6 +427,9 @@ def SHD(true_model, est_model):
     >>> SHD(dag1, dag2)
     2
     """
+    if set(dag_true.nodes()) != set(dag_est.nodes()):
+        raise ValueError("The graphs must have the same number of nodes.")
+
     dag_true = nx.DiGraph(true_model.edges())
     dag_true.add_nodes_from(true_model.nodes())
     adj_mat_true = nx.adjacency_matrix(dag_true).todense()
@@ -434,9 +437,6 @@ def SHD(true_model, est_model):
     dag_est = nx.DiGraph(est_model.edges())
     dag_est.add_nodes_from(est_model.nodes())
     adj_mat_est = nx.adjacency_matrix(dag_est).todense()
-
-    if set(dag_true.nodes()) != set(dag_est.nodes()):
-        raise ValueError("The graphs must have the same number of nodes.")
 
     shd = np.sum(adj_mat_true != adj_mat_est)
     return shd
--- a/pgmpy/metrics/metrics.py
+++ b/pgmpy/metrics/metrics.py
@@ -1,9 +1,9 @@
 import math
 from itertools import combinations
 
+import networkx as nx
 import numpy as np
 import pandas as pd
-import networkx as nx
 from scipy import stats
 from sklearn.metrics import f1_score
 from tqdm import tqdm
@@ -423,18 +423,16 @@ def SHD(true_model, est_model):
     >>> print(SHD(dag1, dag2))
     >>> 2
     """
-    dag_true = nx.DiGraph(true_model.edges())  
-    dag_true.add_nodes_from(true_model.nodes())  
-    adj_mat_true = nx.adjacency_matrix(dag_true).todense()  
-    
-    dag_est = nx.DiGraph(est_model.edges())  
-    dag_est.add_nodes_from(est_model.nodes()) 
-    adj_mat_est = nx.adjacency_matrix(dag_est).todense()  
+    dag_true = nx.DiGraph(true_model.edges())
+    dag_true.add_nodes_from(true_model.nodes())
+    adj_mat_true = nx.adjacency_matrix(dag_true).todense()
+
+    dag_est = nx.DiGraph(est_model.edges())
+    dag_est.add_nodes_from(est_model.nodes())
+    adj_mat_est = nx.adjacency_matrix(dag_est).todense()
 
     if adj_mat_true.shape != adj_mat_est.shape:
         raise ValueError("The graphs must have the same number of nodes.")
 
     shd = np.sum(adj_mat_true != adj_mat_est)
     return shd
-
-
--- a/pgmpy/metrics/metrics.py
+++ b/pgmpy/metrics/metrics.py
@@ -3,6 +3,7 @@ from itertools import combinations
 
 import numpy as np
 import pandas as pd
+import networkx as nx
 from scipy import stats
 from sklearn.metrics import f1_score
 from tqdm import tqdm
@@ -394,3 +395,46 @@ def fisher_c(model, data, ci_test, show_progress=True):
     C = -2 * np.log(cis.loc[:, "p_value"]).sum()
     p_value = 1 - stats.chi2.cdf(C, df=2 * cis.shape[0])
     return p_value
+
+
+def SHD(true_model, est_model):
+    """
+    Computes the Structural Hamming Distance between `true_model` and `est_model`.
+    SID is defined as total number of basic operations: adding edges, removing
+    edges, and reversing edges required to transform one graph to the other. It
+    is a symmetrical measure.
+    Parameters
+    ----------
+    true_model: pgmpy.base.DAG or pgmpy.base.CPDAG or pgmpy.models.BayesianNetwork
+        The first model to compare.
+    est_model: pgmpy.base.DAG or pgmpy.base.CPDAG or pgmpy.models.BayesianNetwork
+        The first model to compare.
+    Returns
+    -------
+    int or tuple: If both true_model and est_model are DAGs or Bayesian Networks returns
+        an integer. If either or both of them are CPDAG returns a tuple of best and worst
+        SHD values from all possible orientations.
+    Examples
+    --------
+    >>> from pgmpy.metrics import SHD
+    >>> from pgmpy.models.BayesianNetwork import BayesianNetwork
+    >>> dag1 = BayesianNetwork([(1, 2), (2, 3)])
+    >>> dag2 = BayesianNetwork([(2, 1), (2, 3)])
+    >>> print(SHD(dag1, dag2))
+    >>> 2
+    """
+    dag_true = nx.DiGraph(true_model.edges())  
+    dag_true.add_nodes_from(true_model.nodes())  
+    adj_mat_true = nx.adjacency_matrix(dag_true).todense()  
+    
+    dag_est = nx.DiGraph(est_model.edges())  
+    dag_est.add_nodes_from(est_model.nodes()) 
+    adj_mat_est = nx.adjacency_matrix(dag_est).todense()  
+
+    if adj_mat_true.shape != adj_mat_est.shape:
+        raise ValueError("The graphs must have the same number of nodes.")
+
+    shd = np.sum(adj_mat_true != adj_mat_est)
+    return shd
+
+
--- a/pgmpy/metrics/metrics.py
+++ b/pgmpy/metrics/metrics.py
@@ -400,28 +400,32 @@ def fisher_c(model, data, ci_test, show_progress=True):
 def SHD(true_model, est_model):
     """
     Computes the Structural Hamming Distance between `true_model` and `est_model`.
+
     SID is defined as total number of basic operations: adding edges, removing
     edges, and reversing edges required to transform one graph to the other. It
     is a symmetrical measure.
+
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
     C = -2 * np.log(cis.loc[:, "p_value"]).sum()
     p_value = 1 - stats.chi2.cdf(C, df=2 * cis.shape[0])
     return p_value
+
+
+def SHD(true_model, est_model):
+    """
+    Computes the Structural Hamming Distance between `true_model` and `est_model`.
```
ankurankan, line 402: Ideally, I would structure the method explanation as: A single line about what the function does. Leave an empty line after and the give the full description. For example, I would change you current explanation as follows:
```
Computes the Structural Hamming Distance between `true_model` and `est_model`.

SHD is defined as total number of basic operations: adding edges, removing
edges, and reversing edges required to transform one graph to the other. It
is a symmetrical measure.
```

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    SID is defined as total number of basic operations: adding edges, removing
+    edges, and reversing edges required to transform one graph to the other. It
+    is a symmetrical measure.
+    Parameters
```
ankurankan, line 412: There should be an empty line between each section, otherwise it wouldn't render properly on the online documentation. We follow numpydoc style in general, you can have a look at details here: https://numpydoc.readthedocs.io/en/latest/format.html

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    ----------
+    true_model: pgmpy.base.DAG or pgmpy.base.CPDAG or pgmpy.models.BayesianNetwork
+        The first model to compare.
+    est_model: pgmpy.base.DAG or pgmpy.base.CPDAG or pgmpy.models.BayesianNetwork
+        The first model to compare.
+    Returns
```
ankurankan, line 419: Empty line above.

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    -------
+    int or tuple: If both true_model and est_model are DAGs or Bayesian Networks returns
+        an integer. If either or both of them are CPDAG returns a tuple of best and worst
+        SHD values from all possible orientations.
+    Examples
```
ankurankan, line 425: Empty line above.

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    is a symmetrical measure.
+    Parameters
+    ----------
+    true_model: pgmpy.base.DAG or pgmpy.base.CPDAG or pgmpy.models.BayesianNetwork
+        The first model to compare.
+    est_model: pgmpy.base.DAG or pgmpy.base.CPDAG or pgmpy.models.BayesianNetwork
+        The first model to compare.
+    Returns
+    -------
+    int or tuple: If both true_model and est_model are DAGs or Bayesian Networks returns
```
ankurankan: The description of the return value should be on a new line. Ref: https://numpydoc.readthedocs.io/en/latest/format.html#returns

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    """
+    dag_true = nx.DiGraph(true_model.edges())
+    dag_true.add_nodes_from(true_model.nodes())
+    adj_mat_true = nx.adjacency_matrix(dag_true).todense()
+
+    dag_est = nx.DiGraph(est_model.edges())
+    dag_est.add_nodes_from(est_model.nodes())
+    adj_mat_est = nx.adjacency_matrix(dag_est).todense()
+
+    if adj_mat_true.shape != adj_mat_est.shape:
```
ankurankan: This check would only work if the number of variables is different in the two graphs. If they have the same number of variables but with different variable names, this won't be able to catch that. I would rather check something like:  `set(dag_true.nodes()) == set(dag_est.nodes())`

ankurankan (changes requested):

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+        The first model to compare.
+    Returns
+    -------
+    int or tuple: If both true_model and est_model are DAGs or Bayesian Networks returns
+        an integer. If either or both of them are CPDAG returns a tuple of best and worst
+        SHD values from all possible orientations.
+    Examples
+    --------
+    >>> from pgmpy.metrics import SHD
+    >>> from pgmpy.models.BayesianNetwork import BayesianNetwork
```
ankurankan: This should be `from pgmpy.models import BayesianNetwork`.

I would also use the DAG class in the example here. It works similar to Bayesian Netowrk. You can do something like:
```python
from pgmpy.base import DAG

dag1 = DAG[('A', 'B'), ('B', 'C')]
....
```

Nimish-4:
@ankurankan I am working on this!

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,44 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    >>> dag1 = BayesianNetwork([(1, 2), (2, 3)])
+    >>> dag2 = BayesianNetwork([(2, 1), (2, 3)])
+    >>> print(SHD(dag1, dag2))
+    >>> 2
```
ankurankan: You should put `>>>` only in front of inputs and not for outputs.

Nimish-4:
@ankurankan I have reformatted the code. But I guess I messed up a new commit from the dev branch that was updated late in my branch. Is it an issue? What should I do?

ankurankan:
@Nimish-4 It should be fine. You can keep adding your changes to your branch.

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,48 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    """
+    dag_true = nx.DiGraph(true_model.edges())
+    dag_true.add_nodes_from(true_model.nodes())
+    adj_mat_true = nx.adjacency_matrix(dag_true).todense()
+
+    dag_est = nx.DiGraph(est_model.edges())
+    dag_est.add_nodes_from(est_model.nodes())
+    adj_mat_est = nx.adjacency_matrix(dag_est).todense()
+
+    if set(dag_true.nodes()) != set(dag_est.nodes()):
```
ankurankan: I would also move this to the top of this function before creating the adjacency matrix.

Nimish-4:
@ankurankan Okay, on it!

pgmpy/tests/test_metrics/test_metrics.py
```diff
@@ -8,6 +8,7 @@
 from pgmpy.base import DAG
 from pgmpy.estimators.CITests import chi_square
 from pgmpy.metrics import *
```
ankurankan: @Nimish-4 I didn't realize this line was already there. Could you please change it to explicitly import only the functions that we are using in the tests?

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,48 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+
+    dag_true = nx.DiGraph(true_model.edges())
+    dag_true.add_nodes_from(true_model.nodes())
+    adj_mat_true = nx.adjacency_matrix(dag_true).todense()
+
+    dag_est = nx.DiGraph(est_model.edges())
+    dag_est.add_nodes_from(est_model.nodes())
+    adj_mat_est = nx.adjacency_matrix(dag_est).todense()
+
+    shd = np.sum(adj_mat_true != adj_mat_est)
```
ankurankan: I don't think this would be enough to compute the SHD. This would count the SHD as 2 for a single edge reversal. For example, with graph g1 = (A -> B) and g2 = (A <- B), the SHD should be 1 as it is a single operation of edge reversal. But with the current implementation, this would get counted as 2.

ankurankan (changes requested):

Nimish-4:
@ankurankan On it!

pgmpy/metrics/metrics.py
```diff
@@ -425,18 +429,40 @@ def SHD(true_model, est_model):
     >>> dag1 = BayesianNetwork([(1, 2), (2, 3)])
     >>> dag2 = BayesianNetwork([(2, 1), (2, 3)])
     >>> SHD(dag1, dag2)
-    2
+    1
     """
     if set(true_model.nodes()) != set(est_model.nodes()):
         raise ValueError("The graphs must have the same number of nodes.")
```
ankurankan: The error message should not say "the same number of nodes", should rather be "the same nodes" as we are testing for the node names as well.

pgmpy/metrics/metrics.py
```diff
@@ -425,18 +429,40 @@ def SHD(true_model, est_model):
...
 
     dag_est = nx.DiGraph(est_model.edges())
-    dag_est.add_nodes_from(est_model.nodes())
-    adj_mat_est = nx.adjacency_matrix(dag_est).todense()
+    adj_mat_est = nx.adjacency_matrix(dag_est, nodelist=nodes_list).todense()
+
+    shd = 0
+
+    m1 = np.array(adj_mat_true)
+    m2 = np.array(adj_mat_est)
```
ankurankan: You shouldn't need to do this. adj_mat_true and adj_mat_est should already be numpy arrays.

pgmpy/metrics/metrics.py
```diff
@@ -425,18 +429,40 @@ def SHD(true_model, est_model):
...
+
+    s1 = m1 + m1.T
+    s2 = m2 + m2.T
+
+    s1[s1 == 2] = 1
+    s2[s2 == 2] = 1
```
ankurankan: Would we ever have any values of s1 and s2 as 2 in a DAG? Because we can always represent the adjacency matrix of a DAG as an upper triangular matrix, I don't think we would ever get a value of 2 in s1 or s2. Am I missing something here?

pgmpy/metrics/metrics.py
```diff
@@ -425,18 +429,40 @@ def SHD(true_model, est_model):
...
+
+    ds = s1 - s2
```
ankurankan, line 451: I would add a few comments in this function. For example, specify that this block of code deals with edge deletions from m1, the next block of code deals with edge additions to m1, and the last block deals with flips.

ankurankan (changes requested):

pgmpy/metrics/metrics.py
Nimish-4: Yep, I think this would only be true when two nodes have edges pointing to each other, which would go against the definition of a DAG. I guess the R implementation is for a general graph.

pgmpy/metrics/metrics.py
Nimish-4: I have added the same in the commented description of the function at the beginning.

pgmpy/metrics/metrics.py
```diff
@@ -394,3 +395,74 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    dag_true = nx.DiGraph(true_model.edges())
+    adj_mat_true = nx.adjacency_matrix(dag_true, nodelist=nodes_list).todense()
+
+    dag_est = nx.DiGraph(est_model.edges())
+    adj_mat_est = nx.adjacency_matrix(dag_est, nodelist=nodes_list).todense()
+
+    shd = 0
+
+    m1 = adj_mat_true
+    m2 = adj_mat_est
```
ankurankan: I don't think these two lines add anything. Why not just directly assign m1 and m2 in lines 440 and 443?

ankurankan (changes requested):
@Nimish-4 I think the PR looks good now. I have added a minor comment, and it should be ready to be merged.

pgmpy/metrics/metrics.py
Nimish-4: True, my bad.

ankurankan:
@Nimish-4 Thanks for the PR.

PR #1841, opened 2024-10-16
Refining the examples and XMLBIF for small error 

Nuna7:
1. Remove e which is not defined in except section of importing pyparsing. 

2. Add import numpy as np on structure learning in bayesian network example. 

3. In Inference in Discrete Bayesian Networks example, fix typo, include state_names as the parameter of TabularCPD and add import statement of TabularCPD. 

4. In extending pgmpy example, add constructor which call _initialize_structures() function.


pgmpy/readwrite/XMLBIF.py
```diff
@@ -10,8 +10,7 @@
     import pyparsing as pp
 except ImportError:
     raise ImportError(
-        e.message()
-        + ". pyparsing is required for using read/write methods. Please install using: pip install pyparsing."
+        ". pyparsing is required for using read/write methods. Please install using: pip install pyparsing."
```
ankurankan: I don't think we need the full stop here as we are not appending to the error message anymore. Could you remove it?

examples/Extending pgmpy.ipynb
```diff
@@ -33,6 +33,10 @@
     "\n",
     "\n",
     "class SimpleInference(Inference):\n",
+    "    def __init__(self,model):\n",
+    "        super(DBNInference, self).__init__(model)\n",
```
ankurankan: Shouldn't this be `super(SimpleInference, self) .... ` ?

ankurankan (changes requested):
@Nuna7 Thanks for the PR. I have added a few minor  comments.

ankurankan (approved):

ankurankan:
@Nuna7 Thanks for the PR.

PR #1846, opened 2024-10-18
Modified .predict() method for more inference methods.

### LINKED ISSUE #314: Predict method should have an argument to provide algorithm

Nimish-4:

- Fixes #314

- Modified BayesianNetwork.predict() method
- Modified a test in test_BayesianNetwork

### DIFF
--- a/pgmpy/inference/ApproxInference.py
+++ b/pgmpy/inference/ApproxInference.py
@@ -2,6 +2,7 @@ import itertools
 
 from pgmpy.factors.discrete import DiscreteFactor
 from pgmpy.models import BayesianNetwork, DynamicBayesianNetwork
+from pgmpy.utils import compat_fns
 
 
 class ApproxInference(object):
@@ -67,6 +68,9 @@ class ApproxInference(object):
             Else, returns a dict with marginal distribution of each variable in
             `variables`.
         """
+        if isinstance(variables, (set, tuple)):
+            variables = list(variables)
+
         if joint == True:
             return self._get_factor_from_df(
                 samples.groupby(variables).size() / samples.shape[0], state_names
@@ -145,7 +149,9 @@ class ApproxInference(object):
         {'HISTORY': <DiscreteFactor representing phi(HISTORY:2) at 0x7f92dc61eb50>,
          'CVP': <DiscreteFactor representing phi(CVP:3) at 0x7f92d915ec40>}
         """
+
         # Step 1: If samples are not provided, generate samples for the query
+
         if samples is None:
             if isinstance(self.model, BayesianNetwork):
                 samples = self.model.simulate(
@@ -196,3 +202,81 @@ class ApproxInference(object):
         return self.get_distribution(
             samples, variables=variables, state_names=state_names, joint=joint
         )
+
+    def map_query(
+        self,
+        variables,
+        n_samples=int(1e4),
+        samples=None,
+        evidence=None,
+        virtual_evidence=None,
+        state_names=None,
+        show_progress=True,
+        seed=None,
+    ):
+        """
+        Computes the MAP Query using approx inference. Returns the
+        highest probable state in the joint distribution of `variables`.
+
+        Parameters
+        ----------
+        variables: list
+            List of variables for which the probability distribution needs to be calculated.
+
+        n_samples: int
+            The number of samples to generate for computing the distributions. Higher `n_samples`
+            results in more accurate results at the cost of more computation time.
+
+        samples: pd.DataFrame (default: None)
+            If provided, uses these samples to compute the distribution instead
+            of generating samples. `samples` **must** conform with the provided
+            `evidence` and `virtual_evidence`.
+
+        evidence: dict (default: None)
+            The observed values. A dict key, value pair of the form {var: state_name}.
+
+        virtual_evidence: list (default: None)
+            A list of pgmpy.factors.discrete.TabularCPD representing the virtual/soft
+            evidence.
+
+        state_names: dict (default: None)
+            A dict of state names for each variable in `variables` in the form {variable_name: list of states}.
+            If None, inferred from the data but is possible that the final distribution misses some states.
+
+        show_progress: boolean (default: True)
+            If True, shows a progress bar when generating samples.
+
+        seed: int (default: None)
+            Sets the seed for the random generators.
+
+        Examples
+        --------
+        >>> from pgmpy.utils import get_example_model
+        >>> from pgmpy.inference import ApproxInference
+        >>> model = get_example_model("alarm")
+        >>> infer = ApproxInference(model)
+        >>> print(infer.map_query(variables=["HISTORY", "CVP"]))
+        {'HISTORY': 'FALSE', 'CVP': 'NORMAL'}
+        """
+
+        final_distribution = self.query(
+            variables,
+            n_samples=n_samples,
+            samples=samples,
+            evidence=evidence,
+            virtual_evidence=virtual_evidence,
+            joint=True,
+            state_names=state_names,
+            show_progress=show_progress,
+            seed=seed,
+        )
+
+        argmax = compat_fns.argmax(final_distribution.values)
+        assignment = final_distribution.assignment([argmax])[0]
+
+        map_query_results = {}
+        for var_assignment in assignment:
+            var, value = var_assignment
+            map_query_results[var] = value
+
+        return map_query_results
--- a/pgmpy/inference/ApproxInference.py
+++ b/pgmpy/inference/ApproxInference.py
@@ -149,9 +149,7 @@ class ApproxInference(object):
         {'HISTORY': <DiscreteFactor representing phi(HISTORY:2) at 0x7f92dc61eb50>,
          'CVP': <DiscreteFactor representing phi(CVP:3) at 0x7f92d915ec40>}
         """
-
         # Step 1: If samples are not provided, generate samples for the query
-
         if samples is None:
             if isinstance(self.model, BayesianNetwork):
                 samples = self.model.simulate(
@@ -215,8 +213,8 @@ class ApproxInference(object):
         seed=None,
     ):
         """
-        Computes the MAP Query using approx inference. Returns the
-        highest probable state in the joint distribution of `variables`.
+        Finds the most probable state in the joint distribution of variables. Calculates the
+        result by generating samples and calculating most probable states based on the probabilities.
 
         Parameters
         ----------
@@ -258,10 +256,16 @@ class ApproxInference(object):
         --------
         >>> from pgmpy.utils import get_example_model
         >>> from pgmpy.inference import ApproxInference
+        >>> from pgmpy.factors.discrete import State,TabularCPD
         >>> model = get_example_model("alarm")
         >>> infer = ApproxInference(model)
         >>> print(infer.map_query(variables=["HISTORY", "CVP"]))
         {'HISTORY': 'FALSE', 'CVP': 'NORMAL'}
+        >>> virtual_evidence_history = TabularCPD(variable='HISTORY', variable_card=2,values=[[0.99],[0.01]],
+                                          state_names={"HISTORY": ["TRUE", "FALSE"]})
+        >>> evidence = {'CVP':'NORMAL'}
+        >>> print(infer.map_query(variables=["HISTORY"],evidence=evidence,virtual_evidence=[virtual_evidence_history]))
+        {'HISTORY': 'TRUE'}
         """
         final_distribution = self.query(
             variables,
--- a/pgmpy/inference/ApproxInference.py
+++ b/pgmpy/inference/ApproxInference.py
@@ -249,6 +249,11 @@ class ApproxInference(object):
         seed: int (default: None)
             Sets the seed for the random generators.
 
+        Returns
+        -------
+        MAP values: dict
+            The most probable state of provided `variables` given the evidence.
+
         Examples
         --------
         >>> from pgmpy.utils import get_example_model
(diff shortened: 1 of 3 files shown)


pgmpy/models/BayesianNetwork.py
```diff
@@ -641,7 +646,14 @@ def predict(self, data, stochastic=False, n_jobs=-1):
             raise ValueError("Data has variables which are not in the model")
 
         missing_variables = set(self.nodes()) - set(data.columns)
-        model_inference = VariableElimination(self)
+
+        if algo is None:
+            algo = VariableElimination
+        else:
+            if not issubclass(algo, Inference):
+                raise TypeError("Algorithm should be a valid pgmpy inference method.")
```
ankurankan: You can modify the error message to be a bit more informative by adding the type detected in the code. Something like adding "Got {type(algo)}" at the end of the message.

pgmpy/tests/test_models/test_BayesianNetwork.py
```diff
@@ -1018,7 +1019,12 @@ def test_predict(self):
         titanic.add_edges_from([("Sex", "Survived"), ("Pclass", "Survived")])
         titanic.fit(self.titanic_data2[500:])
 
-        p1 = titanic.predict(self.titanic_data2[["Sex", "Pclass"]][:30])
+        p1_variable_elimination = titanic.predict(
+            self.titanic_data2[["Sex", "Pclass"]][:30]
+        )
+        p1_belief_propagation = titanic.predict(
+            self.titanic_data2[["Sex", "Pclass"]][:30], algo=BeliefPropagation
+        )
```
ankurankan, line 1029: Could you add a test for approx inference as well? The values would be different, and you would need to test it using `assertAlmostEqual`.

pgmpy/models/BayesianNetwork.py
```diff
@@ -585,7 +585,7 @@ def fit_update(self, data, n_prev_samples=None, n_jobs=1):
         self.add_cpds(*cpds)
         logger.disabled = False
 
-    def predict(self, data, stochastic=False, n_jobs=-1):
+    def predict(self, data, algo=None, stochastic=False, n_jobs=-1, **kwargs):
```
ankurankan, line 588: The kwargs arguments provided here need to be passed to the Inference method initialization and the query calls later. Have a look at the various options in the query method for each of the inference and specify that in the docstring for the method.

pgmpy/models/BayesianNetwork.py
```diff
@@ -585,7 +585,7 @@ def fit_update(self, data, n_prev_samples=None, n_jobs=1):
...
         """
         Predicts states of all the missing variables.
```
ankurankan, line 590: Docstring will need to be updated for the new arguments.

ankurankan (changes requested):

Nimish-4:
On it.

pgmpy/tests/test_models/test_BayesianNetwork.py
Nimish-4: @ankurankan The test is failing for some reason. 
`p2_approx_inference = titanic.predict( self.titanic_data2[["Survived", "Pclass"]][:30], algo=ApproxInference )`
is showing the type error for not being a subclass of Inference.
`TypeError: Algorithm should be a valid pgmpy inference method. Got <class 'type'> instead.`

ankurankan: As you can see here: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/inference/ApproxInference.py, ApproxInference is not a subclass of Inference, so you would need to handle it separately.

Nimish-4: @ankurankan Any other cases to be accounted for now? Or only ApproxInference, VariableElim, and BeliefProp?

ankurankan: No

Nimish-4: The .predict() method [calls for map_query()](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/BayesianNetwork.py#L676) when stochastic is False (default case) but there is no .map_query() for ApproxInference. So should this combination of parameters be disallowed or something else?

ankurankan: You can add a map_query method to ApproxInference. The map_query is similar to the query method. The difference is that it returns the state with the highest probability instead of returning the probability distribution. For the map_query method, you would need to call the query method and then return the variables' states with the highest probability.

pgmpy/models/BayesianNetwork.py
Nimish-4: It does not look like there are keyword arguments for just the initialization of Inference methods. Is this not the case?

ankurankan: @Nimish-4 That is correct. The query methods for some of the algorithms do accept additional arguments. I would suggest passing it to them.

pgmpy/inference/ApproxInference.py
```diff
@@ -196,3 +202,81 @@ def query(
...
+        state_names: dict (default: None)
+            A dict of state names for each variable in `variables` in the form {variable_name: list of states}.
+            If None, inferred from the data but is possible that the final distribution misses some states.
+
+        show_progress: boolean (default: True)
+            If True, shows a progress bar when generating samples.
+
+        seed: int (default: None)
+            Sets the seed for the random generators.
+
```
ankurankan, line 249: @Nimish-4 Could you please add a "Returns" section as well?

pgmpy/inference/ApproxInference.py
```diff
@@ -196,3 +202,81 @@ def query(
...
+        Examples
+        --------
+        >>> from pgmpy.utils import get_example_model
+        >>> from pgmpy.inference import ApproxInference
+        >>> model = get_example_model("alarm")
+        >>> infer = ApproxInference(model)
+        >>> print(infer.map_query(variables=["HISTORY", "CVP"]))
+        {'HISTORY': 'FALSE', 'CVP': 'NORMAL'}
+        """
+
```
ankurankan: Please try to avoid unnecessary blank lines. Usually, throughout the codebase, we have tried to use blank lines to denote different stages (or blocks) or code.

ankurankan (changes requested):
@Nimish-4 I have added a couple of very minor comments. Let me know when the PR is ready for a review.

Nimish-4:
@ankurankan I think it is ready for a review. Inform me in case more changes need to be accommodated in the docstring or code!

pgmpy/inference/ApproxInference.py
```diff
@@ -196,3 +202,85 @@ def query(
...
+        n_samples=int(1e4),
+        samples=None,
+        evidence=None,
+        virtual_evidence=None,
+        state_names=None,
+        show_progress=True,
+        seed=None,
+    ):
+        """
+        Computes the MAP Query using approx inference. Returns the
```
ankurankan: I think using MAP query in the docstring could be confusing to a lot of users (plus is repeated information as the method name) as they might not know what it means. I would suggest having the first line as: "Finds the highest probable state in the joint distribution of `variables`. Then the next lines can describe how the method computes the values.

pgmpy/inference/ApproxInference.py
```diff
@@ -145,7 +149,9 @@ def query(
         {'HISTORY': <DiscreteFactor representing phi(HISTORY:2) at 0x7f92dc61eb50>,
          'CVP': <DiscreteFactor representing phi(CVP:3) at 0x7f92d915ec40>}
         """
+
```
ankurankan: Please remove the blank line here.

pgmpy/inference/ApproxInference.py
```diff
@@ -145,7 +149,9 @@ def query(
...
         # Step 1: If samples are not provided, generate samples for the query
+
```
ankurankan: Same.

pgmpy/inference/ApproxInference.py
```diff
@@ -196,3 +202,85 @@ def query(
...
+            The most probable state of provided `variables` given the evidence.
+
+        Examples
+        --------
+        >>> from pgmpy.utils import get_example_model
+        >>> from pgmpy.inference import ApproxInference
+        >>> model = get_example_model("alarm")
+        >>> infer = ApproxInference(model)
+        >>> print(infer.map_query(variables=["HISTORY", "CVP"]))
+        {'HISTORY': 'FALSE', 'CVP': 'NORMAL'}
```
ankurankan, line 263: I would add another example showing how to use evidence and virtual evidence arguments.

pgmpy/models/BayesianNetwork.py
```diff
@@ -594,6 +594,9 @@ def predict(self, data, stochastic=False, n_jobs=-1):
         data: pandas DataFrame object
             A DataFrame object with column names same as the variables in the model.
 
+        algo: pgmpy Inference instance
```
ankurankan: A more informative description would be "a subclass of pgmpy.inference.Inference"

pgmpy/models/BayesianNetwork.py
```diff
@@ -602,11 +605,75 @@ def predict(self, data, stochastic=False, n_jobs=-1):
         n_jobs: int (default: -1)
             The number of CPU cores to use. If -1, uses all available cores.
 
+        **kwargs
+            Optional keyword arguments specific to the selected algorithm.
+
+            - Variable Elimination:
+
+              - virtual_evidence: list (default:None)
```
ankurankan: virtual_evidence wouldn't be valid here.

pgmpy/models/BayesianNetwork.py
```diff
@@ -602,11 +605,75 @@ def predict(self, data, stochastic=False, n_jobs=-1):
...
+                refer https://pgmpy.org/exact_infer/ve.html#module-pgmpy.inference.EliminationOrder
+                for details.
+
+              - joint: boolean (should only be used with stochastic=True i.e. when not calculating MAP)
+                If True, returns a Joint Distribution over `variables`.
+                If False, returns a dict of distributions over each of the `variables`.
+
+            - Belief Propagation:
+
+              - virtual_evidence: list (default:None)
```
ankurankan: Same

pgmpy/models/BayesianNetwork.py
```diff
@@ -602,11 +605,75 @@ def predict(self, data, stochastic=False, n_jobs=-1):
...
+                Sets the seed for the random generators.
+
+              - joint: boolean (should only be used with stochastic=True i.e. when not calculating MAP)
+                If True, returns a Joint Distribution over `variables`.
+                If False, returns a dict of distributions over each of the `variables`.
+
+        Returns
+        -------
+        Inference results: Pandas DataFrame
+            If `stochastic` is True, returns the distribution of the queried variables.
```
ankurankan: stochastic=True doesn't return the distribution, it returns a sample from the distribution.

ankurankan (changes requested):

pgmpy/inference/ApproxInference.py
Nimish-4: On it.

pgmpy/models/BayesianNetwork.py
Nimish-4: Understood. Is it because we are only interested in the most likely value without unnecessary computation? Also, remove `virtual_evidence` for `ApproxInference` too?

ankurankan: No, because we do not currently have a way to specify virtual evidence for prediction in the `predict` method. The dataframe that we pass to the method defines the `evidence` argument for inference. Specifying virtual evidence in the current API would mean that every query call would get the same virtual evidence argument.

pgmpy/inference/ApproxInference.py
```diff
@@ -258,10 +256,16 @@ def map_query(
         --------
         >>> from pgmpy.utils import get_example_model
         >>> from pgmpy.inference import ApproxInference
+        >>> from pgmpy.factors.discrete import State,TabularCPD
         >>> model = get_example_model("alarm")
         >>> infer = ApproxInference(model)
         >>> print(infer.map_query(variables=["HISTORY", "CVP"]))
         {'HISTORY': 'FALSE', 'CVP': 'NORMAL'}
+        >>> virtual_evidence_history = TabularCPD(variable='HISTORY', variable_card=2,values=[[0.99],[0.01]],
+                                          state_names={"HISTORY": ["TRUE", "FALSE"]})
```
ankurankan: @Nimish-4 Line continuation in docstring need `...`. Example: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/BayesianNetwork.py#L151

pgmpy/inference/ApproxInference.py
```diff
@@ -258,10 +256,16 @@ def map_query(
...
+        >>> evidence = {'CVP':'NORMAL'}
+        >>> print(infer.map_query(variables=["HISTORY"],evidence=evidence,virtual_evidence=[virtual_evidence_history]))
```
ankurankan: Space after `,` in code.

ankurankan (changes requested):
@Nimish-4 Have added a couple of very minor comments. Everything else looks good.

pgmpy/inference/ApproxInference.py
Nimish-4: Done.

pgmpy/inference/ApproxInference.py
Nimish-4: Done.

ankurankan (approved):
Looks good. Thanks for the PR :)

ankurankan:
@Nimish-4 I missed that the tests were failing. It was because no seed value was specified for Approx Inference and hence the values were different in each run. I have fixed that now.

Nimish-4:
My bad for missing it!

Issue #1848, opened 2024-10-20
Function combining all the identification methods


ankurankan:
For causal inference, the user currently has to try out each identification method to figure out which method works for their model. Add a new method to the CausalInference class such that it can iteratively apply the different identification methods, and return a combined result.

Vishwaraj839:
@ankurankan Can you assign me this task

ankurankan:
Fixed in #1995

PR #1851, opened 2024-10-24
issue #1848 function identify_causal_effect implemented

Vishwaraj839:

- Fixes #

- 
-
-


pgmpy/inference/CausalInference.py
```diff
@@ -47,6 +47,73 @@ class CausalInference(object):
...
+    {'backdoor': 0.25, 'frontdoor': None, 'instrumental': None}
+
+    References
+    ----------
+    Pearl, Judea. "Causality: models, reasoning and inference." Cambridge university press, 2009.
+        """
+        results = {}
+
+        # Try Backdoor Adjustment
+        try:
```
ankurankan, line 89: @Vishwaraj839 The idea here is just to get the adjustment sets, not to estimate the effect (so no need to call the query methods). This means that you need to call the `get_minimal_adjustment_set`, `get_all_backdoor_adjustment_sets`, and `get_all_frontdoor_adjustment_sets` methods one by one and collect the results from them. The instrumental variable method is only implemented for Structural Equation Models right now, and would need to be first ported to the Causal Inference class. But this should be a separate issue / PR.

PR #1849, opened 2024-10-21
Issue/1787 add continuous variable models

B3CODER:
### Your checklist for this pull request
Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [x] Make sure you are making a pull request against the **dev branch** (left side). Also you should start *your branch* off *our dev*.
- [x] Check the commit's or even all commits' message styles matches our requested structure.

### Issue number(s) that this pull request fixes
- Fixes #1787

### List of changes to the codebase in this pull request
- Implemented continuous variable models for Bayesian networks.
- Fixed the sorting of import statements in 'pgmpy/utils/utils.py' (isort).
- Removed trailing whitespace in 'pgmpy/models/BayesianNetwork.py' (pre-commit hooks).

PR #1852, opened 2024-10-24
Issue/1787 add continuous variable models updated

### LINKED ISSUE #1787: Add continuous variable models to examples
The `pgmpy.utils.get_example_model` function is able to load discrete variable models from the bnlearn repository: https://www.bnlearn.com/bnrepository/. However, loading the Gaussian Bayesian Networks is not supported yet. Add support for continuous variables as well.

B3CODER:

- Fixes #1787 

- Updated CPD construction for model handling:
Modified the code in pgmpy/utils/utils.py to handle the distinction between categorical and continuous models more clearly by introducing predefined sets of models.
The new structure uses cat_models and cont_models variables to classify models and select the correct processing method based on whether the model is categorical or continuous.

-Updated model handling logic in utils.py:
Previous logic was based solely on file extensions (e.g., ".bif.gz"), which was insufficient.
Introduced new logic to check the model type based on predefined lists of cat_models and cont_models.
Raised a custom error if the model is not found in either set.

-Added tests to test_utils.py:

Added unit tests in pgmpy/tests/test_utils/test_utils.py to verify that both categorical and continuous models are being handled correctly.
Tests include:
Verifying that known categorical models are processed correctly.
Verifying that known continuous models are processed correctly.


ankurankan:
@B3CODER The PR looks good to me. Can you add the Json files for other gaussian models?

B3CODER:
Certainly, I'd be happy to add the JSON files for the other Gaussian models.

B3CODER:
Hello,
Just small question you just requested to close the issue.

I have little confusion about that.

ankurankan:
@B3CODER To clarify, I thought #1849 and #1852 are duplicates as both of them seem to implement the same feature. Hence, I wanted to check which branch you plan to continue adding the new models to? And I would close the other.

B3CODER:
No, I did changes as per your request in new one

ankurankan:
Okay, but you don't need to create new PRs for changes. You can keep adding commits to the same one, which makes it easier to track new changes. And with duplicated PR I mean that all the changes that you have in this PR is also in the other PR, hence if we merge this one, the other would not be needed. Both the PRs address the same changes, and this is the updated version compared to the other.

B3CODER:
Okay,  understood and yes this is the updated one.

model.bif
```diff
@@ -0,0 +1,32 @@
+network unknown {
```
ankurankan: What is this file used for?

pgmpy/utils/utils.py
```diff
@@ -98,25 +103,74 @@ def get_example_model(model):
...
+            ref = files("pgmpy") / path
+            with gzip.open(ref) as f:
+                content = f.read()
+            reader = BIFReader(string=content.decode("utf-8"))
+            return reader.get_model()
+    elif model in cont_models:
+        from pgmpy.factors.continuous import LinearGaussianCPD
+        from pgmpy.models import LinearGaussianBayesianNetwork
+
+        # Handle reading of continuous models
```
ankurankan: We don't need this if statement. Because it's only continuous model, it will always be a JSON.

ankurankan (changes requested):
Have added couple of minor comments but looks good to me otherwise.

ankurankan:
@B3CODER Sorry for the delay on this. I think I found a minor mistake in the current JSON file format that you are using. The `variance` values are currently a list but the variance would always be a float value. Because of this, we get an error when we try to print the CPD, for example:
```python
In [9]: model = get_example_model('ecoli70')                                                  
                                                                                              
In [10]: model.cpds 
```
Could you please update the JSON files?

PR #1853, opened 2024-10-25
Implement reduce functionality in Independencies class.

Nuna7:
1. Add reduce functionality for Independencies class.
2. Add test case for the functionality

Assumption : If assertion A entails assertion B and assertion B don't entails assertion A, then assertion A is consider to be more informative and while assertion B is removed.

I am not aware of other situation to be taken care of. Please mention any other situation to consider.

Another way of implementing the checking of entailment between two assertion is to take the length of each assertion and assume the longer assertion provide more information.

### DIFF
--- a/pgmpy/independencies/Independencies.py
+++ b/pgmpy/independencies/Independencies.py
@@ -326,13 +326,48 @@ class Independencies(object):
         """
         return self.entails(other) and other.entails(self)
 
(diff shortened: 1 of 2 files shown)


ankurankan:
@Nuna7 Looks good. Thanks for the PR.

PR #1854, opened 2024-10-28
Modified Bayesian Network's predict method to handle NaNs

### LINKED ISSUE #1817: Multiple missing values in BayesianNetwork.predict throws IndexError
### Subject of the issue
When there are multiple missing values in the DataFrame passedi nto the `predict` function for BayesianNetwork, it throws an error.

### Your environment
* pgmpy version: dev
* Python version: 3.10.14
* Operating System: Red Hat Enterprise Linux 8.8 (Ootpa)

### Steps to reproduce
Using the documentation example:

```py
import numpy as np
import pandas as pd
from pgmpy.models import BayesianNetwork

values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)),
                      columns=['A', 'B', 'C', 'D', 'E'])

train_data = values[:800]
predict_data = values[800:]
model = BayesianNetwork([('A', 'B'), ('C', 'B'), ('C', 'D'), ('B', 'E')])
model.fit(train_data)
predict_data = predict_data.copy()
predict_data.drop('E', axis=1, inplace=True)

# randomly add some 
mask = np.random.choice([True, False], size=predict_data.shape)
mask[mask.all(1),-1] = 0
predict_data = predict_data.mask(mask)

# predict throws error
y_pred = model.predict(predict_data)
```
### Expected behaviour
Shouldn't throw an error.

### Actual behaviour
Throws an IndexError when it gets down to `DiscreteFactor`:

```
WARNING:pgmpy:Found unknown state name. Trying to switch to using all state names as state numbers

Cell In[1], line 21
     18 predict_data = predict_data.mask(mask)
     20 # predict throws error
---> 21 y_pred = model.predict(predict_data)

File [~/.conda/envs/bnlearn/lib/python3.10/site-packages/pgmpy/models/BayesianNetwork.py:730](https://ood-mccleary.ycrc.yale.edu/node/r102u31n01.mccleary.ycrc.yale.edu/37837/lab/tree/Documents/Uncertainty_Network/~/.conda/envs/bnlearn/lib/python3.10/site-packages/pgmpy/models/BayesianNetwork.py#line=729), in BayesianNetwork.predict(self, data, stochastic, n_jobs)
    727 pred_values = []
    729 # Send state_names dict from one of the estimated CPDs to the inference class.
--> 730 pred_values = Parallel(n_jobs=n_jobs)(
    731     delayed(model_inference.map_query)(
    732         variables=missing_variables,
    733         evidence=data_point.to_dict(),
    734         show_progress=False,
    735     )
    736     for index, data_point in tqdm(
    737         data_unique.iterrows(), total=data_unique.shape[0]
    738     )
    739 )
    741 df_results = pd.DataFrame(pred_values, index=data_unique.index)
    742 data_with_results = pd.concat([data_unique, df_results], axis=1)
...
File [~/.conda/envs/bnlearn/lib/python3.10/site-packages/joblib/parallel.py:1918](https://ood-mccleary.ycrc.yale.edu/node/r102u31n01.mccleary.ycrc.yale.edu/37837/lab/tree/Documents/Uncertainty_Network/~/.conda/envs/bnlearn/lib/python3.10/site-packages/joblib/parallel.py#line=1917), in Parallel.__call__(self, iterable)
   1916     output = self._get_sequential_output(iterable)
   1917     next(output)
-> 1918     return output if self.return_generator else list(output)
   1920 # Let's create an ID that uniquely identifies the current call. If the
   1921 # call is interrupted early and that the same instance is immediately
   1922 # re-used, this id will be used to prevent workers that were
   1923 # concurrently finalizing a task from the previous call to run the
   1924 # callback.
   1925 with self._lock:
File [~/.conda/envs/bnlearn/lib/python3.10/site-packages/pgmpy/factors/discrete/DiscreteFactor.py:569](https://ood-mccleary.ycrc.yale.edu/node/r102u31n01.mccleary.ycrc.yale.edu/37837/lab/tree/Documents/Uncertainty_Network/~/.conda/envs/bnlearn/lib/python3.10/site-packages/pgmpy/factors/discrete/DiscreteFactor.py#line=568), in DiscreteFactor.reduce(self, values, inplace, show_warnings)
    567 phi.cardinality = phi.cardinality[var_index_to_keep]
    568 phi.del_state_names([var for var, _ in values])
--> 569 phi.values = phi.values[tuple(slice_)]
    571 if not inplace:
    572     return phi

IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

### Solution

This is an easy fix. I just changed the following line: https://github.com/pgmpy/pgmpy/blob/5f52e039eaa5834c2fd0d9761a0a73764b4fae79/pgmpy/models/BayesianNetwork.py#L710
 
to
```py
evidence=data_point[~data_point.isna()].to_dict(),
```

Two things to note:

1. It needs to be changed for both the stochastic and non-stochastic cases, as well as in `predict_probability`, I'm assuming.
2. We probably want to put in the documentation that we will use available evidence, but will only be predicting for the `missing_variables`, not all missing variables. This could alternatively be changed to fill in any values that are missing for the entire DataFrame.

Nimish-4:

- Fixes #1817

- Added support for calculating NaN values in the DataFrame passed used for BayesianNetwork.predict()

### DIFF
--- a/pgmpy/models/BayesianNetwork.py
+++ b/pgmpy/models/BayesianNetwork.py
@@ -712,15 +712,17 @@ class BayesianNetwork(DAG):
         model_inference = algo(self)
 
         if stochastic:
-            data_unique_indexes = data.groupby(list(data.columns)).apply(
+            data_unique_indexes = data.groupby(list(data.columns), dropna=False).apply(
                 lambda t: t.index.tolist()
             )
             data_unique = data_unique_indexes.index.to_frame()
 
             pred_values = Parallel(n_jobs=n_jobs)(
                 delayed(model_inference.query)(
-                    variables=missing_variables,
-                    evidence=data_point.to_dict(),
+                    variables=missing_variables.union(
+                        set(data_point.index[data_point.isna()])
+                    ),
+                    evidence=data_point[~data_point.isna()].to_dict(),
                     show_progress=False,
                     **kwargs,
                 )
@@ -728,22 +730,29 @@ class BayesianNetwork(DAG):
                     data_unique.iterrows(), total=data_unique.shape[0]
                 )
             )
+
+            all_columns = data.columns.tolist() + [col for col in missing_variables]
             predictions = pd.DataFrame()
             for i, row in enumerate(data_unique_indexes):
-                p = pred_values[i].sample(n=len(row))
-                p.index = row
-                predictions = pd.concat((predictions, p), copy=False)
+                samples = pred_values[i].sample(n=len(row))
+                initial_variables = data_unique.iloc[[i]]
+                known_variables = initial_variables.dropna(axis=1)
+                known_df = known_variables.loc[known_variables.index.repeat(len(row))]
+                known_df.index = samples.index
 
-            return predictions.reindex(data.index)
+                complete_data = pd.concat([samples, known_df], axis="columns")
+                complete_data.index = row
+                complete_data = complete_data.reindex(columns=all_columns)
+                predictions = pd.concat((predictions, complete_data), copy=False)
+
+            return predictions.sort_index()
 
         else:
             data_unique_indexes = data.groupby(list(data.columns), dropna=False).apply(
                 lambda t: t.index.tolist()
             )
             data_unique = data_unique_indexes.index.to_frame()
-            pred_values = []
 
-            # Send state_names dict from one of the estimated CPDs to the inference class.
             pred_values = Parallel(n_jobs=n_jobs)(
                 delayed(model_inference.map_query)(
                     variables=missing_variables.union(
@@ -757,23 +766,21 @@ class BayesianNetwork(DAG):
                     data_unique.iterrows(), total=data_unique.shape[0]
                 )
             )
-            df_results = pd.DataFrame(pred_values, index=data_unique.index)
 
-            all_columns = data_unique.columns.tolist() + [
-                col for col in df_results.columns if col not in data_unique.columns
-            ]
-            df_complete_results = df_results.reindex(columns=all_columns)
-            data_with_results = df_complete_results.combine_first(data_unique)
-
+            all_columns = data.columns.tolist() + [col for col in missing_variables]
             predictions = pd.DataFrame()
             for i, row in enumerate(data_unique_indexes):
-                unique_row = data_with_results.iloc[[i]]
-                duplicate_rows = unique_row.loc[unique_row.index.repeat(len(row))]
-                duplicate_rows.index = row
-                predictions = pd.concat((predictions, duplicate_rows), copy=False)
+                predicted = pd.DataFrame(pred_values[i], index=[0])
+                initial_variables = data_unique.iloc[[i]].reset_index(drop=True)
+                known_variables = initial_variables.dropna(axis=1)
 
-            predictions = predictions.sort_index()
-            return predictions
+                complete_data = pd.concat([predicted, known_variables], axis="columns")
+                complete_data = complete_data.loc[complete_data.index.repeat(len(row))]
+                complete_data.index = row
+                complete_data = complete_data.reindex(columns=all_columns)
+                predictions = pd.concat([predictions, complete_data], copy=False)
+
+            return predictions.sort_index()
 
     def predict_probability(self, data):
         """
--- a/pgmpy/models/BayesianNetwork.py
+++ b/pgmpy/models/BayesianNetwork.py
@@ -672,7 +672,7 @@ class BayesianNetwork(DAG):
         >>> predict_data.drop('E', axis=1, inplace=True)
         >>> approx_inf_parameters = {'n_samples':int(1e3),'seed':42}
         >>> y_pred = model.predict(predict_data,algo=ApproxInference,**approx_inf_parameters)
-        >>> y_pred
+        >>> y_pred['E']
             E
         800 1
         801 0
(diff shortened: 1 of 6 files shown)


pgmpy/models/BayesianNetwork.py
```diff
@@ -753,8 +755,45 @@ def predict(self, data, algo=None, stochastic=False, n_jobs=-1, **kwargs):
...
+            # converting results dict to dataframe, only the predicted data points are not NaN
+            #        E    B    D    C    A
+            #  800  1.0  1.0  1.0  NaN  NaN
+            #  801  0.0  NaN  NaN  1.0  NaN
+
+            all_columns = data_unique.columns.tolist() + [
+                col for col in df_results.columns if col not in data_unique.columns
+            ]
+            df_complete_results = df_results.reindex(columns=all_columns)
+            print(df_complete_results)
```
ankurankan: I can't think of an easy way to get the final dataframe using merge when the Nan values are present. But you can do something similar to the logic used in the `stochastic=True` case: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/BayesianNetwork.py#L715. This line creates a mapping of duplicate columns and then utilizes this mapping to create the final results data frame. The main difference in this case would be that we won't do sampling as done at: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/BayesianNetwork.py#L733 but rather it would be the same value for each duplicated row.

Nimish-4:
@ankurankan Waiting for your comments. Also wondering about the unit test. Just using the titanic dataset with a mask for the NaNs?

ankurankan:
@Nimish-4 I don't think the current implementation is correct. Only the missing column gets returned with the current code. Am I missing something?
```python
In [1]: import numpy as np
   ...: import pandas as pd
   ...: from pgmpy.models import BayesianNetwork
   ...: 
   ...: values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)),
   ...:                       columns=['A', 'B', 'C', 'D', 'E'])
   ...: 
   ...: train_data = values[:800]
   ...: predict_data = values[800:]
   ...: model = BayesianNetwork([('A', 'B'), ('C', 'B'), ('C', 'D'), ('B', 'E')])
   ...: model.fit(train_data)
   ...: predict_data = predict_data.copy()
   ...: predict_data.drop('E', axis=1, inplace=True)
   ...: 
   ...: # randomly add some 
   ...: mask = np.random.choice([True, False], size=predict_data.shape)
   ...: mask[mask.all(1),-1] = 0
   ...: predict_data = predict_data.mask(mask)
   ...: 
   ...: # predict throws error
   ...: y_pred = model.predict(predict_data)
 {'A': 'N', 'B': 'N', 'C': 'N', 'D': 'N', 'E': 'N'}
100%|█████████████████████████████████████████████████████████| 68/68 [00:03<00:00, 17.59it/s]

In [3]: y_pred
Out[3]: 
       E
800  0.0
801  0.0
802  0.0
803  1.0
804  1.0
..   ...
995  1.0
996  1.0
997  1.0
998  1.0
999  0.0

[200 rows x 1 columns]


```

Nimish-4:
@ankurankan I was just returning the missing variables for now, changed it to return the entire dataframe.

Nimish-4:
Hmmm  the test cases are failing now that the code returns all the variables instead of just the missing variables. Maybe changing the test cases or adding another parameter to choose which variables to return?

ankurankan:
@Nimish-4 You can change the test cases as the behavior of the function has changed. You would also need to make similar changes to the `stochastic=True` case in the `predict` function. We would similarly want to fill in values for all Nan's in the dataframe.

Nimish-4:
On it!

pgmpy/tests/test_models/test_BayesianModel.py
```diff
@@ -855,9 +855,9 @@ def test_predict(self):
             ]
         )
 
-        np_test.assert_array_equal(p1.values.ravel(), p1_res)
```
ankurankan, line 858: All the tests need to go to BayesianNetwork class. The BayesianModel class will be removed in future.

pgmpy/models/BayesianNetwork.py
```diff
@@ -712,39 +712,53 @@ def predict(self, data, algo=None, stochastic=False, n_jobs=-1, **kwargs):
...
+                complete_data = complete_data.reindex(columns=all_columns)
+                predictions = pd.concat((predictions, complete_data), copy=False)
 
-            return predictions.reindex(data.index)
+            return predictions.sort_index()
 
         else:
-            data_unique = data.drop_duplicates()
-            pred_values = []
+            data_unique_indexes = data.groupby(list(data.columns), dropna=False).apply(
```
ankurankan: Because the log for both stochastic=True or False is now same, with the only difference that in one we fill the dataframe with samples and in the other with the value. Can we simplify the method here as there seems to be a lot of duplicated code?

ankurankan (changes requested):

pgmpy/tests/test_models/test_BayesianModel.py
Nimish-4: That makes sense. I made the change only because it was failing with the new predict method. As the test cases are pretty much identical in both BayesianNetwork and BayesianModel (BayesianNetwork has more comprehensive ones), for now should I change anything?

pgmpy/models/BayesianNetwork.py
Nimish-4: Understood, I'll reduce the unnecessary code.

pgmpy/tests/test_models/test_BayesianNetwork.py
```diff
@@ -1142,14 +1162,38 @@ def test_predict(self):
...
+            )
+        )
+        self.assertTrue(
+            p2_variable_elimination.sort_index(axis=1).equals(
+                p2_complete.sort_index(axis=1)
+            )
+        )
 
-        np_test.assert_array_equal(p3.values.ravel(), p3_res)
+        # TODO: Tests for dataframes with NaN data?
```
ankurankan: @Nimish-4 Do you still need to add some tests here?

Nimish-4: @ankurankan Kind of wanted your opinion if it is needed to test it currently. Like simply test whether the predicted NaNs are within an acceptable range (>60% correct or something like that). Beyond that, I think it is ready.

ankurankan: @Nimish-4 For the `stochastic=False` case, we should get the exact same results everytime, so we can write tests for that checking that we get the correct dataframe output. For the `stochastic=True` case, I think we should add a `seed` argument to the `predict` method, and use that to get reproducible results that we can write tests for.

pgmpy/tests/test_models/test_BayesianNetwork.py
```diff
@@ -1039,6 +1039,12 @@ def test_predict(self):
         )
 
         p3 = titanic.predict(self.titanic_data2[["Survived", "Sex"]][:30])
+        mask = np.random.choice(
+            [True, False], size=self.titanic_data2[["Survived", "Sex"]][:30].shape
+        )
+        p3_with_NaNs = titanic.predict(
```
Nimish-4: This test case with a NaN mask is failing. The values predicted in place of NaNs are mostly equal but in some cases they differ, so they are not exactly equal.

pgmpy/tests/test_models/test_BayesianNetwork.py
Nimish-4: @ankurankan The test case for `stochastic=False` is failing when the NaN mask is applied.

ankurankan:
@Nimish-4 I have added a few tests for the Nan case.

Nimish-4:
Thanks for that!

Issue #1622, opened 2023-03-27
Update the documentation for discretizing methods


lmz123321:
Hi, I am trying to use the discretizer in pgmpy. 

I follow the documentation [here](https://pgmpy.org/factors/discretize.html#) and encounter the same bug [here](https://github.com/pgmpy/pgmpy/issues/1513#issue-1184011820).

How should I use the discretization methods? Could you update the documentation?

Thanks a lot.

pranav33317:
@lmz123321 Sir is this issue open?

ankurankan:
@pranav33317 Sorry, we have removed the discretization methods in the last release. We somehow forgot to close this issue.

PR #1857, opened 2024-10-29
Refined documentation of discretize.py, Add cdf and integrate minimal_dseperator with get_independencies()

Nuna7:
#1622

1. Refined documentation of discretize.py 

2. Add cumulative density function calculation for ContinuousFactor as it is being called several times in discretize (Without the cdf function, none of the discretize class examples and all need the cdf function). Add test case for it. 

3. Integrate minimal_dseperators with get_independencies()

### DIFF
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -574,7 +574,7 @@ class DAG(nx.DiGraph):
         else:
             return False
 
-    def minimal_dseparator(self, start, end, include_latents):
+    def minimal_dseparator(self, start, end, include_latents=False):
         """
         Finds the minimal d-separating set for `start` and `end`.
 
@@ -585,6 +585,9 @@ class DAG(nx.DiGraph):
 
         end: node
             The second node.
+
+        include_latents: boolean (default: False)
+            If true, latent variables are consider for minimal d-seperator.
 
         Examples
         --------
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -394,7 +394,7 @@ class DAG(nx.DiGraph):
 
     def get_independencies(self, latex=False, include_latents=False):
         """
-        Computes independencies in the DAG, by checking d-seperation.
+        Computes independencies in the DAG, by checking minimal d-seperation.
 
         Parameters
         ----------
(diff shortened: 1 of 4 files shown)


pgmpy/base/DAG.py
```diff
@@ -554,7 +538,7 @@ def get_immoralities(self):
                     immoralities.add(tuple(sorted(parents)))
         return immoralities
 
-    def is_dconnected(self, start, end, observed=None):
+    def is_dconnected(self, start, end, observed=None, include_latents=False):
```
ankurankan, line 541: I wouldn't add the `include_latents` argument here. It can lead to misleading results. For example, if either the start or end variable is latent but the user passes `include_latents=False`, it would give incorrect results.

pgmpy/base/DAG.py
```diff
@@ -580,12 +564,17 @@ def is_dconnected(self, start, end, observed=None):
...
+            in self.active_trail_nodes(
+                variables=start, observed=observed, include_latents=include_latents
+            )[start]
+        ):
             return True
         else:
             return False
 
-    def minimal_dseparator(self, start, end):
+    def minimal_dseparator(self, start, end, include_latents=False):
```
ankurankan, line 580: I don't think we want to add the `include_latents` argument here either. I think we always want the separating set to be observed (to be usable in practice). I don't think this would have any practical use case and would be confusing for users.

ankurankan (changes requested):

pgmpy/base/DAG.py
Nuna7: @ankurankan I have one question. This may be because I am a beginner in Bayesian inference. But if we always want the separating set to be observed, why do we expect the get_independencies to return an Independencies object representing X and Z are independent given Y (being latent) in test_BayesianModel.py line 230. ? Is this still correct ?

```python
fork = BayesianModel([("Y", "X"), ("Y", "Z")], latents=["Y"])
self.assertEqual(
            fork.get_independencies(include_latents=True),
            Independencies(("X", "Z", "Y")),
        )
```

Further, this exact test case fail when I remove include_latents from minimal_dseparator as Y is removed from the separator for being latent and resulting the separator set  to be empty which then cause the minimal_separator to return None and hence the test case fail.

```python
# If any of the parents were latents, take the latent's parent
while len(separator.intersection(self.latents)) != 0:
           separator_copy = separator.copy()
            for u in separator:
                if u in self.latents:
                    separator_copy.remove(u)
                    separator_copy.update(set(self.predecessors(u)))
            separator = separator_copy

        # Remove the start and end nodes in case it reaches there while removing latents.
        separator.difference_update({start, end})

        # If the initial set is not able to d-separate, no d-separator is possible.
        if an_graph.is_dconnected(
            start, end, observed=separator
        ):
            return None
```

Nuna7: @ankurankan Regarding my above question, as the latent are removed as candidate for minimal_dseparator, and the test case expect to return X and Z are independent given Y (latent), how do we handle this specific condition ? Should we just return Independencies instance with the assertion that x and y are independent given all latent variable or only latent variable which are parent of x or y or latent variable with path from x or y ? Or any correct handling ? Can you please help in this solution.

ankurankan: @Nuna7 Sorry for the late reply. On second thought, we can keep the `include_latents` argument for both methods. I still can't think of any use case for it though, but I think there's no harm in having that option with the default being set to False.

Nuna7: @ankurankan I add the include_latents and its documentation back. Anymore changes ?

ankurankan (approved):

ankurankan:

Issue #1736, opened 2024-02-27
Add option to simulate missing data


ankurankan:
Introduce an option in BayesianNetwork.simulate that allows for the simulation of data with certain values missing, including the ability to define the scheme for such missingness.

Ref: https://papers.nips.cc/paper_files/paper/2013/file/0ff8033cf9437c213ee13937b1c4c455-Paper.pdf

ankurankan:
Partially solved in #1850 (specifically covers the MCAR case)

PR #1874, opened 2024-11-28
Implement sampling mechanism for MAR and MNAR.

Nuna7:
#1736

@ankurankan I am trying to implement the MAR and MNAR cases to sample missing values. I don't know if they satisfy the desired functionality.

### DIFF
--- a/pgmpy/models/BayesianNetwork.py
+++ b/pgmpy/models/BayesianNetwork.py
@@ -1208,9 +1208,7 @@ class BayesianNetwork(DAG):
         partial_samples=None,
         seed=None,
         show_progress=True,
-        missing_scheme=None,
-        missing_prob=0.1,
-        missing_columns=None,
+        missing_prob=None,
     ):
         """
         Simulates data from the given model. Internally uses methods from
@@ -1253,20 +1251,9 @@ class BayesianNetwork(DAG):
         show_progress: bool
             If True, shows a progress bar when generating samples.
 
-        include_missing: str (default: None)
-            If None, no missing value.
-            MCAR : missing completely at random.
-            MAR : missing at random.
-            MNAR : missing not at random.
-
-        missing_prob: float, TabularCPD (default: 0.1)
-            The probability that there is missing values in the samples (CPD in case of MAR and MNAR).
+        missing_prob: float, list, TabularCPD (default: 0.1)
+            The probability that there is missing values in the samples (CPD in case of MAR).
             The variable of the CPD is the missing column while value of remaining columns of samples are to be specified as evidence.
-
-        missing_columns: list (default: None)
-            The list of columns where there will be missing values in the samples.
-            If None, then all columns could contain the missing values.
-            Not necessary for MNAR and MAR
 
         Returns
         -------
@@ -1375,7 +1362,51 @@ class BayesianNetwork(DAG):
                 model.add_cpds(new_cpd)
                 evidence[new_var] = 0
 
-        # Step 3: If no evidence do a forward sampling
+        # Step 3: If missing_prob; include missing values in samples.
+        if missing_prob:
+            if isinstance(missing_prob, list):
+                for cpd in missing_prob:
+                    if not isinstance(cpd, TabularCPD):
+                        raise ValueError(
+                            "Missing probability should only contain TabularCPD."
+                        )
+            else:
+                if isinstance(missing_prob, TabularCPD):
+                    missing_prob = [missing_prob]
+                else:
+                    raise ValueError("Missing probability should be TabularCPD.")
+
+            for cpd in missing_prob:
+                variable = cpd.variables[0]
+
+                if "*" not in variable:
+                    raise ValueError(
+                        "TabularCPD variable should end with * symbol to represent missingnness variable."
+                    )
+
+                if variable.split("*")[0] not in model.nodes:
+                    raise ValueError("TabularCPD variable not in model nodes.")
+
+                if cpd.cardinality[0] != 2:
+                    raise ValueError(
+                        "Tabular CPD variable should have 2 possible states : Missing (1) and Not Missing (0)"
+                    )
+
+                model.add_node(variable)
+
+                if len(cpd.variables) > 1:
+                    evidences = cpd.variables[1:]
+                    for node in evidences:
+                        if node not in model.nodes():
+                            raise ValueError(
+                                f"TabularCPD evidence {node} not in model nodes."
+                            )
+                        else:
+                            model.add_edge(node, variable)
+
+                model.add_cpds(cpd)
+
+        # Step 4: If no evidence do a forward sampling
         if len(evidence) == 0:
             samples = BayesianModelSampling(model).forward_sample(
                 size=n_samples,
@@ -1385,7 +1416,7 @@ class BayesianNetwork(DAG):
                 partial_samples=partial_samples,
             )
 
-        # Step 4: If evidence; do a rejection sampling
+        # Step 5: If evidence; do a rejection sampling
         else:
             samples = BayesianModelSampling(model).rejection_sample(
                 size=n_samples,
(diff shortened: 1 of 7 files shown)


ankurankan:
@Nuna7 I haven't looked at the current implementation in much detail yet but I was thinking wouldn't it be easier to simply modify the network based on the provided missingness mechanism and the probability values (generate models similar to Fig. 1 in the paper, i.e. m-graphs)? Essentially, add binary X* variables (which essentially would be the mask for X) for each of the variables which has missing values, and add the edges based on the provided CPD. This would allow us to simply call the `simulate` method itself on the modified network and that should generate all the required information. We would simply need to then post-process the variables using the masks. What do you think? I haven't looked at the paper in much detail either, so I could be missing something here.

Nuna7:
@ankurankan I will take a closer look on the paper for the idea of modifying the network. I simply use the current approach when I hear the idea of user providing CPD.

ankurankan:
@Nuna7 Let me know when it is ready for review :)

Nuna7:
@ankurankan I think it can be reviewed.

pgmpy/models/BayesianNetwork.py
```diff
@@ -1364,7 +1363,51 @@ def simulate(
...
+            else:
+                if isinstance(missing_prob, TabularCPD):
+                    missing_prob = [missing_prob]
+                else:
+                    raise ValueError("Missing probability should be TabularCPD.")
+
+            for cpd in missing_prob:
+                variable = cpd.variables[0]
+
+                if "*" not in variable:
```
ankurankan, line 1383: Should this be `variable.endswith('*')` ?

ankurankan: Should this be variable.endswith('*') ?

pgmpy/models/BayesianNetwork.py
```diff
@@ -1253,15 +1251,9 @@ def simulate(
...
-            If True, include missing values in the samples.
-
-        missing_prob: float (default: 0.1)
-            The probability that there is missing values in the samples.
-
-        missing_columns: list (default: None)
-            The list of columns where there will be missing values in the samples.
-            If None, then all columns could contain the missing values.
+        missing_prob: TabularCPD, list  (default: None)
+            The probability of missing value for the variable of TabularCPD.
```
ankurankan, line 1256: We should specify what the variable names in these CPDs should look like such as the variable name and the state names.

pgmpy/models/BayesianNetwork.py
```diff
@@ -1296,7 +1288,14 @@ def simulate(
         >>> model.simulate(n_samples, virtual_intervention=virt_intervention)
 
         Simulation with missing values:
-        >>> model.simulate(n_samples, include_missing=True, missing_prob=0.4, missing_columns=['MINVOLSET', 'VENTLUNG'])
+        >>> from pgmpy.factors.discrete.CPD import TabularCPD
```
ankurankan, line 1297: I think we should show examples of different types of missingness here.

pgmpy/models/BayesianNetwork.py
```diff
@@ -1364,7 +1363,51 @@ def simulate(
                 model.add_cpds(new_cpd)
                 evidence[new_var] = 0
 
-        # Step 3: If no evidence do a forward sampling
+        # Step 3: If missing_prob; include missing values in samples.
+        if missing_prob:
```
ankurankan: I would suggest being more explicit here. Something like `if missing_prob is not None`

pgmpy/models/BayesianNetwork.py
```diff
@@ -1364,7 +1363,51 @@ def simulate(
...
+            if isinstance(missing_prob, list):
+                for cpd in missing_prob:
+                    if not isinstance(cpd, TabularCPD):
+                        raise ValueError(
+                            "Missing probability should only contain TabularCPD."
```
ankurankan: The error message can be a bit more precise. Something like `f"missing_prob must be a list of TabularCPD objects. Got {type(cpd)}"`

pgmpy/models/BayesianNetwork.py
```diff
@@ -1364,7 +1363,51 @@ def simulate(
...
+                    missing_prob = [missing_prob]
+                else:
+                    raise ValueError("Missing probability should be TabularCPD.")
+
+            for cpd in missing_prob:
+                variable = cpd.variables[0]
+
+                if "*" not in variable:
+                    raise ValueError(
+                        "TabularCPD variable should end with * symbol to represent missingnness variable."
```
ankurankan: I would suggest trying to add the "Got {}" part (as in the comment above). I think it makes the error message a bit more informative. For example, here we can just add something like: `Got {variable}`

pgmpy/tests/test_models/test_BayesianNetwork.py
```diff
@@ -1807,23 +1805,77 @@ def test_missing_scheme_sample(self):
...
-        samples = self.con_model.simulate(
-            n_samples=1000, include_missing=True, missing_prob=0.9
+        cpd = TabularCPD(
+            "Z*",
+            2,
+            [[0.5], [0.5]],
+        )
+        samples = self.con_model.simulate(n_samples=1000, missing_prob=cpd)
+        self.assertTrue(samples["Z"].isnull().values.any())
+        self.assertFalse(samples.drop(columns=["Z"]).isnull().values.any())
```
ankurankan, line 1817: In the tests, I would suggest testing for the number of missing values to make sure that the method is working as expected. As this is the MCAR case with 0.5 probability, we would expect around 50% of the values to be missing. We can test that by, for example, testing that the fraction of missing values are between 0.45 to 0.55.

pgmpy/tests/test_models/test_BayesianNetwork.py
```diff
@@ -1807,23 +1805,77 @@ def test_missing_scheme_sample(self):
...
+
+        cpd = TabularCPD(
+            "Z*", 2, [[0.5, 0.5, 0.5, 0.5], [0.5, 0.5, 0.5, 0.5]], ["X", "Y"], [2, 2]
+        )
+        samples = self.con_model.simulate(n_samples=1000, missing_prob=cpd)
+        self.assertTrue(samples["Z"].isnull().values.any())
+        self.assertFalse(samples.drop(columns=["Z"]).isnull().values.any())
```
ankurankan: Similarly, we can test the expected number of missing values with non MCAR cases. But as the number of missing values depends on some other variable now, we would need to take subsamples corresponding to the states of the parent variables, and then test for the expected number of missing values in that.

ankurankan (changes requested):
@Nuna7 I have added a few comments but overall the PR seems mostly ready.

Nuna7:
@ankurankan Please mention any change necessary.

pgmpy/tests/test_models/test_BayesianNetwork.py
```diff
@@ -1807,23 +1805,92 @@ def test_missing_scheme_sample(self):
         samples = self.con_model.simulate(n_samples=1000)
         self.assertFalse(samples.isnull().values.any())
 
-        samples = self.con_model.simulate(
-            n_samples=1000, include_missing=True, missing_prob=0.9
+        cpd = TabularCPD(
+            "Z*",
+            2,
+            [[0.5], [0.5]],
```
ankurankan: I would suggest using values other than 0.5 everywhere because we would end up with 0.5 probability of missingness if the missingness is random and doesn't use the specified CPD.

pgmpy/tests/test_models/test_BayesianNetwork.py
```diff
@@ -1807,23 +1805,92 @@ def test_missing_scheme_sample(self):
...
+        cpd = TabularCPD(
+            "Z*", 2, [[0.5, 0.5, 0.5, 0.5], [0.5, 0.5, 0.5, 0.5]], ["X", "Y"], [2, 2]
+        )
+        samples = self.con_model.simulate(n_samples=1000, missing_prob=cpd)
+        grouped = samples.groupby(["X", "Y"], observed=False)["Z"]
+        for (x, y), group in grouped:
+            missing_fraction = group.isnull().mean()
+            expected_missing_fraction = cpd.values[0][x, y]
+            self.assertAlmostEqual(
+                missing_fraction, expected_missing_fraction, delta=0.15
```
ankurankan: Would suggest increasing the sample size to reduce the delta required here. Would ideally want to keep the delta under 0.1

ankurankan (changes requested):
@Nuna7 I have added a couple of very minor comments. The test function is also quite large now making it a bit hard to read. I would suggest splitting the test over multiple tests.

Nuna7:
@ankurankan Please mention any change necessary. Further, I couldn't actually find a sweet spot to split the test function and would be great if you make suggestion (if you still feel necessary).

ankurankan:
@Nuna7 One potential way to split would be based on missingness type: `test_simulate_missing_mcar`, `test_simulate_missing_mar` ...

pgmpy/models/BayesianNetwork.py
```diff
@@ -1440,15 +1444,28 @@ def simulate(
...
                 samples.loc[samples[variable] == 1, variable.split("*")[0]] = np.nan
                 samples.drop(columns=[variable], inplace=True)
 
+                if return_full:
+                    full_samples.drop(columns=[variable], inplace=True)
+
         # Step 7: Postprocess and return
         if include_latents:
+            if return_full:
+                return samples.astype("category"), full_samples
```
ankurankan: My idea here was to return a single `samples` dataframe that contains both the missing and original columns.

ankurankan (changes requested):

Nuna7:
@ankurankan Any suggestions on the name for the column. For example, if Z is missing, name for Z where there is missing and when there is no missing value

ankurankan:
@Nuna7 Maybe `variable` for the missing column and something like `variable_full` or `variable_original` for the full values one?

Nuna7:
@ankurankan  Please mention any change necessary.

ankurankan (approved):
Looks good. Thanks :)

PR #1881, opened 2024-12-11
Implementation of a class for expert/prior knowledge

### LINKED ISSUE #1828: Modularize expert knowledge specification for causal discovery algorithms
Currently, causal discovery algorithms have arguments to specify expert knowledge. We want to modularize this interface by creating an ExpertKnowledge class that the user can define. This ExpertKnowledge object can then be interchangeably used between different causal discovery algorithms.

Nimish-4:

- Fixes #1828 

- Created ExpertKnowledge class for pgmpy.estimators
- Modified usage of expert knowledge in different files
- Modified tests


pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -0,0 +1,41 @@
+class ExpertKnowledge:
+    """
+    Class to specify expert knowledge for causal discovery algorithms.
```
ankurankan, line 3: We follow the numpydoc style for docstrings: https://numpydoc.readthedocs.io/en/latest/format.html. The standard format for the description should be a one sentence summary, then blank line, and a paragraph of details.

ankurankan, line 3: We follow the numpydoc style for docstrings: https://numpydoc.readthedocs.io/en/latest/format.html. The standard format for the description should be a one sentence summary, then blank line, and a paragraph of details.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -0,0 +1,41 @@
+class ExpertKnowledge:
+    """
+    Class to specify expert knowledge for causal discovery algorithms.
+    Currently, expert knowledge can provide information about edges that have to be present/absent in
+    the final learned graph and a limited search space for edges.
+    Parameters
```
ankurankan, line 10: Empty line above "Parameters"

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -0,0 +1,41 @@
...
+            If a list of edges is provided as `black_list`, they are excluded from the search
+            and the resulting model will not contain any of those edges. Default: None
+    fixed_edges: iterable
+            A list of edges that will always be there in the final learned model.
+            The algorithm will add these edges at the start of the algorithm and
+            will never change it.
+    """
+
+    def _validate_edges(self, edge_list):
+        if not hasattr(edge_list, "__iter__"):
```
ankurankan, line 29: I would also suggest checking if an iterable object other than `set` is passed here. In that case, we can covert that to a set.

pgmpy/estimators/HillClimbSearch.py
```diff
@@ -283,24 +272,33 @@ def estimate(
...
+            else:
+                fixed_edges = set(expert_knowledge.fixed_edges)
+                start_dag.add_edges_from(fixed_edges)
+                if not nx.is_directed_acyclic_graph(start_dag):
+                    raise ValueError(
+                        "fixed_edges creates a cycle in start_dag. Please modify either fixed_edges or start_dag."
+                    )
+
+            # Step 1.4: Check black list and white list
+            black_list = (
```
ankurankan: These prepocessing steps can be moved to  `ExpertKnowlege._validate_edges` ?

pgmpy/estimators/HillClimbSearch.py
```diff
@@ -283,24 +272,33 @@ def estimate(
...
-                )
-
-        # Step 1.4: Check black list and white list
-        black_list = set() if black_list is None else set(black_list)
-        white_list = (
-            set([(u, v) for u in self.variables for v in self.variables])
-            if white_list is None
-            else set(white_list)
-        )
+            white_list = set([(u, v) for u in self.variables for v in self.variables])
```
ankurankan: I would suggest avoiding this and leave it as None. Because we will have to check whether an edge is in the white_list set in every iteration of the algorithm, it can be computationally expensive. It would be faster to just check if the value is None and only check if the edge is in white_list if it is not None.

ankurankan (changes requested):

pgmpy/estimators/HillClimbSearch.py
Nimish-4: Can be done for black list and fixed edges. For white edges, we cannot move it because because we need the variables of dataset to build the white list in the case when the user does not provide it. Unless, as you suggested we get rid of building a white list set (when the user does not provide one) and simply make it None.

pgmpy/estimators/HillClimbSearch.py
Nimish-4: I guess for this I will have to slightly change the [legal operation calculation step](https://github.com/pgmpy/pgmpy/blob/cf8d0f12e2e5be62b01ff8fded85f3f64eab1e84/pgmpy/estimators/HillClimbSearch.py#L61). I'll make the change.

pgmpy/estimators/ExpertKnowledge.py
Nimish-4: I think that case would automatically be considered valid, since we are only rejecting non-iterables? A relevant example I saw was [this part of Mmhc estimator](https://github.com/pgmpy/pgmpy/blob/cf8d0f12e2e5be62b01ff8fded85f3f64eab1e84/pgmpy/estimators/MmhcEstimator.py#L88). Here the white list passed is of type <class 'networkx.classes.reportviews.OutEdgeView'>. So, any iterator is being accepted for each edge information right now. What do you think?

ankurankan: Indeed it would work, but we do convert everything to a `set` in HillClimbSearch: https://github.com/pgmpy/pgmpy/blob/cf8d0f12e2e5be62b01ff8fded85f3f64eab1e84/pgmpy/estimators/HillClimbSearch.py#L299 . Doing that conversion here would just separate out this preprocessing logic.

pgmpy/estimators/HillClimbSearch.py
ankurankan: Yes, I think it's a better idea to keep white_list as None.

pgmpy/estimators/HillClimbSearch.py
ankurankan: Hmm, I didn't realize that's how white_list was getting handled when None. I would change checks like this: https://github.com/pgmpy/pgmpy/blob/cf8d0f12e2e5be62b01ff8fded85f3f64eab1e84/pgmpy/estimators/HillClimbSearch.py#L98 to `(white_list is not None) and ((X, Y) in white_list)`.

Nimish-4: Will make the changes today!

pgmpy/estimators/HillClimbSearch.py
Nimish-4: Do you think instead of keeping white_list as None, we just keep it as an empty set like the fixed edges and blacklist? Won't be an issue, right? Just would look more consistent I think.

ankurankan (changes requested):
@Nimish-4 For the `ExpertKnowledge.__init__` method, I think the code logic is correct. However, I would suggest changing the code a bit to improve readability.
1. I would suggest changing checks to `if white_list is not None`. It is more explicit.
2. I would include the `_convert_to_set` method in the `_validate_edges` method. Ideally, we want to separate out some substantial amount of code/logic in a separate function/method. Here, I don't think it's worth separating it out as it just makes it harder to read as we have to jump around.
3. I would return the converted set from the  `_validate_edges` method, this would simplify our code as: `if white_list is not None: self.white_list = self._validate_edges(white_list)`.

pgmpy/estimators/HillClimbSearch.py
Nimish-4: @ankurankan 

> Do you think instead of keeping white_list as None, we just keep it as an empty set like the fixed edges and blacklist? Won't be an issue, right? Just would look more consistent I think.

ankurankan: @Nimish-4 I think setting it to empty set would be bit confusing for the user because an empty set would imply that all edges are possible.

ankurankan (approved):
@Nimish-4 Looks good to me. I have made a few minor changes as well.

PR #1844, opened 2024-10-17
PDAG cycle issue 

kollisaisiddartha:
### Your checklist for this pull request
Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

- [ ] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [ ] Make sure you are making a pull request against the **dev branch** (left side). Also you should start *your branch* off *our dev*.
- [ ] Check the commit's or even all commits' message styles matches our requested structure.

### Issue number(s) that this pull request fixes
- Fixes #1300 

### List of changes to the codebase in this pull request
- Made changes to resolve PDAG cycle issue.
- Added a test case to check the condition.
-

PR #1882, opened 2024-12-16
Add code to check pdag cycle

Nuna7:
#1844


ankurankan:
Not required anymore. PDAG can have cycles.

PR #1887, opened 2024-12-18
Add ExpertKnowledge support for PC algorithm

Nimish-4:

### Changes up for discussion - 
1. Implementing checks like these, as [highlighted in previous PR](https://github.com/pgmpy/pgmpy/pull/1813/commits/c315fd2f9012c753fe9cb25814edc899c8b5a09c#diff-21d6ba646ed3ffcff694dbcec44c92fe40c2b7717f8bf14755127e8ba3bdbc89R194).
2. Changing test cases, as [highlighted in the previous PR](https://github.com/pgmpy/pgmpy/pull/1813/commits/e206132b41d736bdde4b56ff7027854cd09abb5e). Looks like they were implemented to reduce sample size/time taken for testing PC.

### DIFF
--- a/pgmpy/estimators/ExpertKnowledge.py
+++ b/pgmpy/estimators/ExpertKnowledge.py
@@ -23,6 +23,10 @@ class ExpertKnowledge:
             A list of edges that will always be there in the final learned
             model. The algorithm will add these edges at the start of the
             algorithm and will never change it.
+
+    max_cond_vars: int
+            The maximum number of conditional variables to be used for statistical
+            independce tests (e.g. Peter Clark algorithm). Default: 5
     """
 
     def _validate_edges(self, edge_list):
@@ -35,7 +39,14 @@ class ExpertKnowledge:
         else:
             return edge_list
 
-    def __init__(self, white_list=None, black_list=None, fixed_edges=None, **kwargs):
+    def __init__(
+        self,
+        white_list=None,
+        black_list=None,
+        fixed_edges=None,
+        max_cond_vars=5,
+        **kwargs,
+    ):
         self.white_list = (
             self._validate_edges(white_list) if white_list is not None else None
         )
@@ -45,6 +56,7 @@ class ExpertKnowledge:
         self.fixed_edges = (
             self._validate_edges(fixed_edges) if fixed_edges is not None else set()
         )
+        self.max_cond_vars = max_cond_vars
 
     def check_against_dag(self):
         pass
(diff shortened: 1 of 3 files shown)


Nimish-4:
I also have a doubt about integration of expert knowledge with PC. Imagine the user passes only one fixed edge to PC, say (u,v). Since there would be no white list, we consider all edges in initial graph. [This part of the algorithm](https://github.com/pgmpy/pgmpy/pull/1887/files#diff-21d6ba646ed3ffcff694dbcec44c92fe40c2b7717f8bf14755127e8ba3bdbc89R299) means that both (u,v) and (v,u) would remain in the graph instead of just (u,v). Basically, this integration of fixed edges does not seem to be present currently, unless I misunderstood something.

ankurankan:
@Nimish-4 The PC algorithm starts with an undirected graph, removes edges from it, and then tries to orient the edges to make the final graph directed. Currently, the changes that you have made is for the first part of the algorithm. We would also need to make changes to the skeleton_to_pdag method: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/PC.py#L372. For fixed_edges we would need to first orient the edges as provided by fixed_edges, and then let the other orientation rules work after that.

pgmpy/estimators/PC.py
```diff
@@ -206,6 +213,7 @@ def estimate(
 
     def build_skeleton(
         self,
+        expert_knowledge=None,
         ci_test="chi_square",
         max_cond_vars=5,
```
ankurankan: I would suggest moving `max_cond_vars` to ExpertKnowledge as well.

pgmpy/estimators/PC.py
```diff
@@ -206,8 +207,8 @@ def estimate(
 
     def build_skeleton(
         self,
+        expert_knowledge=None,
```
ankurankan: I would suggest moving this argument after `variant` as the others are more commonly used and more important.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -23,6 +23,10 @@ class ExpertKnowledge:
             A list of edges that will always be there in the final learned
             model. The algorithm will add these edges at the start of the
             algorithm and will never change it.
+
+    max_cond_vars: int
+            The maximum number of conditional variables to be used for statistical
+            independce tests (e.g. Peter Clark algorithm). Default: 5
```
ankurankan: I would suggest using PC algorithm everywhere instead of Peter Clark. I don't think most users would know the full name.

ankurankan (changes requested):
Mostly looks good to me. I haven't tried to run any of the code yet. Is the PR ready or any other changes still need to be made?

Nimish-4:
> Mostly looks good to me. I haven't tried to run any of the code yet. Is the PR ready or any other changes still need to be made?

I'm still reading up on this a little bit and completing the requested changes. Beyond that I think its fine. Maybe 1 or 2 days to finalize.

PR #1889, opened 2025-01-09
Expert Knowledge for Structure Learning 

Nimish-4:
- Renamed variables in ExpertKnowledge class
- Modified PC according to above changes

### List of changes pending
- Incorporate ExpertKnowledge into PC
- Implement temporal PC


pgmpy/estimators/PC.py
```diff
@@ -258,6 +260,11 @@ def build_skeleton(
                     f"ci_test must either be one of {list(CI_TESTS.keys())}, or a function. Got: {ci_test}"
                 )
 
+        if expert_knowledge is not None:
```
ankurankan: I would suggest moving this to `ExpertKnowledge.__init__` method.

Nimish-4: @ankurankan  I am not sure if I get it ... I think in cases where there is no expert knowledge, we just specify `max_cond_vars` to the default value. But if any expert knowledge is specified, `ExpertKnowledge.__init__` already does take care of it.

ankurankan: @Nimish-4 Sorry, I had missed that this is for the case when expert_knowledge is None. But even for that case, I would suggest initializing an expert_knowledge object with default values at the start of the `estimate` method. That would help us avoid doing such ad hoc checks.

Nimish-4: > @Nimish-4 Sorry, I had missed that this is for the case when expert_knowledge is None. But even for that case, I would suggest initializing an expert_knowledge object with default values at the start of the `estimate` method. That would help us avoid doing such ad hoc checks.

Do checkout this change. I did it to both the `estimate() `and `build_skeleton()` methods as in the `test_PC.py`, we are calling `build_skeleton` directly without the `estimate` method. So a user may do the same.

Nimish-4:
Before you review the code itself, I think there is another issue. In the changes, one change for `skeleton_to_pdag` method is 
a new keyword `skip_v_structures` to skip the process of creating v-structures.
There is another variable like this `r4`. Overall it looks messy.

To improve this, what do you think about separating the v-structure part from the Meek's rules? I think this should be already required as the cases for PDAG and CPDAG are considered the same, which I think is inaccurate. We should get a PDAG just by the v-structure orientation, and the CPDAG after applying Meek's rules, if I'm not mistaken.

ankurankan:
@Nimish-4 I don't think we need to separate PDAG and CPDAG. While technically they are different, in practice I don't think there are any use cases where PDAG can be directly used. Moreover, because the difference is subtle, I think it would just be confusing for the user as many of them wouldn't know the difference. So, by default I would suggest to return the CPDAG. And if you would like, add a boolean argument (something like `return_pdag`), that returns the PDAG.

Nimish-4:
> @Nimish-4 I don't think we need to separate PDAG and CPDAG. While technically they are different, in practice I don't think there are any use cases where PDAG can be directly used. Moreover, because the difference is subtle, I think it would just be confusing for the user as many of them wouldn't know the difference. So, by default I would suggest to return the CPDAG. And if you would like, add a boolean argument (something like `return_pdag`), that returns the PDAG.

@ankurankan In that case, I think only a few minor changes and updates in docstring are pending; I need only a review of the logic. Should I mention the Meek's rules paper in the comment for rule no. 4, like references are mentioned in other parts of code?

ankurankan:
@Nimish-4 Great!!

> Should I mention the Meek's rules paper in the comment for rule no. 4, like references are mentioned in other parts of code?

Yes, please.

Nimish-4:
@ankurankan  It is ready for a review. One final doubt - in [rule 3 and 4](https://arxiv.org/pdf/1302.4972) of Meek's, [X and Y](https://github.com/pgmpy/pgmpy/pull/1889/files#diff-21d6ba646ed3ffcff694dbcec44c92fe40c2b7717f8bf14755127e8ba3bdbc89R510) are required to be non adjacent. Do we need to take care of this? I may be wrong but it does not look like this is being taken into account.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -9,20 +9,15 @@ class ExpertKnowledge:
...
-
-    fixed_edges: iterable
-            A list of edges that will always be there in the final learned
-            model. The algorithm will add these edges at the start of the
-            algorithm and will never change it.
+    forbidden_edges: iterable
+            The set of directed edges that must be absent in the final
+            graph structure. Defaults to None.
+    required_edges: iterable
+            The set of directed edges that must be present in the final
```
ankurankan: I don't think "must be present" is correct anymore with the new definition?

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -9,20 +9,15 @@ class ExpertKnowledge:
...
-            those edges. Default: None
-
-    fixed_edges: iterable
-            A list of edges that will always be there in the final learned
-            model. The algorithm will add these edges at the start of the
-            algorithm and will never change it.
+    forbidden_edges: iterable
+            The set of directed edges that must be absent in the final
+            graph structure. Defaults to None.
+    required_edges: iterable
```
ankurankan: Need to add empty lines between each arguemnt

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -35,16 +30,25 @@ def _validate_edges(self, edge_list):
...
-        self.white_list = (
-            self._validate_edges(white_list) if white_list is not None else None
-        )
-        self.black_list = (
-            self._validate_edges(black_list) if black_list is not None else set()
+    def __init__(
+        self,
+        forbidden_edges=None,
+        required_edges=None,
+        temporal_order=None,
```
ankurankan, line 70: Missing docstring for `temporal_order`.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -9,20 +9,15 @@ class ExpertKnowledge:
...
-            algorithm and will never change it.
+    forbidden_edges: iterable
+            The set of directed edges that must be absent in the final
+            graph structure. Defaults to None.
+    required_edges: iterable
+            The set of directed edges that must be present in the final
+            graph structure. Defaults to None.
+    max_cond_vars: int
+            The maximum number of conditional variables to be used for statistical
+            independce tests (e.g. PC algorithm). Default is 5.
```
ankurankan: Could you please also add an example and reference section?

pgmpy/estimators/PC.py
```diff
@@ -54,10 +54,10 @@ def estimate(
         self,
         variant="stable",
         ci_test="chi_square",
-        max_cond_vars=5,
         return_type="dag",
         significance_level=0.01,
         n_jobs=-1,
+        expert_knowledge=None,
```
ankurankan, line 60: I would suggest moving this above `n_jobs` argument.

pgmpy/estimators/PC.py
```diff
@@ -174,24 +175,66 @@ def estimate(
...
             **kwargs,
         )
 
         if return_type.lower() == "skeleton":
             return skel, separating_sets
 
-        # Step 2: Orient the edges based on build the PDAG/CPDAG.
-        pdag = self.skeleton_to_pdag(skel, separating_sets)
+        # Step 2: Orient the edges based on Meek's rules to build the PDAG/CPDAG.
+        pdag = self.skeleton_to_pdag(skel, separating_sets, expert_knowledge)
```
ankurankan: Using the additional arguments in the `skeleton_to_pdag` method looks a bit hacky to me. I think a cleaner way would be to instead split the `skeleton_to_pdag` method into smaller functions: 1) `orient_colliders`, 2) `apply_orientation_rules` (with a boolean argument `apply_r4`). What do you think?

pgmpy/estimators/PC.py
```diff
@@ -174,24 +175,66 @@ def estimate(
...
+
+        # Step 3: Either return the CPDAG, integrate expert knowledge or fully orient the edges to build a DAG.
+        if (
+            expert_knowledge.required_edges != set()
+            or expert_knowledge.forbidden_edges != set()
+        ):
+            # Algorithm to check consistency of background knowledge with the learned graph. Phase II'' in  https://doi.org/10.48550/arXiv.1302.4972
+            progress = True
```
ankurankan: Can we move the consistency check code to a new method in the ExpertKnowledge class? We can pass the pdag object to the method.

pgmpy/estimators/PC.py
```diff
@@ -174,24 +175,66 @@ def estimate(
...
+            while progress:
+                for edge in expert_knowledge.forbidden_edges:
+                    u, v = edge
+                    if pdag.has_edge(u, v):
+                        raise RuntimeError(
```
ankurankan: I would suggest throwing a warning here saying that the edge is incompatible, and ignore that edge. From a user perspective, it would be difficult to come up with consistent expert knowledge, and might need a few iterations of refining the specified expert knowledge. Throwing a warning would provide this information to the user while not forcing them to fix the expert knowledge immediately.

pgmpy/estimators/PC.py
```diff
@@ -174,24 +175,66 @@ def estimate(
...
+                    pdag.remove_edge(v, u)
+                pdag = self.skeleton_to_pdag(
+                    pdag,
+                    separating_sets,
+                    expert_knowledge,
+                    skip_v_structures=True,
+                    r4=True,
+                )
+                # Terminate when all required edges have been added
+                if expert_knowledge.required_edges == set():
```
ankurankan: Can we not iterate over `expert_kowledge.required_edges` instead of using the `while progress` loop? I think it would make the code simpler to read.

ankurankan (changes requested):

pgmpy/estimators/PC.py
Nimish-4: @ankurankan In that case, if the user only provides forbidden edges, we would not be able to check for them as the loop won't even start. That's why I used a loop like that, to check both forbidden and required edges at least once. What do you think?

pgmpy/estimators/PC.py
Nimish-4: @ankurankan This would require having the Meek's rules in the ExpertKnowledge class, won't it? Because we apply them in each iteration after checking the required and forbidden edges. So we would have to  duplicate them there, right?

pgmpy/estimators/PC.py
Nimish-4: > To improve this, what do you think about separating the v-structure part from the Meek's rules? 

That's kind of what I was wanting to do here to improve the readability ... will do!

pgmpy/estimators/ExpertKnowledge.py
Nimish-4: @ankurankan An example on how to use it? I'm not getting it.

pgmpy/estimators/ExpertKnowledge.py
Nimish-4: @ankurankan So something like "directed edges that are forbidden from prior knowledge of the causal structure" seems better?

ankurankan: I think we can say something like: If the specified required edges are compatible with the learned CPDAG, they will be oriented as specified in required_edges. If an edge is not compatible, a warning will be issued, and the edge will be ignored. What do you think?

pgmpy/estimators/ExpertKnowledge.py
ankurankan: Yes, just a few basic examples of how to initialize the class. You can have a look at the Example sections in other classes for example: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/BayesianNetwork.py

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -1,28 +1,62 @@
+import warnings
```
ankurankan: Please use the `pgmpy.global_vars.logger` to allow users to control logging behavior. Example: https://github.com/pgmpy/pgmpy/blob/ec5c973132e72a38050900d668890f6e5bb74c4d/pgmpy/models/BayesianNetwork.py#L262

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -1,28 +1,62 @@
...
-            from the search and the resulting model will not contain any of
-            those edges. Default: None
-
-    fixed_edges: iterable
-            A list of edges that will always be there in the final learned
-            model. The algorithm will add these edges at the start of the
-            algorithm and will never change it.
+    forbidden_edges: iterable
+            The set of directed edges that are to be absent in the final
+            graph structure. If present, a warning will be issued and
```
ankurankan: The `if present, ...` part now depends on the value of `enforce_expert_knowledge`. Right? I would remove the second sentence here, and specify that for the `enforce_expert_knowledge` argument.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -1,28 +1,62 @@
...
+            the edge will be ignored. Defaults to None.
+
+    required_edges: iterable
+            The set of directed edges that are to be present in the final
+            graph structure. If absent, a warning will be issued and
```
ankurankan: Same as above.

pgmpy/estimators/ExpertKnowledge.py
ankurankan: Could also add: "Refer to the algorithm documentation for details on what to expect from the output."

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -35,16 +69,66 @@ def _validate_edges(self, edge_list):
...
+        self.max_cond_vars = max_cond_vars
+
+    def check_against_pdag(self, pdag):
+        """
+        Method to check consistency and orient edges in a graph based on expert knowledge.
+
+        The required and forbidden edges, if specified by the user, are correctly
+        oriented in the graph object passed. In case of any conflict between the
+        graph structure and a required/forbidden edge, the edge is ignored and
+        a user warning is raised at the end. Ref:https://doi.org/10.48550/arXiv.2306.01638
```
ankurankan: For references create a new section in the docstring. This would render properly in the online documentation. Example: https://github.com/pgmpy/pgmpy/blob/ec5c973132e72a38050900d668890f6e5bb74c4d/pgmpy/estimators/CITests.py#L79

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -35,16 +69,66 @@ def _validate_edges(self, edge_list):
...
+        for edge in self.required_edges:
+            u, v = edge
+
+            if pdag.has_edge(u, v) and pdag.has_edge(v, u):
+                pdag.remove_edge(v, u)
+            elif pdag.has_edge(u, v) is False:
+                flag = True
+
+        if flag is True:
+            warnings.warn(
```
ankurankan: It would also be good to specify which specified edge is inconsistent. I suggest putting the warning inside the loop so that a warning is thrown for each inconsistent edge.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -35,16 +69,66 @@ def _validate_edges(self, edge_list):
...
-        self.fixed_edges = (
-            self._validate_edges(fixed_edges) if fixed_edges is not None else set()
+        self.required_edges = (
+            self._validate_edges(required_edges)
+            if required_edges is not None
+            else set()
         )
+        self.max_cond_vars = max_cond_vars
+
+    def check_against_pdag(self, pdag):
```
ankurankan: I would rename this method to something like `orient_pdag` as it is not only checking the pdag.

pgmpy/estimators/PC.py
```diff
@@ -174,24 +183,38 @@ def estimate(
...
+
+        # Step 3: Either return the CPDAG, integrate expert knowledge or fully orient the edges to build a DAG.
+        if (
+            expert_knowledge.required_edges != set()
+            or expert_knowledge.forbidden_edges != set()
+        ) and enforce_background_knowledge is False:
+            pdag = expert_knowledge.check_against_pdag(pdag)
+            pdag = self.apply_orientation_rules(pdag, apply_r4=True)
+
+            return pdag
```
ankurankan: I don't think the `return` statement should be here? The following post-processing still needs to be done before returning.

pgmpy/estimators/PC.py
```diff
@@ -174,24 +183,38 @@ def estimate(
...
 
-        # Step 2: Orient the edges based on build the PDAG/CPDAG.
-        pdag = self.skeleton_to_pdag(skel, separating_sets)
+        # Step 2: Orient the edges based on Meek's rules to build the PDAG/CPDAG.
+        pdag = self.orient_colliders(skel, separating_sets)
+        pdag = self.apply_orientation_rules(pdag)
+
+        # Step 3: Either return the CPDAG, integrate expert knowledge or fully orient the edges to build a DAG.
+        if (
+            expert_knowledge.required_edges != set()
```
ankurankan: Can we move the `!= set()` checks to `check_against_pdag` method? I think it would be much cleaner.

pgmpy/estimators/PC.py
```diff
@@ -369,9 +411,9 @@ def _parallel_fun(u, v):
...
     @staticmethod
-    def skeleton_to_pdag(skeleton, separating_sets):
-        """Orients the edges of a graph skeleton based on information from
-        `separating_sets` to form a DAG pattern (DAG).
+    def orient_colliders(skeleton, separating_sets):
+        """Orients the edges that form v-structures in a graph skeleton
+        based on information from `separating_sets` to form a DAG pattern (PDAG).
 
         Parameters
         ----------
```
ankurankan, line 434: The example here needs to be updated.

ankurankan (changes requested):
I think it mostly looks good. I have added a few minor comments.

ankurankan (approved):
I have made a few small changes, and looks good to me.

Issue #1843, opened 2024-10-17
Extend Structure Learning notebook to include continuous and mixed variable data


ankurankan:
The structure learning tutorial notebook: https://pgmpy.org/examples/Structure%20Learning%20in%20Bayesian%20Networks.html is outdated and only considers discrete data. However, both PC and Hill Climb Search now supports continuous and mixed data. Extend this tutorial to include these cases as well.

rohiitgit:
Hello, I would like to take up this issue and work on it. Can you assign it to me?

jihyeseo:
I'd like to take this. Could you assign me? Thanks.

jihyeseo:
I can add examples with the continuos data. 
However, for mixed one, we'd need to wait.

PR #1893, opened 2025-01-21
Add fit method to allow SVI and MCMC fitting on FunctionalBayesianNetwork

Nuna7:
- Add fit method
- Std for x2 in the test case didn't work yet
- SVI didn't work yet
-  Change how FunctionalCPD input function (fn).


Nuna7:
@ankurankan Sorry for the conflict and large number of commits. No idea why previous commit are include even though git status only show my current change in FunctionalCPD.

ankurankan:
@Nuna7 
> No idea why previous commit are include even though git status only show my current change in FunctionalCPD.

You might have created this branch before your previous PR was merged into dev. I have updated the branch, and fixed the merge conflicts. Make sure to pull to your local before making any changes. 

I saw that you have changed all the examples from using numpy to pyro. Do the FunctionalCPD and FunctionalBN still work with numpy functions? Ideally, we would like to have both options.

Nuna7:
@ankurankan I think I can change functions like the sample function to support both pyro and numpy, but I don't think (could be wrong) fit method would support numpy function because there seems like some underlying function which is use in SVI and MCMC (like I got the error "no attribute log_prob" when using numpy random distributions which seem to be embedded in pyro distributions).

ankurankan:
@Nuna7 Yes, when defined using numpy the current fitting method would not work. But because it's easier to define CPDs using numpy methods, I think it would still be useful for users who want to either do simulations or inference. We can later on figure out ways to do fitting for it.

Nuna7:
@ankurankan If you look at my commit, I put a very subtle change but very important. The following change in sample method :

```
self.fn(parent_sample)
```
become
```
self.fn(parent_sample)()
```

This addition of () is important. Because when fitting, we need to access the pyro distributions as class. For example, dist.Normal(0, 1). So when user specify distributions in FunctionalCPD, they can just put as below :

```
cpd = FunctionalCPD("x1", lambda _ : dist.Normal(0, 1))
```

So when we call self.fn(parent_sample), it return dist.Normal() [Doesn't return sample value]. So basically, the addition of () allow user to specify the distributions in simple way like dist.Normal() and this specification of distributions works both in FunctionalCPD sample and fit method of FunctionalBayesianNetwork. However, when using specification of distributions in numpy as below.

```
cpd = FunctionalCPD("x1", lambda _ : np.random.normal(0, 1))
```

self.fn()() doesn't work as the calling self.fn() also perform the sampling. So, I just want to explain this design decision. And in order to allow user to specify distributions in numpy, the thing which we can do is to call self.fn() once and if the return value is float, then assume it is numpy specification and accordingly only called self.fn() once in sampling instead of self.fn()().

Please provide any better design and whether it is a good idea for checking whether the specification is in numpy or pyro.

Nuna7:
@ankurankan I change the code to account for parent + children relationship. Is it the desired functionality ? Like I guess SVI aren't working yet, user couldn't specify which params to fit and how the fitted params to be reused (unless they add cpd) and all the problems you mentioned earlier.

Nuna7:
@ankurankan The inference function should be ignored for now.

ankurankan:
@Nuna7 Would it be possible for you to move the inference function to a separate PR? It would make it easier for me to review, and we can also merge it before finishing inference as it blocks: #1843

Nuna7:
@ankurankan Just remove it

pgmpy/models/FunctionalBayesianNetwork.py
```diff
@@ -191,3 +192,156 @@ def simulate(self, n_samples=1000, seed=None):
...
+                loss=pyro.infer.Trace_ELBO(),
+            )
+
+            for step in range(num_steps):
+                loss = svi.step()
+                if step % 50 == 0:
+                    print(f"Step {step} | Loss: {loss:.4f}")
+
+            params = pyro.get_param_store()
+            return {name: params[name].detach().numpy() for name in params.keys()}
```
ankurankan: No need to detach here. I think it's okay to return tensors.

pgmpy/tests/test_models/test_FunctionalBayesianNetwork.py
```diff
@@ -162,6 +170,146 @@ def test_simulate_different_distributions(self):
...
+
+        params = self.model.fit(data, method="SVI", learning_rate=0.05, num_steps=2000)
+
+        self.assertIn("x1_mu", params)
+        self.assertIn("x1_sigma", params)
+        self.assertIn("x2_mu", params)
+        self.assertIn("x2_sigma", params)
+        self.assertIn("x2_alpha", params)
+
+        self.assertAlmostEqual(params["x1_mu"].mean(), 1, delta=0.2)
```
ankurankan: I don't think we need to use mean here. The params would be single values (the MAP value) as we are using param to define them.

pgmpy/tests/test_models/test_FunctionalBayesianNetwork.py
```diff
@@ -162,6 +170,146 @@ def test_simulate_different_distributions(self):
         self.assertTrue(np.all(samples["lognormal"] > 0))
         self.assertTrue(np.all(samples["gamma"] > 0))
 
+    def test_svi_fit_normal(self):
+        alpha = 0.23
+        x1 = np.random.normal(1, 2, size=5000)
```
ankurankan: I would suggest reducing the sample size to `1000` to make it faster. Else, it would slow down our github actions.

pgmpy/tests/test_models/test_FunctionalBayesianNetwork.py
```diff
@@ -162,6 +170,146 @@ def test_simulate_different_distributions(self):
...
+            )
+            alpha = pyro.param("x2_alpha", torch.tensor(1.0))
+            return dist.Normal(mu + (parent["x1"] * alpha), sigma)
+
+        cpd1 = FunctionalCPD("x1", fn=lambda _: x1_prior())
+        cpd2 = FunctionalCPD("x2", fn=lambda parent: x2_prior(parent), parents=["x1"])
+
+        self.model.add_cpds(cpd1, cpd2)
+
+        params = self.model.fit(data, method="SVI", learning_rate=0.05, num_steps=2000)
```
ankurankan: The `num_steps` is quite high for a simple model like this. Would suggest changing it to `100` to speed up things.

pgmpy/tests/test_models/test_FunctionalBayesianNetwork.py
```diff
@@ -162,6 +170,146 @@ def test_simulate_different_distributions(self):
...
+        self.assertIn("x1_sigma", params)
+        self.assertIn("x2_mu", params)
+        self.assertIn("x2_sigma", params)
+        self.assertIn("x2_alpha", params)
+
+        self.assertAlmostEqual(params["x1_mu"].mean(), 1, delta=0.2)
+        self.assertAlmostEqual(params["x1_sigma"].mean(), 2, delta=0.2)
+        self.assertAlmostEqual(params["x2_mu"].mean(), 5, delta=0.2)
+        self.assertAlmostEqual(params["x2_sigma"].mean(), 1, delta=0.2)
+        self.assertAlmostEqual(params["x2_alpha"].mean(), 0.23, delta=0.2)
```
ankurankan: delta is a bit too high for a simple model.  0.1 (or even lower) should be good.

pgmpy/models/FunctionalBayesianNetwork.py
```diff
@@ -191,3 +192,156 @@ def simulate(self, n_samples=1000, seed=None):
...
+            raise ValueError("Current implementation only support SVI or MCMC.")
+
+        sort_nodes = list(nx.topological_sort(self))
+
+        if method == "SVI":
+
+            def model():
+                with pyro.plate(f"data", data.shape[0]):
+                    for node in self.nodes:
+                        if node in data.columns:
```
ankurankan: I think we should move this check outside of the fitting loop. We should throw an error if there are missing nodes in the data at the start of the method.

pgmpy/models/FunctionalBayesianNetwork.py
```diff
@@ -191,3 +192,156 @@ def simulate(self, n_samples=1000, seed=None):
...
+
+            params = pyro.get_param_store()
+            return {name: params[name].detach().numpy() for name in params.keys()}
+        else:
+            nuts_kwargs = nuts_kwargs or {}
+            mcmc_kwargs = mcmc_kwargs or {}
+
+            def combined_model():
+                for node in sort_nodes:
+                    if node in data.columns:
```
ankurankan: Same here.

pgmpy/models/FunctionalBayesianNetwork.py
```diff
@@ -191,3 +192,156 @@ def simulate(self, n_samples=1000, seed=None):
...
+                        cpd = self.get_cpds(node)
+                        parents = cpd.parents
+
+                        if len(parents) > 0:
+                            parent_sample = {
+                                parent: torch.tensor(data[parent].values).float()
```
ankurankan: We need to be a bit more careful here as values might not always be continuous. So, the type of variable needs to be decided based on the dataframe's dtypes.

pgmpy/models/FunctionalBayesianNetwork.py
```diff
@@ -191,3 +192,156 @@ def simulate(self, n_samples=1000, seed=None):
...
+            mcmc_kwargs = mcmc_kwargs or {}
+
+            def combined_model():
+                for node in sort_nodes:
+                    if node in data.columns:
+                        cpd = self.get_cpds(node)
+                        parents = cpd.parents
+
+                        if len(parents) > 0:
+                            parent_sample = {
```
ankurankan: I think this conversion is a preprocessing step and should be done outside the algorithm code block. So, for this method as a whole, we should try to have the first block of code for checking all inputs, second for any kind of pre-processing / data structure initialization, and then we should implement the main algorithm.

pgmpy/models/FunctionalBayesianNetwork.py
```diff
@@ -191,3 +192,156 @@ def simulate(self, n_samples=1000, seed=None):
...
+                if step % 50 == 0:
+                    print(f"Step {step} | Loss: {loss:.4f}")
+
+            params = pyro.get_param_store()
+            return {name: params[name].detach().numpy() for name in params.keys()}
+        else:
+            nuts_kwargs = nuts_kwargs or {}
+            mcmc_kwargs = mcmc_kwargs or {}
+
+            def combined_model():
```
ankurankan: Can this combined_model be reused in the case of svi and mcmc? So essentially define this combined model first then do an if-else statement for the fitting method.

pgmpy/tests/test_models/test_FunctionalBayesianNetwork.py
```diff
@@ -162,6 +170,146 @@ def test_simulate_different_distributions(self):
         self.assertTrue(np.all(samples["lognormal"] > 0))
         self.assertTrue(np.all(samples["gamma"] > 0))
 
+    def test_svi_fit_normal(self):
```
ankurankan, line 184: Can we also add a test using the dataset that I had used in the prototype? That is a more realistic scenario with different types of variables connections (v-structures, confounding, etc.). It would help us in making sure that we are not missing something.

pgmpy/models/FunctionalBayesianNetwork.py
```diff
@@ -191,3 +192,156 @@ def simulate(self, n_samples=1000, seed=None):
...
+        method="SVI",
+        learning_rate=1e-4,
+        num_steps=100,
+        nuts_kwargs=None,
+        mcmc_kwargs=None,
+    ):
+        """
+        Fit the Bayesian network to data using Pyro's stochastic variational inference.
+
+        Parameters:
```
ankurankan: We will need to add more details here and use the format that we use in pgmpy.

ankurankan (changes requested):
@Nuna7 Sorry for the delay in reviewing this. I think the implementation looks good. I see that the implementation still does fit the whole model together (not (node + parents) at a time as we had discussed), I don't know if this is what you intended? I have added a few (mostly cosmetic) comments.

pgmpy/models/FunctionalBayesianNetwork.py
Nuna7: @ankurankan Regarding this suggestion, shouldn't there be a node which we are not fitting ? Like as follows :

```
x1 = np.random.normal(1, 2, size=5000)
x2 = np.random.normal(5 + x1, 1)
x3 = np.random.normal(3 + x2, 2)

data = pd.DataFrame({"x1": x1, "x2": x2})

...

model.add_cpds(cpd1, cpd2, cpd3)
params = model.fit(data, method="SVI", learning_rate=0.05, num_steps=1000)
```

ankurankan: @Nuna7 No, when fitting the model we are trying to learn the parameters for all the variables. So, we would need data for all of them. There would be missing variables (as in your example) only for inference.

pgmpy/models/FunctionalBayesianNetwork.py
ankurankan: Also need to use the values from pgmpy.global_vars.Config to determine dtype and device.

Nuna7:
@ankurankan I implement the local fitting as per my understanding

ankurankan (dismissed):

ankurankan:
@Nuna7 Could you review the changes?

Nuna7 (commented):
@ankurankan I think all changes are good and ready for merge.

ankurankan (approved):

PR #1909, opened 2025-01-28
Initialize DAG with lavaan parser

### LINKED ISSUE #1769: Add ways to specify DAGs using simpler notations
Such as with dagitty syntax: https://www.dagitty.net/manual-3.x.pdf#page=3.58 , lavaan syntax: https://lavaan.ugent.be/tutorial/syntax1.html, adjacency matrix

jihyeseo:

- Partially fixes #1769 

- 
-
-

### DIFF
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -76,7 +76,7 @@ class DAG(nx.DiGraph):
         if lavaan_str:
             ebunch, latents, _, _ = parse_lavaan(lavaan_str)
 
-        super(DAG, self).__init__(ebunch=ebunch)
+        super(DAG, self).__init__(ebunch)
         self.latents = set(latents)
         cycles = []
         try:
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -72,11 +72,11 @@ class DAG(nx.DiGraph):
     3
     """
 
-    def __init__(self, ebunch=None, latents=set(), lavaan_str=None, syntax=None):
-        if syntax == "lavaan":
+    def __init__(self, ebunch=None, latents=set(), lavaan_str=None):
+        if lavaan_str:
             ebunch, latents, _, _ = parse_lavaan(lavaan_str)
 
-        super(DAG, self).__init__(ebunch)
+        super(DAG, self).__init__(ebunch=ebunch)
         self.latents = set(latents)
         cycles = []
         try:
@@ -89,10 +89,6 @@ class DAG(nx.DiGraph):
             out_str += "".join([f"({u},{v}) " for (u, v) in cycles])
             raise ValueError(out_str)
 
-            ebunch, latents, err_corr, err_var = parse_lavaan(kwargs["lavaan_str"])
-
-            # Call the parent __init__ with the arguments
-            super(DAG, self).__init__(ebunch=ebunch, latents=latents, err_corr=err_corr)
 
     @classmethod
     def from_lavaan(cls, string=None, filename=None):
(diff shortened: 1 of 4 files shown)


pgmpy/base/DAG.py
```diff
@@ -71,7 +72,10 @@ class DAG(nx.DiGraph):
     3
     """
 
-    def __init__(self, ebunch=None, latents=set()):
+    def __init__(self, ebunch=None, latents=set(), lavaan_str=None, syntax=None):
+        if syntax == "lavaan":
```
ankurankan: I would suggest removing the `syntax` argument and instead check if `lavaan_str` is specified or not.

pgmpy/base/DAG.py
```diff
@@ -85,6 +89,38 @@ def __init__(self, ebunch=None, latents=set()):
             out_str += "".join([f"({u},{v}) " for (u, v) in cycles])
             raise ValueError(out_str)
 
+            ebunch, latents, err_corr, err_var = parse_lavaan(kwargs["lavaan_str"])
```
ankurankan: Is this needed as we are already parsing in line 77?

pgmpy/base/DAG.py
```diff
@@ -85,6 +89,38 @@ def __init__(self, ebunch=None, latents=set()):
...
+
+            # Call the parent __init__ with the arguments
+            super(DAG, self).__init__(ebunch=ebunch, latents=latents, err_corr=err_corr)
```
ankurankan: I don't think nx.DiGraph accepts `err_corr`. Does it just ignore it?

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -340,6 +341,56 @@ def tearDown(self):
...
+        model_str = """# %load model.lav
+                       # measurement model
+                         ind60 =~ x1 + x2 + x3
+                         dem60 =~ y1 + y2 + y3 + y4
+                         dem65 =~ y5 + y6 + y7 + y8
+                       # regressions
+                         dem60 ~ ind60
+                         dem65 ~ ind60 + dem60
+                       # residual correlations
+                         y1 ~~ y5
```
ankurankan: I think we should throw a warning that residual correlations would be ignored. And the user should use the SEM class if they want to use it.

ankurankan (changes requested):
@jihyeseo Thanks for the PR. I have added a few comments.

pgmpy/base/DAG.py
jihyeseo: Thanks for the catch. I will delete the line. It was accidentally there. :-0

pgmpy/base/DAG.py
jihyeseo: This line was there accidentally. I deleted it now. 
In fact, we need to pass the ebunch as `incoming_graph_data` (or just unnamed as default).

pgmpy/tests/test_base/test_DAG.py
jihyeseo: Fixed in b61dcf63fc5aa1a24dbbedd07b1b3a5f4cf788b0

pgmpy/base/DAG.py
jihyeseo: Fixed in 65a808846a5740fe2004b51b83c334e39c04484b

pgmpy/base/DAG.py
jihyeseo: Fixed it 573c6208a4f97892560d4ba6e6894267de903ebd

jihyeseo:
Finished implementing the suggestions. 

Sorry for too many small commits. I was planning to use "rebase"... Got a bit messy.

ankurankan:
@jihyeseo Thanks for the PR. I have made a small change to use the logger module instead of warnings to generate the warning. This gives the user a package-wide control over the behavior of logging messages. And had to also remove the test for warning because I couldn't find an easy way to capture the logger warning.

PR #1917, opened 2025-02-01
Adds support for temporal PC estimator

Nimish-4:

### DIFF
--- a/pgmpy/estimators/ExpertKnowledge.py
+++ b/pgmpy/estimators/ExpertKnowledge.py
@@ -1,3 +1,5 @@
+from itertools import chain
+
 from pgmpy.global_vars import logger
 
 
@@ -22,11 +24,12 @@ class ExpertKnowledge:
             graph structure. Refer to the algorithm documentation for details
             on how the argument is handled.
 
-    temporal order: list of lists (default: None)
+    temporal order: iterator (default: None)
             The temporal ordering of variables according to prior knowledge.
-            Each list in the list of lists contains variables with the same
-            temporal significance; the more prior (parental) variables (list) are at
-            the start while the priority decreases as we go down the list.
+            Each list/structure in the (2 dimensional) iterator contains
+            variables with the same temporal significance; the more prior
+            (parental) variables are at the start while the priority decreases
+            as we go move towards the end of the structure (iterator).
 
     Examples
     --------
@@ -95,9 +98,7 @@ class ExpertKnowledge:
         if self.temporal_order == [[]]:
             return
 
-        tier_set = set()
-        for tier in self.temporal_order:
-            tier_set = tier_set.union(tier)
+        tier_set = set(chain(*self.temporal_order))
 
         if tier_set != set(nodes):
             raise ValueError(
@@ -114,7 +115,7 @@ class ExpertKnowledge:
 
         Parameters
         ----------
-        temporal_order: list of lists
+        temporal_order: iterator
             The temporal ordering of variables according to prior knowledge.
 
         Returns
@@ -122,14 +123,18 @@ class ExpertKnowledge:
         temporal_ordering: dict
             Dictionary with the tier (0, 1, 2, 3 etc.) for each node.
         """
-        if type(temporal_order) != list:
+        if not hasattr(temporal_order, "__iter__"):
             raise TypeError(
-                f"Expected list type for temporal order. Got {type(temporal_order)} instead."
+                f"Expected iterator type for temporal order. Got {type(temporal_order)} instead."
             )
 
         temporal_ordering = dict()
         for order, tier in enumerate(self.temporal_order):
             for node in tier:
+                if node in temporal_ordering:
+                    raise ValueError(
+                        f"Variable {node} present in multiple tiers. Aborting"
+                    )
                 temporal_ordering[node] = order
 
         return temporal_ordering
--- a/pgmpy/estimators/ExpertKnowledge.py
+++ b/pgmpy/estimators/ExpertKnowledge.py
@@ -33,7 +33,10 @@ class ExpertKnowledge:
     Import an example model from pgmpy.utils
 
     >>> from pgmpy.utils import get_example_model
+    >>> from pgmpy.estimators import ExpertKnowledge, PC
+    >>> from pgmpy.sampling import BayesianModelSampling
     >>> asia_model = get_example_model("asia")
+    >>> cancer_model = get_example_model("cancer")
 
     **Required and forbidden edges**
 
@@ -43,8 +46,21 @@ class ExpertKnowledge:
 
     **Use during structure learning**
 
-    >>> from pgmpy.estimators import PC
     >>> data = BayesianModelSampling(asia_model).forward_sample(size=int(1e4))
+    >>> est = PC(data)
+    >>> est.estimate(
+    ...         variant="stable",
+    ...         expert_knowledge=expert_knowledge,
+    ...         show_progress=False,
+    ...     )
+
+    **Temporal order**
+
+    >>> expert_knowledge = ExpertKnowledge(temporal_order=[["Pollution", "Smoker"], ["Cancer"], ["Dyspnoea", "Xray"]])
+
+    **Use during structure learning**
+
+    >>> data = BayesianModelSampling(cancer_model).forward_sample(size=int(1e4))
     >>> est = PC(data)
     >>> est.estimate(
     ...         variant="stable",
@@ -87,8 +103,6 @@ class ExpertKnowledge:
             raise ValueError(
                 f"Missing nodes in temporal order - {set(nodes) - tier_set}"
             )
-
-        return
 
     def _get_temporal_ordering(self, temporal_order):
         """
@@ -139,7 +153,6 @@ class ExpertKnowledge:
                     forbidden_edges.append((node, neighbor))
 
         self.forbidden_edges = self.forbidden_edges.union(forbidden_edges)
-        return
 
     def __init__(
         self,
--- a/pgmpy/estimators/ExpertKnowledge.py
+++ b/pgmpy/estimators/ExpertKnowledge.py
@@ -81,8 +81,7 @@ class ExpertKnowledge:
             else set()
         )
 
-        if temporal_order is not None:
-            raise ValueError(f"Specification of temporal order isn't supported yet.")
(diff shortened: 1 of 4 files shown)


pgmpy/estimators/PC.py
```diff
@@ -313,6 +315,30 @@ def build_skeleton(
                                 graph.remove_edge(u, v)
                                 break
 
+            elif variant == "temporal":
```
ankurankan: @Nimish-4 I don't think we need to add this as a variant. If temporal ordering is specified, we should simply modify the search space for the current variants.

Nimish-4: On it.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +63,20 @@ def _validate_edges(self, edge_list):
...
 
+    def _check_complete_temporal_order(self, pdag_nodes):
+        if self.temporal_order == [[]]:
+            return
+        tier_set = set()
+
+        for tier in self.temporal_order:
+            tier_set = tier_set.union(tier)
+
+        if tier_set != set(pdag_nodes):
```
ankurankan: I wouldn't make this assumption here. If there are missing nodes in the temporal order, I would suggest throwing an error.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +63,20 @@ def _validate_edges(self, edge_list):
         else:
             return edge_list
 
+    def _check_complete_temporal_order(self, pdag_nodes):
```
ankurankan: We can rename `pdag_nodes` -> `nodes` as it doesn't matter where the nodes are coming from for this check. Would make it clearer to read.

ankurankan: Missing docstring

pgmpy/estimators/PC.py
```diff
@@ -392,7 +418,43 @@ def _parallel_fun(u, v):
 
         if show_progress and config.SHOW_PROGRESS:
             pbar.close()
-        return graph, separating_sets
+        return graph, separating_sets, temporal_ordering
+
+    @staticmethod
```
ankurankan, line 400: It would be better to put this in the ExpertKnowledge class or can already compute it and store in a variable when the ExpertKnowledge class is initialized.

pgmpy/estimators/PC.py
```diff
@@ -287,15 +304,24 @@ def build_skeleton(
...
+                    temporal_neighbours = PC._get_temporal_separating_set(
+                        u, v, temporal_ordering, temporal_order
+                    )
                     if (enforce_expert_knowledge is False) or (
                         (u, v) not in expert_knowledge.required_edges
                     ):
                         for separating_set in chain(
                             combinations(
-                                set(graph.neighbors(u)) - set([v]), lim_neighbors
+                                set(graph.neighbors(u))
```
ankurankan: I would suggest renaming _get_temporal_separating_set to something like _get_separating_superset and get the final set of nodes directly. Would make it much cleaner.

pgmpy/estimators/PC.py
```diff
@@ -392,7 +418,43 @@ def _parallel_fun(u, v):
...
+        for tier in range(max_order + 1, overall_max_order + 1):
+            separating_set = separating_set.union(set(temporal_order[tier]))
+
+        separating_set.discard(u)
+        separating_set.discard(v)
+
+        return separating_set
+
+    @staticmethod
+    def _orient_temporal_forbidden_edges(skel, temporal_ordering):
```
ankurankan: Can we move it to ExpertKnowledge as well and directly modify the ExpertKnowledge.forbidden_edges ?

pgmpy/estimators/PC.py
```diff
@@ -392,7 +418,43 @@ def _parallel_fun(u, v):
 
         if show_progress and config.SHOW_PROGRESS:
             pbar.close()
-        return graph, separating_sets
+        return graph, separating_sets, temporal_ordering
```
ankurankan: Do we need to return temporal_ordering here?

ankurankan (changes requested):
I think the algorithm should be fine (I haven't properly tested it yet). However, my main feedback would be to restructure the code. Ideally, we wanted to think of classes as separate entities and try to keep the functionality as separate as possible between them to make it modular. A good design should minimize the dependency between these two classes, i.e., the interaction between these two classes should be clean and we should need to know the internals of either. I have added a few comments toward that here but I think there are more potential improvements that can be made towards this.

pgmpy/estimators/PC.py
Nimish-4: Maybe not really ... I'll try to fix the overall structure.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +63,84 @@ def _validate_edges(self, edge_list):
...
+        tier_set = set()
+        for tier in self.temporal_order:
+            tier_set = tier_set.union(tier)
+
+        if tier_set != set(nodes):
+            raise ValueError(
+                f"Missing nodes in temporal order - {set(nodes) - tier_set}"
+            )
+
+        return
```
ankurankan: We don't need to put return statements here, python by default returns None.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +63,84 @@ def _validate_edges(self, edge_list):
...
+            for neighbor in graph.neighbors(node):
+                if self.temporal_ordering[neighbor] < self.temporal_ordering[node]:
+                    forbidden_edges.append((node, neighbor))
+
+        self.forbidden_edges = self.forbidden_edges.union(forbidden_edges)
+        return
+
     def __init__(
         self,
         forbidden_edges=None,
```
ankurankan, line 167: Could you also add an example of `temporal_order` in the docstring?

pgmpy/estimators/PC.py
```diff
@@ -394,6 +410,45 @@ def _parallel_fun(u, v):
...
+        separating_set: set
+            Set containing the superset of separating set of u, v.
+        """
+        separating_set = set(graph.neighbors(u))
+        separating_set.discard(v)
+        if temporal_ordering == dict():
+            return separating_set
+
+        for neigh in list(separating_set):
+            if temporal_ordering[neigh] > temporal_ordering[u]:
```
ankurankan: I think we need to check whether the variable is in a lower or same `temporal_ordering` to both u and v. So, we would first need to determine which is lower between u and v and then compare `temporal_ordering[neigh]` to that.

pgmpy/estimators/PC.py
```diff
@@ -292,10 +302,16 @@ def build_skeleton(
                     ):
                         for separating_set in chain(
                             combinations(
-                                set(graph.neighbors(u)) - set([v]), lim_neighbors
+                                PC._get_separating_superset(
```
ankurankan: We would need to make this change for stable and parallel variants as well. After which, I think this block of code gets repeated often, so we could potentially move the whole thing to `_get_separating_superset` function and rename it to `_get_potential_sepsets`

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +63,84 @@ def _validate_edges(self, edge_list):
...
+            return
+
+        forbidden_edges = []
+        for node in graph.nodes:
+            for neighbor in graph.neighbors(node):
+                if self.temporal_ordering[neighbor] < self.temporal_ordering[node]:
+                    forbidden_edges.append((node, neighbor))
+
+        self.forbidden_edges = self.forbidden_edges.union(forbidden_edges)
+        return
```
ankurankan: Same

ankurankan (changes requested):

pgmpy/estimators/PC.py
Nimish-4: Correct. Making the change.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +79,81 @@ def _validate_edges(self, edge_list):
...
+
+        Parameters
+        ----------
+        nodes: iterable
+            A collection of nodes present in a dataset/graph object.
+        """
+        if self.temporal_order == [[]]:
+            return
+
+        tier_set = set()
```
ankurankan: Can simplify this as: tier_set = set(chain(*self.temporal_order))

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +79,81 @@ def _validate_edges(self, edge_list):
...
+        for tier in self.temporal_order:
+            tier_set = tier_set.union(tier)
+
+        if tier_set != set(nodes):
```
ankurankan: Would also need to check that the same node is not present in multiple tiers.

pgmpy/estimators/ExpertKnowledge.py
```diff
@@ -63,6 +79,81 @@ def _validate_edges(self, edge_list):
...
+        ----------
+        temporal_order: list of lists
+            The temporal ordering of variables according to prior knowledge.
+
+        Returns
+        --------
+        temporal_ordering: dict
+            Dictionary with the tier (0, 1, 2, 3 etc.) for each node.
+        """
+        if type(temporal_order) != list:
```
ankurankan: Other types should also be allowed. A better way is to test for `hasattr(temporal_order, "__iter__")`

ankurankan (changes requested):
@Nimish-4 I have added a few minor comments but looks good to me. I would suggest adding a few more tests with a bit larger network and test that there are no edges in the network that are not consistent with the provided temporal order. For example, could do the sachs model:
```python
In [1]: from pgmpy.utils import get_example_model                                             
In [2]: model = get_example_model('sachs')                                                    
                                                                                              
In [3]: df = model.simulate(int(1e3))  
In [4]: from pgmpy.estimators import PC, ExpertKnowledge

In [5]: expert = ExpertKnowledge(temporal_order=[['PKC', 'Plcg'], ['PKA', 'Raf', 'Jnk', 'P38',
   ...:  'PIP3', 'PIP2', 'Mek', 'Erk',], ['Akt']])

In [6]: PC(df).estimate(ci_test='chi_square', expert_knowledge=expert)

```

ankurankan (approved):

PR #1937, opened 2025-02-10
Fixes error message in ExpertKnowledge 

Nimish-4:

- Fixes #

- 
-
-


ankurankan:
@Nimish-4 Could you add a test to reproduce the case that was failing?

Nimish-4:
@ankurankan No test case was failing, because we are using the correct temporal orders only in the test. If someone does give an incomplete temporal order, then the error part would be executed ... I just fixed the message. Basically,[ the variable used](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/ExpertKnowledge.py#L108) in the message is not defined. We don't need a test case for this, right?

ankurankan:
Sorry, I misunderstood this. Looks good.

PR #1942, opened 2025-02-12
Modifies the example notebook for structure estimation

Nimish-4:

- Fixes #

- 
-
-


jihyeseo:
Seems related to (partially fixes) https://github.com/pgmpy/pgmpy/issues/1843

ankurankan:
@Nimish-4 Thanks for adding expert knowledge to the notebook. But I think it would be better to have a separate notebook for expert knowledge. I think we could keep this notebook for only "non expert-knowledge" structure learning, so handling different algorithms, data types, and hyperparameters. I think @jihyeseo wanted to make these changes. And we could create a new notebook that goes into more details about specifying expert knowledge for structure learning where we can show examples of different types of expert knowledge. We can always link that notebook from this one. What do you think?

Nimish-4:
> @Nimish-4 Thanks for adding expert knowledge to the notebook. But I think it would be better to have a separate notebook for expert knowledge. I think we could keep this notebook for only "non expert-knowledge" structure learning, so handling different algorithms, data types, and hyperparameters. I think @jihyeseo wanted to make these changes. And we could create a new notebook that goes into more details about specifying expert knowledge for structure learning where we can show examples of different types of expert knowledge. We can always link that notebook from this one. What do you think?

Didn't know that! Making a separate notebook for expert knowledge sounds like the better option. Closing this.

PR #1951, opened 2025-02-14
Remove state_names from (map)_query

### LINKED ISSUE #1948: Remove `state_names` arguemnt from ApproxInference.query
The `ApproxInference.query` method currently has a `state_names` argument that allows the user to specify all possible states of the queried variable. This helps in generating a full distribution on all the possible states of the queried variable if some of these states were missing in the generated samples. We should remove this argument, and automatically read this information for the specified model itself.

Ref #1941

jihyeseo:

- Fixes #1948 

- 
-
-


pgmpy/inference/ApproxInference.py
```diff
@@ -184,17 +179,14 @@ def query(
...
-                    var: list(samples.loc[:, var].unique()) for var in variables
-                }
-            elif isinstance(self.model, DynamicBayesianNetwork):
-                state_names = {
-                    var: list(samples.loc[:, [var]].iloc[:, 0].unique())
-                    for var in variables
-                }
+        # Step 2: Infer state_names from samples.
+        if isinstance(self.model, BayesianNetwork):
+            state_names = {var: list(samples.loc[:, var].unique()) for var in variables}
```
ankurankan, line 184: @jihyeseo We actually want to infer this from the model instead of the samples. Inferring from the samples can lead to problems when some states of a variable don't end up being in the samples (for example when the probability of that state is too low). The `state_names` argument was there to handle this specific case to allow users to specify all possible states.

To infer the state names from the model, we can simply use the `states` attribute for BayesianNetwork.
```python
In [1]: from pgmpy.utils import get_example_model

In [2]: model = get_example_model("alarm")

In [3]: model.states
Out[3]: 
{'ANAPHYLAXIS': ['TRUE', 'FALSE'],
 'VENTALV': ['ZERO', 'LOW', 'NORMAL', 'HIGH'],
 'ARTCO2': ['LOW', 'NORMAL', 'HIGH'],
 'CO': ['LOW', 'NORMAL', 'HIGH'],
 'TPR': ['LOW', 'NORMAL', 'HIGH'],
 'BP': ['LOW', 'NORMAL', 'HIGH'],
 'INSUFFANESTH': ['TRUE', 'FALSE'],
 'SAO2': ['LOW', 'NORMAL', 'HIGH'],
 ...
```
For dynamic BN, we might need to implement something like this.

PR #1952, opened 2025-02-14
Compute mutual information for inference and nodes

### LINKED ISSUE #1801: Mutual Information between Nodes
### Subject of the issue
I think entropy and mutual information would be very useful to include in performing a sensitivity analysis or quantifying relative impact of nodes in relation to one another. 

### Your environment
* pgmpy version: 0.1.25
* Python version: 3.10.14 
* Operating System: Red Hat Enterprise Linux, VERSION: 8.8 

### Expected behaviour
It would be useful to understand the node-level contributions of particular diagnostic nodes on query nodes, similar to a CP query. The `bnmonitor` library has implemented `mutual information` and several other influence metrics: https://github.com/manueleleonelli/bnmonitor?tab=readme-ov-file#overview

### Proposed Solution
I've implemented mutual information and % change in entropy here: https://gist.github.com/vsocrates/2ba5caa42ae27a92c30c9cdb020d5630

I'm not sure if there's a place for this in `pgmpy` so please let me know if not!

jihyeseo:

- Fixes #1801 

- 
-
-

### DIFF
--- a/pgmpy/metrics/__init__.py
+++ b/pgmpy/metrics/__init__.py
@@ -6,6 +6,8 @@ from .metrics import (
     implied_cis,
     log_likelihood_score,
     structure_score,
+    mi,
+    mutual_info_with_percents,
 )
 
 __all__ = [
@@ -15,4 +17,6 @@ __all__ = [
     "implied_cis",
     "fisher_c",
     "SHD",
+    "mi",
+    "mutual_info_with_percents",
 ]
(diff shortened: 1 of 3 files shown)


ankurankan (changes requested):
@jihyeseo Since mutual information is a property between two distributions, I think we should change the `mi` implementation to accept two CPD objects instead of infer and node names. For the `mutual_info_with_percents` function, I think a better place to put it would be as a method in https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/inference/base.py#L19. This would expose the method to all inference algorithms. What do you think?

PR #1965, opened 2025-02-19
Adds support for NOTEARS estimator

### LINKED ISSUE #1342: [ENH] NOTEARS Algorithm
### Subject of the issue
Requesting for adding a new structural learning algorithm (NOTEARS) for learning large networks.
This has been added to causalnex package.

Historically, structure learning has been a very hard problem. We are interested in looking for the optimal directed acyclic graph (DAGs) that describes the conditional dependencies between variables. However, the search space for this is combinatorial and scales super-exponentially with the number of nodes. NOTEARS algorithm cleverly introduces a new optimisation heuristic and approach to solving this problem, where the runtime for this is no longer exponential but scales cubically with the number of nodes.

References:
https://github.com/xunzheng/notears
https://causalnex.readthedocs.io/en/latest/05_resources/05_faq.html#why-notears-algorithm-over-other-structure-learning-methods

### Your environment
Enhancement request.

### Steps to reproduce
Enhancement Request.

### Expected behaviour
Enhancement request.

### Actual behaviour
Enhancement request.

Nimish-4:
Reference implementation - https://github.com/xunzheng/notears

- Fixes #1342 

### List of changes remaining
- Add tests
- Formatting, input checks and docstrings
- Logging and progress bars
- Discrete/non-numeric data (possibly)
- License issue?


ankurankan:
@Nimish-4 The test is failing because of circular import. You will need to revert the __init__ file, the order of imports needs to be the same as before. You can also specify `--no-verify` when committing to not let black auto format it again

ankurankan:
@Nimish-4 Does this work with torch backend? Could you please add tests for that? Example: https://github.com/pgmpy/pgmpy/blob/3c0889084a2675e2a85b8495ed53e856be3abbb1/pgmpy/tests/test_estimators/test_MaximumLikelihoodEstimator.py#L209

ankurankan:
I think another possibility to implement expert knowledge here would be to add another constraint to the optimization problem (would be another Lagrange term in the objective function). We could design a smooth function that goes to 0 when the adjacency matrix aligns with the expert knowledge mask. This should help the optimizer to converge faster.

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        M = X @ W
+        if loss_type == "l2":
+            R = X - M
+            loss = 0.5 / X.shape[0] * (R**2).sum()
+            G_loss = -1.0 / X.shape[0] * X.T @ R
+
+        elif loss_type == "logistic":
+            loss = 1.0 / X.shape[0] * (np.logaddexp(0, M) - X * M).sum()
+            G_loss = 1.0 / X.shape[0] * X.T @ (sigmoid(M) - X)
```
ankurankan: @Nimish-4 Could you also have a look at the codecov warnings and see if you can add tests such that the warning lines get executed when running tests?

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+    ----------
+    Xun Zheng, Bryon Aragam, Pradeep Ravikumar, and Eric P. Xing. 2018. DAGs with NO TEARS: continuous
+    optimization for structure learning. In Proceedings of the 32nd International Conference on Neural
+    Information Processing Systems (NIPS'18). Curran Associates Inc., Red Hook, NY, USA, 9492–9503.
+    """
+
+    def __init__(self, data, **kwargs):
+        super(NOTEARS, self).__init__(data=data, **kwargs)
+
+    def _loss_grad(self, loss_type, X, W):
```
ankurankan: I would suggest using better names instead of `X` and `W`.

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+
+        W: numpy.ndarray
+            The weighted matrix for the DAG.
+
+        Returns
+        -------
+        loss: float
+            The value of the loss function for given data.
+
+        G_loss: numpy.ndarray
```
ankurankan: Better name.

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+    def _constraint_grad(self, W, n):
+        """
+        Calculate the value of the acyclicity constraint and its gradient.
+
+        The acyclicity constraint is given by  -
+            h(W) = trace(exp(W*W)) - d
+                where W is the weighted matrix and d is the number of nodes in graph.
+
+        Parameters
+        ----------
```
ankurankan, line 109: Missing docstring

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+            S = np.exp(M)
+            loss = 1.0 / X.shape[0] * (S - X * M).sum()
+            G_loss = 1.0 / X.shape[0] * X.T @ (S - X)
+
+        else:
+            raise ValueError("unknown loss type")
+
+        return loss, G_loss
+
+    def _constraint_grad(self, W, n):
```
ankurankan: Better names

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+        h = np.trace(E) - n
+        G_h = E.T * W * 2
+        return h, G_h
+
+    def _adj(self, w, n):
+        """
+        Convert doubled array ([2 d^2] array) back to original variables ([d, d] matrix).
+
+        Parameters
+        ----------
```
ankurankan, line 131: Missing docstring.

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+
+        g_h: numpy.ndarray
+            The gradient of the objective function for given parameters.
+        """
+        E = slin.expm(W * W)
+        h = np.trace(E) - n
+        G_h = E.T * W * 2
+        return h, G_h
+
+    def _adj(self, w, n):
```
ankurankan: Better names

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+            The weighted matrix in original dimensions (n*n)
+        """
+        return (w[: n * n] - w[n * n :]).reshape([n, n])
+
+    def _objective_function(self, w, alpha, rho, lambda1, n, loss_type, X):
+        """
+        Evaluate value and gradient of augmented Lagrangian for doubled variables ([2 d^2] array).
+
+        Parameters
+        ----------
```
ankurankan, line 247: Missing docstring

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+        max_iter=20,
+        h_tol=1e-8,
+        rho_max=1e16,
+        w_threshold=0.3,
+        c=0.25,
+        expert_knowledge=None,
+        show_progress=True,
+    ):
+        """
+        Solve min_W L(W; X) + lambda1 ‖W‖_1 s.t. h(W) = 0 using augmented Lagrangian.
```
ankurankan, line 315: Should use latex here: https://numpydoc.readthedocs.io/en/latest/format.html#notes

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,275 @@
...
+
+        c: float
+            progress rate in the range (0, 1)
+
+        Returns
+        -------
+        Estimated model: pgmpy.base.DAG
+            The estimated model with (local) minimum value of objective function.
+
+        Examples
```
ankurankan, line 347: Missing docstring

ankurankan (changes requested):

ankurankan:
@Nimish-4 About defining constraints for expert knowledge, we can define two mask matrices. 1) `forbidden_mask`: value of 1 corresponding to each forbidden edge, 0 otherwise, 2) `fixed_mask`: value of 1 for each fixed edge, 0 otherwise. We can then define constraint for each of these masks. Both of these go to 0 when the constraints are satisfied:
1. For `forbidden_mask`: `sum( abs(adjacency_matrix * fobidden_mask))`
2. For `fixed_mask`, we can use a variant of the bump function: ` sum( f (adjacency_matrix * fixed_mask) )` where `f(x) = e^(- 1 / (w_threshold - x)^(1/alpha) )` if `x > w_threshold`, otherwise `f(x) = 0`. `alpha` would be another parameter here that controls how strictly the constraint should be enforced, a higher `alpha` would be stricter.

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,420 @@
...
+            for i in range(n)
+            for j in range(n)
+        ]
+
+        # Why are we normalizing only for l2 - stabilze by reducing value of squared loss
+        if loss_type == "l2":
+            data = data - self.backend.mean(data, axis=0, keepdims=True)
+
+        # Step 2: Iterate until max_iter iterations
+        adjacency_matrix_new, h_new = None, None
```
ankurankan: This line was inside the `for _ in iteration` loop. I have moved it outside, as I wasn't sure of it's purpose inside. Am I missing something here?

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,420 @@
...
+        for _ in iteration:
+
+            # Step 2.(a): Solve for new values of W and h
+            while rho < rho_max:
+                sol = sopt.minimize(
```
ankurankan: This still doesn't work for pytorch backend as scipy would convert all tensors to numpy. Will need to use pytorch.optim.LBFGS for pytorch backend or can add a new function in compat_fns.py file.

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,420 @@
...
+        # Why are we normalizing only for l2 - stabilze by reducing value of squared loss
+        if loss_type == "l2":
+            data = data - self.backend.mean(data, axis=0, keepdims=True)
+
+        # Step 2: Iterate until max_iter iterations
+        adjacency_matrix_new, h_new = None, None
+        for _ in iteration:
+
+            # Step 2.(a): Solve for new values of W and h
+            while rho < rho_max:
```
ankurankan: I am not sure why we need this nested optimization. Can we simplify this into a single optimization run.

pgmpy/tests/test_estimators/test_NOTEARS.py
```diff
@@ -0,0 +1,195 @@
...
+            W_est, forbidden_mask
+        )
+
+        self.assertEqual(fixed_penalty, 0)
+        self.assertEqual(forbidden_penalty, 0)
+        self.assertTrue(fixed_jac.all() == np.zeros((5, 5)).all())
+        self.assertTrue(forbidden_jac.all() == np.zeros((5, 5)).all())
+
+        # Satisfies forbidden edges but not required edges
+        W_est = np.array(
```
ankurankan, line 104: The fixed penalty doesn't seem to be working. Need to adjust the function.

pgmpy/estimators/NOTEARS.py
```diff
@@ -0,0 +1,420 @@
...
+        lambda1,
+        lambda2=10,
+        lambda3=10,
+        loss_type="l2",
+        max_iter=20,
+        h_tol=1e-8,
+        rho_max=1e16,
+        w_threshold=0.3,
+        c=0.25,
+        alpha_2=1,
```
ankurankan: We will need to reduce the number of hyperparameters. This is not user friendly and would be very difficult for anyone to use it.

pgmpy/tests/test_estimators/test_NOTEARS.py
```diff
@@ -0,0 +1,195 @@
...
+
+        expert_knowledge = ExpertKnowledge(
+            forbidden_edges=[("D", "C"), ("E", "B")],
+            fixed_edges=[("A", "B"), ("C", "D")],
+        )
+        est_dag = self.est_rand.estimate(
+            lambda1=0.01, expert_knowledge=expert_knowledge, show_progress=False
+        )
+
+    def test_estimate_ecoli(self):
```
ankurankan, line 194: Expert knowledge tests are missing for ecoli

pgmpy/tests/test_estimators/test_NOTEARS.py
```diff
@@ -0,0 +1,195 @@
...
+        expert_knowledge = ExpertKnowledge(fixed_edges=[("A", "B"), ("C", "D")])
+        est_dag = self.est_rand.estimate(
+            lambda1=0.01, expert_knowledge=expert_knowledge, show_progress=False
+        )
+
+        expert_knowledge = ExpertKnowledge(
+            forbidden_edges=[("D", "C"), ("E", "B")],
+            fixed_edges=[("A", "B"), ("C", "D")],
+        )
+        est_dag = self.est_rand.estimate(
```
ankurankan, line 188: Add temporal order tests.

ankurankan (changes requested):
@Nimish-4 I have added a few comments. Could you start with the simpler issues (docstring fixes, backend, tests, etc.)? I will also have a look at the more algorithmic issues when I get a chance.

pgmpy/estimators/NOTEARS.py
Nimish-4: My bad, fixed it.

Nimish-4: I misunderstood what you said ... yeah, I don't think there is any point in keeping it inside the `for _ in iteration` loop.

Issue #1089, opened 2019-04-13
Separate direct effect and total effect Instrumental variable


ankurankan:
### Subject of the issue
__This code is current not in dev.__
The current methods for finding conditional/ancestral Instrumental Variables (IVs) `SEMGraph.get_conditional_ivs` compute IVs for the parameter between the variables passed. Therefore the IV is valid only when the direct effect is being computed and would fail sometimes when total effect is needed. It will specifically fail in the case when the conditional IV contains a conditional variable which lies on one of the effect paths as conditioning on that variable would bias the total effect computation.

For example in the case of union sentiment model for the direct effect between `deferenc` and `unionsen`, a conditional IV is `age | yrsmill, laboract` but this IV is not valid for computing total effect between them as `laboract` lies on one to the effect paths and conditioning on it would block that path.

### Your environment
* pgmpy version: All
* Python version: All
* Operating System: All

### Expected behaviour
There should be another method that would compute the IVs in the case when total effect needs to be computed. This can be easily done by marking all the variables in all effect paths as latents and then call the original conditional IV method.

Nishita-Jha:
Hey !! I’d like to take this up. Planning to add a method for conditional IVs for total effect. Let me know if that works :)

dhruvd-1:
hii can this be assigned to me pls

Nishita-Jha:
@ankurankan I’m currently tied up with other tasks could you please reassign this to @dhruvd-1?

Varshith-Yadav:
Hi @ankurankan ,  I noticed this issue has been inactive for a while . If the current assignee is no longer available, I would love to pick this up . Please let me know if I can go ahead and start working on a PR

ankurankan:
@Varshith-Yadav, I have assigned the issue to you. @DARHWOLF is working on a refactor of the current Instrumental variable implementation (#2433). This issue would most likely be another variant in that implementation, but we will need to discuss the design of this.

Varshith-Yadav:
@ankurankan  Thanks for assigning this to me. I'll check out PR #2433 to understand the new structure.

Since my changes depend on that refactor, should I wait for #2433 to be merged before starting implementation? Or would you prefer I draft a design proposal here based on the new architecture in the meantime?

amankum2004:
@ankurankan hello,  I noticed this issue has been inactive for a while . If the current assignee is no longer available, I would love to pick this up . Please let me know if I can go ahead and start working on a PR

PR #1989, opened 2025-03-21
Shift IV implementation, total effect on conditional IV and identification method to try out all methods

Nuna7:
1. Shift IV implementation from SEMGraph to CausalInference and test on both SEMGraph to BayesianNetwork
2. Prototype (new function name 'identification_method') which try out all identification method. (#1848)
3. New function name 'get_total_conditional_ivs' to account for total effect passing the test case of  union sentiment model. (#1089)

@ankurankan Please review as I am kind of not aware of the technicality for 2 and 3 implementation I include so that I am not too diverge.

Documentation is not yet performed....


Nuna7:
@ankurankan Please review and provide any mistake so that the implementation doesn't diverge from the indent functionality.

pgmpy/inference/CausalInference.py
```diff
@@ -58,10 +56,17 @@ def __init__(self, model, set_nodes=None):
...
-        )
+
+        if isinstance(self.model, BayesianNetwork):
+            self.observed_variables = frozenset(self.model.nodes()).difference(
+                model.latents
+            )
+            self.graph = self.model.to_directed()
+        else:
+            self.observed_variables = frozenset(self.model.observed)
+            self.graph = self.model.graph
```
ankurankan, line 67: @Nuna7 I would suggest doing `self.graph = self.model.full_graph_struct`. The `self.graph` can then be treated as a standard DAG model, allowing us to use methods such as `DAG.active_trail_nodes` and `DAG.moralize` instead of reimplmenting them here.

ankurankan (changes requested):
@Nuna7 I have added a comment. I think that is the main change needed for this PR. I plan to further split the current CausalInference class into CausalIdentification and CausalInference classes, but I think it would be easiest if I do it myself.

Additionally, I think you have accidentally pushed your branch to the main pgmpy repo. You will need to push this branch to your forked repo and create a new PR.

PR #1990, opened 2025-03-23
Adds read and write upport for .xdsl file format

### LINKED ISSUE #1967: save the pgmpy.models.BayesianNetwork into a xdsl file that can be opened in GeNIe: https://support.bayesfusion.com/docs/GeNIe/.
Here is my implementation, please consider to add this code into pgmpy: 
```
def export_to_xdsl(bn, filename, network_id="MyNetwork"):
    """
    Export a Bayesian Network (pgmpy model) to a .xdsl file that can be imported into GeNIe.
    
    Parameters:
        bn : BayesianNetwork
            The pgmpy Bayesian network model.
        filename : str
            The file path to save the .xdsl file.
        network_id : str
            Identifier for the network (used in the XML 'id' attribute).
    """
    # Create the root <smile> element with required attributes.
    root = ET.Element("smile", {
        "version": "1.0",
        "id": network_id,
        "numsamples": "10000",
        "discsamples": "10000"
    })
    
    # Create the <nodes> element to hold all CPT definitions.
    nodes_elem = ET.SubElement(root, "nodes")
    
    # Create a dictionary mapping variable names to their CPD for quick lookup.
    cpd_dict = {cpd.variable: cpd for cpd in bn.get_cpds()}
    
    # Obtain nodes in topological order from the Bayesian network.
    topological_nodes = list(nx.topological_sort(bn))
    
    # Process each node (and its associated CPD) in topological order.
    for node in topological_nodes:
        # Retrieve the CPD corresponding to the node.
        cpd = cpd_dict[node]
        
        # Create a <cpt> element with id equal to the variable name.
        cpt_elem = ET.SubElement(nodes_elem, "cpt", {"id": cpd.variable})
        
        # Determine state names: if available in cpd.state_names, use them; otherwise, default to string indices.
        if hasattr(cpd, 'state_names') and cpd.state_names is not None and cpd.variable in cpd.state_names:
            states = cpd.state_names[cpd.variable]
        else:
            states = [str(i) for i in range(cpd.variable_card)]
        
        # Add a <state> element for each state.
        for st in states:
            ET.SubElement(cpt_elem, "state", {"id": st})
        
        # Use the network structure to determine the parents in the correct order.
        evidence = cpd.get_evidence()
        if evidence:
            parents_str = " ".join(evidence)
            parents_elem = ET.SubElement(cpt_elem, "parents")
            parents_elem.text = parents_str
        
        # Add the <probabilities> element.
        probs_elem = ET.SubElement(cpt_elem, "probabilities")
        values = np.array(cpd.get_values())
        # Flatten in column-major order so that for each parent configuration the probabilities for all states are listed.
        flat_values = values.flatten(order='F')
        probs_elem.text = " ".join("{:.17f}".format(x) for x in flat_values)
    
    # Create the <extensions> block with a minimal <genie> element for layout information.
    extensions_elem = ET.SubElement(root, "extensions")
    genie_elem = ET.SubElement(extensions_elem, "genie", {
        "version": "1.0",
        "app": "GeNIe 5.0.4830.0 ACADEMIC",
        "name": network_id
    })
    
    # Provide default layout for each node.
    pos_x, pos_y = 100, 100
    for node in bn.nodes():
        node_elem = ET.SubElement(genie_elem, "node", {"id": node})
        
        # Set the node name.
        name_elem = ET.SubElement(node_elem, "name")
        name_elem.text = node
        
        # Appearance details (colors, font).
        ET.SubElement(node_elem, "interior", {"color": "e5f6f7"})
        ET.SubElement(node_elem, "outline", {"color": "000080"})
        ET.SubElement(node_elem, "font", {"color": "000000", "name": "Arial", "size": "8"})
        
        # Set node position (x1, y1, x2, y2).
        pos_elem = ET.SubElement(node_elem, "position")
        pos_elem.text = f"{pos_x} {pos_y} {pos_x+72} {pos_y+48}"
        
        # Add a default barchart element.
        ET.SubElement(node_elem, "barchart", {"active": "true", "width": "128", "height": "128"})
        
        # Increment positions for a simple layout.
        pos_x += 100
        pos_y += 50
    
    # Convert the ElementTree to a pretty-printed XML string.
    xml_str = ET.tostring(root, encoding="utf-8")
    parsed = minidom.parseString(xml_str)
    pretty_xml_str = parsed.toprettyxml(indent="    ", encoding="UTF-8")
    
    # Write the XML string to the specified file.
    with open(filename, "wb") as f:
        f.write(pretty_xml_str)
       
```

Nimish-4:
### Issue number(s) that this pull request fixes (and reference implementation)
- Fixes #1967 

### List of changes remaining
- Handling extensions (spacing)
- Handling deterministic elements in the .xdsl file
- Testing, docstrings

### DIFF
--- a/pgmpy/readwrite/XDSL.py
+++ b/pgmpy/readwrite/XDSL.py
@@ -1,0 +1,252 @@
+import xml.dom.minidom as md
+import xml.etree.ElementTree as etree
+from io import BytesIO
+from itertools import chain
+
+import networkx as nx
+import numpy as np
+
+from pgmpy.factors.discrete import State, TabularCPD
+from pgmpy.models import BayesianNetwork
+
+
+class XDSLReader(object):
+    def __init__(self, path=None, string=None):
+        if path:
+            self.network = etree.ElementTree(file=path).getroot()  # .find("smile")
+        elif string:
+            self.network = etree.fromstring(string.decode("utf-8"))  # # .find("smile")
+        else:
+            raise ValueError("Must specify either path or string")
+        self.network_name = self.network.attrib["id"]
+        self.variables = self.get_variables()
+        self.variable_parents = self.get_parents()
+        self.edge_list = self.get_edges()
+        self.variable_states = self.get_states()
+        self.variable_CPD = self.get_values()
+        # self.variable_property = self.get_property()
+        # self.state_names = self.get_states()
+
+    def get_parents(self):
+        variable_parents = {}
+        for node in self.network.find("nodes").findall("cpt"):
+
+            parents = node.find("parents")
+            if parents is not None:
+                variable_parents[node.attrib["id"]] = parents.text.split(" ")
+            else:
+                variable_parents[node.attrib["id"]] = []
+
+        return variable_parents
+
+    def get_variables(self):
+        nodes = self.network.find("nodes")
+        variables = [variable.attrib["id"] for variable in nodes.findall("cpt")]
+        return variables
+
+    def get_edges(self):
+        edge_list = [
+            [value, key]
+            for key in self.variable_parents
+            for value in self.variable_parents[key]
+        ]
+        return edge_list
+
+    def get_states(self):
+        nodes = self.network.find("nodes").findall("cpt")
+        variable_states = {}
+        for cpt in nodes:
+            variable_states[cpt.attrib["id"]] = [
+                state.attrib["id"] for state in cpt.findall("state")
+            ]
+        return variable_states
+
+    def get_values(self):
+        variable_CPD = {}
+        nodes = self.network.find("nodes").findall("cpt")
+        for cpt in nodes:
+
+            combined_prob = cpt.find("probabilities")
+            num_states = len([state for state in cpt.findall("state")])
+            cpd_arr = [[] for k in range(num_states)]
+            prob_values = combined_prob.text.split(" ")
+
+            for j in range(num_states):
+                for i in range(j, len(prob_values), num_states):
+                    cpd_arr[j].append(prob_values[i])
+
+            variable_CPD[cpt.attrib["id"]] = cpd_arr
+        return variable_CPD
+
+    def get_model(self, state_name_type=str):
+        model = BayesianNetwork()
+        model.add_nodes_from(self.variables)
+        model.add_edges_from(self.edge_list)
+        model.name = self.network_name
+
+        tabular_cpds = []
+        for var, values in self.variable_CPD.items():
+            evidence_card = [
+                len(self.variable_states[evidence_var])
+                for evidence_var in self.variable_parents[var]
+            ]
+
+            cpd = TabularCPD(
+                var,
+                len(self.variable_states[var]),
+                values,
+                evidence=self.variable_parents[var],
+                evidence_card=evidence_card,
+                state_names={
+                    var: list(map(state_name_type, self.variable_states[var]))
+                    for var in chain([var], self.variable_parents[var])
+                },
+            )
+            tabular_cpds.append(cpd)
+
+        model.add_cpds(*tabular_cpds)
+
+        """for node, properties in self.variable_property.items():
+            for prop in properties:
+                if prop is not None:
+                    prop_name, prop_value = map(lambda t: t.strip(), prop.split("="))
+                    model.nodes[node][prop_name] = prop_value"""
+
+        return model
+
+
+class XDSLWriter(object):
+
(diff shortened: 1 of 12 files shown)


pgmpy/tests/test_readwrite/test_xdsl.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+import xml.etree.ElementTree as etree
+
+import numpy as np
+import numpy.testing as np_test
+
+from pgmpy import config
+from pgmpy.factors.discrete import TabularCPD
+from pgmpy.models import BayesianNetwork
+from pgmpy.readwrite.XDSL import XDSLReader, XDSLWriter
```
ankurankan: You would need to add both these classes to `pgmpy/readwrite/__init__.py` to make the import path a bit short as: `from pgmpy.readwrite import XDSLReader, XDSLWriter`.

pgmpy/readwrite/XDSL.py
```diff
@@ -0,0 +1,252 @@
...
+        xml_str = etree.tostring(self.root, encoding="utf-8")
+        parsed = md.parseString(xml_str)
+        pretty_xml_str = parsed.toprettyxml(indent="    ", encoding="UTF-8")
+
+        # Write the XML string to the specified file.
+        if filename is not None:
+            with open(filename, "wb") as f:
+                f.write(pretty_xml_str)
+
+        return pretty_xml_str
```
ankurankan: I don't think we should return this.

pgmpy/readwrite/XDSL.py
```diff
@@ -0,0 +1,252 @@
...
+        variable_tag = {}
+        nodes_elem = etree.SubElement(self.root, "nodes")
+
+        for var in list(nx.topological_sort(self.model)):
+            variable_tag[var] = etree.SubElement(nodes_elem, "cpt", {"id": var})
+            # etree.SubElement(variable_tag[var], "NAME").text = var
+
+        return variable_tag
+
+    def get_states(self):
```
ankurankan: This function isn't just gettting states

ankurankan (changes requested):
@Nimish-4 There seems to be a problem with the order in which the values are getting read from the file. You can reproduce it using:
```python
In [39]: from pgmpy.utils import get_example_model                                                                                                                                            
                                                                                                                                                                                              
In [40]: model = get_example_model('alarm')                                                                                                                                                   
                                                                                                                                                                                              
In [41]: from pgmpy.readwrite.XDSL import XDSLReader, XDSLWriter                                                                                                                              
                                                                  
In [42]: XDSLWriter(model).write_xdsl('alarm.xdsl')

In [43]: read_model = XDSLReader('alarm.xdsl').get_model()

In [45]: read_model == model
Out[45]: False

In [51]: for node in read_model.nodes():
    ...:     orig_cpd = model.get_cpds(node)
    ...:     read_cpd = read_model.get_cpds(node)
    ...:     print(orig_cpd == read_cpd)
    ...: 
True
True
True
True
True
True
True
True
True
True
True
True
True
False
False
True
True
False
True
True
True
False
False
False
False
False
False
True
False
False
False
True
False
False
False
False
False

In [52]: for node in read_model.nodes():
    ...:     orig_cpd = model.get_cpds(node)
    ...:     read_cpd = read_model.get_cpds(node)
    ...:     if not (orig_cpd == read_cpd):
    ...:         print(orig_cpd.variables)
    ...:         print(read_cpd.variables)
    ...:         print(orig_cpd.values)
    ...:         print(read_cpd.values)
    ...:         break
    ...: 
['LVEDVOLUME', 'HYPOVOLEMIA', 'LVFAILURE']
['LVEDVOLUME', 'LVFAILURE', 'HYPOVOLEMIA']
[[[0.95 0.01]
  [0.98 0.05]]

 [[0.04 0.09]
  [0.01 0.9 ]]

 [[0.01 0.9 ]
  [0.01 0.05]]]
[[[0.95 0.01]
  [0.98 0.05]]

 [[0.04 0.09]
  [0.01 0.9 ]]

 [[0.01 0.9 ]
  [0.01 0.05]]]
   
```
As you can see in the example above, the order of variables is different but both of them have the same values array which should not be the case.

I would also suggest adding more tests to explicitly check for expected values to catch these cases.

ankurankan:
@Nimish-4 Looks correct now. Should be ready once the other comments and tests are added.

ankurankan (dismissed):

ankurankan (dismissed):

PR #2004, opened 2025-04-06
Enhances XDSL readwrite methods and tests

Nimish-4:


pgmpy/readwrite/XDSL.py
```diff
@@ -361,7 +360,7 @@ def get_cpds(self):
 
             # Add the <probabilities> element.
             probs_elem = etree.SubElement(cpt_elem, "probabilities")
-            values = np.array(cpd.get_values())
+            values = compat_fns.to_numpy(cpd.get_values())
```
ankurankan: @Nimish-4 Shouldn't this be dependent on the compute backend? A numpy array if the backend is numpy otherwise a torch Tensor.

pgmpy/utils/compat_fns.py
```diff
@@ -21,7 +21,12 @@ def copy(arr):
         elif isinstance(arr, (int, float)):
             return deepcopy(arr)
     else:
-        return torch.tensor(arr, dtype=config.get_dtype(), device=config.get_device())
+        if isinstance(arr, torch.Tensor):
+            return arr.detach().clone()
```
ankurankan, line 25: I am not sure why this is needed. Is there a case where the previous version breaks? If the problem is that the type and device gets changed, I think a better solution would be to just replace the previous code with `detach().clone()`.

ankurankan (changes requested):

pgmpy/readwrite/XDSL.py
Nimish-4: @ankurankan Yeah, the `cpd.get_values()` should return either a numpy or a Torch array depending on the backend. I just thought there would not be much use of a Torch array here and kept the original behaviour from the code. I'll make the change for consistency.

pgmpy/utils/compat_fns.py
Nimish-4: In the normal `torch.tensor()` case, it threw a user warning when `arr` is a torch tensor itself. So I replaced it with `.detach().clone()` . The second case is to handle when arr is some dtype other than a torch tensor ... I think such a situation did come up in of the tests somewhere.

ankurankan (approved):

PR #2023, opened 2025-04-29
Refactor Meek's rules

### LINKED ISSUE #1911: Refactor PC.apply_orientation_rules
The `PC.apply_orientation_rules` method currently works on the PDAG object that can contain both directed and undirected edges. This makes the algorithm implementation a bit difficult to read because of relying a lot on `successors` and `predecessors` methods. We would like to refactor this to simplify the implementation by utilizing two separate graphs, a directed one and an undirected one.

ankurankan:

- Fixes #1911

- 
-
-

### DIFF
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -1490,14 +1490,8 @@ class PDAG(nx.DiGraph):
                 if len(undirected_nbs) < 3:
                     continue
 
-                for y, z, w in itertools.product(
-                    undirected_nbs, undirected_nbs, undirected_nbs
-                ):
-                    if (
-                        len(set([y, z, w])) == 3  # No repeated variables
-                        and pdag.has_directed_edge(y, w)
-                        and pdag.has_directed_edge(z, w)
-                    ):
+                for y, z, w in itertools.permutations(undirected_nbs, 3):
+                    if pdag.has_directed_edge(y, w) and pdag.has_directed_edge(z, w):
                         pdag.orient_undirected_edge(x, w, inplace=True)
                         changed = True
                         if debug:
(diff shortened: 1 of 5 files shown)


pgmpy/base/DAG.py
```diff
@@ -1327,6 +1376,154 @@ def copy(self):
...
+
+        Returns
+        -------
+        True, if no new V-structures are formed.
+        False, if the orientation u -> v would lead to creation of a new V-structure.
+        """
+        for node in self.directed_parents(v):
+            if (node != u) and (not self.is_adjacent(u, node)):
+                return False
+        return True
```
Nimish-4: I think the return values need to be exchanged, no? Returning `True` when we find a non adjacent node that is also a parent?

ankurankan: @Nimish-4 Thanks. Fixed now.

pgmpy/base/DAG.py
```diff
@@ -1327,6 +1376,183 @@ def copy(self):
...
+                        for d in pdag.directed_parents(c):
+                            if b == d or pdag.is_adjacent(b, d):
+                                continue  # b adjacent d ⇒ rule not applicable
+
+                            # find nodes a that are undirected neighbor to b, c, d
+                            cand = set(pdag.undirected_neighbors(b)).intersection(
+                                pdag.undirected_neighbors(c),
+                                pdag.undirected_neighbors(d),
+                            )
+                            for a in cand:
```
Nimish-4, line 1571: In the [original Meek's work](https://arxiv.org/pdf/1302.4972), `a` and `c` are allowed to have a directed edge between them too ... I don't know how significant this is. Maybe it was slightly inaccurate, or maybe it was refined in more recent work that you have pointed out before.

Nimish-4 (dismissed):

pgmpy/base/DAG.py
ankurankan: Thanks for catching this. I had missed it. Fixed now.

PR #2035, opened 2025-05-04
feat: optimize sum utility

### LINKED ISSUE #2009: Optimization Suggestion: Replace np.sum(arr) with arr.sum() when type is known
https://github.com/pgmpy/pgmpy/blob/a5ebed557bc2c1a9f103417778fd5c8f3e74cef4/pgmpy/utils/compat_fns.py#L134
Hi, I’d like to suggest a small optimization in the following code:
```
if isinstance(arr, np.ndarray):
    return np.sum(arr)
```
This can be safely and more efficiently written as:
```
if isinstance(arr, np.ndarray):
    return arr.sum()

```
Both implementations produce the same result when arr is a NumPy array, but the latter avoids unnecessary overhead. Internally, np.sum(arr) performs additional type dispatching and checks to support array-like objects, duck-typing, and function overrides through __array_function__. Even if the type is already confirmed to be ndarray, this path still incurs a general-purpose ufunc resolution process. In contrast, arr.sum() directly invokes the bound C-level implementation of PyArray_Sum, bypassing all dispatch layers and operating with minimal overhead. Since arr is already guaranteed to be a NumPy array, calling its method directly yields better performance without affecting correctness.

mdrazak2001:

- Fixes #2009 

- changed `np.sum(arr)` to `arr.sum()` to optimize performance

### DIFF
--- a/pgmpy/utils/compat_fns.py
+++ b/pgmpy/utils/compat_fns.py
@@ -139,7 +139,7 @@ def exp(arr):
 
 def sum(arr):
     if isinstance(arr, np.ndarray):
-        return np.sum(arr)
+        return arr.sum()
     else:
         return torch.sum(arr)
 


ankurankan (approved):
Thanks for the PR @mdrazak2001

PR #2036, opened 2025-05-04
Fixes missing data error in `ExpectationMaximization`

### LINKED ISSUE #2012: Fail to using ExpectationMaximization with missing data
https://github.com/pgmpy/pgmpy/blob/2a58db86efc53a69198dab72b5e34fc77e5af9a5/pgmpy/estimators/EM.py#L16-L34

When I tried to use EM with missing data, it failed and get the following report

Image

My original dataframe is shown in the below, which follows the data requirement description in source code.

Image

Prtm2110:

- Refs: #2012

- Drops rows which contain NaN in non completely empty column. 
- I will add tests later

### DIFF
--- a/pgmpy/estimators/EM.py
+++ b/pgmpy/estimators/EM.py
@@ -1,6 +1,5 @@
 from itertools import chain, product
 from math import log
-from warnings import warn
 
 import numpy as np
 import pandas as pd
@@ -11,6 +10,7 @@ from pgmpy import config
 from pgmpy.base import DAG
 from pgmpy.estimators import MaximumLikelihoodEstimator, ParameterEstimator
 from pgmpy.factors.discrete import TabularCPD
+from pgmpy.global_vars import logger
 from pgmpy.models import DiscreteBayesianNetwork
 
 
@@ -68,9 +68,12 @@ class ExpectationMaximization(ParameterEstimator):
 
         # Drop rows that have missing values in those partially missing columns
         if partial_missing_cols:
+            original_rows = data.shape[0]
             data = data.dropna(subset=partial_missing_cols)
-            warn(
-                "Rows with missing values in partially missing columns were dropped from the dataset."
+            remaining_rows = data.shape[0]
+            rows_dropped = original_rows - remaining_rows
+            logger.warning(
+                f"{rows_dropped} Rows with missing values in partially missing columns were dropped from the dataset."
             )
 
         super(ExpectationMaximization, self).__init__(model, data, **kwargs)
--- a/pgmpy/estimators/EM.py
+++ b/pgmpy/estimators/EM.py
@@ -63,27 +63,27 @@ class ExpectationMaximization(ParameterEstimator):
             model_bn.latents = model.latents
             model = model_bn
 
-        # Handle fully missing columns
-        full_missing_cols = [col for col in data.columns if data[col].isnull().all()]
-        for col in full_missing_cols:
-            if col not in model.latents:
-                logger.warning(
-                    f"Column '{col}' has all missing values and is not marked as latent. "
-                    "Treating it as a latent variable."
-                )
-                model.latents.add(col)
+        # Drop fully missing columns and treat them as latent if not already
+        original_cols = set(data.columns)
+        data = data.dropna(axis=1, how="all")
+        dropped_cols = original_cols - set(data.columns)
+        new_latents = [col for col in dropped_cols if col not in model.latents]
 
-        data = data.drop(
-            columns=full_missing_cols
-        )  # Drop fully missing columns from the data (now considered latent)
-
-        # Drop rows that have missing values in those partially missing columns
-        if data.isnull().any().any():
-            original_rows = data.shape[0]
-            data = data.dropna()
-            dropped = original_rows - data.shape[0]
+        if new_latents:
             logger.warning(
-                f"{dropped} Rows with missing values in partially missing columns were dropped from the dataset."
+                f"Columns {new_latents} have all missing values and are not marked as latent. "
+                "Treating them as latent variables."
+            )
+            model.latents.update(new_latents)
+
+        # Drop rows with any missing values in partially observed columns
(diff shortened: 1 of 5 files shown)


pgmpy/estimators/EM.py
```diff
@@ -62,6 +63,16 @@ def __init__(self, model, data, **kwargs):
             model_bn.latents = model.latents
             model = model_bn
 
+        partial_missing_mask = data.isnull().any(axis=0) & ~data.isnull().all(axis=0)
```
ankurankan: @Prtm2110 Can we not just do `data.dropna()` to remove the rows with any nan values? For the warning, we can then compare the size of the dataframe before and after the dropna, and show a warning if the new size is smaller.

pgmpy/estimators/EM.py
```diff
@@ -62,6 +63,16 @@ def __init__(self, model, data, **kwargs):
...
+        partial_missing_cols = partial_missing_mask[partial_missing_mask].index.tolist()
+
+        # Drop rows that have missing values in those partially missing columns
+        if partial_missing_cols:
+            data = data.dropna(subset=partial_missing_cols)
+            warn(
```
ankurankan: We use the logging module in pgmpy to give control to the user on what messages get printed. This line would need to be:
```python
from pgmpy.global_vars import logger
logger.warning(... Message ...)
```

ankurankan: I also think it would be nice to include the number of rows that were dropped in the warning message as well to make it a bit more informative.

ankurankan (changes requested):

pgmpy/estimators/EM.py
Prtm2110: Yes, that would be greate! Thank you for the reviews. 
Should I change this as well https://github.com/Prtm2110/pgmpy/blob/c6faa519a7e97421254e6f7fdca43529d01bdc36/pgmpy/factors/discrete/CPD.py#L7 to use logger?

pgmpy/estimators/EM.py
Prtm2110: > Can we not just do data.dropna() to remove the rows with any nan values?

Yes we can but suppose if we have a entire column full of NaNs then `data.dropna()` will remove all the rows (i.e complete dataframe. That's why I created a mask so that we can ignore those columns (as mentioned in https://github.com/pgmpy/pgmpy/issues/2012#issuecomment-2834579839) and remove just those rows. If I am missing something please let me know :)

ankurankan: On second thought, there are a couple of things to consider here: 1) We do not expect any columns in the data for latent variables (`model.latents`). 2) If a full column is missing but the variable is not latent, we won't be able to  process it as we well need to know the number of possible states of the variable (so essentially need to treat it as latent).

What I think we can do is the following:
1) In the input, first check for any nan columns. If the column is in `model.latents`, it's good; otherwise, throw a warning and add the variable to latents. We can then initialize the number of states for this variable as well in Line 218.
2) The remaining dataframe would only have non-latent variables (so no full column nans). Here, drop any of the rows with missing values with a warning. 

What do you think?

pgmpy/estimators/EM.py
ankurankan: Yes. Thanks for spotting it.

pgmpy/estimators/EM.py
Prtm2110: Yes, I agree I have made the changes - if full nan col is present then check it in latents if it's not in there then add it in latent and then drop all full nan cols and `dropna()` on remaining dataframe. Let me know if you need further modifications.

pgmpy/estimators/EM.py
```diff
@@ -62,6 +63,29 @@ def __init__(self, model, data, **kwargs):
             model_bn.latents = model.latents
             model = model_bn
 
+        # Handle fully missing columns
+        full_missing_cols = [col for col in data.columns if data[col].isnull().all()]
```
ankurankan: Isn't a simpler way to drop Nan columns is to use pandas' dropna function? We can do something like: `data.dropna(axis=1, how='all')`. This will remove any Nan columns. We can then compare this new dataframe to original, and if they are different show the warning and add the latent variables.

pgmpy/estimators/EM.py
```diff
@@ -62,6 +63,29 @@ def __init__(self, model, data, **kwargs):
...
+        for col in full_missing_cols:
+            if col not in model.latents:
+                logger.warning(
```
ankurankan: This would end up giving multiple warnings if there are more than one columns. If we do it the way I have suggested above, we can show all the columns in a single warning.

pgmpy/estimators/EM.py
```diff
@@ -62,6 +63,29 @@ def __init__(self, model, data, **kwargs):
...
+                    "Treating it as a latent variable."
+                )
+                model.latents.add(col)
+
+        data = data.drop(
+            columns=full_missing_cols
+        )  # Drop fully missing columns from the data (now considered latent)
+
+        # Drop rows that have missing values in those partially missing columns
+        if data.isnull().any().any():
```
ankurankan: Here as well, we can first call the `data.dropna()` and then compare it to the previous dataframe. If there is a difference, show the warning.

ankurankan (changes requested):
@Prtm2110 Thanks for the changes. I have added a few comments for mostly improving code readability. Could you also add some test cases for these changes?

pgmpy/estimators/EM.py
Prtm2110: Ah, I wasn't aware of this. Thank you! The code is now more readable. Let me know if you need further modifications.

ankurankan:
@Prtm2110 Could you check why the tests are failing?

Prtm2110:
Yes, fixed it was due to the change to the logger in CPD.

ankurankan (approved):
Looks good now. Thanks for the PR @Prtm2110

PR #2037, opened 2025-05-04
Fix: Include example_models data files in package-data

Susmita331:
###  Why Not Just Use `"utils/example_models/*"`?

Although `utils/example_models/*` seems like it should include everything in that folder, in practice it **doesn’t reliably package files like `.bif.gz`** when building and installing the library.

####  The Problem

The `*` wildcard is interpreted during the packaging process, but:

- It **doesn’t always catch compressed files** like `.bif.gz` or other non-standard extensions
- `setuptools` may skip files it doesn’t recognize as package data unless they are **explicitly listed**
- Behavior can be **inconsistent across environments and tools** (`pip install`, `python -m build`, etc.)

So even though the files are present in the source tree, they may be **excluded from the final wheel (.whl)**, causing issues for downstream users who install the library with pip and expect those files to be available.

#### The Fix

To make the inclusion consistent and reliable, I updated the `pyproject.toml` with **explicit file patterns**:

```toml
[tool.setuptools.package-data]
pgmpy = [
  "utils/example_models/*.bif.gz",
  "utils/example_models/*.json",
  "utils/example_models/*.txt"
]

[tool.setuptools]
include-package-data = true
```
This tells setuptools exactly which files to include, avoiding any ambiguity with wildcard expansion or compressed file types.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,97 @@
+[project]
+name = "pgmpy"
+version = "1.0.0"
+description = "Python Library for Causal and Probabilistic Modeling using Bayesian Networks"
+readme = "README.md"
+license = { file = "LICENSE" }
+authors = [
+  { name = "Ankur Ankan", email = "ankurankan@gmail.com" }
+]
+requires-python = ">=3.10,<3.14"
+classifiers = [
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Intended Audience :: Education",
+  "Intended Audience :: Science/Research",
+  "Operating System :: Unix",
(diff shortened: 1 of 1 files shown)


ankurankan (changes requested):
@Susmita331 Thanks for the PR but I think you have accidentally committed additional files on your branch. We only need to make the changes in the `pyproject.toml`. Can you remove the other files?

Susmita331:
okay…I am fixing that

Susmita331:
@ankurankan I’ve cleaned up the `fix/include-example-models` branch — it now contains only the intended `pyproject.toml` change to include the `example_models` directory in the wheel. I rebased it properly on `master`, removed all unrelated files, and force-pushed the cleaned branch to my fork.

Could you please confirm if everything looks good on your end? Just want to make sure the updated PR reflects the clean state now. Thanks again for your guidance!

PR #2044, opened 2025-05-05
Todo Fix - Add warning when state names are modified in XMLBIF writer

Akhilesh-max:

- Fixed todo in XMLBIF.py via adding warning when state names are modified in XMLBIF writer.

### DIFF
--- a/pgmpy/readwrite/XMLBIF.py
+++ b/pgmpy/readwrite/XMLBIF.py
@@ -17,6 +17,7 @@ except ImportError as e:
 from pgmpy.factors.discrete import State, TabularCPD
 from pgmpy.models import DiscreteBayesianNetwork
 from pgmpy.utils import compat_fns
+from pgmpy.global_vars import logger
 
 
 class XMLBIFReader(object):
@@ -396,7 +397,6 @@ class XMLBIFWriter(object):
         XMLBIF states must start with a letter an only contain letters,
         numbers and underscores.
         """
-        # TODO: Throw a warning that the state names are going to be modified instead of silently modifying it.
         s = str(state_name)
         s_fixed = (
             pp.CharsNotIn(pp.alphanums + "_")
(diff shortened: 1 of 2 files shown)


pgmpy/readwrite/XMLBIF.py
```diff
@@ -3,6 +3,7 @@
 import xml.etree.ElementTree as etree
 from io import BytesIO
 from itertools import chain
+import warnings
```
Nimish-4: pgmpy uses the global (in terms of configuration) pgmpy logger for warnings instead of directly using the warnings module. For an example, you can look into the XDSL class or just searh for warnings in the codebase.

Nimish-4 (changes requested):

pgmpy/readwrite/XMLBIF.py
```diff
@@ -405,6 +405,13 @@ def _make_valid_state_name(self, state_name):
         )
         if not s_fixed[0].isalpha():
             s_fixed = s_fixed
+
+        if s != s_fixed:
+            warnings.warn(
```
Nimish-4: You should also write a test to ensure that the warning is triggered.

ankurankan: @Akhilesh-max You can add a simple test for the `make_valid_state_name` function in TestXMLBIFReaderMethods: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/tests/test_readwrite/test_XMLBIF.py#L129. You can use a `state_name` such that it triggers the warning.

Nimish-4:
@Akhilesh-max I think a test case is still missing ... it should also fix the codecov warning about test coverage.

Akhilesh-max:
Hi @Nimish-4 @ankurankan, thank you for the reviews. It took me a while to go through the codecov reviews. 
Please take a look when get a chance. 

Thanks!!

pgmpy/readwrite/XMLBIF.py
Akhilesh-max: done

ankurankan (approved):
Looks good to me. Thanks @Akhilesh-max for the PR.

Issue #1983, opened 2025-03-20
Reduce dependencies for the package


ankurankan:
1. pyro can be moved to optional dependencies as only used for FunctionalBN
2. Very few functions of statsmodels are used, can replace them with our own implementations? 
3. Very few functions of sklearn are used, can replace them with our own implementations? 
4. Check if pytorch is needed if the chosen backend is numpy. Then pytorch can be moved to optional dependencies if using torch backend.

StefanNeu:
In the last days I did a quick, dirty fix on my fork of pgmpy, where I tried to remove PyTorch as dependency, but I am not sure how to do this properly.

Should the presence of a PyTorch installation be saved in pgmpy.config? And how should imports then be handled? Still with a try-except or something like 
```
if config.HAS_TORCH():
    import torch
```
?

fkiraly:
In `sktime`, we have a suite of tools around soft dependency management (part of `scikit-base`), we have spent a lot of time thinking about soft dependency management and isolation. One utility is used idiomatically via:

```python
if _check_soft_dependencies("torch"):
    import torch
```

If you need `torch` to be an object, or another object (e.g., if using in inheritance or decorator), another option is a more recent utility `_safe_import`, which is used
```python
nn = _safe_import("torch.nn")
```
which will return a mock-like object that can be used anywhere without raising exceptions.

Here is the full docstring for `_check_soft_dependencies`:

```
def _check_soft_dependencies(
    *packages,
    severity="error",
    obj=None,
    msg=None,
    normalize_reqs=True,
):
    """Check if required soft dependencies are installed and raise error or warning.

    Parameters
    ----------
    packages : str or list/tuple of str nested up to two levels
        str should be package names and/or package version specifications to check.
        Each str must be a PEP 440 compatible specifier string, for a single package.
        For instance, the PEP 440 compatible package name such as ``"pandas"``;
        or a package requirement specifier string such as ``"pandas>1.2.3"``.
        arg can be str, kwargs tuple, or tuple/list of str, following calls are valid:

        * ``_check_soft_dependencies("package1")``
        * ``_check_soft_dependencies("package1", "package2")``
        * ``_check_soft_dependencies(("package1", "package2"))``
        * ``_check_soft_dependencies(["package1", "package2"])``
        * ``_check_soft_dependencies(("package1", "package2"), "package3")``
        * ``_check_soft_dependencies(["package1", "package2"], "package3")``
[...]

PR #2045, opened 2025-05-05
Added `_check_soft_dependencies` in `dependencies` module. 

Prtm2110:

- Related to:  #1983

- [x] I added `_check_soft_dependencies` function greatly inspired from sktime.
- [ ] Use `_check_soft_dependencies` wherever necessary and possible - any suggestions are most welcome. 
- [ ] Add tests for the new module and its functions. I will add them once we confirm these changes.

### DIFF
--- a/pgmpy/utils/dependencies.py
+++ b/pgmpy/utils/dependencies.py
@@ -1,0 +1,351 @@
+"""Utility to check soft dependency imports, and raise warnings or errors."""
+
+import sys
+import warnings
+from functools import lru_cache
+from importlib.util import find_spec
+from inspect import isclass
+
+from packaging.requirements import InvalidRequirement, Requirement
+from packaging.specifiers import Specifier, SpecifierSet
+from packaging.version import InvalidVersion, Version
+
+
+def _check_soft_dependencies(
+    *packages,
+    severity="error",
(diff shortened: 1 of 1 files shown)


fkiraly (changes requested):
This is copy-pasted from `sktime`, so at least credits should be included (in line with the `sktime` BSD-3 license which requires attribution).

An alternative approach which you could also consider is making `scikit-base` a dependency - it does not itself have dependencies, and a number of utilities useful for creating `sklearn`-like interfaces, as well as a maintained instance of this utility - in `skbase.utils.dependencies`.

Prtm2110:
> This is copy-pasted from `sktime`, so at least credits should be included (in line with the `sktime` BSD-3 license which requires attribution).

Yes, I am sorry I completely agree  I was going to do it after asking the licensing part as I am not very aware of it. Thanks for letting me know.

Prtm2110:
@ankurankan the code I provided is from sktime and it comes with a lot of functionalities and maybe if we don't require it if so then I can create a smaller function which just check if the dependency is installed in the current env or else I will add tests to the current implementation in the PR.

ankurankan:
Done in #2324 and #2326. Closing.

PR #2048, opened 2025-05-06
Updated

Susmita331:
This PR addresses a previously unhandled edge case in the VariableElimination.query() method where users could call the method with an empty list of variables:

inference.query(variables=[], evidence={"A": 1})
Such a query is logically invalid, as it attempts to compute a probability distribution over “nothing,” which has no meaningful interpretation.

### DIFF
--- a/pgmpy/tests/test_inference/test_ExactInference.py
+++ b/pgmpy/tests/test_inference/test_ExactInference.py
@@ -3,6 +3,7 @@ import unittest
 
 import numpy as np
 import numpy.testing as np_test
+from pgmpy.utils import get_example_model
 
 from pgmpy.factors.discrete import DiscreteFactor, TabularCPD
 from pgmpy.inference import BeliefPropagation, VariableElimination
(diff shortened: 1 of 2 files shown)


pgmpy/tests/test_inference/test_ExactInference.py
```diff
@@ -47,6 +47,18 @@ def setUp(self):
     # All the values that are used for comparison in the all the tests are
     # found using SAMIAM (assuming that it is correct ;))
 
+    def test_query_raises_for_empty_variables(self):
+        model = DiscreteBayesianNetwork([("A", "B")])
```
ankurankan: An easier way to create a model is to import an existing one. It will also simplify the code. For example:
```
from pgmpy.utils import get_example_modele
model = get_example('earthquake')
```

ankurankan:
@Susmita331 I think this PR looked good. Any reason for closing this?

Susmita331:
It was failing

Susmita331:
I don’t have any idea why all the test cases got canceled

ankurankan:
@Susmita331 It's failing because of code formatting issues. You can run the following to fix it:
```bash
black pgmpy/inference/ExactInference.py
black pgmpy/tests/test_inference/test_ExactInference.py
```

For future commits, you can run `pre-commit install` in the base pgmpy directory to install formatting checks. That will automatically fix these before every commit.

>  I don’t have any idea why all the test cases got canceled

Our github actions is setup to cancel all tests if anything fails (to save compute time). If you click on the one that shows as failed, you will be able to see the logs and find why it failed.

ankurankan:
Thanks for the fix @Susmita331

PR #2050, opened 2025-05-06
Allow ClusterGraph.add_factors to accept subset-scope factors

Susmita331:
This pull request updates the add_factors method in ClusterGraph to support adding factors whose variable scopes are subsets of existing cluster nodes, rather than requiring an exact match. This change makes the behavior consistent with the formal definition of cluster graphs, where each factor only needs to be associated with a cluster node that contains all the variables in its scope.

Previously, the method raised a ValueError if the factor's variable set was not exactly equal to any node in the graph. As a result, valid models where a factor's scope was just a part of a larger cluster (like a factor over ['a', 'b'] for a node ['a', 'b', 'c']) were unnecessarily rejected.

The fix replaces the equality check with a subset check using issubset(). Now, as long as the factor’s variables are included within any existing cluster node, it will be accepted.

In addition to the fix,  I added a corresponding unit test named test_add_factor_with_subset_scope. It checks that the graph correctly accepts multiple factors whose scopes are subsets of a single cluster node.

### DIFF
--- a/pgmpy/models/ClusterGraph.py
+++ b/pgmpy/models/ClusterGraph.py
@@ -154,12 +154,14 @@ class ClusterGraph(UndirectedGraph):
         """
         for factor in factors:
             factor_scope = set(factor.scope())
-            #nodes = [set(node) for node in self.nodes()]
-            found_cluster = any(factor_scope.issubset(set(node)) for node in self.nodes())
-            #if factor_scope not in nodes:
-                #raise ValueError(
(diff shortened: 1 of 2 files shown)


ankurankan:
@Susmita331 While in theory, we should allow adding such factors to ClusterGraphs, this is really not required as ClusterGraph is only used by JunctionTree (which doesn't need such factors) and we do not have any plans to expand thefeatures of ClusterGraph. I would suggest opening an issue and having a discussion before deciding to work on any issue.

PR #2051, opened 2025-05-06
Fix query() crash on invalid variable

rushi-k12:
This PR addresses a bug in the query() method of the VariableElimination class in pgmpy. Previously, if a user passed a variable name that does not exist in the model, it would result in an unhandled exception such as:

makefile
Copy
Edit
ValueError: Variable not found in the model: X
This behavior was confusing and did not provide clear feedback to the user.

Fix:
Added a validation step that checks whether each variable in the variables list exists in the model.

If any unknown variables are found, a ValueError is raised with a helpful message.

Benefits:
Improves user experience with clearer error messages.

Prevents unexpected crashes by handling invalid input early.

Aligns with expected behavior for model query interfaces.

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -248,134 +248,104 @@ class VariableElimination(Inference):
         joint=True,
         show_progress=True,
     ):
-        """
-        Parameters
-        ----------
-        variables: list
-            list of variables for which you want to compute the probability
+    """
+    Parameters
+    ----------
+    variables: list
+        list of variables for which you want to compute the probability
 
-        evidence: dict
-            a dict key, value pair as {var: state_of_var_observed}
-            None if no evidence
+    evidence: dict
+        a dict key, value pair as {var: state_of_var_observed}
+        None if no evidence
 
-        virtual_evidence: list (default:None)
-            A list of pgmpy.factors.discrete.TabularCPD representing the virtual
-            evidences.
+    virtual_evidence: list (default:None)
+        A list of pgmpy.factors.discrete.TabularCPD representing the virtual
+        evidences.
 
-        elimination_order: str or list (default='greedy')
-            Order in which to eliminate the variables in the algorithm. If list is provided,
-            should contain all variables in the model except the ones in `variables`. str options
-            are: `greedy`, `WeightedMinFill`, `MinNeighbors`, `MinWeight`, `MinFill`. Please
-            refer https://pgmpy.org/exact_infer/ve.html#module-pgmpy.inference.EliminationOrder
-            for details.
+    elimination_order: str or list (default='greedy')
+        Order in which to eliminate the variables in the algorithm.
 
-        joint: boolean (default: True)
-            If True, returns a Joint Distribution over `variables`.
-            If False, returns a dict of distributions over each of the `variables`.
+    joint: boolean (default: True)
+        If True, returns a Joint Distribution over `variables`.
 
-        show_progress: boolean
-            If True, shows a progress bar.
+    show_progress: boolean (default: True)
+        If True, shows a progress bar.
+    """
 
-        Examples
-        --------
-        >>> from pgmpy.inference import VariableElimination
-        >>> from pgmpy.models import DiscreteBayesianNetwork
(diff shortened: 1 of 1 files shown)


pgmpy/inference/ExactInference.py
```diff
@@ -248,134 +248,104 @@ def query(
...
-        """
-        Parameters
-        ----------
-        variables: list
-            list of variables for which you want to compute the probability
+    """
+    Parameters
+    ----------
+    variables: list
+        list of variables for which you want to compute the probability
```
ankurankan, line 255: Could you please change this to: `List of variables for which to compute the posterior distribution`

pgmpy/inference/ExactInference.py
```diff
@@ -248,134 +248,104 @@ def query(
...
-            refer https://pgmpy.org/exact_infer/ve.html#module-pgmpy.inference.EliminationOrder
-            for details.
+    elimination_order: str or list (default='greedy')
+        Order in which to eliminate the variables in the algorithm.
 
-        joint: boolean (default: True)
-            If True, returns a Joint Distribution over `variables`.
-            If False, returns a dict of distributions over each of the `variables`.
+    joint: boolean (default: True)
+        If True, returns a Joint Distribution over `variables`.
```
ankurankan, line 269: Could you change this to something like "If True, returns a joint posterior distribution over `variables`. Otherwise, returns marginal posterior distributions over each variable"

pgmpy/inference/ExactInference.py
```diff
@@ -248,134 +248,104 @@ def query(
...
 
-        show_progress: boolean
-            If True, shows a progress bar.
+    show_progress: boolean (default: True)
+        If True, shows a progress bar.
+    """
 
-        Examples
```
ankurankan, line 279: Any reason to remove the example?

pgmpy/inference/ExactInference.py
```diff
@@ -248,134 +248,104 @@ def query(
...
-        >>> import pandas as pd
-        >>> values = pd.DataFrame(np.random.randint(low=0, high=2, size=(1000, 5)),
-        ...                       columns=['A', 'B', 'C', 'D', 'E'])
-        >>> model = DiscreteBayesianNetwork([('A', 'B'), ('C', 'B'), ('C', 'D'), ('B', 'E')])
-        >>> model.fit(values)
-        >>> inference = VariableElimination(model)
-        >>> phi_query = inference.query(['A', 'B'])
-        """
-        evidence = evidence if evidence is not None else dict()
+    evidence = evidence if evidence is not None else dict()
```
ankurankan, line 275: I don't think the indentation of code is correct; also leads to failing tests.

pgmpy/inference/ExactInference.py
```diff
@@ -248,134 +248,104 @@ def query(
...
-            return self.query(
-                variables=variables,
-                evidence={**evidence, **virt_evidence},
-                virtual_evidence=None,
-                elimination_order=elimination_order,
-                joint=joint,
-                show_progress=show_progress,
-            )
+    # ✅ Step 1.5: Check for invalid query variables
+    invalid_vars = [var for var in variables if var not in self.model.nodes()]
```
ankurankan, line 290: It would be better to put it together with `if not variable` as an else statement.

pgmpy/inference/ExactInference.py
```diff
@@ -248,134 +248,104 @@ def query(
...
-            virt_evidence = {"__" + cpd.variables[0]: 0 for cpd in virtual_evidence}
-            return self.query(
-                variables=variables,
-                evidence={**evidence, **virt_evidence},
-                virtual_evidence=None,
-                elimination_order=elimination_order,
-                joint=joint,
-                show_progress=show_progress,
-            )
+    # ✅ Step 1.5: Check for invalid query variables
```
ankurankan, line 289: Please do not use special characters in code.

ankurankan (changes requested):

rushi-k12:
performed changes

ankurankan:
@rushi-k12 I am confused. Should I be looking at this PR or #2055 ?

PR #2052, opened 2025-05-06
Add __slots__ to DiscreteFactor to reduce memory usage

Susmita331:
This pull request introduces __slots__ in the DiscreteFactor class to improve memory efficiency and attribute access performance.

By declaring __slots__, we prevent Python from creating a __dict__ for each instance of the class. This reduces the per-object memory overhead and improves attribute access speed. The optimization is particularly beneficial in applications that involve constructing a large number of factor instances, such as large-scale Bayesian networks or Markov models.


ankurankan:
@Susmita331 Would it be possible for you to do a simple analysis of how much time and memory improvement does using `__slots__` bring?

Susmita331:
yes! I created a benchmark .py file  to assess the performace on my local computer . 
```python
import time
import tracemalloc

NUM_OBJECTS = 100_000

class WithDict:
    def __init__(self, a, b, c):
        self.a = a
        self.b = b
        self.c = c

class WithSlots:
    __slots__ = ("a", "b", "c")
    def __init__(self, a, b, c):
        self.a = a
        self.b = b
        self.c = c

def measure(cls):
    tracemalloc.start()
    start_time = time.perf_counter()
    objs = [cls(1, 2, 3) for _ in range(NUM_OBJECTS)]
    end_time = time.perf_counter()
    current, peak = tracemalloc.get_traced_memory()
    tracemalloc.stop()
    return end_time - start_time, current / 1024, peak / 1024

t_dict, cur_dict, peak_dict = measure(WithDict)
t_slots, cur_slots, peak_slots = measure(WithSlots)

print("=== Without __slots__ ===")
print(f"Time: {t_dict:.4f}s | Current: {cur_dict:.2f} KB | Peak: {peak_dict:.2f} KB")

print("=== With __slots__ ===")
print(f"Time: {t_slots:.4f}s | Current: {cur_slots:.2f} KB | Peak: {peak_slots:.2f} KB")

print("=== Difference ===")
print(f"Memory saved (peak): {peak_dict - peak_slots:.2f} KB")
print(f"Speedup: {(t_dict - t_slots) / t_dict * 100:.2f}%")
```
this is the script

Susmita331:
Screenshot 2025-05-07 at 14 56 13

ankurankan:
@Susmita331 Sorry, I meant benchmarks for DiscreteFactor specifically to understand the impact of the change on our codebase. So, I would like to see how DiscreteFactor operations such as product, marginalize are affected by this change, as well as, how much improvement we see in inference (for example using Variable Elimination).

Susmita331:
### DiscreteFactor Benchmark Results

Ran microbenchmarks and inference on the Asia model comparing DiscreteFactor with and without `__slots__`.

| Operation         | Without `__slots__` | With `__slots__` | Difference |
|------------------|---------------------|------------------|------------|
| product          | 0.000049 sec        | 0.000177 sec     | Slower     |
| reduce           | 0.000013 sec        | 0.000013 sec     | No change  |
| marginalize      | 0.000022 sec        | 0.000155 sec     | Slower     |
| normalize        | 0.000009 sec        | 0.000009 sec     | No change  |
| inference.query  | 0.000383 sec        | 0.000467 sec     | Slower     |

### Observation

In this small-scale benchmark, the version **without `__slots__` was slightly faster** in some operations, which is likely due to the microsecond-level measurements and the fact that Python’s dict access is highly optimized for very small objects.

However, for **large-scale models with thousands of factors**, `__slots__` is expected to offer:
- Reduced per-object memory overhead
- More consistent memory footprint
- Better performance at scale (due to better CPU cache utilization)

Susmita331:
Screenshot 2025-05-08 at 11 12 15

ankurankan:
@Susmita331 I tried to run this benchmark (and simulations) on a much larger network. I do not see much improvement in the results. I have push my larger network code to your branch.

With slots:
```
--- DiscreteFactor micro-benchmarks ---
product: 0.000115 sec
reduce: 0.000048 sec
marginalize: 0.000038 sec

--- Inference benchmark (Variable Elimination) ---
Running inference.query...
inference.query: 0.002961 sec
Simulation Benchmark
Simulation: 0.174256 sec
```
Without slots:
```
--- DiscreteFactor micro-benchmarks ---
product: 0.000122 sec
reduce: 0.000052 sec
marginalize: 0.000053 sec

--- Inference benchmark (Variable Elimination) ---
Running inference.query...
inference.query: 0.002949 sec
Simulation Benchmark
Simulation: 0.181225 sec
```

Susmita331:
@ankurankan Thank you for testing this on your larger network! 

I agree — the time differences in micro-benchmarks (e.g., reduce, marginalize, inference.query) appear small in absolute terms, especially in isolated operations. That makes sense because:
The methods themselves are dominated by NumPy/C-level computation, not Python attribute access.
In our current structure, core attributes like variables, cardinality, and values are stored in BaseFactor, which does not yet define __slots__, so the per-instance overhead remains in effect.
So even though I applied __slots__ to DiscreteFactor, the benefit is masked by inheritance — Python retains the __dict__ because BaseFactor is dynamic.

Susmita331:
@ankurankan 
I request some time to work on this

Susmita331:
@ankurankan Can you please review it now

pgmpy/results.md
```diff
@@ -0,0 +1,106 @@
...
+| reduce        | 0.000028               | 0.000023                  |
+| marginalize   | 0.000027               | 0.000044                  |
+
+>  *Operations are comparable. `marginalize` slightly benefits from `__slots__`.*
+
+---
+
+### Full Model Simulation
+
+| Task               | Time (With `__slots__`) |
```
ankurankan, line 67: @Susmita331  Is there a run time comparision here without __slots__

Susmita331:
Thanks for the comment!

The full model simulation (`inference.query`, `simulate(1000)`) was only run for the slotted class (`pgmpy.DiscreteFactor`) because the non-slotted version (`DiscreteFactorNoSlots`) is a custom class created purely for benchmarking core object and memory behavior. It's not part of the actual model graph or inference engine in `pgmpy`, and thus can't be used directly with `VariableElimination`.

That said, core operations like `product`, `reduce`, and `marginalize` were benchmarked for both versions, and their runtimes are reported side by side.

ankurankan:
@Susmita331  But the runtime is still the same as what I saw in: https://github.com/pgmpy/pgmpy/pull/2052#issuecomment-2862074755 . Correct?

PR #2040, opened 2025-05-05
Update preprocess_data function to show INFO message once

mhmohona:
### Your checklist for this pull request
Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [x] Make sure you are making a pull request against the **dev branch** (left side). Also you should start *your branch* off *our dev*.
- [x] Check the commit's or even all commits' message styles matches our requested structure.

### Issue number(s) that this pull request fixes
- Fixes #2021

### List of changes to the codebase in this pull request

Updated the `preprocess_data`, now keeps track of unique DataFrame structures and only logs a message the first time it sees a new set of columns and data types.

Screenshot from 2025-05-05 07-48-56

PR #2053, opened 2025-05-06
fix: Ensure preprocess_data is called only once during estimation

### LINKED ISSUE #2021: Fix detected datatype INFO when calling estimation methods
When any of the estimation methods are called, an INFO message is shown with the datatypes detected by pgmpy. For example:
```python
In [21]: from pgmpy.models import LinearGaussianBayesianNetwork

In [23]: bn = LinearGaussianBayesianNetwork([('X', 'Y'), ('Z', 'Y')])

In [24]: bn.add_cpds(*bn.get_random_cpds())

In [25]: df = bn.simulate(int(1e4))

In [26]: from pgmpy.estimators import GES

In [27]: GES(df).estimate(scoring_method='bic-g').edges()
 {'X': 'N', 'Z': 'N', 'Y': 'N'}
 {'X': 'N', 'Z': 'N', 'Y': 'N'}
 {'X': 'N', 'Z': 'N', 'Y': 'N'}
Out[27]: OutEdgeView([('X', 'Y'), ('Z', 'Y')])
```
But for some estimators, this message is shown multiple times (as in the example above). This happens because the `pgmpy.utils.preprocess_data` method that shows this message is called multiple times. In the above example, firstly, when the estimator is initialized, and then when the scoring object is initialized. We need to figure out a solution such that this message gets shown only once.

Green-Kedia:

- Fixes #2021

- Modified `pgmpy.estimators.base.BaseEstimator.__init__` to accept optional `_processed_data` and `_processed_dtypes` arguments. If these are provided, the internal call to `preprocess_data` is skipped, and these pre-processed values are used directly.
- Modified `pgmpy.estimators.StructureScore.get_scoring_method` function to accept `**kwargs` (which will include `_processed_data` and `_processed_dtypes`)  and pass it to the `supported_methods` classes and `ScoreCache` class.
- Modified `pgmpy.estimators.ExhaustiveSearch` , `pgmpy.estimators.HillClimbSearch` , and `pgmpy.estimators.GES` files so that whenever they call `get_scoring_method` they pass `self.data` (processed) and
  `self.dtypes` as `_processed_data` and `_processed_dtypes`.
  
This set of changes ensures that `preprocess_data` is called only once by the estimator (e.g., `GES`, `HillClimbSearch`) and not redundantly by the subsequently instantiated scoring objects, addressing the multiple logging issue.

### DIFF
--- a/pgmpy/estimators/ExhaustiveSearch.py
+++ b/pgmpy/estimators/ExhaustiveSearch.py
@@ -40,7 +40,11 @@ class ExhaustiveSearch(StructureEstimator):
     def __init__(self, data, scoring_method="k2", use_cache=True, **kwargs):
         super(ExhaustiveSearch, self).__init__(data, **kwargs)
         _, self.scoring_method = get_scoring_method(
-            scoring_method, self.data, use_cache
+            scoring_method,
+            self.data,
+            use_cache,
+            _processed_data=self.data,
+            _processed_dtypes=self.dtypes,
         )
 
     def all_dags(self, nodes=None):
--- a/pgmpy/estimators/GES.py
+++ b/pgmpy/estimators/GES.py
@@ -152,7 +152,13 @@ class GES(StructureEstimator):
         """
 
         # Step 0: Initial checks and setup for arguments
-        _, score_c = get_scoring_method(scoring_method, self.data, self.use_cache)
+        _, score_c = get_scoring_method(
+            scoring_method,
+            self.data,
+            self.use_cache,
+            _processed_data=self.data,
+            _processed_dtypes=self.dtypes,
+        )
         score_fn = score_c.local_score
 
         # Step 1: Initialize an empty model.
--- a/pgmpy/estimators/HillClimbSearch.py
+++ b/pgmpy/estimators/HillClimbSearch.py
@@ -213,7 +213,13 @@ class HillClimbSearch(StructureEstimator):
         # Step 1: Initial checks and setup for arguments
         # Step 1.1: Check scoring_method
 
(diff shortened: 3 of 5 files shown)


ankurankan:
@Green-Kedia Thanks for the PR, but we already have an open PR for this issue: #2040. I also think the solution on that PR is cleaner than adding new arguments.

Green-Kedia:
Hi @ankurankan ,
Thanks again for the feedback and for pointing out PR #2040.
I understand the preference for its approach of centralizing the logging logic. My PR aimed not only to address the logging but also to:
1. Prevent redundant data transformations by `preprocess_data` (such as type casting) when data has already been suitably prepared by a parent estimator.
2. Ensure `preprocess_data` (and its logging) is invoked based on the explicit state of an estimator (i.e., whether it received already processed data), rather than relying on a global cache of data signatures which might coincidentally match for different datasets.

This provides more predictable control over when preprocessing occurs and avoids repeated operations.
However, I respect your view on the cleaner interface with the solution in #2040. If that's still the preferred direction, please feel free to close this PR.

Thanks

ankurankan:
@Green-Kedia Thanks for explaining your thought process. I still feel the changes make the code much more complex and make the arguments quite confusing without a lot of benefits. 

> Prevent redundant data transformations by preprocess_data (such as type casting) when data has already been suitably prepared by a parent estimator.

This would have an extremely small computational cost when compared to the overall method.

> Ensure preprocess_data (and its logging) is invoked based on the explicit state of an estimator (i.e., whether it received already processed data), rather than relying on a global cache of data signatures which might coincidentally match for different datasets.

 This is indeed possible, but I think we will need to find some another solution for this.

PR #2056, opened 2025-05-07
feature: Added option for custom function for edge orientation in Exp…

### LINKED ISSUE #1953: Allow users to provide custom function for edge orientation in ExpertInLoop algorithm
Currently, the ExpertInLoop algorithm only allows users two options for specifying the edge orientations: 1. Either use an LLM through litellm, 2. The algorithm iteratively asks the user to specify the orientation. To allow more flexibility, we should instead allow users to specify any custom function that takes the two variable names and returns the orientation for them. This would allow for a lot more flexibility, such as using a local LLM, etc.

Green-Kedia:
…ertInLoop algorithm

- Fixes #1953

- Changes only in `pgmpy/estimators/expert.py` to add the flexibility for user to pass a custom function that provides the orientation of a queried edge
- Also add the possible edges from u->v or v->u in `blacklisted_edges` if the `orientation_fn` gives None so that no edge in future is added.
- Updated the docstrings accordingly

### DIFF
--- a/pgmpy/estimators/expert.py
+++ b/pgmpy/estimators/expert.py
@@ -14,6 +14,7 @@ from pgmpy.utils import llm_pairwise_orient, manual_pairwise_orient
 class ExpertInLoop(StructureEstimator):
     def __init__(self, data=None, **kwargs):
         super(ExpertInLoop, self).__init__(data=data, **kwargs)
+        self.orientations_from_fn = set([])
         self.orientations_llm = set([])
 
     def test_all(self, dag):
@@ -57,9 +58,7 @@ class ExpertInLoop(StructureEstimator):
         self,
         pval_threshold=0.05,
         effect_size_threshold=0.05,
-        use_llm=True,
-        llm_model="gemini/gemini-1.5-flash",
-        variable_descriptions=None,
+        orientation_fn=llm_pairwise_orient,
         show_progress=True,
         orientations=set([]),
         use_cache=True,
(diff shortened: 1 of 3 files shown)


pgmpy/estimators/expert.py
```diff
@@ -89,53 +88,91 @@ def estimate(
             And if the effect size for an edge is less than the threshold,
             would suggest to remove the edge.
 
-        use_llm: bool
-            Whether to use a Large Language Model for edge orientation. If
-            False, prompts the user to specify the direction between the edges.
+        orientation_fn: callable, (default: pgmpy.utils.llm_pairwise_orient)
```
ankurankan, line 90: Should mention here that any additional kwargs argument will be passed to this method as well.

pgmpy/estimators/expert.py
```diff
@@ -188,37 +225,58 @@ def estimate(
 
             selected_edge = nonedge_effects.iloc[nonedge_effects.effect.argmax()]
             edge_direction = None
-            if use_llm:
+
+            if orientations:
```
ankurankan, line 238: Could you please add a comment here with "Step 3.2: ..."

pgmpy/estimators/expert.py
```diff
@@ -14,6 +14,7 @@
 class ExpertInLoop(StructureEstimator):
     def __init__(self, data=None, **kwargs):
         super(ExpertInLoop, self).__init__(data=data, **kwargs)
+        self.orientations_from_fn = set([])
         self.orientations_llm = set([])
```
ankurankan: These two are used for caching. Right? Can we get rid of one?

pgmpy/estimators/expert.py
ankurankan: I think would help with readability if you could write the overall if-else flow in the comment.

ankurankan (changes requested):
@Green-Kedia Thanks for the PR. I think it mostly looks good; I have added a few minor comments.

Green-Kedia:
@ankurankan I have done some changes 
- As without `orientations_llm` a test was failing so I adjusted the tests according to whichever cache attribute is present.
- Also as you asked, I have updated the llm and manual pairwise orient functions to accept None as an option. 
- I have also updated the doctorings as you asked

any more suggestions?

pgmpy/estimators/expert.py
```diff
@@ -14,7 +14,7 @@
 class ExpertInLoop(StructureEstimator):
     def __init__(self, data=None, **kwargs):
         super(ExpertInLoop, self).__init__(data=data, **kwargs)
-        self.orientations_llm = set([])
+        self.orientations_from_fn = set([])
```
ankurankan: I think a more readable name could be something like `orientation_cache`

ankurankan (changes requested):
@Green-Kedia Sorry, but can you revert the changes for returning None. I had another look at the algorithm, and I think we will need to make some modifications to the algorithm to allow None. I also think the algorithm is more general without the None option (it usually harder to tell if a causal relationship doesn't exist), and should automatically fix any incorrect edges. Sorry about the confusion.

Green-Kedia:
@ankurankan No problem at all. I’ve reverted the changes and updated a few docstrings accordingly. Apologies for the obvious complications in the attribute names, I’ve corrected them now.
Let me know if there’s anything else that needs to be changed.

ankurankan (approved):
@Green-Kedia Looks good. Thanks again for the PR.

PR #2057, opened 2025-05-08
Implement native PyTorch support for sample_discrete_maps function

Akhilesh-max:

### List of changes to the codebase in this pull request:
- Added native PyTorch tensor support in sample_discrete_maps function
- Removed unnecessary conversion from PyTorch tensors to NumPy arrays
- Implemented backend-specific logic to handle both NumPy and PyTorch
- Used PyTorch's native functions like torch.unique and torch.multinomial
- Preserved device information for PyTorch tensors
- Maintained weight normalization for different backends
- Fixed TODO in mathext.py file

### DIFF
--- a/pgmpy/utils/compat_fns.py
+++ b/pgmpy/utils/compat_fns.py
@@ -165,3 +165,55 @@ def allclose(arr1, arr2, atol):
             arr2,
             atol=atol,
         )
+
+
+def zeros(size, dtype=None):
+    if config.get_backend() == "numpy":
+        return np.zeros(size, dtype=dtype or config.get_dtype())
+    else:
+        return torch.zeros(
+            size, dtype=dtype or config.get_dtype(), device=config.get_device()
+        )
+
+
+def random_choice(values, size=1, p=None, seed=None):
+    if seed is not None:
+        np.random.seed(seed)
+        if torch.is_tensor(p) or torch.is_tensor(values):
+            torch.manual_seed(seed)
+
+    if isinstance(values, np.ndarray) or (
+        isinstance(p, np.ndarray) if p is not None else False
+    ):
+        if isinstance(values, torch.Tensor):
+            values = to_numpy(values)
+        if p is not None and isinstance(p, torch.Tensor):
+            p = to_numpy(p)
+        return np.random.choice(values, size=size, p=p)
+
+    elif isinstance(values, torch.Tensor) or (
+        isinstance(p, torch.Tensor) if p is not None else False
+    ):
+        if not isinstance(values, torch.Tensor):
+            values = torch.tensor(
+                values, dtype=config.get_dtype(), device=config.get_device()
+            )
+
+        if p is None:
+            idx = torch.randint(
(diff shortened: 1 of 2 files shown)


pgmpy/utils/mathext.py
```diff
@@ -179,22 +180,58 @@ def sample_discrete_maps(states, weight_indices, index_to_weight, size=1, seed=N
...
 
-    # TODO: Remove this conversion and find a way to do this natively in torch.
-    states = np.array(states)
-    weight_indices = compat_fns.to_numpy(weight_indices)
-    index_to_weight = {
-        key: compat_fns.to_numpy(value) for key, value in index_to_weight.items()
-    }
-    size = int(size)
+    # Handle PyTorch tensors natively if we're using the torch backend
+    if config.get_backend() == "torch" and isinstance(weight_indices, torch.Tensor):
```
ankurankan: @Akhilesh-max The idea is not to do this if-else block for torch and numpy. Ideally, we would like to handle both in a single block without such if-else logic. To make some of the functions agnostic to the array type we implement functions in `compat_fns.py` that applies the relevant operation depending on whether the array is numpy or torch.

pgmpy/utils/compat_fns.py
```diff
@@ -165,3 +165,55 @@ def allclose(arr1, arr2, atol):
...
+    if config.get_backend() == "numpy":
+        return np.zeros(size, dtype=dtype or config.get_dtype())
+    else:
+        return torch.zeros(
+            size, dtype=dtype or config.get_dtype(), device=config.get_device()
+        )
+
+
+def random_choice(values, size=1, p=None, seed=None):
+    if seed is not None:
```
ankurankan, line 180: Please use `np.random.default_rng` and `torch.random.fork_rng` to set seeds only for this method. Directly calling `np.random.seed` or `manual_seed` would set a global seed for them.

pgmpy/utils/compat_fns.py
```diff
@@ -165,3 +165,55 @@ def allclose(arr1, arr2, atol):
...
+        np.random.seed(seed)
+        if torch.is_tensor(p) or torch.is_tensor(values):
+            torch.manual_seed(seed)
+
+    if isinstance(values, np.ndarray) or (
```
ankurankan, line 185: I don't think we will need this many if-elses. If the backend is numpy, both `values` and `p` arrays should be numpy arrays and torch tensors otherwise. If that is not the case, we should instead try to fix the code when these `values` and `p` arrays are created instead of doing a manual check and conversion here.

ankurankan (changes requested):
@Akhilesh-max I have added a couple of comments. The tests also seem to be failing. Could you have a look at them as well?

pgmpy/utils/compat_fns.py
Akhilesh-max: Both `values` and 'p' are not being set according to the specified backend when they are created. I am writing a fix.

PR #2058, opened 2025-05-08
Self Compatibility score for the validation of causal discovery algorithms

### LINKED ISSUE #2054: [ENH] New methods for DAG validation
Causal discovery methods often make errors when learning DAGs or CPDAGs from data. Since causal discovery is an unsupervised task (i.e., the ground truth causal graph is unknown), quantitatively evaluating the accuracy of these methods is quite difficult. There are still some methods, such as testing for implied CI, comparing models based on their score, etc., to quantify how well the model explains our data, that we implement in our model testing module. Recent research has introduced some more interesting approaches.

Implement the two methods from the following two papers:
1. Faller, Philipp M., et al. "Self-compatibility: Evaluating causal discovery without ground truth." International Conference on Artificial Intelligence and Statistics. PMLR, 2024.
2. Eulig, Elias, et al. "Toward falsifying causal graphs using a permutation-based test." Proceedings of the AAAI Conference on Artificial Intelligence. Vol. 39. No. 25. 2025.

**Update**: This issue has been split into two separate issues, each dealing with one of the above methods. Ref #2115

DARHWOLF:
<h2>List of changes to the codebase in this pull request</h2>
<ul>
  <li><strong><code>pgmpy/metrics/metrics.py</code></strong>
    <ul>
      <li>Added <code>self_compatibility_graphical()</code>, implementing Definition 6 (graphical compatibility) via latent-ADMG projection + SHD.</li>
      <li>Refactored latent-ADMG logic into a private helper <code>_latent_admg()</code> following Definition 5 (Faller et al., AISTATS 2024).</li>
      <li>Ensured existing <code>SHD()</code> is reused to compute per-subset distances.</li>
      <li>Forwarded arbitrary <code>estimator_kwargs</code> into the joint and marginal estimators.</li>
      <li>Enriched docstrings with detailed parameters, return values, references, and usage examples.</li>
    </ul>
  </li>
  <li><strong><code>pgmpy/tests/test_metrics/test_metrics.py</code></strong>
    <ul>
      <li>Added <code>TestGraphicalSelfCompatibility</code> test class:</li>
        <ul>
          <li><code>test_perfect_compatibility</code>: verifies zero SHD when joint = marginals.</li>
          <li><code>test_flip_compatibility</code>: checks mean SHD=2.0 when each marginal is edge-reversed.</li>
          <li><code>test_kwargs_forwarded</code>: confirms custom kwargs reach the estimator.</li>
        </ul>
    </ul>
  </li>
</ul>

Fixes #2054


DARHWOLF:
@ankurankan Please let me know if I have to document the method in an rst file

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,82 @@ def SHD(true_model, est_model):
...
 
     return int(shd)
+
+
+def self_compatibility_score(
+    estimator_class,
+    data: pd.DataFrame,
+    num_resamples: int = 20,
+    with_replacement: bool = True,
+    random_state: int = None,
```
ankurankan: Will also need to add a kwargs argument here to pass additional parameters to the estimate method.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,82 @@ def SHD(true_model, est_model):
...
+) -> float:
+    """
+    Computes the self-compatibility of a structure learner class.
```
ankurankan: I would suggest describing it in much more detail. Its not a very standard method, so users will need to get an idea of what result they will get out of the method.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,82 @@ def SHD(true_model, est_model):
...
+        If True, bootstrap; else subsample.
+    random_state : int or None
+        RNG seed for reproducibility.
+
+    Returns
+    -------
+    float
+        Mean pairwise adjacency‐matrix agreement ∈ [0,1].
+
+    Raises
```
ankurankan: Needs an example section as well.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,82 @@ def SHD(true_model, est_model):
...
+    ------
+    AttributeError
+        If `estimator_class` cannot be instantiated on a DataFrame
+        or does not implement `.estimate()`.
```
ankurankan: Needs a references section as well.

ankurankan (changes requested):
I don't think the implementation here is correct. The algorithm described in the paper samples columns from the data (i.e., choose subsets of variables) to construct multiple DAGs/ CPDAGs and compute a compatibility score over that. The approach implemented here is a bootstrapping approach where samples of rows are taken. These are two completely different methods.

DARHWOLF:
noted

pgmpy/metrics/metrics.py
DARHWOLF: Added

pgmpy/metrics/metrics.py
DARHWOLF: Added example section

pgmpy/metrics/metrics.py
DARHWOLF: Added references section

DARHWOLF:
@ankurankan I have refactored the method to better align with the paper and as per your review. Please have a look when time permits. Thanks!

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,126 @@ def SHD(true_model, est_model):
...
+        subsets.append(S)
+        graphs.append(G)
+
+    # Compute pairwise compatibility over overlaps
+    sims = []
+    for (i, Gi), (j, Gj) in combinations(enumerate(graphs), 2):
+        common = list(set(subsets[i]).intersection(subsets[j]))
+        if len(common) < 2:
+            continue
+        A_i = np.array([[int(Gi.has_edge(u, v)) for v in common] for u in common])
```
ankurankan: @DARHWOLF I don't think this is how compatibility is defined in the paper. I would really like you to first try to understand the paper before trying to implement this.

ankurankan (changes requested):
The implementation is still incorrect.

pgmpy/metrics/metrics.py
DARHWOLF: @ankurankan my apologies for the confusion. I was using simple adjacency matrix agreement for compatibility, thinking that undirected edges would capture stability. Upon re-reading the paper i realized that they define define compatibility over ordered pairs u!=v with a denominator term |s|(|s|-1) and require directionality. 

To implement this, I replace lines 586-588 with: 
<code>
matches = 0  
n = len(common)
for u in common:
    for v in common:
        if u == v:
            continue
        e_i = 1 if G_i.has_edge(u, v) else 0
        e_j = 1 if G_j.has_edge(u, v) else 0
         here.)
        matches += (e_i == e_j)
compat_ij = matches / float(n * (n - 1))
sims.append(compat_ij)
  </code>


please let me know if i'm on the right track. sorry for the confusion...

ankurankan: @DARHWOLF, I still do not think you have looked at the paper. The paper defines two scores for incompatibility: interventional (Definition 9) and graphical (Definition 10). The implementation is not even close to either of those two definitions. If you are using an AI tool to implement the scores, please don't do that because it would not know about these definitions (the paper is the only source where these metrics have been defined). I would also like you to understand that this wastes my time because I end up spending a lot more time trying to figure out how your implementation relates to the actual methods. I would much rather have you ask questions about the paper if you don't understand the method, rather than implementing a random method.

DARHWOLF:
@ankurankan sir, please take a look at the latent admg method when time permits. 
I'll add graphical_compatibility method after this.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,126 @@ def SHD(true_model, est_model):
     shd = shd + (np.sum((d + d.T) > 0) / 2)
 
     return int(shd)
+
+
+def latent_admg(dag: DiscreteBayesianNetwork, observed: list) -> nx.MultiDiGraph:
```
ankurankan: I would make this a private method as this won't have any application outside the compatibility score method. Rename this to `_latent_admg`.

pgmpy/metrics/__init__.py
```diff
@@ -4,6 +4,7 @@
     correlation_score,
     fisher_c,
     implied_cis,
+    latent_admg,
```
ankurankan: Remove this method from `__init__.py`, so that it doens't show as an option to the user.

pgmpy/metrics/metrics.py
ankurankan: This should be `dag: DAG`

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,126 @@ def SHD(true_model, est_model):
...
+    ----------
+    dag : DiscreteBayesianNetwork
+        The full DAG G on variables V (may include latent nodes).
+    observed : list of hashable
+        The subset S of variables to keep (observed nodes).
+
+    Returns
+    -------
+    nx.MultiDiGraph
+        A NetworkX MultiDiGraph over nodes in S where:
```
ankurankan: Doesn't need to be a nx.MultiDiGraph. nx.DiGraph can be used to represent both directed and undirected edges. Undirected edges can be represented using two edges X -> Y and X <- Y. We already do this in `pgmpy.base.PDAG` class.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,126 @@ def SHD(true_model, est_model):
...
+        - Each `kind="directed"` arc (u→v) represents condition (1).
+        - Each pair of opposite arcs with `kind="bidirected"` (u→v and v→u)
+          represents a bidirected edge u↔v per condition (2).
+
+    Example
+    -------
+    >>> from pgmpy.models import DiscreteBayesianNetwork
```
ankurankan, line 501: DiscreteBayesianNetwork -> DAG

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,126 @@ def SHD(true_model, est_model):
...
+    >>> sorted(admg.edges(data=True))
+    [('A', 'B', {'kind': 'bidirected'}), ('B', 'A', {'kind': 'bidirected'})]
+
+    References
+    ----------
+    Definition 5, Faller, P. M., et al. (2024).
+    “Self-compatibility: Evaluating Causal Discovery without Ground Truth.”
+    In AISTATS. arXiv:2307.09552
+    """
+    G = dag.to_directed()
```
ankurankan: Why do we need the `to_directed` call?

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,126 @@ def SHD(true_model, est_model):
...
+    admg = nx.MultiDiGraph()
```
ankurankan: Use nx.DiGraph to simplify.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,126 @@ def SHD(true_model, est_model):
...
+            if G.has_edge(a, b) and G.has_edge(c, b):
+                return False
+        return True
+
+    directed_edges = set()
+    bidirected_edges = set()
+    V = set(G.nodes())
+    latent = V - set(observed)
+
+    # 1) Directed edges via latent-only intermediate nodes
```
ankurankan: A simpler way to do this is to iterate over nodes in `observed` and use the `G.active_trail_nodes` method.
1. The `active_trail_nodes` method would give all the nodes that are reachable through a non-collider path. This would imply that a variable should be connected to each node that is on its active trail in the latent ADMG.
2. Once we have the active trail nodes, we need to classify whether there should be directed edge or bidirected edge between them. We can do a simple check (using nx.has_path) that if there is a directed path between the variable and its active trail node, then there should be a directed edge between them. Otherwise, a bidirected edge.

ankurankan (changes requested):

DARHWOLF:
@ankurankan Sir, I have made the requested changes as per the previous review. Please have a look when time permits. Thanks!

Changes made: 
- Refactored latent-ADMG logic into a private helper `_latent_admg()`
- Uses `nx.has_path()` over induced subgraphs to detect latent-mediated directed paths.
- Applies `active_trail_nodes(..., include_latents=True)` to identify collider-only connections
- Returns a `nx.DiGraph`

-Added `self_compatibility_graphical()` method implementing Definition 6 and unit tests

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,189 @@ def SHD(true_model, est_model):
...
+    >>> isinstance(score, float)
+    True
+
+    References
+    ----------
+    [1] Faller, P. M., et al. (2024).
+        “Self‐compatibility: Evaluating Causal Discovery without Ground Truth.”
+        In AISTATS. arXiv:2307.09552
+    """
+    rng = np.random.RandomState(random_state)
```
ankurankan: Should always accept seed values as function arguments and not set the seed if None is passed.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,189 @@ def SHD(true_model, est_model):
...
+          - A bidirected link u↔v is encoded by having both arcs u→v and v→u.
+
+        References
+        ----------
+        [1] Faller, P. M., et al. (2024).
+        “Self‐compatibility: Evaluating Causal Discovery without Ground Truth.”
+        In AISTATS. arXiv:2307.09552
+        """
+        # 1) Prep
+        full_nx = dag.to_directed()
```
ankurankan: Why do we need the conversion here? DAG objects should already be directed.

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,189 @@ def SHD(true_model, est_model):
...
+        V = set(dag.nodes())
```
ankurankan: Try to use more readable variable names, such as `variables`, `vars`, or `nodes`

pgmpy/metrics/metrics.py
```diff
@@ -465,3 +465,189 @@ def SHD(true_model, est_model):
...
+    [1] Faller, P. M., et al. (2024).
+        “Self‐compatibility: Evaluating Causal Discovery without Ground Truth.”
+        In AISTATS. arXiv:2307.09552
+    """
+    rng = np.random.RandomState(random_state)
+    variables = list(data.columns)
+    p = len(variables)
+    k = max(2, int(np.floor(subset_fraction * p)))
+
+    def _latent_admg(dag: DAG, observed: list) -> nx.DiGraph:
```
ankurankan: I think we should have this function outside of this function. Nested functions make it much harder to read.

pgmpy/tests/test_metrics/test_metrics.py
```diff
@@ -212,3 +214,123 @@ def test_shd(self):
     def test_shd_unequal_graphs(self):
         with self.assertRaises(ValueError, msg="The graphs must have the same nodes."):
             SHD(self.dag_4, self.dag_5)
+
+
+class TestGraphicalSelfCompatibility(unittest.TestCase):
```
ankurankan, line 403: Can you also add separate tests for `_latent_admg` method?

pgmpy/tests/test_metrics/test_metrics.py
```diff
@@ -212,3 +214,123 @@ def test_shd(self):
...
+        rng = np.random.RandomState(0)
+        cls.data = pd.DataFrame(
+            {
+                "A": rng.randint(2, size=100),
+                "B": rng.randint(2, size=100),
+                "C": rng.randint(2, size=100),
+            }
+        )
+
+    def test_perfect_compatibility(self):
```
ankurankan, line 420: We will need to test this on a real algorithm and not just on a mocked function.

pgmpy/metrics/metrics.py
```diff
@@ -467,3 +467,189 @@ def SHD(true_model, est_model):
...
+        admg.add_nodes_from(observed)
+        admg.add_edges_from(directed_edges)
+        admg.add_edges_from(bidirected_edges)
+        return admg
+
+    # 1) Fit joint model on all variables
+    joint_learner = estimator_class(data)
+    joint = joint_learner.estimate(**estimator_kwargs)
+    if not isinstance(joint, DAG):
+        joint = DiscreteBayesianNetwork(list(joint.edges()))
```
ankurankan: This is not correct. If the returned type is CPDAG, use the `PDAG.to_dag` method to convert it to a DAG.

Also, currently, this implementation converts PDAGs to DAGs to compute the score; it should ideally work directly on the PDAG (see section A9 in the supplementary material of the paper). But for now let's first get it working for DAGs and then we can expand it to other graph types.

pgmpy/metrics/metrics.py
```diff
@@ -467,3 +467,189 @@ def SHD(true_model, est_model):
...
+    # 2) For each random subset...
+    for i in range(num_subsets):
+        # a) sample subset of columns
+        S = rng.choice(variables, size=k, replace=False).tolist()
+        sub_data = data[S]
+        # b) fit marginal model
+        marg_learner = estimator_class(sub_data)
+        marginal = marg_learner.estimate(**estimator_kwargs)
+        # c) project joint onto S
+        joint_proj = _latent_admg(joint, S)
```
ankurankan: Would save a lot of compute if we put this line outside the for loop as `joint_proj` would be the same in each loop.

pgmpy/tests/test_metrics/test_metrics.py
```diff
@@ -224,3 +226,123 @@ def test_shd_unequal_graphs(self):
 
         with self.assertRaises(ValueError, msg="The graphs must have the same nodes."):
             SHD(dag1, dag2)
+
+
+class TestGraphicalSelfCompatibility(unittest.TestCase):
```
ankurankan, line 403: For the tests, I would suggest the following:
1. Write tests for the _latent_admg method. Give it some known DAG and check if the output is correct.
2. For the compatibility score: a) use a fake estimator to give the correct sub-DAG for each estimator, and check if the incompatibility score is 0. b) Use a known model from example_models, simulate some data from it, and check the incompatibility score; it should be low as the data has been simulated from the model.

ankurankan (changes requested):
@DARHWOLF Thanks. I had a quick look at the implementation, it mostly looks good to me. I have added a few comments, and once you have added the tests, I can have another detailed look.

pgmpy/metrics/metrics.py
DARHWOLF: Hi, sorry I have a bit of a confusion here. Do you want me to move `_latent_admg()` function outside of `self_compatibility_graphical()`? And also, to test `_latent_admg()` function, I will have to add it to `__init__.py`. Will that be correct? I'll proceed as per your suggestion. Thanks!

ankurankan: Yes, you can move it out of `self_compatibility_graphical`. No, you don't need to add it to `__init__.py`, the only thing it does is that it changes the import path. If you don't add it to `__init__.py`, you will have to import it like: `from pgmpy.metrics.metrics import _latent_admg`

DARHWOLF: Got it... Thanks!

pgmpy/metrics/metrics.py
DARHWOLF: Would it be correct to check whether the estimator’s output has a .to_dag() method and if it does, i invoke joint = joint.to_dag() 
may i approach it as follows
`if not isinstance(joint, DAG) and hasattr(joint, "to_dag"):`
`    joint = joint.to_dag()`
`elif isinstance(joint_model, DAG):`
`    joint = joint_model # in the case where it is already a DAG`
`else:`
   ` joint = DiscreteBayesianNetwork(list(joint_model.edges())) #in case of CPDAG or other graph object` 
   
   and likewise for each marginal fit....

ankurankan: I would rather check the output type of the estimator. It would be either a `pgmpy.base.DAG` or `pgmpy.base.PDAG`. If it's a PDAG we should call the `to_dag` method to convert it to a DAG.

DARHWOLF: okay so I'll check directly on the output
`if isinstance(joint, PDAG): `
    `joint = joint.to_dag()`
`elif isinstance(joint_model, DAG):`
    `joint = joint_model`
`else:`
    `joint = DiscreteBayesianNetwork(list(joint_model.edges())) #Is this part correct?`

ankurankan: You don't need the last else statement. There are only two possible outputs from the estimator methods: a PDAG or a DAG.

DARHWOLF: okay got it, I had a doubt there. Thanks for clarifying

pgmpy/metrics/metrics.py
DARHWOLF: Sir, I have some confusion here. I tried putting `joint_proj` outside the loop to compute the latent ADMG of the full joint: 
`full_joint_proj = _latent_admg(joint, variables)  #renamed joint_proj to full_joint_proj here`
`    for()...`

Then inside the loop, do you want me to just restrict that fixed graph to each subset? Could you point me in the right direction here please? 
Thanks

ankurankan: My understanding is that we compare the full structure on all the variables with a bunch of structures learned on subsets of variables. So, the full structure remains constant in each of these comparisons, so we can move it outside of the for loop.

DARHWOLF: Since each iteration's S is different, shouldn't we call `joint_proj = latent_admg(joint, S)` inside the loop so that the joint-DAG projection would know which S we’re projecting onto? Because S is different for each iteration, wouldn't it yield a different projection for that particular iteration's subset? I have a slight confusion here.

ankurankan: Yes, you are correct. I think I misunderstood this step.

pgmpy/metrics/metrics.py
DARHWOLF: @ankurankan Sir, I am facing a challenge while trying to use active_trail_nodes method. 

Upon introducing the tests, I am running into errors since `active_trail_nodes` returns every node d‐connected to a source, but doesn’t distinguish whether that connection is a fully‐directed chain or a pure‐collider path. To patch this, we would have to traverse the paths manually to separate directed vs collider links which is drastically increasing the number of lines of code.
Also, since `active_trail_nodes` require graphs to be converted to DAG and the future implementation of `self_compatibility_graphical()` method should ultimately operate on all graph types, we would have to refactor this `_latent_admg()` method (and `self_compatibility_graphical()` method) which might be more work. 
If we use nx.all_simple_paths, it could make the future expansion much easier since it accepts all graph types. It would also allow us to directly inspect each path’s edge directions to classify it as “latent‐only chain” vs. “latent‐only collider.”

Could you share some advice? I'll proceed as per your feedback. 
Thanks!

DARHWOLF:
@ankurankan Sir, I have made the requested changes as per your previous review. Please take a look when time permits. 
 
 List of changes made:-

-  moved `_latent_admg()` helper out of `self_compatibility_graphical()`
- Added handling for partially directed graphs by converting any returned PDAG to a DAG in `self_compatibility_graphical()`, both for the joint fit and for each marginal fit
- Used more readable variable names
- Added 3 tests for `_latent_admg()` helper where we give it known dags and check if output is correct; covering chain, collider, and mixed‐structure cases.
- Added two additional tests for `self_compatibility_graphical()` : 
  - use a fake estimator to give the correct sub-DAG for each estimator, and check if the incompatibility score is 0.
  - simulating data from `example_model's` `child` model, and asserting that the resulting self-compatibility score is reliably low
- Added docstrings and examples for `_latent_admg()`

pgmpy/tests/test_metrics/test_metrics.py
```diff
@@ -224,3 +228,248 @@ def test_shd_unequal_graphs(self):
...
+    def test_latent_chain_projects_to_directed(self):
+        """
+        A → H → B projects to A → B when H is latent.
+        """
+        admg = _latent_admg(self.dag_chain, observed=["A", "B"])
+        self.assertEqual(set(admg.edges()), {("A", "B")})
+
+    def test_latent_collider_projects_to_bidirected(self):
+        """
+        A → H ← B projects to A ↔ B (i.e., both A→B and B→A) when H is latent.
```
ankurankan: Is this correct? I thought that because A and B are d-separated here, they shouldn't have an edge between them in the projection ADMG

pgmpy/metrics/metrics.py
```diff
@@ -467,3 +467,242 @@ def SHD(true_model, est_model):
...
+
+    # 1) Preserve original observed→observed
+    for u, v in dag.edges():
+        if u in observed_set and v in observed_set:
+            directed_edges.add((u, v))
+
+    # 2) Rule 1: latent-only directed chains via active_trail_nodes
+    for u, v in combinations(observed_set, 2):
+        # Condition on all other observed nodes so only paths via latents remain
+        conditioning_set = list(observed_set - {u, v})
```
ankurankan: Conditioning wouldn't do this. Rather, you would have to remove these edges. For example, conditioning on a collider can open a path.

ankurankan (changes requested):
@DARHWOLF I think the `_latent_admg` implementation isn't correct. I also found the original implementation for the paper that you can reference: https://github.com/amazon-science/causal-self-compatibility/. They also have a bunch of tests that you can also implement to check your implementation: https://github.com/amazon-science/causal-self-compatibility/blob/9084041f9c1e238250ce281868c4f6d2971d5e0c/test/test_admg_marginalisation.py#L14

pgmpy/metrics/metrics.py
```diff
@@ -474,3 +474,236 @@ def SHD(true_model, est_model):
...
+    dag : DAG
+        The full DAG G on variables V (may include latent nodes).
+    observed : list
+        Subset observed_set ⊂ V to project onto (observed variables).
+
+    Returns
+    -------
+    nx.DiGraph
+    An ADMG over the observed nodes, where:
+        - X→Y encodes a latent‐only directed chain X→…→Y
```
ankurankan: Please try not to use special characters. You can denote an arrow using `->`

pgmpy/metrics/metrics.py
```diff
@@ -474,3 +474,236 @@ def SHD(true_model, est_model):
...
+    directed_edges = set()
+    bidirected_edges = set()
+
+    # 1) Preserve original observed→observed edges
+    for u, v in dag.edges():
+        if u in observed_set and v in observed_set:
+            directed_edges.add((u, v))
+
+    # 2) Add latent‐only directed chains via active trail + directed‐path check
+    for u, v in combinations(observed_set, 2):
```
ankurankan: Wouldn't this be much simpler by iterating over variables in `observed_set` instead of the node combinations.
Pseudocode:
```
for u in observed_set:
    d_connected_nodes = set(dag.active_trail_nodes(u))
    d_connected_observed = d_connected_nodes.intersection(observed_set)
    for v in d_connected_observed:
         if nx.has_path(dag, u, v):
               add directed edge u to v.
         else:
              add bidirected edge u to v.
```
Am I missing something here?

ankurankan (changes requested):

pgmpy/metrics/metrics.py
ankurankan: @DARHWOLF Sorry for the late reply. The

pgmpy/metrics/metrics.py
```diff
@@ -474,3 +474,236 @@ def SHD(true_model, est_model):
...
+        #  take all the observed nodes except u and v
+        observed_excluding_pair = list(observed_set - {u, v})
+        reach_u = dag.active_trail_nodes(
+            [u], observed=observed_excluding_pair, include_latents=False
+        )[u]
```
ankurankan: Hey @DARHWOLF, sorry for the late reply. I don't think the current logic here is correct. By setting `observed=observed_excluding_pair`, we are not actually blocking active‐trail paths through observed variables, rather it can open new paths through them. For example, in the model A -> C <- B, if `C` is the observed variable, we shouldn't add any edge between A and B in the ADMG, but we would get an active trail between A and B if we specify C as observed in the activate_trail_nodes method.

What I would suggest is:
1. Build a subgraph containing only the edges: a) between two latent variables b) or between one latent and one observed variable.
2. Run the pseudocode I shared earlier on that subgraph.

That way, we will exclusively trace paths through the latent variables. What do you think?

Nimish-4 (dismissed):

Issue #1684, opened 2023-10-16
Getting the strength for all edges of a network


zzzrbx:
Once a structure is found via hill climb (for example), is it possible to get the strength for all edges in one go, as opposed to compute the bic score for all edges manually?

ankurankan:
Solution: Add a method to DAG class to get the edge strength for every edge in the model. The edge strength can be computed using the effect size measure of ci_pillai test. For an edge X -> Y, the edge strength would be the effect size of the test `X _|_ Y | pa(X) U pa(Y)`.

aniket2405:
Is this open for discussion or do we have a solution for this?
@ankurankan since you mentioned that #913 is related to this.

ankurankan:
@aniket2405 This is still open. The solution is essentially what I have mentioned in the [comment above](https://github.com/pgmpy/pgmpy/issues/1684#issuecomment-2834514149). For the implementation, we would need the following steps:
1. Add a new method to the DAG class, say `edge_strength` or any other suitable name. I think we should allow the users to specify either a edge, a list of edges, or None (implying all edges).
2. For each of the edges, we will need to first check if it has any latent parent nodes. If yes, the method would fail and ask the user to use CausalInference class for advance causal effect estimation.
4. Else we will  run a conditional independence (CI) test (`X _|_ Y | pa(X) U pa(Y)`) to compute the strength. The CI tests are implemented in pgmpy/estimators/CITests.py. The effect size measure of these tests would be the edge strength.

Let me know if you have any questions.

nags0x:
@ankurankan , I'll give it a shot

nags0x:
@ankurankan , noticed that `edge_strength` method is already implemented in the `DAG.py` which seems to fulfill all the requirements mentioned in the above issue. Could u pls verify nd lmk the specifics

ankurankan:
Yes, this got implemented in the PR referenced above.

PR #2059, opened 2025-05-09
Getting the strength for all edges of a network 

Manas-7854:
Issue #1684 
- Added edge_strength function inside DAG.py
- added test to test the function


Nimish-4:
@Manas-7854 The tests are failing due to formatting issues. pgmpy uses `black`, `isort` and a few other formatting checks. You can find them in the ".pre-commit-config.yaml" file in the codebase. Install them locally on your editor, use them to format the code and then the formatting checks will pass.

.gitignore
```diff
@@ -19,6 +19,7 @@ develop-eggs
 .installed.cfg
 lib
 lib64
+myenv
```
Nimish-4: This looks related to your local environment. If yes, I think it doesn't need to be here and should be removed from the changes.

model.bif
```diff
@@ -0,0 +1,31 @@
...
+probability ( C | B ) {
+    ( 0 ) 1.0;
+    ( 1 ) 1.0;
+    ( 2 ) 1.0;
+
+}
+probability ( D | C ) {
+    ( 0 ) 0.11202275075549158, 0.46515371328405486, 0.08913003410201674, 0.33369350185843677;
+
+}
```
Nimish-4: This model file is created by a test that does not clean up the files it created. Should also be removed.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
...
+        >>> # Generate some data
+        >>> data = pd.DataFrame({'X': [0, 1, 0, 1], 'Y': [1, 3, 0, 2], 'Z': [1, 1, 0, 0]})
+        >>> # Compute edge strengths
+        >>> dag.edge_strength(data)
+        {('X', 'Y'): 1.2260921400386593e-07, ('Z', 'Y'): 5.140313027451882e-07}
+        >>> # Compute strength for a specific edge
+        >>> dag.edge_strength(data, edges=('X', 'Y'))
+        {('X', 'Y'): 1.2260921400386593e-07}
+        """
+        from pgmpy.estimators.CITests import pillai_trace
```
Nimish-4, line 1414: Better to import it at the top with all the other imports.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
...
+
+            # Get parents of x and y
+            pa_X = list(self.predecessors(x))
+            pa_Y = list(self.predecessors(y))
+
+            # Combine parents for conditioning set (excluding x and y themselves)
+            conditioning_set = set(pa_X + pa_Y) - {x, y}
+
+            # Run CI test and get effect size
+            result = pillai_trace(
```
Nimish-4: If any other tests are suitable, or were discussed previously, you can create a new argument for the user to choose a CI test.

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -360,6 +360,39 @@ def test_dag_fit(self):
...
+            data={"a": [0, 1, 0, 1], "b": [1, 3, 0, 2], "c": [1, 1, 0, 0]}
+        )
+        with self.assertRaises(ValueError):
+            dag_latent.edge_strength(data)
+
+    def test_edge_strength_empty_dag(self):
+        empty_dag = DAG()
+        data = pd.DataFrame({"a": [0, 1, 0, 1], "b": [1, 3, 0, 2], "c": [1, 1, 0, 0]})
+        self.assertEqual(empty_dag.edge_strength(data), {})
+
```
Nimish-4, line 680: I think adding a test where the output of the function is tested would be great. In other words, some numerical comparison of the output and expected output.

.gitignore
Manas-7854: is there a way i can use `git add .` and the myenv folder doesn't get commited or do i need to add all the files i have changed manually?

Nimish-4: @Manas-7854 You could look it up, but I don't think it can be done like that. I prefer to do it manually, also helps in keeping track of what is being committed.

pgmpy/base/DAG.py
Manas-7854: > Solution: Add a method to DAG class to get the edge strength for every edge in the model. The edge strength can be computed using the effect size measure of ci_pillai test. For an edge X -> Y, the edge strength would be the effect size of the test `X _|_ Y | pa(X) U pa(Y)`.

in the issue only pilliai_test was discussed, however i can allow the user to select the test if that is required?

Nimish-4: In that case, it's fine! @ankurankan will let us know if anything else is required.

.gitignore
jgyasu: > is there a way i can use git add . and the myenv folder doesn't get commited or do i need to add all the files i have changed manually?

you can also create a `.gitignore` file and add the files and folders there that you don't want to commit

pgmpy/base/DAG.py
Manas-7854: Hi, importing at the top is giving the following error `test_DAG.py - TypeError: module() takes at most 2 arguments (3 given)`.
Also there are other functions that import inside the function like for example `fit`.

Nimish-4: @Manas-7854 Keep your original implementation, in that case!

.gitignore
avinab-neogy: The `myenv` directory is already included in the `.gitignore` file, so it shouldn't be an issue right? @Nimish-4

Nimish-4: @avinabneogy23 It can be present in your local project, but it doesn't need to be pushed to the pgmpy repo is what I'm saying.

avinab-neogy: I see, thanks @Nimish-4

ankurankan: @avinabneogy23 In general, it is always a better idea to name each file (as @Nimish-4 also mentioned) when doing a git add: `git add <filename1> <filename2> ...`. This makes sure that you are adding the correct changes. Additionally, you can always run `git diff --cached` after doing add so that you can double check which changes are going to get committed.

pgmpy/base/DAG.py
ankurankan: Importing outside might be resulting in a circular import.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
         dag.add_nodes_from(self.nodes())
         return dag
 
+    def edge_strength(self, data, edges=None):
+        """
+        Computes the strength of edges in the DAG using conditional independence tests.
```
ankurankan: This will need to be expanded by a lot. Pillai's trace is not a standard method (the research paper still hasn't been published), so we will need to give the users a better idea of what numbers they are seeing. The following things need to be mentioned in the documentation:

1. Uses a residualization based approach (similar to partial correlation) and uses the XGBoost estimator instead of linear regression.
2. Computes the Pillai Trace effect size based on canonical correlations instead of using correlation (as is the case with partial correlation)
3. Measure of linear relationship between the residuals.
1. Works for any mixture of categorical and continuous variables.
2. The value is bounded between 0 and 1.

Interpretation:

1. If the edge is between two continuous variables, the strength would be equivalent to partial correlation.
4. If the edge is between two categorical variables, the strength would be analogous to Cramers' V.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
...
+
+        Returns
+        -------
+        dict
+            A dictionary mapping edges to their strength values. Each strength value
+            is based on the effect size measure of the ci_pillai test.
+
+        Examples
+        --------
+        >>> import pandas as pd
```
ankurankan: Please use a Linear Gaussian BN for the Example here.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
...
+        else:
+            edges_to_compute = edges
+
+        strengths = {}
+
+        for edge in edges_to_compute:
+            x, y = edge
+
+            # Check if either x or y is a latent node
+            if x in self.latents or y in self.latents:
```
ankurankan: The check needs to be:
1. Fail if either X or Y are latents
2. Fail if any of the parents of X and Y are latents.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
...
+                raise ValueError(
+                    f"Edge {edge} involves latent variables. Use CausalInference class for "
+                    "advanced causal effect estimation."
+                )
+
+            # Get parents of x and y
+            pa_X = list(self.predecessors(x))
```
ankurankan: Use the `DAG.get_parents` method here. The definition of predecessors in case of causal inference and graph theory are different. In causal inference predecessor means: parents + parents of parents + their parents ... whereas in standard graph theory predecessors are defined as just the parents. 

The get_parents method makes a call to the predecessors as well, but using `get_parents` makes it more explicit here.

pgmpy/base/DAG.py
ankurankan: Yeah, I would keep it just as pillai trace for now as it can work with all types of data. We also do not have an effect size measure for all the tests. In future, we can expand this.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
...
+            # Combine parents for conditioning set (excluding x and y themselves)
+            conditioning_set = set(pa_X + pa_Y) - {x, y}
+
+            # Run CI test and get effect size
+            result = pillai_trace(
+                X=x, Y=y, Z=list(conditioning_set), data=data, boolean=False
+            )
+
+            # Store the edge strength
+            strengths[edge] = result[1]
```
ankurankan: I think result is (effect_size, p_value), and the effect size represents the strength, so this should be `result[0]`.

pgmpy/tests/test_base/test_DAG.py
ankurankan: I agree, we can use a LinearGaussian BN, simulate data from it and try to estimate the edge strengths. The edge strength should match the alpha parameter of the LinearGaussian CPD

ankurankan: We should also add some tests that cover the latent variable case.

ankurankan (changes requested):
@Manas-7854 Thanks for the PR. I think most of the code structure is here now, but the code and some logic still needs work.

pgmpy/base/DAG.py
Manas-7854: do i need to use Linear Gaussian BN for tests as well?

pgmpy/tests/test_base/test_DAG.py
Manas-7854: hi I am having a little hard time understanding how to generate data from the linear gaussian, can you help me with that?
here is what i have tried to implement but its not working since the edge_strengths from `DAG.edge_strength` do not match with alpha parameter. 

```  
def test_edge_strength_basic(self):
        dag = DAG([("X", "Y"), ("Z", "Y")])

        num_samples = 1000
        seed=42

        np.random.seed(seed)

        # Root nodes: X and Z
        X = np.random.normal(loc=0, scale=1, size=num_samples)
        Z = np.random.normal(loc=0, scale=1, size=num_samples)

        # Child node: Y = 0.9 + (-2.0 * X) + (1.5 * Z) + noise
        noise = np.random.normal(loc=0, scale=1.0, size=num_samples)
        Y = 0.9 + (-2.0 * X) + (1.5 * Z) + noise

        data= pd.DataFrame({"X": X, "Z": Z, "Y": Y})

        # Define true CPD with known beta
        cpd = LinearGaussianCPD("Y", beta=[0.9, -2.0, 1.5], std=1.0, evidence=["X", "Z"])

        # Ground truth edge strengths from beta (skip intercept)
        true_strengths = {}
        for i, parent in enumerate(cpd.evidence):
            true_strengths[(parent, cpd.variable)] = cpd.beta[i + 1]  # skip intercept at index 0


        # Estimated edge strengths from data
        estimated_strengths = dag.edge_strength(data=data)

        tolerance = 0.1
        for edge, true_val in true_strengths.items():
            est_val = estimated_strengths[edge]
            assert abs(abs(est_val) - abs(true_val)) <= tolerance, f"Edge {edge}: expected ~{true_val}, got {est_val}"
```

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,82 @@ def copy(self):
...
+                    f"Edge {edge} involves latent variables. Use CausalInference class for "
+                    "advanced causal effect estimation."
+                )
+
+            # Get parents of x and y
+            pa_X = list(self.predecessors(x))
+            pa_Y = list(self.predecessors(y))
+
+            # Combine parents for conditioning set (excluding x and y themselves)
+            conditioning_set = set(pa_X + pa_Y) - {x, y}
```
ankurankan: @Manas-7854 Sorry, slight change. This should be only be `pa_Y - {x, y}`

pgmpy/tests/test_base/test_DAG.py
ankurankan: I think I have found an issue with `pillai_trace`, so the values wouldn't match for now (#2066). You can write the test with the values that you are getting now and we will fix the tests after #2066 is fixed.

ankurankan (changes requested):

pgmpy/tests/test_base/test_DAG.py
ankurankan: You can simulate data from a linear model as:
```python
In [47]: from pgmpy.estimators.CITests import pearsonr, pillai_trace

In [48]: from pgmpy.models import LinearGaussianBayesianNetwork as LGBN

In [49]: model = get_example_model("alarm")

In [50]: linear_model = LGBN([("X", "Y"), ("Z", "Y")])

In [51]: x_cpd = LinearGaussianCPD(variable='X', beta=[0], std=1)

In [52]: y_cpd = LinearGaussianCPD(variable='Y', beta=[0, 0.4, 0.6], std=1, evidence=['X', 'Z'])

In [53]: z_cpd = LinearGaussianCPD(variable='Z', beta=[0], std=1)

In [54]: linear_model.add_cpds(x_cpd, y_cpd, z_cpd)

In [55]: df = linear_model.simulate(n_samples=int(1e4))

```

Manas-7854: just to clarify, should i keep the tests as i had initially implemented?

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,107 @@ def copy(self):
...
+        from pgmpy.estimators.CITests import pillai_trace
+
+        # If edges is None, compute for all edges in the DAG
+        if edges is None:
+            edges_to_compute = list(self.edges())
+        # If edges is a single edge tuple
+        elif isinstance(edges, tuple) and len(edges) == 2:
+            edges_to_compute = [edges]
+        # If edges is a list of edge tuples
+        else:
```
ankurankan: I would suggest checking there that the argument here is a list of 2-tuples. And add another else, which throws an ValueError that the input by the user is not correct with some information on what we expect.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,107 @@ def copy(self):
...
+
+            # Get parents of x and y using get_parents instead of predecessors
+            pa_X = self.get_parents(x)
+            pa_Y = self.get_parents(y)
+
+            # Check if either x or y is a latent node
+            if (
+                x in self.latents
+                or y in self.latents
+                or any(parent in self.latents for parent in pa_X + pa_Y)
```
ankurankan: This would need to be just pa_Y. `pa_X` is not needed.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,107 @@ def copy(self):
...
+        else:
+            edges_to_compute = edges
+
+        strengths = {}
+
+        for edge in edges_to_compute:
+            x, y = edge
+
+            # Get parents of x and y using get_parents instead of predecessors
+            pa_X = self.get_parents(x)
```
ankurankan: We don't need pa_X.

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,107 @@ def copy(self):
...
+            # Combine parents for conditioning set (excluding x and y themselves)
+            conditioning_set = set(pa_Y) - {x, y}
+
+            # Run CI test and get effect size
+            result = pillai_trace(
+                X=x, Y=y, Z=list(conditioning_set), data=data, boolean=False
+            )
+
+            # Store the edge strength
+            strengths[edge] = result[0]
```
ankurankan: I think we should also store the computed strengths in the graph itself. This could help us with plotting functionality. For each edge, we can do something like:
```
self.edges[edge]['strength'] = compute_value
```

pgmpy/base/DAG.py
```diff
@@ -1257,6 +1257,107 @@ def copy(self):
...
+
+        for edge in edges_to_compute:
+            x, y = edge
+
+            # Get parents of x and y using get_parents instead of predecessors
+            pa_X = self.get_parents(x)
+            pa_Y = self.get_parents(y)
+
+            # Check if either x or y is a latent node
+            if (
```
ankurankan, line 1442: In the case when we are computing the strength for a list of edges, it might be better to show a warning for the edge that we can't compute the strength for and compute for the remaining. What do you think?

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -360,6 +361,52 @@ def test_dag_fit(self):
...
+        dag = DAG([("X", "Y"), ("Z", "Y")])
+        data = pd.DataFrame({"X": [0, 1, 0, 1], "Y": [1, 3, 0, 2], "Z": [1, 1, 0, 0]})
+        strengths = dag.edge_strength(data, edges=[("X", "Y"), ("Z", "Y")])
+
+        # Test structure
+        self.assertEqual(set(strengths.keys()), {("X", "Y"), ("Z", "Y")})
+
+        # Test numerical values from docstring
+        self.assertAlmostEqual(strengths[("X", "Y")], 0.999999754781587, places=15)
+        self.assertAlmostEqual(strengths[("Z", "Y")], 0.9999989719376587, places=15)
```
ankurankan: Did the LinearGaussian BN tests still didn't work when comparing to the square root of strength?

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -360,6 +361,52 @@ def test_dag_fit(self):
     def tearDown(self):
         del self.graph
 
+    def test_edge_strength_basic(self):
```
ankurankan, line 469: Could you also add some tests with latent variables?

ankurankan (changes requested):

pgmpy/base/DAG.py
Manas-7854: yeah that would be better, will do that

ankurankan (approved):
@Manas-7854 Looks good to me. I have added a few minor changes in docstring and code.

PR #2060, opened 2025-05-09
Raise error for invalid evidence variables in VariableElimination

Susmita331:
This PR improves the robustness of VariableElimination by adding an explicit check for invalid evidence variables during query execution.
Fix: Modified _get_working_factors() to raise a ValueError if any evidence variable is not part of the model.
Test Added: test_query_invalid_evidence_variable_with_example_model using get_example_model("alarm") to ensure error is raised for realistic usage with standard models.
Previously, providing an invalid variable in the evidence (e.g., {"Z": 1} where Z is not in the model) caused a KeyError. This fix provides a clearer, user-friendly ValueError with an informative message.

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -53,6 +53,10 @@ class VariableElimination(Inference):
         # Dealing with evidence. Reducing factors over it before VE is run.
         if evidence:
             for evidence_var in evidence:
+                if evidence_var not in working_factors:
+                    raise ValueError(
+                        f"Evidence variable '{evidence_var}' is not present in the model."
+                    )
                 for factor, origin in working_factors[evidence_var]:
                     factor_reduced = factor.reduce(
                         [(evidence_var, evidence[evidence_var])], inplace=False
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -53,8 +53,6 @@ class VariableElimination(Inference):
         # Dealing with evidence. Reducing factors over it before VE is run.
         if evidence:
             for evidence_var in evidence:
-                if evidence_var not in working_factors:
-                    raise ValueError(f"Node {evidence_var} not in graph")
                 for factor, origin in working_factors[evidence_var]:
                     factor_reduced = factor.reduce(
                         [(evidence_var, evidence[evidence_var])], inplace=False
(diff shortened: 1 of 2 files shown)


pgmpy/inference/ExactInference.py
```diff
@@ -53,6 +53,8 @@ def _get_working_factors(self, evidence):
         # Dealing with evidence. Reducing factors over it before VE is run.
         if evidence:
             for evidence_var in evidence:
+                if evidence_var not in working_factors:
```
ankurankan: I think this should go with other checks in the `query` and `map_query` methods.

ankurankan (changes requested):
@Susmita331 Please have a look at the inline comment

Susmita331:
@ankurankan can you please tell me why one test case is falling even though I have added unit test for change I made

Susmita331:
@ankurankan kindly review it

pgmpy/inference/ExactInference.py
```diff
@@ -310,10 +313,13 @@ def query(
                 f"Can't have the same variables in both `variables` and `evidence`. Found in both: {common_vars}"
             )
 
-        if not variables:
-            raise ValueError(
-                "The `variables` argument to query() must contain at least one variable."
-            )
+        if isinstance(self.model, DiscreteBayesianNetwork):
```
ankurankan, line 316: We don't need this check again. Right? There's already a check for this in line 300.

pgmpy/inference/ExactInference.py
```diff
@@ -310,10 +313,13 @@ def query(
...
+            for evidence_var in evidence:
```
ankurankan, line 317: Why is this check not at the top of the function around line 298?

ankurankan (changes requested):

pgmpy/inference/ExactInference.py
Susmita331: You're right — the check for empty variables was already there near the top (around line 298), so having it again at line 314 was redundant. I actually moved the check upwards on purpose, which leads to the next point.

I moved it earlier in the function so that we can catch the missing variables case right away before doing anything else. Since variables is a required argument, it’s better to raise an error as soon as possible instead of doing extra work like checking evidence or model type. It also avoids confusion if any later error hides the real issue of missing variables.

The check if evidence_var not in self.model.nodes() depends on evidence being a dictionary. But evidence could be None, so we first convert it to an empty dictionary (evidence = evidence or {}) before doing this check. That’s why the check comes a little later — it has to wait until evidence is safely defined.

Thanking you

Susmita331:
@ankurankan any feedback?

pgmpy/inference/ExactInference.py
ankurankan: I didn't mean the checks for the variable. I meant the check for `self.model` being a `DiscreteBayesianNetwork`. We already check that in lines 299-301.

ankurankan (changes requested):
@Susmita331 It seems like you are using AI tools for making the changes, which is leading to a few obvious mistakes. While it’s absolutely fine to use such tools as a starting point, I do expect that you fully understand and review the changes before opening a PR. Otherwise, it just makes the review process more difficult and time-consuming on my end.

PR #2067, opened 2025-05-09
Getting rid of skipping tests and ensured tests use custom function

### LINKED ISSUE #2065: [TEST] Add tests for ExpertInLoop using mock
Currently, we are skipping the tests for ExpertInLoop algorithm in Continuous Integration on realistic data, as it makes API calls to Google Gemini. #2056 has modularized the implementation of the estimate method such that a custom function can be passed to query the pairwise orientation. We can now also utilize this to write tests without making calls to the LLM.

@Green-Kedia Since you worked on the ExpertInLoop algorithm, would you like to take this up?

Green-Kedia:

- Fixes #2065

- 
-
-

### DIFF
--- a/pgmpy/tests/test_estimators/test_expert.py
+++ b/pgmpy/tests/test_estimators/test_expert.py
@@ -1,6 +1,7 @@
 import os
 import unittest
 
+import networkx as nx
 import pandas as pd
 import pytest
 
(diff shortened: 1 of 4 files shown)


Green-Kedia:
Hi @ankurankan, sorry about the failed tests, I’m currently travelling and will look into it soon. From a quick glance, it seems like the remove_edge method doesn’t fully remove the edge but just deletes it. I’ll need to investigate further.
Thanks!

pgmpy/tests/test_estimators/test_expert.py
```diff
@@ -74,6 +75,54 @@ def setUp(self):
             ("Age", "Education"),
         }
 
+    def test_estimate(self):
+        """Test the basic functionality of estimate method"""
+
+        # Use a simple orientation function that doesn't require LLM
+        def simple_orient(var1, var2, **kwargs):
```
ankurankan: For the full model test, I would suggest using a "true" model that you can use to answer the orientation questions. So, I would suggest first defining a DAG on the variables that we can use as the true model and then in each call of the orientation function we can check the edge direction in this "true dag" and return the result accordingly. At the end, we can then test if our algorithm was able to recover the "true dag" that we had defined.

ankurankan (changes requested):

Green-Kedia:
Hi @ankurankan, sorry for the delay, but I think this is what you suggested?

ankurankan (approved):
Looks good to me @Green-Kedia . I had made some changes to the ExpertInLoop algorithm, so I have added a few commits to this PR to adjust the tests according to the changes.

PR #2070, opened 2025-05-11
fix: refactor do method for correct causal semantics [WIP]

### LINKED ISSUE #2007: Incorrect Marginalization Logic in `DiscreteBayesianNetwork.remove_node`
Hi pgmpy team,

I've identified a potential issue regarding how node removal is handled in the `DiscreteBayesianNetwork` class, specifically concerning the use of the `marginalize` method on children's CPDs.

**Context:**

*   **File:** `pgmpy/models/DiscreteBayesianNetwork.py`
*   **Specific Line (Problematic):** [https://github.com/pgmpy/pgmpy/blob/eb4d592c0e3e8448fcff10ee34045549673afbe2/pgmpy/models/DiscreteBayesianNetwork.py#L174](https://github.com/pgmpy/pgmpy/blob/eb4d592c0e3e8448fcff10ee34045549673afbe2/pgmpy/models/DiscreteBayesianNetwork.py#L174)
*   **Related Function:** `pgmpy/factors/discrete/DiscreteFactor.py#L387` (`marginalize`)
*   **Commit Hash:** `eb4d592c0e3e8448fcff10ee34045549673afbe2`

**Problem Description:**

The `remove_node(node)` method is intended to remove a specified `node` from the Bayesian Network. Part of this process involves updating the CPDs of the nodes that were children of the removed `node`. The current implementation does this around line 174:

```python
def remove_node(self, node):
    # ... (other code) ...
    affected_nodes = [v for u, v in self.edges() if u == node] # Find children
    for affected_node in affected_nodes: # Iterate through children
        node_cpd = self.get_cpds(node=affected_node) # Get child's CPD
        if node_cpd:
            # This line seems probabilistically incorrect:
            node_cpd.marginalize([node], inplace=True)
    # ... (rest of the code) ...
```

This code attempts to modify the child's CPD (`node_cpd`, representing e.g., `P(Child | RemovedNode, OtherParents)`) by calling `marginalize` to sum out the `RemovedNode`.

**Why this is incorrect:**

Removing a node `X` from a Bayesian Network requires marginalizing `X` out of the *full joint probability distribution* `P(V)` to obtain the correct joint distribution `P(V \ {X})` over the remaining variables. The factorization of this new joint distribution involves updated conditional dependencies.

The operation performed by the code, `sum_x P(Child | X, OtherParents)`, where `X` is the removed node, does **not** generally yield the correct conditional probability distribution `P(Child | OtherParents)` required for the network *after* `X` has been marginalized out from the overall joint distribution. The `marginalize` method correctly sums out a variable within a given factor (the CPD), but applying it this way to a child's CPD when removing a parent does not correctly update the model's probabilistic semantics.

**Example:**

Consider a simple graph `X -> Y`. The joint distribution is $P(X, Y) = P(Y | X) P(X)$.
If we want to remove node `X`, the goal is to obtain the marginal distribution $P(Y)$. The correct calculation is $P(Y) = \sum_x P(Y, X) = \sum_x P(Y | X) P(X)$.
The current `remove_node` code, when removing `X`, takes the CPD $P(Y | X)$ and computes $\sum_x P(Y | X)$. This result is generally not equal to the true marginal $P(Y)$ unless $P(X)$ was uniform and certain other conditions hold, which is not the general case. The code incorrectly modifies the factor $P(Y | X)$ instead of performing the correct marginalization involving all relevant factors.

**Suggested Correction (Conceptual):**

Correctly removing a node and obtaining the marginalized model often requires more involved operations than just modifying the children's CPDs locally. It might involve:
1.  Calculating the full joint distribution (or relevant parts) using all original CPDs.
2.  Performing marginalization (e.g., via variable elimination) on this joint factor to sum out the `node`.
3.  Potentially re-factorizing the resulting marginal joint distribution if a Bayesian Network representation is still desired (though this can be complex).

Alternatively, if the goal is just to allow inference *without* the removed node, inference algorithms should be applied to the original model structure while summing out the target node during the query process, rather than permanently altering the model's CPDs in this potentially incorrect way beforehand.

**Related Concerns in Other Methods:**

This specific misuse of `marginalize` (applying it to a factor in a way that doesn't reflect the intended global probabilistic operation) might indicate a pattern. It's worth reviewing other parts of the codebase where model structure or distributions are modified:

1.  **`do` operation:** As noted in a potential previous discussion (linking to the code below), the `do` method also uses `marginalize` on the *intervened node's* CPD to sum out its parents. This is conceptually different but also seems like an incorrect application of marginalization in the context of a do-intervention.
    *   [https://github.com/pgmpy/pgmpy/blob/eb4d592c0e3e8448fcff10ee34045549673afbe2/pgmpy/models/DiscreteBayesianNetwork.py#L1195](https://github.com/pgmpy/pgmpy/blob/eb4d592c0e3e8448fcff10ee34045549673afbe2/pgmpy/models/DiscreteBayesianNetwork.py#L1195)
2.  **`predict_probability`:** This method uses inference (`VariableElimination`) which internally relies heavily on factor operations including marginalization. While the specific line highlighted below seems correct (marginalizing a query result), it's crucial that the underlying inference engine's use of marginalization during the elimination process is sound, especially if it relies on similar logic to `remove_node`.
    *   [https://github.com/pgmpy/pgmpy/blob/eb4d592c0e3e8448fcff10ee34045549673afbe2/pgmpy/models/DiscreteBayesianNetwork.py#L827](https://github.com/pgmpy/pgmpy/blob/eb4d592c0e3e8448fcff10ee34045549673afbe2/pgmpy/models/DiscreteBayesianNetwork.py#L827)

Could you please review the marginalization logic within `remove_node` and potentially related methods like `do`? The current approach in `remove_node` appears to be probabilistically incorrect.

Thank you!

mdrazak2001:

- Fixes #2007

- removed `marginalize` method in do operation
- setting a probability `1` for the state specified by the user, thus eliminating the effect of parents caused marginalization of them.

### DIFF
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1144,7 +1144,7 @@ class DiscreteBayesianNetwork(DAG):
         else:
             return cpds
 
-    def do(self, nodes, inplace=False):
+    def do(self, nodes, states, inplace=False):
         """
         Applies the do operation. The do operation removes all incoming edges
         to variables in `nodes` and marginalizes their CPDs to only contain the
@@ -1181,6 +1181,16 @@ class DiscreteBayesianNetwork(DAG):
         else:
             nodes = list(nodes)
 
+        if isinstance(states, (str, int)):
+            states = [states]
+        else:
+            states = list(states)
+
+        if len(nodes) != len(states):
+            raise ValueError(
+                f"Length of nodes and states should be same. Got {len(nodes)} and {len(states)}"
+            )
+
         if not set(nodes).issubset(set(self.nodes())):
             raise ValueError(
                 f"Nodes not found in the model: {set(nodes) - set(self.nodes)}"
@@ -1190,9 +1200,28 @@ class DiscreteBayesianNetwork(DAG):
         adj_model = DAG.do(model, nodes, inplace=inplace)
 
         if adj_model.cpds:
-            for node in nodes:
+            for index, node in enumerate(nodes):
                 cpd = adj_model.get_cpds(node=node)
-                cpd.marginalize(cpd.variables[1:], inplace=True)
+                state_names = cpd.state_names[node]
+                if states[index] not in state_names:
+                    raise ValueError(
+                        f"State {states[index]} not defined for {node} in the model."
+                    )
+                idx = state_names.index(states[index])
+                card = cpd.variable_card
+                new_vals = [[1.0 if i == idx else 0.0] for i in range(card)]
+
+                # Drop the old CPD
+                adj_model.remove_cpds(cpd)
+                # Create new CPD by setting the values to 1.0 for the state in `states`
+                spike_cpd = TabularCPD(
+                    variable=node,
+                    variable_card=card,
+                    values=new_vals,
+                    state_names={node: state_names},
+                )
+                adj_model.add_cpds(spike_cpd)
+
         return adj_model
 
     def simulate(
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1358,16 +1358,24 @@ class DiscreteBayesianNetwork(DAG):
 
         # Step 1: If do or virtual_intervention is specified, modify the network structure.
         if (do != {}) or (virtual_intervention != []):
-            # Create a combined intervention dictionary
+            # Create a combined intervention dictionary for hard interventions
             combined_interventions = do.copy()
 
-            # Add virtual interventions to the combined dictionary
+            # Handle virtual interventions separately
             for cpd in virtual_intervention:
                 var = cpd.variables[0]
-                # For virtual interventions, we need to find the state with highest probability
-                state_idx = np.argmax(cpd.values)
-                state = cpd.state_names[var][state_idx]
-                combined_interventions[var] = state
+                if var not in model.nodes():
+                    raise ValueError(
+                        f"Virtual intervention variable {var} not in model"
+                    )
+
+                # Remove incoming edges to the variable
+                for parent in list(model.predecessors(var)):
+                    model.remove_edge(parent, var)
+
+                # Replace the variable's CPD with the virtual intervention CPD
+                model.remove_cpds(model.get_cpds(var))
+                model.add_cpds(cpd)
 
             # Apply the combined interventions
             model = model.do(combined_interventions)
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1359,8 +1359,19 @@ class DiscreteBayesianNetwork(DAG):
 
         # Step 1: If do or virtual_intervention is specified, modify the network structure.
         if (do != {}) or (virtual_intervention != []):
-            virt_nodes = [cpd.variables[0] for cpd in virtual_intervention]
-            model = model.do(list(do.keys()) + virt_nodes)
+            # Create a combined intervention dictionary
+            combined_interventions = do.copy()
+            
+            # Add virtual interventions to the combined dictionary
+            for cpd in virtual_intervention:
+                var = cpd.variables[0]
+                # For virtual interventions, we need to find the state with highest probability
+                state_idx = np.argmax(cpd.values)
+                state = cpd.state_names[var][state_idx]
+                combined_interventions[var] = state
+            
+            # Apply the combined interventions
+            model = model.do(combined_interventions)
             evidence = {**evidence, **do}
             virtual_evidence = [*virtual_evidence, *virtual_intervention]
 
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1181,14 +1181,13 @@ class DiscreteBayesianNetwork(DAG):
                 "`do` expects a dict of the form {variable: state}, "
                 f"got {type(interventions)}"
             )
-        
-        nodes  = list(interventions.keys())
+
+        nodes = list(interventions.keys())
         states = list(interventions.values())
 
         missing_nodes = set(nodes) - set(self.nodes())
         if missing_nodes:
             raise ValueError(f"Intervention variable(s) not in model: {missing_nodes}")
-        
 
         for var, forced_state in interventions.items():
             cpd = self.get_cpds(node=var)
@@ -1196,7 +1195,7 @@ class DiscreteBayesianNetwork(DAG):
                 raise ValueError(
                     f"State {forced_state!r} invalid for variable {var!r}; "
                     f"allowed states: {cpd.state_names[var]}"
-                )     
+                )
 
         model = self if inplace else self.copy()
         adj_model = DAG.do(model, nodes, inplace=inplace)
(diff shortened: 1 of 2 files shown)


mdrazak2001:
@ankurankan I have created this PR to remove marginalization of the intervened nodes parents and this is still in WIP.
Could you please let me know if the refactor of `do` method looks good?
Any other feedback is also much appreciated!

Thanks!

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1144,7 +1144,7 @@ def get_random_cpds(self, n_states=None, inplace=False, seed=None):
         else:
             return cpds
 
-    def do(self, nodes, inplace=False):
+    def do(self, nodes, states, inplace=False):
```
ankurankan: I think it would be better to accept a dict of the form `{node: state}`

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1190,9 +1200,28 @@ def do(self, nodes, inplace=False):
         adj_model = DAG.do(model, nodes, inplace=inplace)
 
         if adj_model.cpds:
-            for node in nodes:
+            for index, node in enumerate(nodes):
                 cpd = adj_model.get_cpds(node=node)
-                cpd.marginalize(cpd.variables[1:], inplace=True)
+                state_names = cpd.state_names[node]
+                if states[index] not in state_names:
```
ankurankan: We usually try to do all the checks at the beginning of functions; it can help with readability as it separates outs checks and actual code logic. So, I would suggest putting this somewhere at the start of the method with the other checks.

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1181,6 +1181,16 @@ def do(self, nodes, inplace=False):
         else:
             nodes = list(nodes)
 
+        if isinstance(states, (str, int)):
```
ankurankan: I think these checks will get quite a bit simplified after accepting dict instead of two arguments.

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1190,9 +1200,28 @@ def do(self, nodes, inplace=False):
...
-            for node in nodes:
+            for index, node in enumerate(nodes):
                 cpd = adj_model.get_cpds(node=node)
-                cpd.marginalize(cpd.variables[1:], inplace=True)
+                state_names = cpd.state_names[node]
+                if states[index] not in state_names:
+                    raise ValueError(
+                        f"State {states[index]} not defined for {node} in the model."
+                    )
+                idx = state_names.index(states[index])
```
ankurankan: TabularCPD object has a method `get_state_no(var, state_name)` that you can use here instead. It utilizes pre-computed state_name to number mapping to save compute.

ankurankan (changes requested):
@mdrazak2001 The PR mostly looks good. I have added a few minor comments.

ankurankan:
@mdrazak2001 Some of the tests will need to be fixed because of the changes.

mdrazak2001:
@ankurankan: Have made all the requested changes.
Also Had to refactor `DiscreteBayesianNetwork.simulate` that invoked the new `do` method, I did this by copying the provided do param and approximated to each `virtual_intervention`'s state to 1 whose probability is maximum.
does the change with respect to `virtual_intervention` look good?

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1176,23 +1176,50 @@ def do(self, nodes, inplace=False):
         OutEdgeView([('asia', 'tub'), ('tub', 'either'), ('smoke', 'lung'), ('lung', 'either'),
```
ankurankan, line 1176: We will need to update the examples and the docstring.

ankurankan (changes requested):
@mdrazak2001 Sorry for the delay in reviewing this. I think the `do` method looks good. I am still unsure about the correct logic with virtual intervention / do-operation. I will need some more time to think about this and will get back to this PR.

ankurankan:
@mdrazak2001 I thought about this, and the logic for virtual intervention should be very similar to the do operator. With the standard do operation, we remove all the incoming edges and assign the full probability weight to the variable state (spike) that the intervened variable is set to. In case of virtual intervention, we are not selecting a single state, rather assigning a distribution to the variable. The virtual intervention CPD defines this distribution (as it doesn't have any parents), so we should similarly remove all incoming edges to the variable and use the virtual intervention CPD as the new CPD of that variable.

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1176,23 +1176,50 @@ def do(self, nodes, inplace=False):
...
-        if isinstance(nodes, (str, int)):
-            nodes = [nodes]
-        else:
-            nodes = list(nodes)
-
-        if not set(nodes).issubset(set(self.nodes())):
+        if not isinstance(interventions, dict):
             raise ValueError(
-                f"Nodes not found in the model: {set(nodes) - set(self.nodes)}"
+                "`do` expects a dict of the form {variable: state}, "
```
ankurankan, line 1181: `do` should be `interventions` here?

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1176,23 +1176,50 @@ def do(self, nodes, inplace=False):
...
+        nodes = list(interventions.keys())
+        states = list(interventions.values())
+
+        missing_nodes = set(nodes) - set(self.nodes())
+        if missing_nodes:
+            raise ValueError(f"Intervention variable(s) not in model: {missing_nodes}")
+
+        for var, forced_state in interventions.items():
+            cpd = self.get_cpds(node=var)
+            if forced_state not in cpd.state_names[var]:
```
ankurankan, line 1194: Can we move this check to the beginning of the method? I think it makes the main part of the logic a bit cleaner.

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1331,8 +1358,27 @@ def simulate(
...
         if (do != {}) or (virtual_intervention != []):
-            virt_nodes = [cpd.variables[0] for cpd in virtual_intervention]
-            model = model.do(list(do.keys()) + virt_nodes)
+            # Create a combined intervention dictionary for hard interventions
+            combined_interventions = do.copy()
+
+            # Handle virtual interventions separately
+            for cpd in virtual_intervention:
+                var = cpd.variables[0]
+                if var not in model.nodes():
```
ankurankan, line 1367: Same here, can we move this check to the first part of the method.

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1331,8 +1358,27 @@ def simulate(
 
         # Step 1: If do or virtual_intervention is specified, modify the network structure.
         if (do != {}) or (virtual_intervention != []):
-            virt_nodes = [cpd.variables[0] for cpd in virtual_intervention]
-            model = model.do(list(do.keys()) + virt_nodes)
+            # Create a combined intervention dictionary for hard interventions
+            combined_interventions = do.copy()
```
ankurankan, line 1362: `combined_interventions` seems a confusing name to me. Maybe `hard_intervention` or just keep it as `do` ?

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1176,23 +1176,50 @@ def do(self, nodes, inplace=False):
...
-                f"Nodes not found in the model: {set(nodes) - set(self.nodes)}"
+                "`do` expects a dict of the form {variable: state}, "
+                f"got {type(interventions)}"
             )
 
+        nodes = list(interventions.keys())
+        states = list(interventions.values())
+
+        missing_nodes = set(nodes) - set(self.nodes())
+        if missing_nodes:
```
ankurankan, line 1189: Might be a good use case for Python's walrus operator.

ankurankan (changes requested):
@mdrazak2001 I think this looks good now. I have added a few comments for mostly cosmetic changes.

PR #2072, opened 2025-05-11
Fixes #2021: Fix duplicate logging of datatype detection messages

### LINKED ISSUE #2021: Fix detected datatype INFO when calling estimation methods

Akhilesh-max:

- Fixes #2021 

- Added a DuplicateFilter to the global logger in pgmpy/global_vars.py. This filter keeps track of messages that have been logged and prevents identical messages from being shown more than once.

### DIFF
--- a/pgmpy/global_vars.py
+++ b/pgmpy/global_vars.py
@@ -7,6 +7,22 @@ logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger("pgmpy")
 
 
+class DuplicateFilter(logging.Filter):
+    def __init__(self):
+        super().__init__()
+        self.msgs = set()
+
+    def filter(self, record):
+        msg = record.getMessage()
+        is_new = msg not in self.msgs
+        if is_new:
+            self.msgs.add(msg)
+        return is_new
+
+
+logger.addFilter(DuplicateFilter())
+
+
 class Config:
     def __init__(self):
         """
--- a/pgmpy/global_vars.py
+++ b/pgmpy/global_vars.py
@@ -8,15 +8,20 @@ logger = logging.getLogger("pgmpy")
 
 
 class DuplicateFilter(logging.Filter):
+    """
+    A logging filter that prevents duplicate consecutive log messages.
+    This filter only allows a message to pass through if it differs from the previous message.
+    """
+
     def __init__(self):
         super().__init__()
-        self.msgs = set()
+        self.last_msg = None
 
     def filter(self, record):
         msg = record.getMessage()
-        is_new = msg not in self.msgs
+        is_new = msg != self.last_msg
         if is_new:
-            self.msgs.add(msg)
+            self.last_msg = msg
         return is_new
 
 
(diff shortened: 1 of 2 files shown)


Akhilesh-max:
@ankurankan Attempted a fix for #2021. PTAL when get a chance. 
Thanks!

pgmpy/global_vars.py
```diff
@@ -7,6 +7,22 @@
...
 
 
+class DuplicateFilter(logging.Filter):
+    def __init__(self):
+        super().__init__()
+        self.msgs = set()
+
+    def filter(self, record):
+        msg = record.getMessage()
+        is_new = msg not in self.msgs
```
ankurankan: I would suggest changing it to only look at the last message instead of all messages.

pgmpy/global_vars.py
```diff
@@ -7,6 +7,22 @@
 logger = logging.getLogger("pgmpy")
 
 
+class DuplicateFilter(logging.Filter):
```
ankurankan, line 10: Should add a docstring explaining what the class does.

ankurankan (changes requested):

pgmpy/global_vars.py
Akhilesh-max: done

pgmpy/global_vars.py
Akhilesh-max: Added.

Akhilesh-max:
@ankurankan Updated as per the review. PTAL when get a chance. 
Thank you!

Nimish-4:
@Akhilesh-max Looks good, was adding a testcase also discussed, to ensure its working?

Akhilesh-max:
@ankurankan @Nimish-4 I've added a test-case too. PTAL.

Akhilesh-max:
@ankurankan Please take a look when get a chance. Thanks!

ankurankan (approved):

ankurankan:
@Akhilesh-max Looks good to me. Thanks for the PR.

PR #2074, opened 2025-05-12
Implementation of an alternate (faster) algorithm to check if two nodes are d-connected or not.

### LINKED ISSUE #1066: Alternate algorithms for finding d-separation
### Subject of the issue
The current implementation of finding d-connected nodes is based on [1] which is great when we want to find all the nodes which are d-connected to some given node. But is slow in the case when we want to just check if two nodes are d-connected or not. [2] and [3] propose faster algorithms (O(|E|) and O(|V|^2) respectively) for checking whether two nodes are d-connected. Implement one of these algorithms to check if d-separation between two given nodes.

[1] (Algorithm 3.1 page 84) Koller, D., Friedman, N., & Bach, F. (2009). Probabilistic graphical models: principles and techniques. MIT press.
[2] Geiger, D., Verma, T., & Pearl, J. (1990). d-separation: From theorems to algorithms. In Machine Intelligence and Pattern Recognition (Vol. 10, pp. 139-148). North-Holland.
[3] Lauritzen, S. L., Dawid, A. P., Larsen, B. N., & Leimer, H. G. (1990). Independence properties of directed Markov fields. Networks, 20(5), 491-505.

### Your environment
* pgmpy version: All
* Python version: All
* Operating System: All

aniket2405:

- Fixes #1066

1. Implementation of a new (faster) algorithm to check if two nodes are d-connected or not. This algorithm is based on Geiger, et al. from the paper -
d-separation: From theorems to algorithms. In Machine Intelligence and Pattern Recognition (Vol. 10, pp. 139-148). North-Holland
- 
-
-

### DIFF
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -619,6 +619,98 @@ class DAG(nx.DiGraph):
             immoralities[node] = parent_pairs
         return immoralities
 
+    def is_dconnected_fast(self, start, end, observed=None, include_latents=False):
+        """
+        Returns True if there is an active trail (i.e. d-connection) between
+        `start` and `end` node given that `observed` is observed.
+
+        Time complexity: O(E)
+
+        Parameters
+        ----------
+        start, end : int, str, any hashable python object.
+            The nodes in the DAG between which to check the d-connection/active trail.
+
+        observed : list, array-like (optional)
+            If given the active trail would be computed assuming these nodes to
+            be observed.
+
+        include_latents: boolean (default: False)
+            If true, latent variables are included in the active trail analysis.
+
+        Returns
+        -------
+        bool
+            True if there's an active trail between start and end, False otherwise.
+        """
+        from collections import deque
+
+        # Handle None for observed
+        if observed is None:
+            observed = set()
+        else:
+            observed = set(observed)
+
+        # Initialize data structures
+        visited = set()  # Tracks (node, direction) pairs
+        active_nodes = set()  # Nodes in the active trail
+        queue = deque()
+
+        # Get ancestors of observed nodes for v-structure handling
+        ancestors_list = self._get_ancestors_of(observed)
+
+        # Initialize queue with start node in both directions
+        queue.append((start, "up"))
+        queue.append((start, "down"))
+
+        while queue:
+            current, direction = queue.popleft()
+
+            # Skip if already visited in this direction
+            if (current, direction) in visited:
+                continue
+            visited.add((current, direction))
+
+            # Add to active nodes if not observed
+            if current not in observed:
+                active_nodes.add(current)
+
+            # Determine next nodes to visit based on d-separation rules
+            if direction == "up" and current not in observed:
+                # Coming from child to this node, not observed
+                # Can go to parents (moving up)
+                for parent in self.predecessors(current):
+                    queue.append((parent, "up"))
+
+                # Can go to children (moving down)
+                for child in self.successors(current):
+                    queue.append((child, "down"))
+
+            elif direction == "down":
+                # Coming from parent to this node
+                if current not in observed:
+                    # If not observed, can go to children
+                    for child in self.successors(current):
+                        queue.append((child, "down"))
+
+                # If node is in ancestors of observed, can go up (v-structure)
+                if current in ancestors_list:
+                    for parent in self.predecessors(current):
+                        queue.append((parent, "up"))
+
+        # Check if end is in active nodes
+        if end in active_nodes:
+            # If include_latents is False, need to check if there's a path without latents
+            if not include_latents:
+                # If end node is a latent, it's not d-connected when include_latents=False
+                if end in self.latents:
+                    return False
+
+                # Here we just check if end node is in the active nodes after removing latents
+                return end in (active_nodes - self.latents)
+            return True
+        return False
+
     def is_dconnected(self, start, end, observed=None, include_latents=False):
         """
         Returns True if there is an active trail (i.e. d-connection) between
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -619,34 +619,33 @@ class DAG(nx.DiGraph):
             immoralities[node] = parent_pairs
         return immoralities
 
-    def is_dconnected_moral(self, start, end, observed=None, include_latents=False):
+    def is_dconnected(self, start, end, observed=None):
         """
-        Returns True if `start` and `end` are d-connected given `observed`
-        according to moral‐graph test.
-
-        Time complexity: O(V + E) for sparse / bounded–in-degree graphs.
+        Returns True if `start` and `end` are d-connected when conditioned on
+        `observed` variables.
 
         Parameters
         ----------
         start, end : hashable
             The two nodes X, Y whose d-connection we want to test.
         observed : iterable (optional)
-            The conditioning set Z.  Treated as “removed” in the moral graph.
-        include_latents : bool (default=False)
-            If False, latent nodes are also treated like observed (i.e. removed).
+            The conditioning set Z.
 
         Returns
         -------
         bool
             True if there's an active trail between start and end, False otherwise.
+
+        References
+        ----------
+        [1] Theorem 4.1 from 2009 Modeling and Reasoning with Bayesian Networks (562s,Adnan Darwiche).
+        Please refer: https://shorturl.at/UgQ6F
         """
         # 1. Build the observed set Z
         if observed is None:
             Z = set()
         else:
             Z = set(observed)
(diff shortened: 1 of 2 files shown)


aniket2405:
@ankurankan Please review and let me know what changes would you like to see/discuss about. Thanks for the help!

ankurankan:
@aniket2405 Isn't this the same algorithm that we have currently implemented in `DAG.active_trail_nodes` method?

ankurankan:
@aniket2405 The algorithm that we need is the one described in Theorem 4.1 of https://github.com/alphabetakappa/Probabilistic-Graphical-Models-Materials/blob/master/2009%20Modeling%20and%20Reasoning%20with%20Bayesian%20Networks%20(562s%2CAdnan%20Darwiche)Good.pdf

aniket2405:
Thanks for the review @ankurankan 

Yes, it is a similar algorithm but we're not looking for all the nodes there. So, the time complexity is O(E). I gathered that will make it more efficient.

Will take a look again at the algorithm you linked and let you know.

ankurankan:
> Yes, it is a similar algorithm but we're not looking for all the nodes there. 

Thanks for clarifying that. I had missed it.

While the other algorithm that I have linked is also O(E) but I think in practice it will still be much faster. It is also much simpler to implement.

aniket2405:
@ankurankan I've implemented the algorithm based on the one you had linkedin (Theorem 4.1).
Could you please take a look and review this? Thanks!

pgmpy/base/DAG.py
```diff
@@ -619,6 +619,88 @@ def get_immoralities(self):
...
+        observed : iterable (optional)
+            The conditioning set Z.  Treated as “removed” in the moral graph.
+        include_latents : bool (default=False)
+            If False, latent nodes are also treated like observed (i.e. removed).
+
+        Returns
+        -------
+        bool
+            True if there's an active trail between start and end, False otherwise.
+        """
```
ankurankan, line 655: Add a reference section here with a reference to the book

pgmpy/base/DAG.py
```diff
@@ -619,6 +619,88 @@ def get_immoralities(self):
             immoralities[node] = parent_pairs
         return immoralities
 
+    def is_dconnected_moral(self, start, end, observed=None, include_latents=False):
```
ankurankan: I don't think we need to `include_latents` argument here. The idea with `include_latents` in `active_trail_nodes` is to specify whether we want to return all variables that are d-connected or just observed variables.

pgmpy/base/DAG.py
```diff
@@ -619,6 +619,88 @@ def get_immoralities(self):
...
+        """
+        Returns True if `start` and `end` are d-connected given `observed`
```
ankurankan: I would suggest a slight change here: `... are d-connected when conditioned on `observed` variables`. I would also remove the moral graph part, as it is a detail related to the algorithm implemented and has nothing to do with the method's functionality.

pgmpy/base/DAG.py
```diff
@@ -619,6 +619,88 @@ def get_immoralities(self):
...
+        according to moral‐graph test.
+
+        Time complexity: O(V + E) for sparse / bounded–in-degree graphs.
```
ankurankan: No need to add this as I don't think it adds anything for the user.

pgmpy/base/DAG.py
```diff
@@ -619,6 +619,88 @@ def get_immoralities(self):
...
+        -------
+        bool
+            True if there's an active trail between start and end, False otherwise.
+        """
+        # 1. Build the observed set Z
+        if observed is None:
+            Z = set()
+        else:
+            Z = set(observed)
+        if not include_latents:
```
ankurankan: I think this not needed.

pgmpy/base/DAG.py
```diff
@@ -619,6 +619,88 @@ def get_immoralities(self):
...
+        ancestors = set()
+        while to_visit:
+            node = to_visit.pop()
+            if node not in ancestors:
+                ancestors.add(node)
+                for p in self.predecessors(node):
+                    to_visit.append(p)
+
+        # 3. Induce subgraph on A, and
+        # 4. Moralize it → build an undirected adjacency U: dict[node] -> set(neighbors)
```
ankurankan: There is already a moralize method `DAG.moralize` implemented. Please use that instead.

pgmpy/base/DAG.py
ankurankan: This method should replace the current `DAG.is_dconnected` method.

ankurankan (changes requested):

ankurankan:
@aniket2405 Thanks for the PR. I have added a few comments, but the algorithm looks correct otherwise.

ankurankan:
Hi @aniket2405, are you still working on this?

aniket2405:
Hi @ankurankan, yes - I'm working on this. I got busy with some personal work, which took more time than expected.
I'll update the patch and raise a request.

pgmpy/base/DAG.py
```diff
@@ -648,43 +648,83 @@ def is_dconnected(
...
-            be observed.
-
-        include_latents: boolean (default: False)
-            If true, latent variables are return as part of the active trail.
+        Returns
+        -------
+        bool
+            True if there's an active trail between start and end, False otherwise.
 
-        Examples
```
ankurankan, line 639: We need to keep the examples.

pgmpy/base/DAG.py
```diff
@@ -648,43 +648,83 @@ def is_dconnected(
...
-            )[start]
-        ):
-            return True
+        # 1. Build the observed set Z
+        if observed is None:
+            Z = set()
         else:
+            Z = set(observed)
+
+        # 2. Compute A = Ancestors({start, end} ∪ Z)
```
ankurankan: Check `DAG._get_ancestors_of`

pgmpy/base/DAG.py
```diff
@@ -648,43 +648,83 @@ def is_dconnected(
...
+        ancestors = set()
+        while to_visit:
+            node = to_visit.pop()
+            if node not in ancestors:
+                ancestors.add(node)
+                for p in self.predecessors(node):
+                    to_visit.append(p)
+
+        # 3. Create subgraph induced by ancestors
+        subgraph = self.__class__()  # Create instance of same class as self
```
ankurankan: Check `DAG.get_ancestral_graph`

ankurankan (changes requested):

pgmpy/base/DAG.py
aniket2405: Ack

pgmpy/base/DAG.py
aniket2405: Ack

pgmpy/base/DAG.py
aniket2405: Ack

pgmpy/base/DAG.py
aniket2405: Ack

pgmpy/base/DAG.py
aniket2405: Ack

pgmpy/base/DAG.py
aniket2405: Ack

pgmpy/base/DAG.py
aniket2405: Ack

pgmpy/base/DAG.py
aniket2405: `get_ancestral_graph ` internally uses `_get_ancestors_of `.
So, we don't have to use this, if I'm not mistaken?

pgmpy/base/DAG.py
aniket2405: Done, using this now.

pgmpy/base/DAG.py
aniket2405: Done

ankurankan (changes requested):
@aniket2405 I am looking at the Theorem 4.1 in the book again, and the implementation doesn't seem correct. The first step of the theorem is compute the ancestral graph, but the second step doesn't require moralization. It just about deleting all outgoing edges from observed variables. What am I missing here?

pgmpy/base/DAG.py
```diff
@@ -619,22 +619,22 @@ def get_immoralities(self):
             immoralities[node] = parent_pairs
         return immoralities
 
-    def is_dconnected(self, start, end, observed=None, include_latents=False):
+    def is_dconnected(self, start, end, observed=None):
```
ankurankan, line 622: We need the `include_latents` argument.

Issue #930, opened 2017-11-05
Add tests for to_junction_tree method of MarkovModel


ankurankan:
`to_junction_tree` needs more tests covering the issue #889. It has been fixed in #929.

Romio1310:
Hi @ankurankan! I’m new to open source and very interested in working on this issue. Could you please assign it to me, or let me know if I can give it a try? 
I would also appreciate any guidance or tips you might have for getting started. Thank you!

ankurankan:
@Romio1310 This has already been fixed. Sorry, we missed to close the issue.

PR #929, opened 2017-11-05
Fix error in to_junction_tree method of MarkovModel 

ankurankan:
In MarkovModel.py line:555, the clique_factors are sometimes empty but because whenan empty list is passed to factor_product it throws an error. This commit adds an extra check
to see it empty or not.

### Your checklist for this pull request
🚨Please review the [guidelines for contributing](CONTRIBUTING.md) to this repository.

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side). Don't request your master!
- [x] Make sure you are making a pull request against the **dev branch** (left side). Also you should start *your branch* off *our dev*.
- [x] Check the commit's or even all commits' message styles matches our requested structure.
- [x] Check your code additions will fail neither code linting checks nor unit test.

### Fixes #889 . (if there is no issue for this, please create one)

PR #2075, opened 2025-05-13
Added test with empty clique factor in `to_junction_tree` method in `DiscreteMarkovNetwork.py`

Manas-7854:
Addressing issue #930 
 - added test for PR #929, testing function `to_junction_tree` with an empty clique factor


ankurankan (approved):

ankurankan:
@Manas-7854 Thanks for the PR.

PR #2087, opened 2025-05-19
Add unit tests for type validation in VariableElimination.query()

Susmita331:
Just added two quick tests to catch common input mistakes:

Passing a string instead of a list to variables
Using something other than a dict for evidence
Helps ensure clear error messages and avoids silent bugs.


ankurankan:
@Susmita331 Tests are failing. Can you have a look?

Susmita331:
@ankurankan While improving input validation for forward_sample() in BayesianModelSampling, I added a TypeError check to ensure variables is an iterable. This worked as expected, and the test passed.

I attempted a similar check for evidence in query() (ExactInference.py), but the test test_query_common_var expects a ValueError instead of TypeError for invalid input. To avoid failing existing tests, I’ve commented out the evidence check for now.

Issue #1984, opened 2025-03-20
Add support for multiple versions of documentation


ankurankan:
The current documentation website is built using the latest dev branch. Improve this such that the documentation is available for multiple version of the package. 

https://github.com/sphinx-contrib/multiversion

mhmohona:
I would like to work on this issue @ankurankan.

MeetInCode:
i would like to work on this 
could you please elaborate on requirements

ankurankan:
@MeetInCode The current online documentation on pgmpy.org is generated using the latest version of the code (the dev branch). This can make it difficult for people working with older versions (or a release version) of the package, as the online documentation might not match the version they are using.  

We would like to have an option on the online website to select the package version to show the documentation for. This should be possible using the multiversion plugin for sphinx (https://github.com/sphinx-contrib/multiversion). Instructions for generating our current documentation can be found here: https://github.com/sphinx-contrib/multiversion

@arnavk23, you had opened a PR for this. Are you still working on it? If yes, @MeetInCode I would suggest choosing a different issue to avoid duplicating efforts.

arnavk23:
Yes, I am still working on it.

mhmohona:
I did some research on it. Without changing the current theme, it can also be done using custom CSS. 

I tried, and havent completed the task.

MeetInCode:
Alright then I will pick some other issue.

PR #2091, opened 2025-05-22
Feature/multiversion docs

arnavk23:
Added support for multiple versions of documentation as asked in #1984


arnavk23:
@ankurankan please review the changes made in this pr.

ankurankan:
@arnavk23 Thanks for the PR. This also has the same issue as the other PR. I would suggest closing this, deleting this branch, and opening a new PR (after following the steps that I mentioned in the other PR).

In general, you should always avoid doing a `git add *` as it ends up adding unnecessary files in the PR. I am also not sure how you are generating your ssh keys, but it should usually be in the `.ssh` directory.

PR #2093, opened 2025-05-22
feat: Added `force_required_edges` parameter to enforce expert knowledge edge directions

### LINKED ISSUE #1913: Enforce required edges from ExpertKnowledge class
Currently, PC and Hill Climb search cannot enforce the direction of an edge in some cases. In these cases, the edge may be present but may not be guaranteed to be in the direction specified. This needs to be looked into and the resulting changes (if any) should be mentioned in the documentation.

Gmin2:

- Fixes #1913 

Added a new `force_required_edges` parameter to structure learning algorithms that:
- Forces addition of required edges even if they're not found in the skeleton
- Forces correct orientation of edges even if they exist in the opposite direction  
- Can be reverted back with default `force_required_edges=False`

### DIFF
--- a/pgmpy/estimators/ExpertKnowledge.py
+++ b/pgmpy/estimators/ExpertKnowledge.py
@@ -212,7 +212,7 @@ class ExpertKnowledge:
 
         self.forbidden_edges = self.forbidden_edges.union(forbidden_edges)
 
-    def apply_expert_knowledge(self, pdag):
+    def apply_expert_knowledge(self, pdag, force_required_edges=False):
         """
         Method to check consistency and orient edges in a graph based on expert knowledge.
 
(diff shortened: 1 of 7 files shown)


Nimish-4:
Hi, thanks for the PR. I am not sure about the validity of the changes. Especially in PC, I think changing the skeleton to enforce expert knowledge would mess up the causal structure of the result. We may need to look into literature for any information/approach related to this. It could also be not possible to enforce this, in which case the current implementation would suffice. @ankurankan Thoughts?

ankurankan:
@Gmin2 @Nimish-4 Indeed, I don't think the changes are valid; we cannot directly force the edge directions. It can lead to invalid CPDAGs.  We'll need to either find relevant literature or design an algorithm that handles this properly.

@Gmin2 For future contributions, it would be helpful to discuss your approach on the GitHub issue thread or on Discord before starting implementation. This can help avoid unnecessary effort and situations like this.

Closing this for now.

PR #2099, opened 2025-05-23
Feature: Expanded EM algorithm to support missing values

### LINKED ISSUE #2018: Expand EM algorithm to support missing values
The current implementation of the EM algorihtm only works for missing variables (i.e., no information for the latent variables in the model). We should expand this implementation such that the algorithm also works when some of the values are missing for some of the variables. Code to reproduce:
```python
import numpy as np

from pgmpy.estimators import ExpectationMaximization as EM

model = get_example_model("alarm")
df = model.simulate(1000)
mask = np.random.choice([True, False], size=df.shape, p=[0.2, 0.8])
df = df.mask(mask)

est = EM(model, df)
est.get_parameters()
```
The above code would fail because of the randomly missing values.

Related: #2012

Green-Kedia:

- Fixes #2018

- Added support for missing values in the EM algorithm without dropping rows, bu treating them as mini latents
- Updated row frequency counting to handle missing values robustly
- updated test of missing values, and added a new test


.gitignore
```diff
@@ -42,3 +42,6 @@ docs/examples/*
 
 # Macos files
 .DS_Store
+
```
ankurankan: @Green-Kedia Can you remove these changes?

test_em_missing_data.py
```diff
@@ -0,0 +1,125 @@
+import matplotlib.pyplot as plt
```
ankurankan: This file should not be committed.

ankurankan (changes requested):
@Green-Kedia Can you also give me a high level explanation of how the missing values are handled now?

Green-Kedia:
> @Green-Kedia Can you also give me a high level explanation of how the missing values are handled now?

Hi @ankurankan , the missing values are now handled by treating them as mini-latent variables specific to each row.
Edit: apologies for such complex implementation, I have kept the changes to the minimum now

Green-Kedia:
@ankurankan just a reminder for this pr

pgmpy/estimators/EM.py
```diff
@@ -76,14 +76,10 @@ def __init__(self, model, data, **kwargs):
...
 
-        # Drop rows with any missing values in partially observed columns
-        original_rows_count = data.shape[0]
-        data = data.dropna()
-        dropped_rows_count = original_rows_count - data.shape[0]
-
-        if dropped_rows_count:
-            logger.warning(
-                f"{dropped_rows_count} rows with missing values in partially missing columns were dropped from the dataset."
+        # Alarm rows have missing values in partially observed columns and will be treated as mini latents
```
ankurankan, line 79: I don't think we need this logging info as this is kind of the expected usecase.

pgmpy/estimators/EM.py
```diff
@@ -113,24 +109,36 @@ def _get_log_likelihood(self, datapoint):
...
     ):
         cache = []
 
         for i in range(offset, min(offset + batch_size, data_unique.shape[0])):
-            v = list(product(*[range(card) for card in latent_card.values()]))
-            latent_combinations = np.array(v, dtype=int)
-            df = data_unique.iloc[[i] * latent_combinations.shape[0]].reset_index(
-                drop=True
+            missing_cols = data_unique.columns[data_unique.iloc[i].isna()]
+            latent_card_copy = latent_card.copy()
```
ankurankan, line 118: This can already be initialized in the `get_parameters` method with the cardinality of all the variables. This method is computationally very expensive, so we need to be very careful about what is in this method.

pgmpy/estimators/EM.py
```diff
@@ -113,24 +109,36 @@ def _get_log_likelihood(self, datapoint):
...
+                    *[
+                        val if isinstance(val, list) else np.arange(val, dtype=int)
+                        for val in latent_card_copy.values()
+                    ]
+                )
             )
-            for index, latent_var in enumerate(latent_card.keys()):
-                df[latent_var] = latent_combinations[:, index]
+            latent_combinations = list(zip(*v))
+            n_count_ = n_counts[
```
ankurankan, line 131: Can't this be computed outside as well, the same way it was being done before?

pgmpy/estimators/EM.py
```diff
@@ -142,13 +150,20 @@ def _compute_weights(self, n_jobs, latent_card, batch_size):
...
-        n_counts = (
-            self.data.groupby(list(self.data.columns), observed=True).size().to_dict()
-        )
+
+        n_counts = {}
+        for _, row in self.data.iterrows():
+            row_key = tuple(["__NA__" if pd.isna(x) else str(x) for x in row])
+            n_counts[row_key] = n_counts.get(row_key, 0) + 1
+
+        unique_cats = {
```
ankurankan, line 159: I think this is more of a preprocessing step and should be somewhere in Step 2 of the get_parameters method.

ankurankan (changes requested):

ankurankan (changes requested):
Could you please also do some timing benchmarks to see how much slower does this get?

PR #2105, opened 2025-05-25
"Add ExpertKnowledge support to ExpertInLoop estimator"

### LINKED ISSUE #2100: [ENH] Allow users to specify expert knowledge in ExpertInLoop algorithm
We have an expert knowledge class: `pgmpy.estimators.ExpertKnowledge` that allows users to specify expert knowledge for causal discovery algorithms. Using expert knowledge users can specify `forbidden_edges`: edges that should not be present in the final model,  `required_edges`: edges that must be present in the final model, `temporal_order`: The temporal / topological ordering of the variables in the final model. All of the causal discovery algorithms implemented in pgmpy supports expert knowledge except ExpertInLoop. We would like to add this functionality to `ExpertInLoop` as well. The behaviour should be:

1. If edge in `forbidden_edges`, never try to add this edge, and initialize `blacklisted_edgse` with this set of edges.
2. If edge in `required_edges`, initialize the DAG with this set of edges. These can however get removed in the prune operation of the algorithm. The documentation should specify this.
3. Before asking the user or LLM for edge orientation, check if the orientation of the edge can be resolved using the specified temporal order. If yes, use the temporal order instead.

KumarADITHYA123:
### Summary

This PR adds `ExpertKnowledge` support to the `ExpertInLoop` estimator, enabling it to respect:

- `forbidden_edges`: edges never added.
- `required_edges`: initialized and protected (unless strongly contradicted).
- `temporal_order`: used to resolve edge directions before consulting LLMs/experts.

### Changes Made
- Added `expert_knowledge` parameter to `__init__` and `estimate()`
- Integrated handling logic in the edge addition and direction phase
- Added `_resolve_edge_direction()` method
- Updated docstrings and added an example using the "cancer" dataset
- Added unit tests for each expert knowledge component

This enhancement brings the `ExpertInLoop` class in line with other pgmpy causal estimators that already support expert knowledge.

Fixes: https://github.com/pgmpy/pgmpy/issues/2100

### DIFF
--- a/pgmpy/estimators/expert.py
+++ b/pgmpy/estimators/expert.py
@@ -1,7 +1,5 @@
-import sys
 from itertools import combinations
 
-import networkx as nx
 import pandas as pd
 
 from pgmpy import config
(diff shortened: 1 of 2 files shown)


ankurankan:
@KumarADITHYA123 Was this generated using a GenAI tool? There are way more changes than what’s necessary and introduce unnecessary complexity to the code. The algorithm doesn't work anymore; it's stuck in an infinite loop.

KumarADITHYA123:
I appreciate your candid and constructive feedback 🙏

I understand now that I took a less streamlined approach and made excessive unnecessary changes. It is becoming evident to me that the algorithm does not terminate correctly, and I will get on that right away.

I will now:

- As much as possible, cut away any elements of the code that lie outside the defined boundaries of the problem at hand.

- Address the logic to prevent infinite loops.

- Maintain consistency with the coding conventions already established in the codebase.

- Implement new set conditions to cover the additional logic without excessive, unfocused tests.

Your time and thorough review are incredibly helpful, and I am learning so much from this. I will adjust the PR in no time!

PR #2116, opened 2025-05-25
Adds an example notebook for expert knowledge

Nimish-4:


ankurankan:
@Nimish-4 I have added some changes to make the language a bit more precise. Please have a look. I would also suggest removing the part with the combined PC and Hill-Climb Search. The combined method performs worse (has a higher SHD). Also, the combined approach is what the MMHC algorithm is, so maybe we can instead add its example, but not necessary. 

Also, about the last example on the forbidden edge, a more practical example would be to have the forbidden list as all incoming edges to `Disease` (except the one from `BirthAsphyxia`). Usually, in practice, we know certain exogenous variables (such as Disease in this example or Age in the adult income dataset), and these exogenous variables can only cause other variables and therefore can't have incoming edges.

ankurankan:
Another practical example can be from this paper: https://openreview.net/pdf?id=1QWcueC0lfZ. Look at section 4.4, I am not sure if the data is available publically though.

Nimish-4:
@ankurankan The dataset they used is not publicly available. Any other ideas of a practical example or use case? I was struggling with coming up with something convincing.

Spinachboul:
@Nimish-4  @ankurankan 
Maybe the Asia dataset works, though it is used in the R language!!
Link: https://www.bnlearn.com/documentation/man/asia.html

Nimish-4:
@Spinachboul The datasets in the bnlearn repo are present in pgmpy, so that's not a problem. The issue is creating a convincing practical example about the use of expert knowledge ... so would have to look for something else.

Spinachboul:
> @Spinachboul The datasets in the bnlearn repo are present in pgmpy, so that's not a problem. The issue is creating a convincing practical example about the use of expert knowledge ... so would have to look for something else.

Oh okay! Then other possible sources would be `paperswithcode` which I am aware of at the moment
And I found something similar , not sure if this would fulfill the requirements of the dataset
Paper Link : https://paperswithcode.com/paper/datgan-integrating-expert-knowledge-into-deep
Github Dataset Link : https://github.com/glederrey/DATGAN/tree/master/example/data

@ankurankan  FYI

Nimish-4:
Thanks, will check it out!

ankurankan:
@Nimish-4 @Spinachboul In my opinion, we can split the idea of tutorial notebooks into two: 1. Give examples of the features offered by pgmpy, 2. Show more practical examples by for example recreating some studies. I think the examples in this notebook already achieve No. 1. I would however, suggest restructuring this notebook such that it is more aligned with the different ways in which we can specify expert knowledge in pgmpy. So ideally, 3 sections each focusing on the 3 ways (or combination of them) in which we can specify expert knowledge. As we have already mentioned in advance that we can specify expert knowledge for all algorithms in pgmpy, I don't think we need to focus on different algorithms in this notebook.

Regarding No. 2, we will have to find some real-world examples that we can recreate. @Nimish-4 had found this paper: https://arxiv.org/pdf/2108.13395 that we can use to already recreate some of the examples. I also have this list of papers that use causal discovery algorithms: https://docs.google.com/spreadsheets/d/1aaV8wGRN18RnMTGwtWSnoohrkHO1H7olQidhjYjLqvI/edit?gid=0#gid=0, we can try to recreate some of the interesting examples from them.

ankurankan:
@Nimish-4 I think we need to figure out what's going on with the `enforce_expert_knowlege` argument as it seems to be failing for multiple models.

Nimish-4:
@ankurankan Do you mean the key error, or some new issue? I have created an issue for the key errors, it should be a minor thing #2138

ankurankan:
@Nimish-4 The PC algorithm on asia model was getting stuck when `enforce_expert_knowledge=True`.

Nimish-4:
@ankurankan Can you share the code, or the arguments used?

ankurankan:
@Nimish-4 Hmm, on second look, I think it wasn't getting stuck, rather the progress bar stops if the algorithm exits early. The performance when we specify `search_space` is quite bad though (see the asia example in the notebook). I think we will need to investigate what is happening there.

Nimish-4:
@ankurankan That seems to be happening because there was both a `search_space` and `enforce_expert_knowedge` in the code. Search space is currently handled by indirectly converting it into `forbidden_edges`. Combined with `enforce_expert_knowedge=True`, this means that edges are simply removed from the skeleton. You can choose not to enforce expert knowledge and see a better result. 
Of course, the entire handling of search space could be improved, but I don't think there are any 'buggy' parts right now. Though you should probably take a look too.

ankurankan:
I tried updating the asia model example to:
```python
from pgmpy.metrics import SHD

search_space = list(asia.edges())

search_space = search_space + [
    ("smoke", "cancer"),
    ("smoke", "lung"),
    ("smoke", "tub"),
    ("tub", "asia"),
    ("asia", "tub"),
    ("tub", "either"),
    ("lung", "either"),
    ("bronc", "either"),
    ("tub", "xray"),
    ("tub", "dysp"),
    ("lung", "x-ray"),
    ("lung", "dysp"),
    ("bronc", "xray"),
    ("bronc", "dysp"),
    ("either", "xray"),
    ("either", "dysp"),
    ("tub", "dysp"),
]

expert_knowledge = ExpertKnowledge(search_space=search_space)

est_model = PC(asia_samples).estimate(expert_knowledge=expert_knowledge, enforce_expert_knowledge=False)
SHD(asia, est_model)
```
This search space already has all the true edges in it. So, even with enforcing expert knowledge, it should give decent results. Right? The performance does improve when I set enforce to False.

Nimish-4:
@ankurankan Again the issue is handling of search space in PC. It needs to be changed. If `u,v` is a true edge present in the search space, then `v,u` would be added to the forbidden edges.  Then if we enforce expert knowledge the entire v---u edge is removed from the skeleton. Hence its poor performance. I guess this method needs to be revamped a little.

Nimish-4:
The search space isn't really acting like a search space for PC. I guess this was overlooked when the changes were made; fixing it shouldn't be a lot of work.

ankurankan:
@Nimish-4 Could you please create an issue if you have already figured out the problem?

ankurankan:
@Nimish-4 Do we finally have a fix for the error that we were getting in this notebook?

Nimish-4:
@ankurankan Nope, but I guess now I'll just change the tutorial when it is working.

Issue #2115, opened 2025-05-25
[ENH] Method for validating output of causal discovery methods


ankurankan:
This is the second part of the issue described in #2054. Splitting the issue to better track the two separate parts. This issue is for implementing the second method as described in:

[1] Eulig, Elias, et al. "Toward falsifying causal graphs using a permutation-based test." Proceedings of the AAAI Conference on Artificial Intelligence. Vol. 39. No. 25. 2025.

Ayash-Bera:
hello there i would i like to be assigned to this. Thank you so much.

JATAYU000:
Hi @ankurankan I would be picking this up, will be going through the paper and abandoned prs and follow up on my plan by this weekend. thank you.

ankurankan:
@JATAYU000 Would suggest picking up from the existing PR: https://github.com/pgmpy/pgmpy/pull/2119/. I think it was mostly there, except for some bugs, which are also mentioned in the PR.

PR #2119, opened 2025-05-25
DAG falsification using permutation-based test

Ayash-Bera:

- Fixes #
#2115 

- added test_permutation_test.py and test_permutation.py
- modified __init__.py  under the metrics folder 
- did not touch any other files.


ankurankan:
@Ayash-Bera The problem was with the `_test.py` in the filenames which pytest was picking up as a test file. I have renamed the file, and seems to work now. You will have to update some of the variable names.

Ayash-Bera:
> @Ayash-Bera The problem was with the `_test.py` in the filenames which pytest was picking up as a test file. I have renamed the file, and seems to work now. You will have to update some of the variable names.

looking into it

model.bif
```diff
@@ -0,0 +1,35 @@
+network unknown {
```
ankurankan: Need to get rid of this file.

pgmpy/metrics/__init__.py
```diff
@@ -15,4 +16,7 @@
     "implied_cis",
     "fisher_c",
     "SHD",
+    "BayesianModelProbability",
+    "permutation_t",
```
ankurankan: I don't see the point of having two names for the same function?

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+    It tests whether a given DAG is significantly better than random node permutations
+    by comparing Local Markov Condition (LMC) violations.
+
+    The test performs two evaluations:
+    1. Falsifiability: Whether the graph is informative enough to be falsifiable
+    2. Falsification: Whether the graph performs significantly better than random
+
+    Parameters
+    ----------
+    model : pgmpy.base.DAG or pgmpy.models.DiscreteDiscreteDiscreteBayesianNetwork
```
ankurankan: This can be either DAG or any *BayesisanNetwork object

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+        The causal graph to test for falsification.
+
+    data : pandas.DataFrame
+        Observational data to test the graph against. Should contain all variables
+        present in the model.
+
+    ci_test : str, default='chi_square'
```
ankurankan: We have support for a lot more test. Have a look at CITests.py and PC.py.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,320 @@
...
+    data : pandas.DataFrame
+        Observational data to test the graph against. Should contain all variables
+        present in the model.
+
+    ci_test : str, default='chi_square'
+        The conditional independence test to use for testing Local Markov Conditions.
+        Options: 'chi_square' (for discrete data), 'pearsonr' (for continuous data).
+
+    significance_level : float, default=0.05
+        Significance level for conditional independence tests. Lower values make
```
ankurankan, line 127: This is not clear. Conservative in which direction, toward rejecting the null or accepting it?

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,320 @@
...
+        the test more conservative.
+
+    n_permutations : int, optional
```
ankurankan, line 130: I think this will need to be a function of the size of the model. For a model on 100 variables, 20 would be too low, whereas for a model on 5 variables, 20 is quite high.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+
+    significance_level : float, default=0.05
+        Significance level for conditional independence tests. Lower values make
+        the test more conservative.
+
+    n_permutations : int, optional
+        Number of random node permutations to generate for the baseline.
+        If None, uses max(20, int(1/significance_level)).
+
+    return_summary : bool, default=False
```
ankurankan: The default should be True.

pgmpy/metrics/permutation_test.py
ankurankan, line 130: This argument should also be higher in the list as it is an important hyperparameter for this method.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+    if not model_nodes.issubset(data_columns):
+        missing_vars = model_nodes - data_columns
+        raise ValueError(f"Data missing variables present in model: {missing_vars}")
+
+    # Set default number of permutations
+    if n_permutations is None:
+        n_permutations = max(20, int(1 / significance_level))
+
+    # Initialize CI test function
+    if ci_test == "chi_square":
```
ankurankan: We have a `get_ci_test` function for getting the function. Have a look at how this is implemented in `PC.py`.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+
+    >>> # Test with wrong model (should be falsified)
+    >>> wrong_model = DiscreteDiscreteBayesianNetwork([('Cancer', 'Smoker'), ('Smoker', 'Pollution')])
+    >>> result_wrong = permutation_based_falsification_test(wrong_model, data)
+    >>> print(f"Wrong model falsified: {result_wrong['falsified']}")
+    """
+
+    # Validate inputs
+    if not isinstance(data, pd.DataFrame):
+        raise TypeError("Data must be a pandas DataFrame")
```
ankurankan: We usually also write the type we got in the error message. Something like "...... be a pandas DataFrame. Got: {type(data)}`.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+from tqdm import tqdm
+
+from pgmpy import config
+from pgmpy.base import DAG
+from pgmpy.estimators.CITests import chi_square, pearsonr
+from pgmpy.global_vars import logger
+
+
+def permutation_t(
+    model,
```
ankurankan: Would suggest renaming this to `dag`

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,320 @@
...
+    >>> # Test with wrong model (should be falsified)
+    >>> wrong_model = DiscreteDiscreteBayesianNetwork([('Cancer', 'Smoker'), ('Smoker', 'Pollution')])
+    >>> result_wrong = permutation_based_falsification_test(wrong_model, data)
+    >>> print(f"Wrong model falsified: {result_wrong['falsified']}")
+    """
+
+    # Validate inputs
+    if not isinstance(data, pd.DataFrame):
+        raise TypeError("Data must be a pandas DataFrame")
+
```
ankurankan: We will need to add a check that the DAG doesn't have any latent variables, otherwise the local Markov condition can't be tested in the data.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+        ci_test_func = pearsonr
+    else:
+        raise ValueError(
+            f"Unsupported CI test: {ci_test}. Use 'chi_square' or 'pearsonr'"
+        )
+
+    logger.info(
+        f"Starting permutation-based falsification test with {n_permutations} permutations"
+    )
+    # Step 1: Count LMC violations in the given graph
```
ankurankan: Use full name for LMC. It's not a common abbreviation.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+    )
+
+    # Step 2: Generate permutations and test them
+    nodes = list(model.nodes())
+    permutation_violations = []
+    same_mec_count = 0
+
+    # Set up progress bar
+    if show_progress and config.SHOW_PROGRESS:
+        pbar = tqdm(total=n_permutations, desc="Testing permutations")
```
ankurankan: The description should be something like: "Constructing Null Distribution"

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+
+    for i in range(n_permutations):
```
ankurankan: We can if-else for pbar to simplify this:
```
if show_progress ....:
     pbar = tqdm( .. .. . . )
else:
     pbar = range(n_permutation)
```

then `for i in range(pbar):`

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+
+        # Count LMC violations in permuted graph
+        lmc_violations_perm = _count_lmc_violations(
+            permuted_model, data, ci_test_func, significance_level
+        )
+
+        permutation_violations.append(lmc_violations_perm)
+
+        # Check if in same Markov equivalence class (0 violations = same structure)
+        if lmc_violations_perm == 0:
```
ankurankan: I don't think this is correct. They can be in the same MEC even with non-zero violations.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,320 @@
...
+                # Handle edge cases (e.g., insufficient data, constant columns)
+                logger.debug(
+                    f"CI test failed for {node} ⊥ {test_node} | {parents}: {e}"
+                )
+                continue
+
+    return violations
+
+
+def _get_non_descendants(model, node):
```
ankurankan: This method needs to go in the DAG class.

pgmpy/tests/test_metrics/test_permutation_t.py
```diff
@@ -0,0 +1,772 @@
+#!/usr/bin/env python
+
+import sys
```
ankurankan: Your tests seem mostly AI-generated. If that is the case, we need to simplify this a lot. This is way too much code for a relatively simple function.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,320 @@
...
+    show_progress=True,
+):
+    """
+    Permutation-based test for falsifying causal graphs using observational data.
+
+    This method implements the permutation-based falsification test from Eulig et al. (2025).
+    It tests whether a given DAG is significantly better than random node permutations
+    by comparing Local Markov Condition (LMC) violations.
+
+    The test performs two evaluations:
```
ankurankan, line 112: The documentation on how to interpret these values needs to be simpler.

ankurankan (changes requested):
I had a quick look at the PR and have added a few comments. Additional things:
1.  I would suggest using some examples from the paper and trying to see if you can recreate the results. 
2. Once we are sure that this is working correctly, we can relax the condition of latent variables and try to use minimal d-separator instead of local Markov condition.
3. There is room for significant performance improvement by permuting the conditional independence statements that we get from the local Markov condition instead of constructing permuted graphs.

pgmpy/tests/test_metrics/test_permutation_t.py
Ayash-Bera: i did have to generate a lot of code just to get the coverage correctly i couldn't get it done by hand ill look into it

ankurankan: If you write tests that check most of the main functionality, it should automatically have very high coverage. The remaining are usually edge cases, that you can write simple tests for.

Nimish-4:
@ankurankan  You can use something like this to see if it works as expected - 

```python3
import numpy as np
import pandas as pd

from pgmpy.metrics import permutation_t


def data_continuous():
    np.random.seed(0)
    n = 1000
    X = np.random.normal(0, 10, n)
    Y = 0.5 * X + np.random.normal(0, 0.05, n)
    Z = 0.7 * Y + np.random.normal(0, 0.03, n)

    A  = 0.4 * X + np.random.normal(0.02, 0.03, n)
    return pd.DataFrame({"X": X, "Y": Y, "Z": Z, 'A': A})

def model_simple():
    return DiscreteBayesianNetwork([("X", "Y"), ("Y", "Z"), ("X", "A")])

df = data_continuous()
model = model_simple()

y = permutation_t( model, df, n_permutations=20, return_summary=True, show_progress=False, ci_test='gcm')

for item in y.items():
    print(item)
```

pgmpy/base/DAG.py
```diff
@@ -206,6 +206,32 @@ def _check_cycles(self):
...
+            if current not in visited:
+                visited.add(current)
+                descendants.add(current)
+                queue.extend(self.successors(current))
+
+        # Non-descendants are all nodes except descendants and the node itself
+        all_nodes = set(self.nodes())
+        non_descendants = all_nodes - descendants - {node}
+
+        return list(non_descendants)
```
ankurankan, line 233: Is this not the same as the ancestors of the node? `DAG.get_ancestors`

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+
+    new_CIs = valid_CIs.copy()
+    new_CIs["u"] = valid_CIs["u"].apply(lambda x: apply_mapping(x))
+    new_CIs["v"] = valid_CIs["v"].apply(lambda x: apply_mapping(x))
+    new_CIs["cond_vars"] = valid_CIs["cond_vars"].apply(lambda x: apply_mapping(x))
+
+    return new_CIs
+
+
+def d_separated_triples(
```
ankurankan: I think we should rename this to `_d_separated_triples` as this shouldn't directly be used by the user.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+
+    if isinstance(
+        dag,
+        (
+            DAG,
+            DiscreteBayesianNetwork,
+            DynamicBayesianNetwork,
+            FunctionalBayesianNetwork,
+            LinearGaussianBayesianNetwork,
+        ),
```
ankurankan: I think just testing for `DAG` should be sufficient as all the other classes inherit `DAG`? I also don't think this would work for `DynamicBayesianNetwork` as we can not compute the implied CIs for it.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+    ):
+        if len(dag.latents) > 0:
+            raise ValueError(
+                f"Found latent variables: {dag.latents}. "
+                "permutation_t does not support latent variables."
+            )
```
ankurankan: I think this should work for latent variable models as well. The `implied_cis` method should automatically handle latent variables.

pgmpy/metrics/permutation_test.py
ankurankan: Ignore this.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+    model_nodes = set(dag.nodes())
+    data_columns = set(data.columns)
+
+    if not model_nodes.issubset(data_columns):
+        missing_vars = model_nodes - data_columns
+        raise ValueError(f"Data missing variables present in model: {missing_vars}")
+
+    # Set default number of permutations
+    if n_permutations is None:
+        n_permutations = max(20, int(1 / significance_level))
```
ankurankan: It might be better to set the default value as 100 in the argument itself.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+
+    # Initialize CI test function
+    ci_test_func = ci_registry.get_test(ci_test, data=data)
+
+    logger.info(
+        f"Starting permutation-based falsification test with {n_permutations} permutations"
+    )
```
ankurankan: Would remove this. Doesn't give much info.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,325 @@
...
+    logger.info(
+        f"Starting permutation-based falsification test with {n_permutations} permutations"
+    )
+
+    # Step 1: Perform calculations for original DAG and generate permutations
+    nodes = list(dag.nodes())
+    permutation_violations = []
+    same_mec_count = 0
+
+    orginal_CIs = implied_cis(dag, data, ci_test_func)
```
ankurankan: Can we not just directly pass the `ci_test` here instead of the `get_test` call in line 238?

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,325 @@
...
+    # Falsifiability test: fraction of permutations in same MEC
+    p_value_falsifiable = same_mec_count / n_permutations
+    falsifiable = p_value_falsifiable <= significance_level
+
+    # Falsification test: fraction of permutations with lesser violations
+    count_less_violations = sum(
+        1 for v in permutation_violations if v <= lmc_violations_given
+    )
+
+    p_value_falsified = count_less_violations / n_permutations
```
ankurankan, line 248: I think we should also report the confidence interval of the p-value. Can be computed using binomial proportion confidence intervals. Some chatGPT code:
```
import statsmodels.stats.proportion as proportion

# Inputs
p_value = 0.045   # Your estimated p-value
n_sims = 1000     # Number of simulations run
count = int(p_value * n_sims) # Convert back to number of "hits"

# Calculate Wilson Interval (95%)
ci_lower, ci_upper = proportion.proportion_confint(
    count, 
    n_sims, 
    alpha=0.05, 
    method='wilson'
)

print(f"95% CI: [{ci_lower:.4f}, {ci_upper:.4f}]")
```

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+    References
+    ----------
+    Eulig, E., Mastakouri, A. A., Blöbaum, P., Hardt, M., & Janzing, D. (2025).
+    Toward falsifying causal graphs using a permutation-based test.
+    Proceedings of the AAAI Conference on Artificial Intelligence, 39(25), 26778-26786.
+
+    Examples
+    --------
+    >>> from pgmpy.models import DiscreteDiscreteBayesianNetwork
+    >>> from pgmpy.metrics import permutation_based_falsification_test
```
ankurankan: I think the import statement is incorrect here?

pgmpy/tests/test_metrics/test_permutation_t.py
```diff
@@ -0,0 +1,192 @@
...
+from pgmpy.estimators.CITests import ci_registry
+from pgmpy.metrics import implied_cis, permutation_t
+from pgmpy.metrics.permutation_t import (
+    _count_lmc_violations,
+)
+from pgmpy.models import DiscreteBayesianNetwork
+
+logger = global_vars.logger
+falsify_graph = permutation_t
+permutation_based_falsification_test = permutation_t
```
ankurankan: Don't think these are used?

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+            "permutation_violations": permutation_violations,
+            "significance_level": significance_level,
+            "ci_test": ci_test,
+            "mean_permutation_violations": np.mean(permutation_violations),
+            "std_permutation_violations": np.std(permutation_violations),
+            "min_permutation_violations": np.min(permutation_violations),
+            "max_permutation_violations": np.max(permutation_violations),
+        }
+
+    logger.info(f"Test completed. Falsifiable: {falsifiable}, Falsified: {falsified}")
```
ankurankan: Would suggest removing this as well.

pgmpy/metrics/permutation_t.py
```diff
@@ -0,0 +1,325 @@
...
+    else:
+        pbar = range(n_permutations)
+
+    for _ in pbar:
+        # Generate random permutation
+        perm = np.random.permutation(nodes)
+        perm_mapping = dict(zip(nodes, perm))
+
+        # Create permuted CIs
+        permuted_CIs = _create_permuted_CIs(valid_CIs, perm_mapping)
```
ankurankan: Would suggest moving this inside the `_create_permuted_CIs` method itself.

ankurankan (changes requested):

pgmpy/base/DAG.py
Nimish-4: I think this one can include nodes that are not ancestors of the node (like existing in another branch of the graph), right? Will have to see if this is required, though.

pgmpy/metrics/permutation_t.py
Nimish-4: Should we think of some other value or condition? I think 100 maybe too much for smaller graphs as we would get repeat permutations.

pgmpy/metrics/permutation_test.py
Nimish-4: I think `implied_cis` expects the test function directly rather than the string.

pgmpy/base/DAG.py
ankurankan: Yes, you are right; `DAG.ancestors` would miss descendants of ancestral variables, which `_get_non_descendants` does find.

pgmpy/metrics/permutation_t.py
ankurankan: Yeah, that's a good point. Maybe let's keep it as is.

pgmpy/metrics/permutation_test.py
ankurankan: Could we modify `implied_cis` to do the selection? I think it would be better to modify `implied_cis` as it is called from other places as well, which should reduce some redundant code.

ankurankan:
@Nimish-4, I have added a few commits simplifying a few things. Could you please have a look? I think there are a couple of bugs in the algorithm implementation; I have added TODOs for them. If you have time, could you please look at it? Otherwise, I will try to fix them later today. 

@Ayash-Bera I don't know if you are still interested in this PR, but let us know if you might want to pick this back up :)

Nimish-4:
@ankurankan From **Definition 2** given in Eulig, Elias, et al., it does seem that we need to consider all possible implied CIs and let the test calculate how many hold. I am making that change, let me know if you think we should keep the current one.

>TODO: This is wrong - need to compare all implied CIs, not just valid ones. Maybe use is_iequivalent method from pgmpy.base.DAG?

We're only working with permuted CIs so I guess this won't really work, right?

PR #2127, opened 2025-05-26
Expand simulation functionality of LGBN

### LINKED ISSUE #2110: [ENH] Expand simulation capability of Linear Gaussian Bayesian Networks
The `LinearGaussianBayesianNetwork.simulate` method currently provides some basic functionality to generate data from the model. Compared to `DiscreteBayesianNetwork`, this function is quite limiting as it doesn't allow users to specify any additional arguments such as evidence, intervention, or missing data. We would like to expand this simulation method to have all the functionality that the DiscreteBN.simulate offers.

- [x] Specifying evidence: Use the predict method to compute the posterior distribution and sample from the posterior distribution.
- [x] Specifying interventions: Modify the network structure and the LinearGaussianCPDs according to the specified intervention and simulate data from the modified model.
- [ ] Specifying virtual interventions: Similar to interventions, but assign the intervened node to the specified CPD.
- [ ] Add `include_latents` argument. If True, returns the full simulated dataframe otherwise drops the latent variables.
- [ ] Missing data: Requires modification to the model by adding new nodes that control the missing values.  See how it is implemented in DiscreteBN.

Sanchay117:

- Fixes #2110 

- Added support for specifying evidence by using the predict method to compute posterior distributions and sample accordingly in `LinearGaussianBayesianNetwork.simulate()`.

- Added handling of interventions by modifying the network structure and updating LinearGaussianCPDs based on the intervention before simulating data in `LinearGaussianBayesianNetwork.simulate()`.

- Added handling of missing data in  `LinearGaussianBayesianNetwork.simulate()` through the addition of new missingness indicator nodes in the model, inspired by the implementation in DiscreteBN.

### DIFF
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -247,6 +247,39 @@ class LinearGaussianBayesianNetwork(DAG):
         # Round because numerical errors can lead to non-symmetric cov matrix.
         return mean.round(decimals=8), implied_cov.round(decimals=8)
 
+    def copy(self):
+        """
+        Returns a copy of the model.
+
+        Returns
+        -------
+        Model's copy: pgmpy.models.LinearGaussianBayesianNetwork
+            Copy of the model on which the method was called.
+
+        Examples
+        --------
+        >>> from pgmpy.models import LinearGaussianBayesianNetwork
+        >>> from pgmpy.factors.continuous import LinearGaussianCPD
+        >>> model = LinearGaussianBayesianNetwork([('A', 'B'), ('B', 'C')])
+        >>> cpd_a = LinearGaussianCPD(variable='A', beta=[1], std=4)
+        >>> cpd_b = LinearGaussianCPD(variable='B', beta=[-5, 0.5], std=4, evidence=['A'])
+        >>> cpd_c = LinearGaussianCPD(variable='C', beta=[4, -1], std=3, evidence=["x2"])
+        >>> model.add_cpds(cpd_a, cpd_b, cpd_c)
+        >>> copy_model = model.copy()
+        >>> copy_model.nodes()
+        NodeView(('A', 'B', 'C'))
+        >>> copy_model.edges()
+        OutEdgeView([('A', 'B'), ('B', 'C')])
+        >>> len(copy_model.get_cpds())
+        3
+        """
+        model_copy = LinearGaussianBayesianNetwork()
+        model_copy.add_nodes_from(self.nodes())
+        model_copy.add_edges_from(self.edges())
+        if self.cpds:
+            model_copy.add_cpds(*[cpd.copy() for cpd in self.cpds])
+        return model_copy
+
     def simulate(
         self, n_samples=1000, do=None, evidence=None, seed=None, missing_prob=None
     ):
@@ -290,19 +323,11 @@ class LinearGaussianBayesianNetwork(DAG):
         >>> model.add_cpds(cpd1, cpd2, cpd3)
         >>> model.simulate(n_samples=500, seed=42)
         """
-        # Step 1: Check if all arguments are specified
+        # Step 1: Check if all arguments are specified and valid
         evidence = {} if evidence is None else evidence
 
         do = {} if do is None else do
 
-        # Step 2: Check if there are any common variables in do and evidence
-        common_vars = set(do.keys()).intersection(set(evidence.keys()))
-        if common_vars:
-            raise ValueError(
-                f"Variable(s) can't be in both do and evidence: {', '.join(common_vars)}"
-            )
-
-        # Step 3: Ensure all variables in the intervention exist in the model's nodes
         nodes = list(do.keys())
 
         if not set(nodes).issubset(set(self.nodes())):
@@ -315,11 +340,15 @@ class LinearGaussianBayesianNetwork(DAG):
                 "Each node in the model should have a CPD associated with it"
             )
 
-        # Step 4: If do is specified, modify the network structure.
+        common_vars = set(do.keys()).intersection(set(evidence.keys()))
+        if common_vars:
+            raise ValueError(
+                f"Variable(s) can't be in both do and evidence: {', '.join(common_vars)}"
+            )
+
+        # Step 2: If do is specified, modify the network structure.
         if do != {}:
-            model = LinearGaussianBayesianNetwork(self.edges())
-            model.add_nodes_from(self.nodes())
-            model.add_cpds(*self.cpds)
+            model = self.copy()
             for var, val in do.items():
                 for parent in list(model.get_parents(var)):
                     model.remove_edge(parent, var)
@@ -360,6 +389,7 @@ class LinearGaussianBayesianNetwork(DAG):
         evidence_var = list(evidence.keys())
         sample_var = [v for v in variables if v not in evidence_var]
 
+        # Step 4: Sample according to evidence
         if len(evidence) == 0:
             df = pd.DataFrame(
                 rng.multivariate_normal(mean=mean, cov=cov, size=n_samples),
@@ -389,6 +419,10 @@ class LinearGaussianBayesianNetwork(DAG):
                 df[mv] = df_missing[mv].values
 
             df = df[variables]
+
+        # Step 5: Add do variables to the final dataframe
+        for do_var, do_val in do.items():
+            df[do_var] = do_val
 
         return df
 
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -340,12 +340,21 @@ class LinearGaussianBayesianNetwork(DAG):
         do = {} if do is None else do
 
         do_nodes = list(do.keys())
+        evidence_nodes = list(evidence.keys())
+        rng = np.random.default_rng(seed=seed)
 
         invalid_nodes = set(do_nodes) - set(self.nodes())
         if not set(do_nodes).issubset(set(self.nodes())):
             raise ValueError(
                 f"The following do-nodes are not present in the model: {invalid_nodes}. "
                 f"do argument contains: {do_nodes}"
+            )
+
+        invalid_nodes = set(evidence_nodes) - set(self.nodes())
+        if not set(evidence_nodes).issubset(set(self.nodes())):
+            raise ValueError(
+                f"The following evidence-nodes are not present in the model: {invalid_nodes}. "
+                f"evidence argument contains: {evidence_nodes}"
             )
 
         if len(self.cpds) != len(self.nodes()):
@@ -400,7 +409,6 @@ class LinearGaussianBayesianNetwork(DAG):
 
         mean, cov = model.to_joint_gaussian()
         variables = list(nx.topological_sort(model))
-        rng = np.random.default_rng(seed=seed)
 
         evidence_var = list(evidence.keys())
         sample_var = [v for v in variables if v not in evidence_var]
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -266,6 +266,11 @@ class LinearGaussianBayesianNetwork(DAG):
         seed: int (default: None)
             Seed for the random number generator.
 
+        missing_prob: LinearGaussianCPD, list  (default: None)
+            In case of missing value for more than one variable, provide list of LinearGaussianCPD.
+            The variable name of each LinearGaussianCPD should end with the name of node in DiscreteBayesianNetwork with * at the end of the name.
+            The state names of each LinearGaussianCPD should be the same as the state names of the corresponding node in DiscreteBayesianNetwork.
+
         Returns
         -------
         pandas.DataFrame: generated samples
@@ -318,13 +323,6 @@ class LinearGaussianBayesianNetwork(DAG):
         else:
             model = self
 
-        mean, cov = model.to_joint_gaussian()
-        variables = list(nx.topological_sort(model))
-        rng = np.random.default_rng(seed=seed)
-
-        evidence_var = list(evidence.keys())
-        sample_var = [v for v in variables if v not in evidence_var]
-
         if missing_prob is not None:
             if isinstance(missing_prob, list):
                 for cpd in missing_prob:
@@ -355,11 +353,6 @@ class LinearGaussianBayesianNetwork(DAG):
                         f"Missingness variable '{missing_var}' refers to base variable '{base_var}', which is not in model nodes."
                     )
 
-                if cpd.cardinality[0] != 2:
-                    raise ValueError(
-                        f"Missingness variable '{missing_var}' must have cardinality 2 (0=observed, 1=missing). Got {cpd.cardinality[0]}."
-                    )
-
                 model.add_node(missing_var)
 
                 if len(cpd.variables) > 1:
@@ -373,6 +366,13 @@ class LinearGaussianBayesianNetwork(DAG):
 
                 model.add_cpds(cpd)
 
+        mean, cov = model.to_joint_gaussian()
+        variables = list(nx.topological_sort(model))
+        rng = np.random.default_rng(seed=seed)
+
+        evidence_var = list(evidence.keys())
+        sample_var = [v for v in variables if v not in evidence_var]
+
         if len(evidence) == 0:
             df = pd.DataFrame(
                 rng.multivariate_normal(mean=mean, cov=cov, size=n_samples),
@@ -381,10 +381,16 @@ class LinearGaussianBayesianNetwork(DAG):
 
         else:
             df = pd.DataFrame([evidence])
-            _, mean_cond, cov_cond = model.predict(data=df)
+            missing_vars, mean_cond, cov_cond = model.predict(data=df)
+            sorted_indices = np.argsort(missing_vars)
+            missing_vars = [missing_vars[i] for i in sorted_indices]
+            mean_cond = mean_cond[:, sorted_indices]
+            cov_cond = cov_cond[sorted_indices][:, sorted_indices]
             df = pd.DataFrame(
-                rng.multivariate_normal(mean=mean_cond, cov=cov_cond, size=n_samples),
-                columns=sample_var,
+                rng.multivariate_normal(
+                    mean=mean_cond[0], cov=cov_cond, size=n_samples
+                ),
+                columns=missing_vars,
             )
 
         if missing_prob is not None:
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -140,7 +140,10 @@ class LinearGaussianBayesianNetwork(DAG):
         P(x1) = N(1; 4)
 
         """
-        return super(LinearGaussianBayesianNetwork, self).remove_cpds(*cpds)
+        for cpd in cpds:
+            if isinstance(cpd, (str, int)):
+                cpd = self.get_cpds(cpd)
+            self.cpds.remove(cpd)
 
     def get_random_cpds(self, loc=0, scale=1, inplace=False, seed=None):
         """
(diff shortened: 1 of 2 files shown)


ankurankan (changes requested):
@Sanchay117 Can you add tests for all of the changes? Apart from checking the types, etc., the tests should also check that the distribution of the data generated from simulations match what we expect analytically.

Sanchay117:
Sure! I'll give it a shot.

ankurankan:
@Sanchay117 Tests are failing.

Sanchay117:
I'm really sorry, I am really new to contributing to open source so I'm making a lot of mistakes. I'll look into why the tests are failing.

ankurankan:
@Sanchay117 I would suggest focusing on making one change at a time, either of `do`, `evidence`, or `missing`. For tests, you can try running the failing tests locally and try to figure out why they might be failing.

Sanchay117:
Also, I noticed that there appears to be a bug in the `LinearGaussianBayesianNetwork` implementation. When calling `.remove_cpds()` during a test, I encountered the following error:

>       >return super(LinearGaussianBayesianNetwork, self).remove_cpds(*cpds)
>       AttributeError: 'super' object has no attribute 'remove_cpds'

Due to this I was unable to add a test for interventions in `.simulate()` as my implementation is using `.remove_cpds()`.

ankurankan:
@Sanchay117 Feel free to add the `remove_cpds` method to LGBN. It's not there because it wasn't required till now with the current functionality.

Sanchay117:
Sure! I’ll go ahead and add it. Let’s see if the tests pass first (hopefully they do). It’s a bit surprising that the functionality wasn’t added earlier, especially since it’s mentioned in the docs.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -253,9 +258,22 @@ def simulate(self, n_samples=1000, seed=None):
         n_samples: int
             The number of samples to draw from the model.
 
+        do: dict (default: None)
+            The interventions to apply to the model. dict should be of the form
+            {variable_name: state}
```
ankurankan: This is continuous data, so should be `{variable: value}`

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -253,9 +258,22 @@ def simulate(self, n_samples=1000, seed=None):
...
+
+        evidence: dict (default: None)
+            Observed evidence to apply to the model. dict should be of the form
+            {variable_name: state}
```
ankurankan: Here as well.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -253,9 +258,22 @@ def simulate(self, n_samples=1000, seed=None):
...
+
         seed: int (default: None)
             Seed for the random number generator.
 
+        missing_prob: LinearGaussianCPD, list  (default: None)
+            In case of missing value for more than one variable, provide list of LinearGaussianCPD.
+            The variable name of each LinearGaussianCPD should end with the name of node in DiscreteBayesianNetwork with * at the end of the name.
```
ankurankan: This needs to be updated: `DiscreteBayesianNetwork` -> `LinearGaussianBayesianNetwork`

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +290,126 @@ def simulate(self, n_samples=1000, seed=None):
         >>> model.add_cpds(cpd1, cpd2, cpd3)
         >>> model.simulate(n_samples=500, seed=42)
         """
+        evidence = {} if evidence is None else evidence
```
ankurankan, line 338: Can you write comments with the main steps of the method? We usually do comments such as: `Step 1: Check all the arguemnts`, `Step 2: If do variables are specified, ... `. Have a look at other algorithm implementations for examples.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +290,126 @@ def simulate(self, n_samples=1000, seed=None):
...
+
+        do = {} if do is None else do
+
+        if set(do.keys()).intersection(set(evidence.keys())):
+            raise ValueError("Variable can't be in both do and evidence")
```
ankurankan: Should specify which variables are common in them.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +290,126 @@ def simulate(self, n_samples=1000, seed=None):
...
                 "Each node in the model should have a CPD associated with it"
             )
 
-        mean, cov = self.to_joint_gaussian()
-        variables = list(nx.topological_sort(self))
+        if do != {}:
+            model = LinearGaussianBayesianNetwork(self.edges())
+            model.add_nodes_from(self.nodes())
+            model.add_cpds(*self.cpds)
+            for var, val in do.items():
```
ankurankan, line 374: I don't think the logic here is correct. For a do operation in LinearGaussianBN, we will need to remove the do variables completely, and for each child variable of the do variable, set the beta parameter corresponding to the do variable to beta[i] * do-value.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +290,126 @@ def simulate(self, n_samples=1000, seed=None):
...
+        sample_var = [v for v in variables if v not in evidence_var]
+
+        if len(evidence) == 0:
+            df = pd.DataFrame(
+                rng.multivariate_normal(mean=mean, cov=cov, size=n_samples),
+                columns=variables,
+            )
+
+        else:
+            df = pd.DataFrame([evidence])
```
ankurankan: You will need to compute the joint distribution and then apply the evidence to compute the posterior joint distribution.

ankurankan (changes requested):
@Sanchay117 I feel that you do not currently understand how these computations should work. I would suggest looking into some theory behind these methods first. I have added comments where the logic is incorrect right now. I would suggest getting rid of the `missing_prob` for now and try to make the other arguments work first. For do argument, you will need to modify the network and the CPDs of the children of do variables. After that you will need to compute the joint distribution over all the remaining variables. If any evidence variables are specified, you will need to then compute the conditional distribution. I would suggest looking at the multivariate Gaussian Wikipedia page for equations on how to compute that.

Sanchay117:
@ankurankan, My apologies for the earlier mistakes. My initial calculations were thinking of this as discrete problem because I was influenced by the `discreteBayesianNetwork.simulation()` implementation, which led to incorrect calculations.

For now, I have removed the calculations for `missing_prob`. Regarding interventions, I am now:

- Removing all incoming edges to the intervened node,
- Updating the CPDs of its children by removing the intervened node from their evidence and adjusting their intercept accordingly,
- Then removing the intervened node from the network (I hope this is the correct approach).

For evidence, I compute the conditional mean and covariance, then after sorting the predicted variables to ensure determinism and avoid intermittent test failures I merge the evidence with the predictions and return the resulting DataFrame.

If anything is still incorrect, my sincerest apologies. Please let me know, and I will make the necessary fixes.

Thank you for your guidance!

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +290,107 @@ def simulate(self, n_samples=1000, seed=None):
...
+        common_vars = set(do.keys()).intersection(set(evidence.keys()))
+        if common_vars:
+            raise ValueError(
+                f"Variable(s) can't be in both do and evidence: {', '.join(common_vars)}"
+            )
+
+        # Step 3: Ensure all variables in the intervention exist in the model's nodes
+        nodes = list(do.keys())
+
+        if not set(nodes).issubset(set(self.nodes())):
```
ankurankan: Shouldn't this check be before the intersection check? We also need to check if evidence variables are in the model.

ankurankan: Also, with the `Step x` comment, all the checks can be in Step 1

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +290,107 @@ def simulate(self, n_samples=1000, seed=None):
...
         if len(self.cpds) != len(self.nodes()):
             raise ValueError(
                 "Each node in the model should have a CPD associated with it"
             )
 
-        mean, cov = self.to_joint_gaussian()
-        variables = list(nx.topological_sort(self))
+        # Step 4: If do is specified, modify the network structure.
+        if do != {}:
+            model = LinearGaussianBayesianNetwork(self.edges())
```
ankurankan: I would suggest creating a new `copy` method for LinearGaussianBN, and then we can have `model_copy = model.copy()` here.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +290,107 @@ def simulate(self, n_samples=1000, seed=None):
...
+
+                    model.remove_cpds(child_cpd)
+                    model.add_cpds(new_cpd)
+
+                model.remove_node(var)
+
+        else:
+            model = self
+
+        mean, cov = model.to_joint_gaussian()
```
ankurankan, line 410: We will need to add the do variables to the final dataframe again with the value set to the do value.

pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
```diff
@@ -106,6 +106,43 @@ def test_simulate(self):
         np_test.assert_array_almost_equal(df_cont.mean(), df_equ.mean(), decimal=1)
         np_test.assert_array_almost_equal(df_cont.cov(), df_equ.cov(), decimal=1)
 
+    def test_simulate_with_evidence(self):
```
ankurankan, line 156: We will need to add tests for some simple models where we can compute the covariance/coefficient under the given do and evidence manually and then compare it with values compared from the simulated data.

ankurankan (changes requested):

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: Yes , you are right. I have fixed this now.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I've done that in the latest commit.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I've done that in the latest commit (at line 424).

pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
Sanchay117: I have added a test where the simulated results are being compared to manual calculations in the latest commit.

Sanchay117:
Hi @ankurankan , I have made the requested changes.
Thanks!

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +323,108 @@ def simulate(self, n_samples=1000, seed=None):
         >>> model.add_cpds(cpd1, cpd2, cpd3)
         >>> model.simulate(n_samples=500, seed=42)
         """
+        # Step 1: Check if all arguments are specified and valid
+        evidence = {} if evidence is None else evidence
+
+        do = {} if do is None else do
+
+        nodes = list(do.keys())
```
ankurankan: I think we should use a better name instead of just `nodes` here. Could you `do_nodes`.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +323,108 @@ def simulate(self, n_samples=1000, seed=None):
...
+
+        if not set(nodes).issubset(set(self.nodes())):
+            raise ValueError(
+                f"Nodes not found in the model: {set(nodes) - set(self.nodes)}"
```
ankurankan: This is not descriptive enough to help in finding the reason. We need to specify nodes specified in the do argument.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +323,108 @@ def simulate(self, n_samples=1000, seed=None):
         >>> model.add_cpds(cpd1, cpd2, cpd3)
         >>> model.simulate(n_samples=500, seed=42)
```
ankurankan: Need to add some examples here with the new arguemnts.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +323,108 @@ def simulate(self, n_samples=1000, seed=None):
...
+
         if len(self.cpds) != len(self.nodes()):
             raise ValueError(
                 "Each node in the model should have a CPD associated with it"
             )
 
-        mean, cov = self.to_joint_gaussian()
-        variables = list(nx.topological_sort(self))
+        common_vars = set(do.keys()).intersection(set(evidence.keys()))
+        if common_vars:
```
ankurankan: Seems like a good place to use Python's walrus operator.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +323,108 @@ def simulate(self, n_samples=1000, seed=None):
...
+                    model.add_cpds(new_cpd)
+
+                model.remove_node(var)
+
+        else:
+            model = self
+
+        mean, cov = model.to_joint_gaussian()
+        variables = list(nx.topological_sort(model))
         rng = np.random.default_rng(seed=seed)
```
ankurankan: This line can go to "Step 1" block.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -272,18 +323,108 @@ def simulate(self, n_samples=1000, seed=None):
...
 
-        mean, cov = self.to_joint_gaussian()
-        variables = list(nx.topological_sort(self))
+        common_vars = set(do.keys()).intersection(set(evidence.keys()))
+        if common_vars:
+            raise ValueError(
+                f"Variable(s) can't be in both do and evidence: {', '.join(common_vars)}"
+            )
+
+        # Step 2: If do is specified, modify the network structure.
```
ankurankan, line 370: Step 2 is doing more than this. The idea with writing the Steps is that we should be able to get the pseudo-code of the overall algorithm by just reading through the Step comments. This is extremely helpful when you look at the code again after a few months, so would suggest improve this a bit. You can also do "Step 2.1" if you want to split it into multiple code blocks.

pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
```diff
@@ -106,6 +106,72 @@ def test_simulate(self):
...
+
+        rng = np.random.default_rng(seed=42)
+        samples = rng.multivariate_normal(mean=mean_cond[0], cov=cov_cond, size=10000)
+        df_equ = pd.DataFrame(samples, columns=missing_vars)
+
+        np_test.assert_array_almost_equal(
+            df.mean()[["x2", "x3"]], df_equ.mean(), decimal=1
+        )
+        np_test.assert_array_almost_equal(
+            df.cov()[["x2", "x3"]].loc[["x2", "x3"]], df_equ.cov(), decimal=1
```
ankurankan: Do we need to use `decimal=1`? I would expect it to be a lot more precise.

ankurankan (changes requested):
@Sanchay117 Thanks for the updates. I think the PR mostly looks good now. I have added a few more comments, but they are mostly cosmetic changes.

pgmpy/tests/test_models/test_LinearGaussianBayesianNetwork.py
Sanchay117: I have updated it to use `decimal=5` in the latest commit.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I have added a few more explanations here.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I have done that in the latest commit.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I have done that in the latest commit.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I have done that in the latest commit.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I have done that in the latest commit.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I don't think it can be moved earlier, since we compute the joint Gaussian and variables based on the model, which may be modified in Step 2 if any interventions (`do`) are applied.

Sanchay117:
@ankurankan Happy to help! If there are any other issues you think I could look into, I'd be glad to contribute.

pgmpy/models/LinearGaussianBayesianNetwork.py
ankurankan: @Sanchay117 I meant the random number generator: `rng = np.random ... `

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -270,20 +321,126 @@ def simulate(self, n_samples=1000, seed=None):
...
         """
+        # Step 1: Check if all arguments are specified and valid
+        evidence = {} if evidence is None else evidence
+
+        do = {} if do is None else do
+
+        do_nodes = list(do.keys())
+
+        invalid_nodes = set(do_nodes) - set(self.nodes())
+        if not set(do_nodes).issubset(set(self.nodes())):
```
ankurankan, line 347: We should have a similar check for `evidence` as well. Right?

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -140,7 +140,10 @@ def remove_cpds(self, *cpds):
         P(x1) = N(1; 4)
 
         """
-        return super(LinearGaussianBayesianNetwork, self).remove_cpds(*cpds)
+        for cpd in cpds:
```
ankurankan, line 143: Could you please add tests for this method as well?

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -270,20 +321,126 @@ def simulate(self, n_samples=1000, seed=None):
...
+
         if len(self.cpds) != len(self.nodes()):
             raise ValueError(
                 "Each node in the model should have a CPD associated with it"
             )
 
-        mean, cov = self.to_joint_gaussian()
-        variables = list(nx.topological_sort(self))
+        if common_vars := set(do.keys()) & set(evidence.keys()):
+            raise ValueError(
```
ankurankan, line 366: Could you add tests for all the codecov annotations?

Sanchay117: Done.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: Sure, I have added one for the `copy()` method as well.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: Yes, I have added it.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: Oh , my apologies. I have moved it now.

ankurankan (approved):
Looks good to me now. Thanks for the PR @Sanchay117

Sanchay117:
Happy to help!

PR #2133, opened 2025-05-27
fix(PC): resolve edge removal and None data bug

Rachel-Lyu:
- Fixed error caused by `data=None` in (estimators/PC.py#L279) due to (estimators/CITests.py#L39-L42).
- Added safe checks in orient_colliders (estimators/PC.py#L514) to prevent accidental edge removal.

### DIFF
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -78,18 +78,7 @@ class DAG(nx.DiGraph):
         self,
         ebunch: Optional[Iterable[tuple[Hashable, Hashable]]] = None,
         latents: set[Hashable] = set(),
-        lavaan_str: Optional[list[str]] = None,
-        dagitty_str: Optional[list[str]] = None,
     ):
-        if lavaan_str:
-            ebunch, latents, err_corr, _ = parse_lavaan(lavaan_str)
-            if err_corr:
-                logger.warning(
-                    f"Residual correlations {err_corr} are ignored in DAG. Use the SEM class to keep them."
-                )
-        elif dagitty_str:
-            ebunch, latents = parse_dagitty(dagitty_str)
-
         super(DAG, self).__init__(ebunch)
         self.latents = set(latents)
         cycles = []
@@ -131,13 +120,20 @@ class DAG(nx.DiGraph):
             lavaan_str = string.split("\n")
         else:
             raise ValueError("Either `filename` or `string` need to be specified")
-
-        return cls(lavaan_str=lavaan_str)
+        ebunch, latents, err_corr, _ = parse_lavaan(lavaan_str)
+        if err_corr:
+            logger.warning(
+                f"Residual correlations {err_corr} are ignored in DAG. Use the SEM class to keep them."
+            )
+        return cls(ebunch=ebunch, latents=latents)
 
     @classmethod
     def from_dagitty(cls, string=None, filename=None) -> "DAG":
         """
         Initializes a `DAG` instance using DAGitty syntax.
+
+        Creates a `DAG` from the dagitty string. If parameter `beta` is specified in the DAGitty
+        string, the method returns a `LinearGaussianBayesianNetwork` instead of a plain `DAG`.
 
         Parameters
         ----------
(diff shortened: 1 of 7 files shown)


pgmpy/estimators/PC.py
```diff
@@ -511,7 +511,11 @@ def orient_colliders(skeleton, separating_sets, temporal_ordering=dict()):
                             (temporal_ordering[Z] >= temporal_ordering[X])
                             and (temporal_ordering[Z] >= temporal_ordering[Y])
                         ):
-                            pdag.remove_edges_from([(Z, X), (Z, Y)])
+                            # delete only if the reverse arc still exists
+                            if pdag.has_edge(Z, X) and pdag.has_edge(X, Z):
+                                pdag.remove_edge(Z, X)
+                            if pdag.has_edge(Z, Y) and pdag.has_edge(Y, Z):
+                                pdag.remove_edge(Z, Y)
```
Nimish-4: I think `remove_edges_from` performs the same thing, silently failing if the edge does not exist.

Nimish-4: @ankurankan In this code block, we can just use `combinations` instead of `permutations`, right? Also, `max_cond_vars` is still unused when it is passed to the `estimate` method.

Rachel-Lyu: Yes but we don’t want to change the skeleton in the orientation stage right? What happens is that edges of both directions might be removed. That’s why we want to check if both directions exist - otherwise we are removing edges without finding a sepset.

Rachel-Lyu: Or say, we want to orient an edge if the edge is undirected (saved as two entries representing both directions), rather than removing the edge when the edge has already been directed (single entry) but some new dsep information conflicts current orientation, simply due to the order dependency of PC algorithm.

Nimish-4: I get it now, though my understanding may not be completely correct. But in case an edge has already been oriented, and we do not want to remove the edge that is remaining, shouldn't the other corresponding edge also be skipped? Otherwise, it seems that we are not really orienting a collider by changing a single edge.

Rachel-Lyu: It can be one way of implementing, or we can reverse the oriented edge. Let me check other packages as well.

Rachel-Lyu: Sorry for the late reply. I saw that other packages like [causal-learn](https://github.com/py-why/causal-learn/blob/9689c1bdc468847729eacf0921b76f598161ae16/causallearn/utils/PCUtils/UCSepset.py) includes those options to choose:
(1) overwriting or prioritizing existing colliders;
(2) orienting bi-directed;
(3) prioritizing stronger colliders (meaning the ones that have a lower p-value or background knowledge).

Rachel-Lyu: The easiest way might be to order colliders by p-values first, then prioritize low-p-value colliders.

Nimish-4: @Rachel-Lyu Thanks for looking it up. Yeah, I agree that since such a situation would be related to statistical inaccuracy, it would be a good idea to prioritize the colliders based on p-values. If this does not work, we could go for a bidirected edge. This may require some work, are you looking to take it up?

Rachel-Lyu: Sure, I can propose some changes, and let's see if they are compatible with other components.

ankurankan: @Rachel-Lyu @Nimish-4 Just wanted to add some more context and references here. The original idea of doing additional CI tests when there is a contradition in collider orientation comes from the conservative PC algorithm [1]. This was later extended to the stable PC algorithm [2], and I think we should focus on matching our implementation to this stable version. Also, a later technical report that summarizes all this is [3], and an even more recent generalization of all these methods is [4]. [4] is however still under peer review and was recently rejected from the UAI conference.

[1] https://arxiv.org/pdf/1206.6843
[2] https://www.jmlr.org/papers/volume15/colombo14a/colombo14a.pdf
[3] https://arxiv.org/pdf/1610.00378
[4] https://www.arxiv.org/pdf/2502.08531

ankurankan: I would also suggest trying to make a list of required changes in a new issue before we start making changes. I will also need to go through these papers again, its been a while since I looked at the them. @Rachel-Lyu Thanks for taking this up, maybe you can create an issue for this?

Rachel-Lyu: Sure, will do soon. Thanks!

PR #2139, opened 2025-05-28
Improve FunctionalCPD sampling by allowing vectorized functions

### LINKED ISSUE #1891: Performance improvements in sampling from FunctionalBN
> Iteratively making a function call can be quite slow when generating a lot of samples. I think we should give the user an option to specify the function in vectorized form. We could possibly add an argument to the __init__ method for the user to specify whether the function is vectorized or not. Using that argument/flag we can make the choice here whether we should do an iterative function call (maybe parallelly) or do a vectorized function call. But lets keep this change for later. Can you please create an issue about it?

Ref #1888 (@Nuna7)

0149Sailesh:

- Fixes #1891 

- Introduced a new flag in FunctionalCPD that indicates if the supplied function allows vectorized input.
- Wrote 4 new tests to validate the working of the change. 

image


ankurankan (changes requested):
@0149Sailesh Thanks for the PR. I think it mostly looks good and is showing significant runtime improvements. I would suggest simplifying the current implementation a bit though. I think we should just get rid of the parallel implementation and just keep two options: 1) vectorized, 2) iteratively sampling.

0149Sailesh:
Got it @ankurankan. Thanks for the suggestion. I'll make the required changes.

0149Sailesh:
Hey @ankurankan 
I have addressed the PR comments. I did not remove the benchmark test as I feel it might help future devs. Let me know if you think otherwise or if the tests can be moved to a different folder where we can club together other benchmark tests.

pgmpy/factors/hybrid/FunctionalCPD.py
```diff
@@ -97,18 +98,40 @@ def sample(self, n_samples=100, parent_sample=None):
                 )
             if len(parent_sample) != n_samples:
                 raise ValueError("Length of `parent_sample` must match `n_samples`.")
+            if self.vectorized:
+                dists = self.fn(parent_sample)
+                if isinstance(dists, list):
```
ankurankan: I don't think dists can ever be a list. Am I missing something here?

pgmpy/factors/hybrid/FunctionalCPD.py
```diff
@@ -97,18 +98,40 @@ def sample(self, n_samples=100, parent_sample=None):
...
-                    ).item()
-                )
         else:
-            for i in range(n_samples):
-                sampled_values.append(
-                    pyro.sample(f"{self.variable}", self.fn(parent_sample)).item()
-                )
+            if self.vectorized:
+                distribution = self.fn(None)
+                if isinstance(distribution, list):
```
ankurankan: Same here.

pgmpy/tests/test_factors/test_hybrid/test_Functional_CPD.py
```diff
@@ -103,3 +104,117 @@ def test_different_distributions(self):
...
+        def iterative_fn(parent_sample):
+            return dist.Normal(0.0, 1.0)
+
+        cpd = FunctionalCPD(variable="z", fn=iterative_fn, parents=[], vectorized=False)
+        samples = cpd.sample(n_samples=1000)
+        self.assertEqual(len(samples), 1000)
+        self.assertTrue(np.isfinite(samples).all())
+
+    # Benchmark test to see performance improvement between vectorized vs iterative. Use -s option to run this test to be able to see prints.
+    def test_sampling_time_vectorized_vs_iterative(self):
```
ankurankan: You can add benchmarks to pgmpy benchmarks (https://github.com/pgmpy/pgmpy-benchmarks), but it should not be in tests.

ankurankan (changes requested):

pgmpy/factors/hybrid/FunctionalCPD.py
0149Sailesh: Ahh yea. To be honest it is an unnecessary check. The only case I can think of is when the supplied function accidentally returns a list. This might cause pyro.sample to throw a TypeError.

pgmpy/tests/test_factors/test_hybrid/test_Functional_CPD.py
0149Sailesh: Sure! I'll move it there

ankurankan (approved):
Looks good to me now. Thanks for the PR @0149Sailesh

Issue #1741, opened 2024-03-04
Refactor Belief Propagation algorithm


ankurankan:
1. Clean the current implementation
2. Better algorithms for creating Junction Trees.
3. Incorporate the Factor Graph BP into the main algorithm. 

Ref #1740

tristantreb:
About 3., IMO it makes sense to keep `BeliefPropagationWithMessageParsing` and `BeliefPropagation` separate because they handle the inference very differently. 

On one hand the `BeliefPropagation` class is tightly bounded to the junction tree algorithm. In fact if you read [this](https://en.wikipedia.org/wiki/Junction_tree_algorithm) article, it is exactly what is happening in the code:
> "One method of exact marginalization in general graphs is called the [junction tree algorithm](https://en.wikipedia.org/wiki/Junction_tree_algorithm), which is simply belief propagation on a modified graph guaranteed to be a tree."

On the other hand the `BeliefPropagationWithMessageParsing` class is tightly bounded to the message parsing method on factor graphs. A relevant extension for this class is to create a `to_factor_graph()` method that transforms any graph into a factor graph, similarly to the existing `to_junction_tree()` method. Also, I'm planning to create an approximate inference class that inherits from `BeliefPropagationWithMessageParsing` to handle loopy cases.

Hence, what do you think of keeping them separated?

tristantreb:
I would even suggest to rename `BeliefPropagation` to `JunctionTree`/`BeliefPropagationOnJunctionTree` and `BeliefPropagationWithMessageParsing` to `BeliefPropagation`/`BeliefPropagationOnFactorGraphs`. I was initially confused when trying to understand what kind of BP algorithm was implemented under the `BeliefPropagation` class. Knowing it was the junction tree algorithm from start would have been very helpful! Happy to help make the transition

ankurankan:
@tristantreb 

> A relevant extension for this class is to create a to_factor_graph() method that transforms any graph into a factor graph, similarly to the existing to_junction_tree() method.

We do have a `to_factor_graph` method here: https://github.com/pgmpy/pgmpy/blob/cc2260e50712c72d71623474483bef17e15aa4a4/pgmpy/models/MarkovNetwork.py#L276

> IMO it makes sense to keep BeliefPropagationWithMessageParsing and BeliefPropagation separate because they handle the inference very differently.

[...]

PR #2149, opened 2025-05-31
Feature/Efficient belief propagation on large graphs

tristantreb:
### Extended features for the recursive belief propagation algorithm on large graphs
Pgmpy struggles with graphs over ~1,000 variables (perhaps also with fewer!). Rather than copying a one‐day network 1,000 times into a huge “meta‐graph,” you can use a technique called slicing:
1. Run belief propagation on the single‐day model with that day’s evidence.
2. Take the resulting posteriors and feed them in as virtual evidence for the next day.
3. Repeat day-by-day using the same graph.

This way, you avoid a million‐node graph and still capture cross-day dependencies.

This PR extends the `BeliefPropagationWithMessagePassing` class to allow creating the slicing layer and run it efficiently.

Implementing slicing requires 3 updates of the BeliefPropagationWithMessagePassing class, the subsequent improvements are also useful in other contexts (such as cutset conditioning):

A. Ability to add virtual evidence. This replicates a feature from the BeliefPropagation class (will help with https://github.com/pgmpy/pgmpy/issues/1741 )
B. Ability to get back all computed messages. This allows improved interpretability and debugging
C. Ability to input precomputed messages. This allows faster inference when the same model evidence are observed across multiple queries

I added tests for every item


tristantreb:
Did not aim to create this PR

tristantreb:
@ankurankan this is good to go

PR #2152, opened 2025-06-01
ImportError

hillhack:
changes:
1. shifted pyparsing to core
2. remove e.msg as ImportError as no attribute msg ,instead use f{e}.


ankurankan:
@hillhack Tests are failing, you will need to format your files using black. In the contributing guide, there are instructions to install a pre-commit hook so that this doesn't happen.

model.bif
```diff
@@ -0,0 +1,30 @@
...
+}
+probability ( C | B ) {
+    ( 0 ) 0.593140087429591, 0.406859912570409;
+
+}
+probability ( D | C ) {
+    ( 0 ) 0.5770513838082355, 0.4229486161917645;
+    ( 1 ) 0.5459904136035945, 0.45400958639640554;
+
+}
```
Nimish-4: You should also delete this `.bif` file. It is created by a test that fails to clean up the test files created.

ankurankan (approved):
Looks good to me.

PR #2166, opened 2025-06-03
Further Expand .simulation() capabilities in LGBN's

### LINKED ISSUE #2110: [ENH] Expand simulation capability of Linear Gaussian Bayesian Networks

Sanchay117:

- Fixes #2110 

- Added support for virtual interventions
- Added a test as well for virtual_interventions

### DIFF
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -285,7 +285,6 @@ class LinearGaussianBayesianNetwork(DAG):
         virtual_intervention=None,
         include_latents=False,
         seed=None,
-        missing_prob=None,
     ):
         """
         Simulates data from the given model.
@@ -313,12 +312,6 @@ class LinearGaussianBayesianNetwork(DAG):
 
         seed: int (default: None)
             Seed for the random number generator.
-
-        missing_prob: LinearGaussianCPD, list  (default: None)
-            In case of missing value for more than one variable, provide list of LinearGaussianCPD.
-            The variable name of each LinearGaussianCPD should end
-              with the name of node in LinearGaussianBayesianNetwork with
-                * at the end of the name.
 
         Returns
         -------
@@ -391,42 +384,6 @@ class LinearGaussianBayesianNetwork(DAG):
                         f"The following nodes are present in the model: {self.nodes()}"
                     )
 
-        if missing_prob is not None:
-            if isinstance(missing_prob, list):
-                for cpd in missing_prob:
-                    if not isinstance(cpd, LinearGaussianCPD):
-                        raise ValueError(
-                            f"missing_prob must be a list of LinearGaussianCPD objects. Got {type(cpd)}"
-                        )
-            else:
-                if isinstance(missing_prob, LinearGaussianCPD):
-                    missing_prob = [missing_prob]
-                else:
-                    raise ValueError(
-                        f"missing_prob should be LinearGaussianCPD. Got {type(missing_prob)}"
-                    )
-
-            for cpd in missing_prob:
-                variable = cpd.variable
-
-                if not variable.endswith("*"):
-                    raise ValueError(
-                        f"Got {variable}. LinearGaussianCPD variable should end with *"
-                        " symbol to represent missingnness variable."
-                    )
-
-                if variable.split("*")[0] not in model.nodes:
-                    raise ValueError(
-                        f"Got {variable}. LinearGaussianCPD variable not in model nodes."
-                    )
-
-                for evidence_node in cpd.evidence:
-                    if evidence_node not in model.nodes:
-                        raise ValueError(
-                            f"LinearGaussianCPD has {evidence_node} as evidence which is not present in the model."
-                            f"CPD: {cpd}"
-                        )
-
         # Step 2: If do is specified, modify the network structure.
         if do != {}:
             for var, val in do.items():
@@ -474,19 +431,10 @@ class LinearGaussianBayesianNetwork(DAG):
             for parent in list(model.get_parents(var)):
                 model.remove_edge(parent, var)
 
-        # Step 4: If missing_prob is not empty, modify the network accordingly
-        if missing_prob is not None:
-            for cpd in missing_prob:
-                var = cpd.variable
-                model.add_node(var)
-                for evidence_node in cpd.evidence:
-                    model.add_edge(evidence_node, var)
-                model.add_cpds(cpd)
-
         mean, cov = model.to_joint_gaussian()
         variables = list(nx.topological_sort(model))
 
-        # Step 5: Sample according to evidence
+        # Step 4: Sample according to evidence
         if len(evidence) == 0:
             df = pd.DataFrame(
                 rng.multivariate_normal(mean=mean, cov=cov, size=n_samples),
@@ -517,34 +465,11 @@ class LinearGaussianBayesianNetwork(DAG):
 
             df = df[variables]
 
-        # Step 6: Add do variables to the final dataframe
+        # Step 5: Add do variables to the final dataframe
         for do_var, do_val in do.items():
             df[do_var] = do_val
 
-        # Step 7: Apply probabilistic missingness masking based on missing_prob CPDs.
-        if missing_prob:
-            for cpd in missing_prob:
-                var_star = cpd.variable
-                var = var_star.rstrip("*")
-
-                # Step 7.1: Use sampled value from the synthetic node as the logit (log-odds) of missingness
-                logits = df[var_star]
-                # Step 7.2: Convert logits to probabilities via sigmoid to get P(var is missing)
-                p_missing = 1 / (1 + np.exp(-logits))
-
-                # Step 7.3: For each row, draw 1 with prob=p_missing → 1 means "missing"
-                is_missing = np.random.binomial(1, p_missing)
-
-                # Step 7.4: Store the unmasked/original value in a new column for reference
-                df[f"{var}_full"] = df[var]
-
-                # Step 7.5: If missing mask the original variable with NaN
-                df.loc[is_missing == 1, var] = np.nan
-
-                # Step 7.6: Drop the synthetic "var*" node used for computing missingness
-                df.drop(columns=[var_star], inplace=True)
-
-        # Step 8: Remove latent variables if specified
+        # Step 6: Remove latent variables if specified
         if not include_latents:
             df = df.drop(columns=self.latents)
 
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -6,7 +6,6 @@ from sklearn.linear_model import LinearRegression
 from pgmpy.base import DAG
 from pgmpy.factors.continuous import LinearGaussianCPD
 from pgmpy.global_vars import logger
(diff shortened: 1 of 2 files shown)


Sanchay117:
Hey @ankurankan, could you please confirm if my approach for handling `virtual_interventions` is correct?
Currently, I’m replacing the CPDs of the specified variables with those provided in the `virtual_interventions` list.
I’m not entirely sure if this is the right way to handle it. If it looks good to you, I’ll go ahead and start implementing the remaining features.

Sanchay117:
Also @ankurankan, there seems to be an inconsistency between the implementation in `LinearGaussianCPD.py` and the tests in `test_LinearGaussianBayesianNetwork.py`:

In the `setUp` method of the tests, we define the CPDs like this:

```python
def setUp(self):
    self.model = LinearGaussianBayesianNetwork([("x1", "x2"), ("x2", "x3")])
    self.cpd1 = LinearGaussianCPD(variable="x1", beta=[1], std=4)
    self.cpd2 = LinearGaussianCPD(
        variable="x2", beta=[-5, 0.5], std=4, evidence=["x1"]
    )
    self.cpd3 = LinearGaussianCPD(
        variable="x3", beta=[4, -1], std=3, evidence=["x2"]
    )
```

Here, we specify the standard deviations directly.
However, in the `test_simulate` method:
```python
def test_simulate(self):
    self.model.add_cpds(self.cpd1, self.cpd2, self.cpd3)
    df_cont = self.model.simulate(n_samples=10000, seed=42)

    # Same model in terms of equations
    rng = np.random.default_rng(seed=42)
    x1 = 1 + rng.normal(0, 2, 10000)
    x2 = -5 + 0.5 * x1 + rng.normal(0, 2, 10000)
    x3 = 4 + -1 * x2 + rng.normal(0, np.sqrt(3), 10000)
    df_equ = pd.DataFrame({"x1": x1, "x2": x2, "x3": x3})

```
We are passing values like `2` and `np.sqrt(3)` as the standard deviations to `rng.normal`, which expects the standard deviation as input.

This suggests that the `std` parameter passed to `LinearGaussianCPD` might actually be interpreted as variance in the implementation, even though it's named `std`. It would be great if you could clarify whether this is intentional or a mismatch between the code and the tests.

I have opened an issue regarding this as well.

Sanchay117:
And, it appears the tests failed because I forgot to run the linting checks. My apologies!

Sanchay117:
Also, I feel that this error check could be more robust:
```python
if len(self.cpds) != len(self.nodes()):
            raise ValueError(
                "Each node in the model should have a CPD associated with it"
            )
```
Instead of checking the lengths of the CPD list and the number of nodes which might still pass even if the CPDs correspond to the wrong variables we could instead verify that each node in the model has exactly one CPD associated with it.

Please let me know if you’d like me to proceed with this change.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -407,6 +422,23 @@ def simulate(
         else:
             model = self
 
+        # Step 3: If virtual_interventions are specified, change the CPD's of intervened variables
+        if virtual_intervention != []:
+            # Step 3.1: Check if CPD's are valid
+            for cpd in virtual_intervention:
+                var = cpd.variable
+                if var not in model.nodes():
```
ankurankan: Please move all checks to the beginning of the method in Step 1.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -407,6 +422,23 @@ def simulate(
...
+                    raise ValueError(
+                        f"Virtual intervention provided for variable which is not in the model: {var}"
+                        f"The following nodes are present in the model: {model.nodes()}"
+                    )
+
+            for cpd in virtual_intervention:
```
ankurankan: We would also need to remove all incoming edges to the variables. This currently is not a valid model as every CPD in the model needs to be `P(var | parents(var))`, however, here `var` has parents but the CPD is just `P(var)`

ankurankan (changes requested):
I have added a comment about removing all incoming edges to the virtual intervention variable. This is similar to the model structure modification that we do for `do` variables.

ankurankan:
> Also, I feel that this error check could be more robust:
> 
> ```python
> if len(self.cpds) != len(self.nodes()):
>             raise ValueError(
>                 "Each node in the model should have a CPD associated with it"
>             )
> ```
> 
> Instead of checking the lengths of the CPD list and the number of nodes which might still pass even if the CPDs correspond to the wrong variables we could instead verify that each node in the model has exactly one CPD associated with it.
> 
> Please let me know if you’d like me to proceed with this change.

Thanks for pointing this out. Please feel free to fix this.

Sanchay117:
> > Also, I feel that this error check could be more robust:
> > ```python
> > if len(self.cpds) != len(self.nodes()):
> >             raise ValueError(
> >                 "Each node in the model should have a CPD associated with it"
> >             )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Instead of checking the lengths of the CPD list and the number of nodes which might still pass even if the CPDs correspond to the wrong variables we could instead verify that each node in the model has exactly one CPD associated with it.
> > Please let me know if you’d like me to proceed with this change.
> 
> Thanks for pointing this out. Please feel free to fix this.

It appears there was a method `check_model` for the same purpose but it wasn't being used so I am using it now.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: Done.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: So now, I am assigning the new CPD's to their respctive variables and removing the incoming edges as well. Please let me know if this is the correct approach?

ankurankan: Indeed. In the case of hard intervention, we had to remove the node and change the CPD for the child nodes because there is no way to define a Gaussian distribution that has the whole probability mass at a single value.

Sanchay117: Okay, thank you for the explanation. I have done this only in the latest commit.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -409,15 +432,21 @@ def simulate(
 
                 model.remove_node(var)
 
-        else:
-            model = self
+        # Step 3: If virtual_interventions are specified, change the CPD's of intervened variables
+        # to specified ones and remove the incoming nodes
+        if virtual_intervention != []:
```
ankurankan: Don't really need this check. Right? If it is an empty list, the code block would just get skipped.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -453,6 +482,10 @@ def simulate(
         for do_var, do_val in do.items():
             df[do_var] = do_val
 
+        # Step 6: Remove latent variables if specified
+        if not include_latents:
+            df = df.drop(columns=self.latents, errors="ignore")
```
ankurankan: Why would we need `errors='ignore'`? I don't think we will ever run into a situation where any of the latents is not present in the dataframe.

ankurankan (changes requested):
I think it looks good. I have added a couple of minor comments.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: You are right, I have removed that now.

pgmpy/models/LinearGaussianBayesianNetwork.py
Sanchay117: I have removed this check now.

Sanchay117:
Hi @ankurankan,

Here is what I have done to handle `missing_prob`:

### For each CPD in `missing_prob`:

- Add the missingness variable (the CPD’s variable, e.g. `X*`) as a **new node** to the model.
- Add edges **from each evidence node to this missingness node**.

### Sampling the data:

- Perform the normal simulation/sampling step to generate data for **all variables, including the missingness variables**.
- This produces a DataFrame containing columns for the original variables (e.g., `X`) **and** their corresponding missingness indicators (e.g., `X*`).

### Applying the missingness mask:

For each missingness variable `var*`:

1. Use the simulated values of `var*` as **logits** (log-odds) for the probability of missingness.
2. Convert logits to probabilities using the sigmoid function:

   p_missing = 1 / (1 + exp(-logit))

3. For each sample, draw from a Bernoulli distribution with probability p_missing to decide if that sample of the original variable `var` should be missing.
4. Create a new column `var_full` that stores the original unmasked values (before applying missingness).
5. Set values of `var` to `NaN` wherever the Bernoulli draw indicates missingness.
6. Drop the missingness indicator column `var*` from the final DataFrame.

---

Please let me know if this is the correct approach.

Sanchay117:
Also, if there’s any other issue you think I could help with, I’d really appreciate it if you could point me to one. I’d love to pick something up, as most open issues seem to be already assigned or have PRs.

ankurankan:
@Sanchay117 can you remove the missingness part from this PR? I am not yet sure how missingness can be handled in the GaussianBN case and will need some time to think about it. The remaining part of the PR looked good, so if you remove the missingness part for now, we can merge the rest.

Sanchay117:
@ankurankan Sure, I'll remove the `missing_prob` part.

ankurankan:
@Sanchay117 I think another missing part in simulations is `virtual_evidence`. This would be similar to the `evidence` argument, where we compute a conditional distribution using the evidence values and sample from this conditional distribution. The difference would be that the evidence variable would need to be set to a distribution instead of a single value. I think there should be straightforward formulas to compute that, but I haven't looked it up. Another thing to consider here would be handling both hard and virtual evidence together. Would you like to take this up? If yes, feel free to create another PR for that, and check if this is already in the list of features in the issue.

Sanchay117:
@ankurankan I'd love to take this up! I'll create a separate PR for it. Just checked #2110, and `virtual_evidence` doesn’t seem to be listed as a feature there.

Sanchay117:
@ankurankan , I have removed the missingness part. You can merge the PR now if everything looks right to you. I'll make another request for `virtual_evidence`.

ankurankan (approved):
Thanks for the PR @Sanchay117

Issue #2162, opened 2025-06-03
[BUG] Handling Missing Optional Dependency (xgboost) in Test Suite


hillhack:
Several test functions in the codebase rely on the optional xgboost package. However, not all of these functions are properly guarded with try-except blocks or conditional skips. As a result, when xgboost is not installed in the environment, these tests raise exceptions such as ImportError or ModuleNotFoundError, rather than being gracefully skipped.

 Example:
A specific example is the function **test_edge_strength_basic** in test_dag.py.
This test calls **dag.edge_strength()**, which internally uses the **pillai_trace** function — a function that depends on xgboost.
If xgboost is not installed, this test fails instead of being skipped.

![Image](https://github.com/user-attachments/assets/cd656a5a-e46e-4c31-bbf2-4506afb7785c)

PR #2168, opened 2025-06-03
Ensure tests requiring xgboost are conditionally skipped

hillhack:
Pr for issue: https://github.com/pgmpy/pgmpy/issues/2162
changes:

1. Add exception handling for xgboost dependency in test functions.
2. Centralize xgboost availability check into optional_dependency.py

### DIFF
--- a/pgmpy/tests/test_base/test_DAG.py
+++ b/pgmpy/tests/test_base/test_DAG.py
@@ -7,6 +7,7 @@ import warnings
 import networkx as nx
 import numpy as np
 import pandas as pd
+import pytest
 
 import pgmpy.tests.help_functions as hf
 from pgmpy.base import DAG, PDAG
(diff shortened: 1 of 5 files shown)


ankurankan:
We shouldn't skip any tests. Only developers would be running tests, and the test environment is expected to have all the dependencies installed. For users who don't have xgboost installed, it throws a clear error asking for installing xgboost.

hillhack:
@ankurankan Sir, I raised an issue and started working on a fix, but later you mentioned the original code was fine. It turned out to be a misunderstanding, and I realize it led to some wasted effort. In such cases, how should we proceed when we think there's a problem — should we confirm with you before opening an issue or PR or any other way you use?

ankurankan:
@hillhack Ideally, you should create an issue and discuss it before starting to work on it.

PR #2171, opened 2025-06-03
fix: update DAG validation and structure scoring logic

### LINKED ISSUE #2148: [ENH] Add a `DAG.validate` method to show a summary of all validation methods.
We are working on expanding the validation functionality for DAGs. To make it easier for users to validate their model, we want to add a `DAG.validate` method that can summarize the results from all the validation methods into a nice tabular format. Metrics to include:

1. Correlation score
2. Log-likelihood.
3. AIC and BIC scores.
4. No. of failing implied Conditional Independence (CIs) and the total number of implied CIs.
5. Fisher C p-value. 
6. RMSEA based on Fisher C.
7. Permutation test p-value (#2119)

D03xen:

- Partially Fixes #2148 

-  Implemented a common LogLikeliHood class and extended to AIC and BIC in StructureScore.py
- Added DAG.validate() with correlation, AIC, BIC and Fisher_c with exception handling


pgmpy/base/DAG.py
```diff
@@ -11,6 +11,7 @@
 from pgmpy.base import UndirectedGraph
 from pgmpy.global_vars import logger
 from pgmpy.independencies import Independencies
+from pgmpy.metrics.metrics import correlation_score, fisher_c, structure_score
```
ankurankan: You might need to put this import inside the validate method. Could be leading to the circular import because of which the tests are failing.

pgmpy/base/DAG.py
```diff
@@ -1583,6 +1584,74 @@ def edge_strength(self, data, edges=None):
...
+                results["aic"] = structure_score(self, data, scoring_method="aic-d")
+            except ValueError:
+                try:
+                    results["aic"] = structure_score(self, data, scoring_method="aic-g")
+                except ValueError as e:
+                    results["aic"] = f"Error: {str(e)}"
+
+        # BIC-score with fail-checks
+        if "bic" in metrics:
+            try:
```
ankurankan: Why do we need try except blocks for scoring methods?

ankurankan (changes requested):

pgmpy/base/DAG.py
D03xen: Got it. That solved the issue for circular imports

pgmpy/base/DAG.py
D03xen: The methods from metrics.py raises ValueError for different checks and I wanted to properly format and store them so that they can be shown in tabulated format. Also I provided an option in case aic-d fails we can switch to aicGauss, same for bic-d and bicGauss

pgmpy/base/DAG.py
```diff
@@ -1583,6 +1583,139 @@ def edge_strength(self, data, edges=None):
...
+            "permutation_test",
+        ]
+
+        # a normal validate call would provide all metric results
+        if metrics is None:
+            metrics = all_metrics
+
+        # Dictionary for holding default parameters
+        params = {
+            "correlation_test": "chi_square",
```
ankurankan: This needs to be automatically detected based on the data. Have a look at: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/CITests.py#L14

pgmpy/base/DAG.py
```diff
@@ -1583,6 +1583,139 @@ def edge_strength(self, data, edges=None):
...
+            "correlation_significance": 0.05,
+            "correlation_score_func": f1_score,
+            "ci_test_permutation": chi_square,
+            "n_permutations": 1000,
+            "ci_test_fischer": chi_square,
```
ankurankan: This and the next line as well.

pgmpy/base/DAG.py
```diff
@@ -1583,6 +1583,139 @@ def edge_strength(self, data, edges=None):
...
+            results["Correlation Score"] = correlation_score(
+                model=self,
+                data=data,
+                test=params["correlation_test"],
+                significance_level=params["correlation_significance"],
+                score=params["correlation_score_func"],
+            )
+
+        # AIC-score with fail checks
+        if "aic" in metrics:
```
ankurankan: Same, needs to be automatically detected. Check: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/StructureScore.py#L15

pgmpy/metrics/metrics.py
```diff
@@ -333,7 +333,7 @@ def implied_cis(model, data, ci_test, show_progress=True):
     return cis
 
 
-def fisher_c(model, data, ci_test, show_progress=True):
+def fisher_c(model, data, ci_test, calculate_rmsea=False, show_progress=True):
```
ankurankan: Please make `calculate_rmsea=True` by default.

pgmpy/metrics/metrics.py
```diff
@@ -403,6 +408,63 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+            f"model must be an instance of DAG or DiscreteBayesianNetwork. Got {type(model)}"
+        )
+    elif not isinstance(data, pd.DataFrame):
+        raise ValueError(f"data must be a pandas.DataFrame instance. Got {type(data)}")
+
+    if random_state is not None:
+        np.random.seed(random_state)
+
+    # Null hypothesis p-value using fischer_c method
+    null_hypothesis_val = fisher_c(
```
ankurankan: Was this now they computed the null distribution in the paper?

ankurankan (changes requested):
I have added a few comments.
1. We would need to get rid of all try-excepts. I don't think there is any case where these methods would fail if we use the correct parameters.
2. We will need to add a lot more tests for the methods.

pgmpy/metrics/metrics.py
D03xen: I had not realised the paper already mentioned in the linked issue and so followed a different paper for implementing this. This is the link: https://link.springer.com/chapter/10.1007/978-3-642-15939-8_21 . I will implement the new permutation test accordingly.

pgmpy/base/DAG.py
```diff
@@ -1639,6 +1639,142 @@ def edge_strength(self, data, edges=None):
 
         return strengths
 
+    def _get_non_descendants(self, node):
```
ankurankan: This should be equivalent to `DAG._get_ancestors_of`. Right?

pgmpy/base/DAG.py
```diff
@@ -1639,6 +1639,142 @@ def edge_strength(self, data, edges=None):
...
+        for node in nodes:
+            parents = list(self.get_parents(node))
+            non_descendants = self._get_non_descendants(node)
+            test_nodes = [
+                nd for nd in non_descendants if nd not in parents and nd != node
+            ]
+
+            for test_node in test_nodes:
+                try:
+                    _, p_value = ci_test_func(
```
ankurankan: How does a CI test work without data?

pgmpy/base/DAG.py
```diff
@@ -1639,6 +1639,142 @@ def edge_strength(self, data, edges=None):
...
+            if current not in visited:
+                visited.add(current)
+                descendants.add(current)
+                queue.extend(self.successors(current))
+
+        all_nodes = set(self.nodes())
+        non_descendants = all_nodes - descendants - {node}
+        return list(non_descendants)
+
+    def count_lmc_violations(self, ci_test_func: Callable, significance_level=0.05):
```
ankurankan: This is not really a property of DAG, so should be in the same file as the `permutation_test`.

ankurankan (changes requested):
Can you please separate out the `pertmutation_test` in a different PR? I think it still needs quite a bit of work.  We can then merge the validate method firsthand then add the permutation test to it.

pgmpy/base/DAG.py
```diff
@@ -1639,6 +1639,142 @@ def edge_strength(self, data, edges=None):
...
+                    if p_value < significance_level:
+                        violations += 1
+                except Exception as e:
+                    logger.debug(
+                        f"CI test failed for {node} ⊥ {test_node} | {parents}: {e}"
+                    )
+                    continue
+        return violations
+
+    def permuted_graph(self, perm_mapping):
```
ankurankan: This should also be with the `permutation_test`.

pgmpy/metrics/metrics.py
```diff
@@ -403,9 +479,98 @@ def fisher_c(model, data, ci_test, show_progress=True):
...
+    if n_permutations is None:
+        n_permutations = 1000
+
+    ci_test_func = get_callable_ci_test(ci_test, data=data)
+
+    # Constructing the Null Hypothesis (LMC = Local Markov Condition)
+    lmc_violations_given = model.count_lmc_violations(ci_test_func, significance_level)
+    # Finding set of d-separated nodes in original model
+    nodes = list(model.nodes())
+    original_dsep_triples = model.d_separated_triples(
```
ankurankan: Can simply use `implied_cis`

pgmpy/base/DAG.py
D03xen: The paper states all of the non descendants of a node i are required when considering _ND(i)_:
Screenshot 2025-06-20 at 16 59 06
I think this would include the _Anc(i)_ as well as the nodes that are not connected to _i_. The `_get_ancestors_of` method only returns the ancestors of _i_ and not the nodes from other branches. Like considering the following example: 
graph
these are the outputs for the 2 functions:
Screenshot 2025-06-20 at 17 11 05

pgmpy/base/DAG.py
D03xen: Will correct that

pgmpy/metrics/PermutationTest.py
```diff
@@ -0,0 +1,160 @@
...
+    else:
+        pbar = range(n_permutations)
+
+    # Running permutations
+    for _ in pbar:
+        perm = np.random.permutation(nodes)
+        perm_mapping = dict(zip(nodes, perm))
+
+        # Creating the permuted model while maintaining causal structure
+        permuted_model = model.create_permuted_graph(perm_mapping)
```
ankurankan: I think we can simplify this further, and we do not really need to explicitly compute the permuted model. Let's say if we have an implied CI `X _|_ Y | Z` in the original DAG, and we use the following mapping to permute the graph `{X: X1, Y: Y1, Z: Z1}`. In this new graph the implied CI would simply be `X1 _|_ Y1 | Z1`. So, we can essentially 1) Compute the implied CI in the original DAG, 2) Generate a mapping, 3) Apply the mapping to the implied CIs, 4) Run the CI tests on the permuted implied CIs, 5) the remaining steps. This would avoid us from doing permutation operations on the graph and then computing implied CIs on the permuted graph, saving a lot of compute time. What do you think?

ankurankan (changes requested):

ankurankan:
@D03xen If possible, I would suggest separating out the Permutation test in another PR. I think the validate method is mostly ready, and we can merge that first. The permutation test will still need a few rounds of review.

pgmpy/metrics/PermutationTest.py
D03xen: That makes sense. I'll first put up a PR for the `validate` and follow up with this implementation in a permutation-test commit

ankurankan:
@D03xen Sorry, I missed to follow up on this. Are you still working on this?

D03xen:
> @D03xen Sorry, I missed to follow up on this. Are you still working on this?

Yes. Are there any changes you suggest or else after merging I can go on to create a separate PR for the Permutation test

pgmpy/base/DAG.py
```diff
@@ -1902,6 +1902,66 @@ def edge_strength(self, data, edges=None):
...
+        if metrics is None:
+            metrics = all_metrics
+        # Dictionary for holding default parameters
+        params = {
+            "test": "chi_square",
+            "significance_level": 0.05,
+            "score": f1_score,
+            "ci_test": chi_square,
+            "calculate_rmsea": True,
+            "show_progress": True,
```
ankurankan: `show_progress` needs to be an argument for the method.

pgmpy/base/DAG.py
```diff
@@ -1902,6 +1902,66 @@ def edge_strength(self, data, edges=None):
...
+        }
+        # For custom parameter updation
+        params.update(kwargs)
+
+        # to store the results of different tests
+        test_results = {}
+
+        for metric_name in metrics:
```
ankurankan: I think most of the logic regarding `get_metrics` should be shifted to the `get_metrics` method itself. For example, if the user doesn't pass any metrics, `get_metrics` should return all existing metrics. We shouldn't be defining `all_metrics` or default `params` here. I think shifting all of this would make the code modular and would make it easier to maintain these lists if they are in a single place in the metrics file.

pgmpy/base/DAG.py
```diff
@@ -1902,6 +1902,66 @@ def edge_strength(self, data, edges=None):
...
+            test_results[metric_name] = get_metrics(
+                metric=metric_name, model=self, data=data, **params
+            )
+
+        metric_vals = {}
+        for t in test_results:
+            if t in ["correlation", "log-likelihood", "aic", "bic"]:
```
ankurankan: I would suggest adding helper methods for some of these metrics in the `metrics.py` file that can return the single value that we expect here.

pgmpy/metrics/metrics.py
```diff
@@ -11,6 +13,65 @@
...
+            )
+
+        metric_info = metrics_and_params[metric]
+        metric_func = metric_info["func"]
+        filtered_kwargs = {
+            k: v for k, v in kwargs.items() if k in metric_info["req_params"]
+        }
+
+        if metric in ["aic", "bic"]:
+            var_type = get_dataset_type(data)
```
ankurankan: If you use the `pgmpy.estimators.StructureScore.get_scoring_method`, it should be able to handle the different scenarios here.
1. If scoring_method is specified as a string, it will return a scoring method.
2. If scoring_method is None, it will automatically select a scoring method based on the data type.

pgmpy/base/DAG.py
```diff
@@ -1902,6 +1902,66 @@ def edge_strength(self, data, edges=None):
...
+        )
+        # a normal validate call would provide all metric results
+        if metrics is None:
+            metrics = all_metrics
+        # Dictionary for holding default parameters
+        params = {
+            "test": "chi_square",
+            "significance_level": 0.05,
+            "score": f1_score,
+            "ci_test": chi_square,
```
ankurankan: We can't default to `chi_square`, it depends on the type of data. You can use the `pgmpy.estimators.CITest.get_callable_ci_test` to automatically choose the correct method.

ankurankan (changes requested):
@D03xen I have added a few comments. Would you mind separating out the Log-Likelihood changes in a separate PR? I think those are ready to be merged, the `validate` method might take some more time.

pgmpy/base/DAG.py
```diff
@@ -1902,6 +1902,66 @@ def edge_strength(self, data, edges=None):
 
         return strengths
 
+    def validate(self, data, metrics: Optional[tuple[str | Callable]] = None, **kwargs):
```
ankurankan: It might also be better to list all the arguments here instead of `kwargs` with some defaults.

D03xen:
> @D03xen I have added a few comments. Would you mind separating out the Log-Likelihood changes in a separate PR? I think those are ready to be merged, the `validate` method might take some more time.

I have separated out the Log-Likelihood changes in PR #2301

ankurankan:
@D03xen Thanks. I’ve merged that PR. One more thought: could we split out the RMSEA function into its own PR? It looks ready and we can merge it quickly. What do you think?

ankurankan:
@D03xen Also, if you are making any changes in this branch, make sure to do a git pull first, as I have resolved some merge conflicts.

D03xen:
> @D03xen Thanks. I’ve merged that PR. One more thought: could we split out the RMSEA function into its own PR? It looks ready and we can merge it quickly. What do you think?

I have added RMSEA with tests in PR #2309

pgmpy/base/DAG.py
D03xen: I assumed the function definition would be too messy if all the parameters were put in it. If it is not an issue I can go with this approach

ankurankan:
@D03xen I have fixed a merge conflict in this PR, so please make sure to do a `git pull` before adding your changes.
Also, please mark the PR as ready for review/ request review (on the top right of this page) when it is ready for another round of review.

pgmpy/base/DAG.py
D03xen: The `show_progress` parameter by default is True in the metric function calls which I am using as a list of in the `validate` function. It can be passed as False if required to the `validate` method itself. The same can be done with other concerned parameters. I have updated the docstring accordingly

pgmpy/base/DAG.py
D03xen: Currently I have changed the `get_metrics` to return back function Callables to the validate itself based on if user is passing None or a set of methods. From the `validate` function the kwargs will get filtered and the respective metrics will be run. This will help maintain the `get_metrics` if further metrics are added. It now holds a dictionary of method names and their method calls

pgmpy/base/DAG.py
D03xen: Have included this in the `validate` method

pgmpy/base/DAG.py
D03xen: Based on the recent revision the user can pass any argument to `DAG.validate()`. The necessary one for CI failure testing would be the significance level which I have explicitly added and defaulted to 0.05 in the method definiton. Other metrics would continue to use the parameters with default values that were already being used in `pgmpy.metrics.metrics`

pgmpy/metrics/metrics.py
D03xen: I am currently using the string method only as I am passing the string based on the dataset type returned by `pgmpy.utils.get_dataset_type()`. The `get_scoring_method()` returns instances of the `StructureScore` based scoring class which can not be passed directly to the Structure Score metric in `pgmpy.metrics.metrics` as a parameter. 

Also this would be AIC and BIC specific and based on the dataset type the `get_scoring_method()` returns the first scoring method it finds from the list. So for example if the dataset type is discrete it would give me a scoring method of K2 whereas i would require either of AIC or BIC types.

Let me know if I maybe understood this wrong or maybe missed something

pgmpy/metrics/metrics.py
```diff
@@ -13,6 +14,43 @@
...
+                    f"Unknown metric method. Available metrics are: {list(metric_call.keys())}"
+                )
+
+            callable_metrics[metric] = metric_call[metric]
+
+        elif callable(metric):
+            metric_name = [k for k, v in metric_call.items() if v == metric]
+            for _ in metric_name:
+                callable_metrics[_] = metric
+
```
ankurankan, line 50: If it is a callable, shouldn't we simply return it? It could, for example, be a custom validation function.

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -906,6 +907,92 @@ def tearDown(self):
...
+                    self.assertIn(name, result["METRIC"].values)
+
+                self.assertEqual(result.shape[0], len(expected_rows))
+
+    def test_output_datatypes(self):
+        result = self.model.validate(data=self.data, compute_rmsea=True)
+        expected_dtypes = {
+            "Correlation": float,
+            "Log-likelihood": float,
+            "Aic": float,
```
ankurankan: `Aic` -> `AIC`

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -906,6 +907,92 @@ def tearDown(self):
...
+            "Bic": float,
```
ankurankan: `Bic` -> `BIC`

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -906,6 +907,92 @@ def tearDown(self):
         del self.graph
 
 
+class TestDAGValidation(unittest.TestCase):
+    def setUp(self):
+        self.model = get_example_model("cancer")
+        self.data = self.model.simulate(int(1e3))
```
ankurankan: Can you pass a `seed` argument to the `simulate` method?

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -906,6 +907,92 @@ def tearDown(self):
...
+            "Correlation",
+            "Log-likelihood",
+            "Aic",
+            "Bic",
+            "Fisher-C p-value",
+            "RMSEA",
+            "Failing CIs / Total CIs",
+        ]
+        for metric_name in expected_metric_names:
+            self.assertIn(metric_name, result["METRIC"].values)
```
ankurankan: Can compare sets instead of the loop. Something like:
```python
self.assertEqual(set(expected_metric_names), set(result["METRIC"]))
```

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+        df_result: pandas.Dataframe
+            A dataframe of all the metric results run on the given causal model
+            using the dataset provided.
+        """
+        # checking if data provided is valid
+        if not isinstance(data, pd.DataFrame) or data is None:
+            raise ValueError(
+                f"data must be a pandas.DataFrame instance. Got {type(data)}"
+            )
+
```
ankurankan, line 1844: We will also need to add a check here that the data columns match the variable names of the model.

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+        from pgmpy.estimators.CITests import get_callable_ci_test
+        from pgmpy.metrics.metrics import get_metrics
+        from pgmpy.utils import get_dataset_type
+
+        # List of method calls generated according to the metric list provided
+        callable_metrics = get_metrics(metrics=metrics)
+
+        # Getting the CI Test method
+        kwargs["ci_test"] = get_callable_ci_test(test=None, data=data)
```
ankurankan: Doesn't `test=None` ignore the `ci_test` passed by the user?

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+
+        # To store final results of the tests
+        metric_vals = {}
+
+        for n, c in callable_metrics.items():
```
ankurankan: Would suggest using more descriptive names here, for example: `n` -> `name` and `c` -> `metric_fn`

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+            sig = inspect.signature(c)
+            valid_params = sig.parameters.keys()
+            filtered_kwargs = {k: v for k, v in kwargs.items() if k in valid_params}
```
ankurankan: Do we need to do filtering here? I think if we pass the full `kwargs` dict, the method would automatically only use the ones that it expects.

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+        from pgmpy.utils import get_dataset_type
+
+        # List of method calls generated according to the metric list provided
+        callable_metrics = get_metrics(metrics=metrics)
+
+        # Getting the CI Test method
+        kwargs["ci_test"] = get_callable_ci_test(test=None, data=data)
+
+        # To store final results of the tests
+        metric_vals = {}
```
ankurankan: I would suggest defining this as a dataframe already, that you add values to in the loop. It will simplify the implementation.

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+
+        **kwargs:
+            Any parameter that needs to be customized in the call to different metric based methods.
+            This includes all parameters that affect the metric calls in pgmpy.metrics.metrics.
+
+        Returns
+        ----------
+        df_result: pandas.Dataframe
+            A dataframe of all the metric results run on the given causal model
+            using the dataset provided.
```
ankurankan: Need to add an examples section.

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+        - Fisher-C p-value
+        - RMSEA based on the Fisher-C statistic
+        - Measure of failing vs. total CIs based on DAG and data fit
+
+        Parameters
+        ----------
+        data: pandas.Dataframe
+            Dataset to be used to run the scoring methods/tests
+
+        metrics: tuple (Callable or strings)
```
ankurankan, line 1776: This currently doesn't say what values it accepts.

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+
+            if n in ["aic", "bic"]:
+                var_type = get_dataset_type(data)
+                if var_type == "continuous":
+                    suffix = "g"
+                elif var_type == "discrete":
+                    suffix = "d"
+                else:
+                    suffix = "cg"
+                filtered_kwargs["scoring_method"] = f"{n}-" + suffix
```
ankurankan: This is a preprocessing step, and should be at the beginning of the method. Try to have all the preprocessing steps out of the core logic of the method. This helps a lot in the readability of the code. You can also add comments with `Steps` to describe the overall logic in the method, have a look at some other classes, such as `PC` or `ExpertInLoop` for examples.

pgmpy/base/DAG.py
```diff
@@ -1828,6 +1829,100 @@ def edge_strength(self, data, edges=None):
...
+                else:
+                    metric_vals["Fisher-C p-value"] = result
+            elif n == "implied-cis":
+                metric_vals["Failing CIs / Total CIs"] = (
+                    f"{(result["p-value"] < significance_level).sum()} / {len(result)}"
+                )
+
+        df_result = pd.DataFrame(
+            {"METRIC": list(metric_vals.keys()), "RESULT": list(metric_vals.values())}
+        )
```
ankurankan: I would suggest making the metric names as the index of the dataframe rather than being a column.

pyproject.toml
```diff
@@ -64,7 +64,7 @@ optional = [
 all = [
   "networkx>=3.0",
   "numpy>=2.0",
-  "scipy>=1.10",
+  "scipy<1.16",
```
ankurankan: I think this needs to be reverted.

ankurankan (changes requested):

pgmpy/metrics/metrics.py
ankurankan: @D03xen For this, I think a better idea could be to modify the `get_scoring_method` itself such that if only `aic` or `bic` is specified, based on the dataset, it returns the correct version of them. This would help in keeping the full logic for scoring method selection in the `get_scoring_method` rather than in two places in the codebase.

pgmpy/metrics/metrics.py
D03xen: My main objective with this was when I am getting the dictionary of Callables from `get_metrics` each of the function call should be paired up with the key value as the corresponding metric name (to make it easier for the final dataframe creation in `validate`). That is also why I am using `metric_call` dictionary. So when I am getting a list of Callables I am first creating that key-value pair in `callable_metrics` dictionary which I am then passing to `DAG.validate()`. 

I mean the current validate method works with the already defined metrics in the metrics.py file. So we would have a choice among those when using the validate method. Could you clarify what you are meaning by custom validation function?

pgmpy/base/DAG.py
D03xen: Will update this

pgmpy/base/DAG.py
D03xen: Yes, I will change this

pgmpy/base/DAG.py
D03xen: The issue is functions `correlation_score`, `fisher_c`, `log_likelihood_score` and `implied_cis` have fixed set of parameters in their function definitions. I did not want to initially change these function definitions to adjust with **kwargs. This is why a TypeError is being shown if I am not filtering due to extra arguments. 

I can change these function definitions but this again causes issues in the function `structure_score` specifically which is using AIC and BIC classes from `pgmpy/estimators/StructureScore.py` and the extra arguments in **kwargs cause TypeError. 

I feel argument filtering should prevent refactoring all of these methods just for validate to work and would rather help if any new metrics are defined in `metrics.py`. What do you suggest?

ankurankan: Got it. I think we can pass filtered arguments for now. I do think we should refactor our metrics at some point to make the signatures more uniform, add some tags, and a registry to simplify some of these implementations, and make metrics easily extensible.

pgmpy/metrics/metrics.py
ankurankan: I meant the case where the user defines their own function that computes some metric. But I think we can ignore that for now.

D03xen:
@ankurankan I have added a draft PR #2373 adding the PermutationTest file back and a helper DAG method

ankurankan:
@D03xen I have added a few changes to this PR. Can you please review it?

D03xen:
> @D03xen I have added a few changes to this PR. Can you please review it?

Looks good to me.

ankurankan (approved):

PR #2183, opened 2025-06-04
[FIX] use DiscreteFactor objects as factor nodes in `to_factor_graph()`

### LINKED ISSUE #1000: Issue in pgmpy/models/MarkovModel.py
Issue in reference to pull request   #998 . 

https://github.com/pgmpy/pgmpy/blob/f9be1709328551450ad6a782b0f49b49d8886a93/pgmpy/models/MarkovModel.py#L265-L296

In line 295, shouldn't it be factor instead of factor_node? Shouldn't the factor node be of type  <class 'pgmpy.factors.discrete.DiscreteFactor.DiscreteFactor'> ? Now on converting Markov Model to factor graph using the above function, factor nodes are of type <class 'str'>.

This issue is rasing a Value error in check_model() function whenever we try to get the factor nodes from factor graph after converting Markov to factor graph. 

https://github.com/pgmpy/pgmpy/blob/f9be1709328551450ad6a782b0f49b49d8886a93/pgmpy/models/FactorGraph.py#L190-L208

Line 207 will always give a Value error since it's  always true because after conversion, factor_node is of type <class 'str> and not of type   <class 'pgmpy.factors.discrete.DiscreteFactor.DiscreteFactor'>

Changing factor_node to factor in line 295 might fix the issue.

Thank  you.

samadpls:

- Fixes #1000

- Fixed `DiscreteMarkovNetwork.to_factor_graph()` to use `DiscreteFactor` objects as factor nodes instead of string identifiers
- Updated `test_factor_graph()` test method to verify factor nodes are `DiscreteFactor` instances rather than strings
- Removed string-based factor node creation (`factor_node = "phi_" + "_".join(scope)`) and directly use factor objects as nodes


samadpls:
Hi @ankurankan, @Nimish-4  just checking in—let me know if there’s anything to update on this PR. Would appreciate your review when you get a chance!

PR #2186, opened 2025-06-05
[BUG]fix to Inconsistent interpretation of std parameter in LinearGaussian…

### LINKED ISSUE #2167: [BUG] Inconsistent interpretation of std parameter in LinearGaussianCPD between implementation and tests
**Describe the bug**

There appears to be a mismatch between the interpretation of the `std` parameter in the `LinearGaussianCPD` implementation and how it is used in the tests. Although the parameter is named `std` (suggesting standard deviation), the test code seems to treat the values as variances.

**To Reproduce**

In the `setUp` method of the tests, we define the CPDs like this:

```python
def setUp(self):
    self.model = LinearGaussianBayesianNetwork([("x1", "x2"), ("x2", "x3")])
    self.cpd1 = LinearGaussianCPD(variable="x1", beta=[1], std=4)
    self.cpd2 = LinearGaussianCPD(
        variable="x2", beta=[-5, 0.5], std=4, evidence=["x1"]
    )
    self.cpd3 = LinearGaussianCPD(
        variable="x3", beta=[4, -1], std=3, evidence=["x2"]
    )
```

Here, we specify the standard deviations directly.
However, in the `test_simulate` method:
```python
def test_simulate(self):
    self.model.add_cpds(self.cpd1, self.cpd2, self.cpd3)
    df_cont = self.model.simulate(n_samples=10000, seed=42)

    # Same model in terms of equations
    rng = np.random.default_rng(seed=42)
    x1 = 1 + rng.normal(0, 2, 10000)
    x2 = -5 + 0.5 * x1 + rng.normal(0, 2, 10000)
    x3 = 4 + -1 * x2 + rng.normal(0, np.sqrt(3), 10000)
    df_equ = pd.DataFrame({"x1": x1, "x2": x2, "x3": x3})

```
We are passing values like `2` and `np.sqrt(3)` as the standard deviations to `rng.normal`, which expects the standard deviation as input.

This suggests that the `std` parameter passed to `LinearGaussianCPD` might actually be interpreted as variance in the implementation, even though it's named `std`. It would be great if you could clarify whether this is intentional or a mismatch between the code and the tests.

**Expected behavior**

If the `std` parameter is truly standard deviation, the test sampling should use the same values directly. If the CPD implementation actually uses the parameter as variance, either the parameter should be renamed (to `var`) or the documentation and tests should reflect that clearly.

Imvedansh:
Pull Request Summary
Fixes CPD std parameter mismatch between implementation and tests

Checklist
 Pulling into dev from a topic branch.

 Commit messages follow repo guidelines.

Issue Fixed
Fixes #2167 

Changes in This PR
Clarified the use of std in LinearGaussianCPD.

Updated test setup to use correct standard deviations instead of variances.

Ensured consistency between test logic and implementation.

Context
There was a mismatch in interpretation of the std parameter in LinearGaussianCPD. While the parameter name suggests it’s a standard deviation, the test was using variance values instead (e.g., std=4 → np.std=2). This PR aligns test cases and implementation under the assumption that std represents the standard deviation.

### DIFF
--- a/pgmpy/factors/continuous/LinearGaussianCPD.py
+++ b/pgmpy/factors/continuous/LinearGaussianCPD.py
@@ -66,19 +66,24 @@ class LinearGaussianCPD(BaseFactor):
     [0.2, -2, 3, 7]
     """
 
-    def __init__(self, variable, beta, std, evidence=[]):
+    def __init__(self, variable, beta, variance, evidence=[]):
         try:
             hash(variable)
         except TypeError:
             raise ValueError(
                 f"`variable` argument must be hashable, Got {type(variable).__name__}"
-            )
-
+            )    
+            
         self.variable = variable
         self.beta = np.array(beta)
-        self.std = std
+        self.variance = variance
         self.evidence = list(evidence)
         self.variables = [variable] + evidence
+
+    @property
+    def std(self):
+        """Standard deviation (derived from variance)"""
+        return np.sqrt(self.variance)
 
     def copy(self):
         """
@@ -101,7 +106,7 @@ class LinearGaussianCPD(BaseFactor):
         copy_cpd = LinearGaussianCPD(
             variable=self.variable,
             beta=self.beta,
-            std=self.std,
+            variance=self.variance,
             evidence=list(self.evidence),
         )
 
@@ -109,7 +114,7 @@ class LinearGaussianCPD(BaseFactor):
 
     def __str__(self):
         mean = self.beta.round(3)
-        std = round(self.std, 3)
+        std = round(np.sqrt(self.variance), 3)
         if self.evidence and list(self.beta):
             # P(Y| X1, X2, X3) = N(-2*X1_mu + 3*X2_mu + 7*X3_mu; 0.2)
             rep_str = "P({node} | {parents}) = N({mu} + {b_0}; {sigma})".format(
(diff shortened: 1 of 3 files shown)


ankurankan:
@Imvedansh I don't think your PR addresses the issue. We don't want to change the argument from std to variance. It should remain as `std`. The problem is that it is getting interpreted as variance somewhere (probably in `to_joint_gaussian`). You will need to:

1. Go through the current tests to make sure that they are correct. Look at @Sanchay117 description in #2167. 
2. Then figure out a fix for it.

De-funkd:
can i try it??? if @Imvedansh has no issues with it

Imvedansh:
@De-funkd I didn't got the time to work on this,sure you can try it out else by this weekend I ll raise the fixed PR

gdreiman1:
Hi I just ran across this, there's an error in [line 257 of LinearGaussianBayesianNetwork](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/LinearGaussianBayesianNetwork.py#L257). The variance matrix (omega) is being populated with standard deviations instead of variances.

```
omega[var_to_index[var], var_to_index[var]] = cpd.std
```
It is a very simple fix:

```
omega[var_to_index[var], var_to_index[var]] = (cpd.std)**2
```

I'm happy to make the fix, and I think that the other changes in this PR would be unnecessary with that fix. Should I just start a new PR for clarity? @Imvedansh @De-funkd @ankurankan

Imvedansh:
> Hi I just ran across this, there's an error in [line 257 of LinearGaussianBayesianNetwork](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/models/LinearGaussianBayesianNetwork.py#L257). The variance matrix (omega) is being populated with standard deviations instead of variances.
> 
> ```
> omega[var_to_index[var], var_to_index[var]] = cpd.std
> ```
> It is a very simple fix:
> 
> ```
> omega[var_to_index[var], var_to_index[var]] = (cpd.std)**2
> ```
> 
> I'm happy to make the fix, and I think that the other changes in this PR would be unnecessary with that fix. Should I just start a new PR for clarity? @Imvedansh @De-funkd @ankurankan
> 
> 
>  
> 

Go ahead!

ankurankan:
@gdreiman1 That would be great. Please feel free to reach out on our Discord if you face any issues in creating the PR. Thanks!!

I am also closing this PR.

PR #2189, opened 2025-06-05
Add net file format to DiscreteBayesianNetwork

### LINKED ISSUE #1618: Add net file format to BayesianNetwork.load and save methods
Net file format support was added in https://github.com/pgmpy/pgmpy/pull/1613, but the option is not available to BayesianNetwork.load and save methods.

### LINKED ISSUE #2187: [BUG] NetWriter accidentally includes truncated/ellipisis in model file
**Describe the bug**
NetWriter fails to save big models, due to a bug during serialization. Values are converted to `str` from `numpy.arrays`, but they get truncated with ellipses (`...`).

**To Reproduce**
1. Use `hailfinder` large model as a test and BayesianDiscreteNetwork as model.
2. Save model to disk
3. Load model

```python
test_model_large = get_example_model("hailfinder")
model.save("model.net")
model.load("model.net")

ValueError: could not convert string to float: '...'
```

**Expected behavior**
Not raised an error and correctly load back the model.

**Versions**
latest

</details>

arocapedro:

- Fixes #1618
- Fixes #2187 

- Refactor `load` and `save` in `DiscreteBayesianNetworks` to avoid code duplication.
- Add NetReader and NetWriter
- Update tests to include "net" file format
- Increase number of decimals serialized in NetWriter to 8 (from 4), since `DiscreteFactor.__eq__` uses atol=1e-08, so the test would fail.


pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1481,41 +1482,36 @@ def save(self, filename, filetype="bif"):
...
-            writer = BIFWriter(self)
-            writer.write_bif(filename=filename)
-
-        elif filetype == "uai":
-            from pgmpy.readwrite import UAIWriter
-
-            writer = UAIWriter(self)
-            writer.write_uai(filename=filename)
+        supported_formats_writer_map = {
+            "bif": ("pgmpy.readwrite", "BIFWriter", "write_bif"),
```
ankurankan: Should we also rename the write methods of these classes to just `write` to make it uniform? I think it would simplify the implementation here as well.

ankurankan (commented):
@arocapedro Thanks for the PR. I have added a comment on renaming the `write_*` methods. Maybe we can include that as well in this PR?

PR #2192, opened 2025-06-05
Enhance and improve zero-count handling in MLE

DarshanCode2005:
Issue Description

The current MaximumLikelihoodEstimator has a problematic approach to handling zero-count states in the estimate_cpd method. The current implementation unconditionally fills zero-count columns with uniform values (1.0), which:

1. Lacks theoretical justification: Pure MLE should handle zero counts differently
2. Provides no user control: Users cannot specify their preferred zero-count handling strategy
3. Is inconsistent with best practices: Modern implementations offer smoothing options
4. May cause unexpected behavior: Silent uniform filling can lead to misleading results

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -186,3 +186,7 @@ cython_debug/
 #  refer to https://docs.cursor.com/context/ignore-files
 .cursorignore
 .cursorindexingignore
+
+myenv/
+
+test_em_missing.ipynb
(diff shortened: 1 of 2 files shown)


ankurankan:
@DarshanCode2005 I don't think this is correct. Please try to discuss the issue before spending time implementing it.

Issue #2140, opened 2025-05-28
[DOC] Build a logic flow diagram of the PC algorithm implementation


ankurankan:
The logic flow of PC algorithm has become a bit complicated with the addition of arguments like `enforce_expert_knowledge`, and it is usually not obvious what would be prioritized in the final output. I think it would be a good idea to create a flowchart/logic flow of how these arguments control the flow and prioritization in the algorithm.

@Nimish-4 Since you were already tinkering with the algorithm, would you mind taking this up?

Nimish-4:
Yep, I'll get on with it.

DebjyotiRay:
Hi @ankurankan !
Is this issue still open for contribution?

Nimish-4:
@DebjyotiRay I have made the change already, so this issue doesn't really need more effort.

ankurankan:
Ref #2116

PR #2193, opened 2025-06-05
Add .codecov.yml to improve coverage reports

nitishmalang:
PR adds a `.codecov.yml` to set an 80% coverage goal (with a small buffer), make PR comments cleaner by only showing up when coverage changes, and wait for all CI jobs to wrap up. It’s just a config file, so it shouldn’t mess with tests or CI.

**What’s in the PR**:
- Added `.codecov.yml` with:
  - 80% coverage targets for the project and PR changes (2% flexibility).
  - PR comment layout with reach, diff, flags, and files for clarity.
  - Comments only when coverage shifts (to cut down on spam).
  - Syncs with CI’s three platforms (Ubuntu, macOS, Windows).

**Checks**:
- I went through Codecov’s docs to make sure the YAML is solid.
- Since it’s a config file, it should breeze through CI (like #2140).
- Happy to tweak anything, like the coverage percentage or comment setup.

Would love to hear what you think! Maybe we could add a quick note about this in `Contributing.md`? Thanks for the chance to contribute and for any guidance!

Best,  
Nitish  
CC: @ankurankan

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,11 +2,11 @@ coverage:
   status:
     project:
       default:
-        target: 80%
-        threshold: 2%
+        target: 90%
+        threshold: 1%
     patch:
       default:
-        target: 80%
+        target: 90%
         threshold: 2%
 comment:
   layout: "reach, diff, flags, files"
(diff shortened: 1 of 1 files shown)


ankurankan:
@nitishmalang Thanks for the PR. Shouldn't we use a much higher target? The current codebase codecov is at 95%

PR #2197, opened 2025-06-06
Fix doctests for estimator submodule files

### LINKED ISSUE #757: Use keyword arguments everywhere in doctests
Using keyword arguments helps in understanding the code.

Vanshitaaa20:
## Your checklist for this pull request

- Fixes #757 and #581 

- Fixed and updated doctests for the following estimator sub-module files : 
        pgmpy/estimators/BayesianEstimator.py
        pgmpy/estimators/CITests.py
        pgmpy/estimators/EM.py
        pgmpy/estimators/ExhaustiveSearch.py
        pgmpy/estimators/ExpertKnowledge.py
        pgmpy/estimators/GES.py
        pgmpy/estimators/base.py
        pgmpy/estimators/MLE.py
 
- Ensured all tests pass with consistent results.
- Replaced positional arguments with keyword arguments for clarity and readability in doctests.
- Skipped memory addresses and adjusted set order outputs to prevent test inconsistencies across environments.


pgmpy/estimators/BayesianEstimator.py
```diff
@@ -97,7 +97,7 @@ def get_parameters(
         ...                       columns=['A', 'B', 'C', 'D'])
         >>> model = DiscreteBayesianNetwork([('A', 'B'), ('C', 'B'), ('C', 'D')])
         >>> estimator = BayesianEstimator(model, values)
-        >>> estimator.get_parameters(prior_type='BDeu', equivalent_sample_size=5)
+        >>> estimator.get_parameters(prior_type='BDeu', equivalent_sample_size=5) # doctest: +SKIP
```
ankurankan: Does the comment specify this line to be skipped?

pgmpy/estimators/ExhaustiveSearch.py
```diff
@@ -115,36 +110,36 @@ def all_scores(self):
         >>> import numpy as np
         >>> from pgmpy.estimators import ExhaustiveSearch, K2
         >>> # create random data sample with 3 variables, where B and C are identical:
-        >>> data = pd.DataFrame(np.random.randint(0, 5, size=(5000, 2)), columns=list('AB'))
+        >>> data = pd.DataFrame(np.random.randint(low=0, high=5, size=(5000, 2)), columns=list('AB'))
```
ankurankan, line 116: I think we will need to specify a seed here, otherwise the results would be different every time.

pgmpy/estimators/BayesianEstimator.py
ankurankan: Why does this need to be skipped?

pgmpy/estimators/EM.py
```diff
@@ -233,7 +233,7 @@ def get_parameters(
         ...                       columns=['A', 'C', 'D'])
         >>> model = DiscreteBayesianNetwork([('A', 'B'), ('C', 'B'), ('C', 'D')], latents={'B'})
         >>> estimator = EM(model, data)
-        >>> estimator.get_parameters(latent_card={'B': 3})
+        >>> estimator.get_parameters(latent_card={'B': 3}) # doctest: +SKIP
```
ankurankan: Why do we need to skip this?

pgmpy/estimators/GES.py
```diff
@@ -145,11 +145,11 @@ def estimate(
 
         >>> # Learn the model structure using GES algorithm from `df`
         >>> from pgmpy.estimators import GES
-        >>> est = GES(data)
+        >>> est = GES(df)
         >>> dag = est.estimate(scoring_method='bic-d')
         >>> len(dag.nodes())
         37
-        >>> len(dag.edges())
+        >>> len(dag.edges()) # doctest: +SKIP
```
ankurankan: Same here?

ankurankan (changes requested):

pgmpy/estimators/BayesianEstimator.py
Vanshitaaa20: > Why does this need to be skipped?

Hey, along with the memory addresses, parts like <TabularCPD representing P(B:2 | C:2, A:2)>  also change slightly ...for example, the parent variable order (C:2, A:2) isn’t guaranteed to be consistent due to internal dictionary/set ordering which change every time the code runs or on different systems. So even if the logic is correct, the doctest fails because of these non-deterministic addresses.

That’s why I used # doctest: +SKIP to avoid false negatives from such non-deterministic outputs, even though the actual values are correct. I can add sorting or use +ELLIPSIS if that would be a better fix!

pgmpy/estimators/ExhaustiveSearch.py
Vanshitaaa20: > I think we will need to specify a seed here, otherwise the results would be different every time.

Yeah, I had actually tried setting a seed using model.seed = 42, and it did make the outputs consistent across runs. I removed it initially thinking it might interfere with the logic or learning process, but looks like using it is the cleaner way to ensure reproducibility in doctests. I’ll go ahead and add it back!

pgmpy/estimators/BayesianEstimator.py
ankurankan: Got it. Yes, using ellipsis might be better. It will at least check for the type of object getting returned.

pgmpy/estimators/GES.py
ankurankan: Does this also learn different number of edges everytime?

Vanshitaaa20: > Does this also learn different number of edges everytime?

Yes. It was varying across runs due to randomness in data generation, so I added `np.random.seed(42)` to ensure consistent results and avoid flaky test results.

Vanshitaaa20:
@ankurankan Also, while working on the doctest for TreeSearch.py with estimator_type='tan', it requires a class_node argument. I wasn’t sure which node would make the most sense to use for this in the example. Is there a preferred choice for the class_node here, or should I just pick any one from the dataset? Keeping both the same raises a ValueError.

ankurankan (approved):

PR #2198, opened 2025-06-07
Fix: Resolve key error in index_to_weight mapping for torch backend

### LINKED ISSUE #1825: Key Error when log_likelihood_score is called when using "torch" backend
### Subject of the issue
Key Error is thrown in the "_log_probability_node"  method, specifically line 80 in " [index_to_weight[state_to_index[tuple(u)]] for u in unique]" , when the log_likelihood_score is called when using the "torch" backend. The specifc key error is "KeyError: tensor(3, device='cuda:0')" 

### Your environment
* pgmpy version : latest dev branch
* Python version: 3.12.0
* Operating System: Windows 11

### Steps to reproduce
* Set the backend to "torch"
* Create a Bayesian Network.  ( I used structure learning and parameter estimation)
* evaluate the model by calling "pgmpy.metrics.log_likelihood_score"
* example: log_likelihood = log_likelihood_score(model, validation_data)

### Expected behaviour
The loglikelihood score should be returned.

### Actual behaviour
A Key Error is raised being: "KeyError: tensor(3, device='cuda:0')". 
Possible problem being the datatype of tensor and needing to use tensor.item() to obtain the value within the tensor instead of the tensor itself.

HarshSaini10:

- Fixes #1825 

- Fixed a KeyError in bn_inference.py::_log_probability_node that occurred when the backend was set to "torch". 
   Cause: The Error was caused due to the index_to_weight dictionary to receive tensors as key.
   Fix: Added .item() to recast the tensors to Python integer
- Added a new test in TestLogLikelihoodScore to validate log-likelihood computation when backend is "torch".
- Identified and fixed a separate einsum bug in sampling/base.py::_reduce_marg, which was surfaced by the new test.


pgmpy/tests/test_metrics/test_metrics.py
```diff
@@ -17,6 +17,7 @@
 )
 from pgmpy.models import DiscreteBayesianNetwork
 from pgmpy.utils import get_example_model
+from pgmpy import config
```
ankurankan, line 20: @HarshSaini10 Can you create separate test classes for torch backend. For examples, have a look at the `test_MaximumLikelihoodEstimator.py` file.

pgmpy/sampling/base.py
```diff
@@ -92,7 +95,19 @@ def _reduce_marg(variable_cpd, reduce_index, sc):
...
+        # Generate einsum string
+        ndim = reduced_values.ndim
+        input_axes = string.ascii_lowercase[:ndim]
+        output_axes = input_axes[0]
+        equation = f"{input_axes}->{output_axes}"
+
+        # Perform marginalization using einsum
+        if config.get_backend() == "torch":
+            reduced_values = torch.tensor(
+                reduced_values, dtype=config.get_dtype(), device=config.get_device()
```
ankurankan: Why is `reduced_values` not already a tensor? If the backend is torch, this should already be a tensor object, if that is not the case, we will need to find the source of this issue.

ankurankan (changes requested):

pgmpy/tests/test_metrics/test_metrics.py
HarshSaini10: Sure, I created a TestLogLikelihoodScoreTorch class for the torch backend, similar to what the test_MaximumLikelihoodEstimator.py file

pgmpy/sampling/base.py
HarshSaini10: Realized that it was due to my fault of switching backends within the function, creating a separate resolved that issue so I removed the unnecessary conversion, apart from that the einsum function call change remains intact as it ensures that it works with the torch backend as well (the previous function call was only supported with numpy backend).

pgmpy/sampling/base.py
```diff
@@ -92,7 +95,16 @@ def _reduce_marg(variable_cpd, reduce_index, sc):
             slice_[index] = sc[i]
 
         reduced_values = variable_cpd.values[tuple(slice_)]
-        marg_values = compat_fns.einsum(reduced_values, range(reduced_values.ndim), [0])
+
+        # Generate einsum string
+        ndim = reduced_values.ndim
+        input_axes = string.ascii_lowercase[:ndim]
+        output_axes = input_axes[0]
+        equation = f"{input_axes}->{output_axes}"
```
ankurankan: This is a work around only for pytorch. Correct? If yes, I think this should be in `compat_fns.py`. 

Another possibility to simplify is to use the einsum operation from `opt_einsum`, that should be agnositc to the type of `reduced_values` and should also work with integer axis values.

pgmpy/tests/test_metrics/test_metrics.py
```diff
@@ -118,6 +119,22 @@ def test_input(self):
...
+class TestLogLikelihoodScoreTorch(TestLogLikelihoodScore):
+    def setUp(self):
+        self.original_backend = config.get_backend()
+        config.set_backend("torch")
+        super().setUp()
+
+    def tearDown(self):
+        config.set_backend(self.original_backend)
+
+    def test_discrete_network(self):
```
ankurankan: We shouldn't need these methods because we are inheriting `TestLogLikelihoodScore`. Right?

ankurankan (changes requested):

pgmpy/sampling/base.py
HarshSaini10: No, it is appropriate for both pytorch and numpy, it's just that the previous way of calling was an outdated version which was working only with numpy, I went with this according to the latest docs for both torch.einsum and np.einsum.

pgmpy/tests/test_metrics/test_metrics.py
HarshSaini10: I felt that explicitly defining them in the sub-class would be better for readability, should i remove them?

ankurankan: I think we can get rid of them.

pgmpy/sampling/base.py
ankurankan: Yes, but it adds extra steps of converting it to the string form. And pytorch is the only one that doesn't support using integer lists. Throughout our codebase, we have used integer lists for einsum, so to keep it uniform, I think it would be better to switch to using `opt_einsum` here.

pgmpy/tests/test_metrics/test_metrics.py
HarshSaini10: ok sure, i'll do that.

pgmpy/sampling/base.py
HarshSaini10: oh yeah that would be a more sensible implementation then, I'll do the required change.

HarshSaini10: Used opt_einsum.contract here instead, is there any other change that should be made?

ankurankan (approved):
Thanks for the PR. Looks good to me.

PR #2201, opened 2025-06-07
Pyro moved to optional dependencies

hillhack:
feat: make pyro an optional dependency; add optional_dependencies.py utility #1983


hillhack:
@ankurankan Sir, can u plz review it once, like as i make the pyro optional so the error message will be appear in failed test cases, same as for xgboost. And i create new  _check_soft_dependencies function instead of doing it with try except block as it  is suggested in the issue.

fkiraly (changes requested):
this looks like copy-paste from `skbase.utils.dependencies` (`scikit-base`), but with any module structure and docstrings removed 

Why would you do this? Is this thrown together by an AI?

PR #2210, opened 2025-06-10
Fix: Prioritize 'weight' in attribute dict in DAG.add_node for NetworkX 2.0 compatibility

DarshanCode2005:
This commit resolves an ambiguity in the `DAG.add_node` method related to handling node weights.

In prior behavior, when both a `weight` keyword argument and a `'weight'` key in the attribute dictionary (as used in NetworkX 2.0-style syntax) were provided, the explicit `weight` parameter would override the dictionary value. This was unintuitive and inconsistent with NetworkX 2.0 expectations.

**Changes made:**

* The `'weight'` from the attribute dictionary is now given priority if both are provided.
* A warning is issued when both values are present, indicating that the dictionary-specified `'weight'` is being used.
* All attributes from the dictionary are preserved and passed to the superclass's `add_node` method to maintain full compatibility with NetworkX 2.0 conventions.

This update ensures more predictable behavior and better alignment with modern NetworkX usage patterns.

### DIFF
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -171,10 +171,15 @@ class DAG(nx.DiGraph):
         Parameters
         ----------
         node: str, int, or any hashable python object.
-            The node to add to the graph.
+            The node to add to the graph. Can also be a 2-tuple `(node_name, attr_dict)`
+            as in NetworkX 2.0 syntax, where `attr_dict` is a dictionary of attributes.
 
-        weight: int, float
+        weight: int, float, optional (default=None)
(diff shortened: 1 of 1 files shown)


DarshanCode2005:
@ankurankan sir could you please review the PR and let me know what do I need to improve?

ankurankan:
@DarshanCode2005 Sorry for the super late reply. This PR somehow got lost in my emails. Can you give me some concrete code examples here or a link to the change in networkx behavior?

ankurankan:
Closing due to inactivity

PR #2215, opened 2025-06-11
Fixes GES implementation

### LINKED ISSUE #2015: GES implementation differs from algorithm in literature
In the literature, the GES algorithm is a greedy search algorithm over equivalence classes of DAGs. In the implementation in this package (https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/GES.py), however, it is implemented as a DAG space hill climber. 

To see the difference, assume one has the equivalence class consisting of DAGs $A \rightarrow B \quad C$ and $A \leftarrow B \quad C$ which may be represented by CPDAG $A - B \quad  C$. Then the neighboring equivalence classes considered by the Insert operator would, e.g., include $A \rightarrow B \leftarrow C$. (Basically, Chickering considers all equivalence classes as neighbors of which a DAG can be reached by adding a single edge to some DAG in the current equivalence class). 

Instead working in DAG space doesn't give the same algorithm because when being at DAG $A \leftarrow B \quad  C$, it is not possible to reach DAG $A \rightarrow B \leftarrow C$ in a single step, even though the corresponding equivalence classes are neighbors according to Chickerings definitions. 

See Def. 12 and 13 in Chickering 2002 for the definition of the Insert and Delete operators (https://www.jmlr.org/papers/volume3/chickering02b/chickering02b.pdf). 

Also see other implementations of GES like https://github.com/juangamella/ges/blob/ad29d0bdc7d186d29b65d43c6b2e180b9c03dc53/ges/main.py, https://github.com/cran/pcalg/blob/master/src/greedy.cpp or https://github.com/mschauer/CausalInference.jl/blob/master/src/ges.jl that use CPDAGs/essential graphs.

Nimish-4:
- Fixes #2015 

- Fixed the implementation according to the CPDAG space algorithm
- The optimizations given in the [original paper](https://www.jmlr.org/papers/volume3/chickering02b/chickering02b.pdf) were referenced from - https://github.com/juangamella/ges


Nimish-4:
@ankurankan Waiting for your review when you get time. I think the changes should be split into two PRs - one for the PDAG class (which had plenty of issues), and then one for GES (for only the forward and backward phase, to start). There's scope for some cleanup and improvement, I'll do it in the actual PR.

PR #2243, opened 2025-06-27
ESOC : ADMG Implementation

Spinachboul:
@ankurankan 
Working on implementation of ADMGS (Acyclic Directed Mixed Graphs)
- Methods not implemented (skipped / work on latere)
1) `m-separation`
2) `get_markov_blanket`

TODO : The **tests** will be added soon for the class.

I would request a review on the methods implementation for this PR.
Thanks!


pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,240 @@
...
+        ----------
+        directed_ebunch : list of tuples
+            List of directed edges in the form (source, target).
+        bidirected_ebunch : list of tuples
+            List of bidirected edges in the form (node1, node2).
+        latents : list, optional
+            List of latent variables (nodes) in the graph.
+        """
+        super().__init__()
+        self.bi_directed_edges = collections.defaultdict(set)
```
ankurankan: Should keep `bi_directed` and `bidirected` consistent. As `bidirected` is word, I would suggest to change this to `self.bidirected_edges`

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,240 @@
...
+
+        if directed_ebunch:
+            self.add_directed_edges(directed_ebunch)
+        if bidirected_ebunch:
+            self.add_bidirected_edges(bidirected_ebunch)
+
+    def add_node(self, node):
+        super().add_node(node)
+
+    def add_nodes_from(self, nodes_for_adding, **attr):
```
ankurankan: Would suggest keeping the arguemnt name consistent with other classes. `nodes_for_adding` -> `nodes`

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,240 @@
...
+        """add multiple nodes to the graph."""
+        return super().add_nodes_from(nodes_for_adding, **attr)
+
+    def add_directed_edge(self, u, v):
+        """
+        Adds a directed edge from node u to node v.
+        Error raised if the edge additions would create a cycle in the graph.
+        """
+        if u not in self.nodes or v not in self.nodes:
```
ankurankan: We shouldn't fail for this. The default behavior in other classes is to add the nodes if they are not already present in the graph. It acts as a shorthand to define models using just the edge list. This is also networkx's default behavior.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,240 @@
...
+            raise ValueError("Both nodes must be present in the graph.")
+
+        # Temporarily add edge using networkx's add_edge method
+        key = self.add_edge(u, v)
+
+        # Check for cycles using networkx's builtin function
+        if not nx.is_directed_acyclic_graph(self):
```
ankurankan: This would result in an incorrect check. Example: A -> B A <-> C. If we try to add the edge A -> C, networkx will find a directed cycle: A -> B -> C -> A, which is not the correct interpretation of bidirected edge. 

Also, a neater way for checking if adding an edge would create a cycle is to check for a directed path. In the example above, we can check if there is a directed path from C to A. If there is adding the edge A -> C would complete the cycle.

In either of the two approaches, I think we will need to implement our own method for finding directed paths or directed cycles.

ankurankan: I missed that bidirected edges weren't getting added to the network. Is that correct?

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,240 @@
...
+            # If a cycle is detected, remove the edge and raise an error
+            self.remove_edge(u, v, key=key)
+            raise ValueError("Adding this edge would create a cycle in the graph.")
+
+        # If no cycle is detected, then the edge has been added
+
+    def add_directed_edges(self, u, v):
```
ankurankan: Typo: `add_directed_edges` -> `add_bidirected_edges`

ankurankan (commented):
Is there a reason to not add the bidirected edges to the graph and maintain a separate dict for it? While this might work, I think it makes the implementation a bit unintuitive.

pgmpy/base/ADMG.py
Spinachboul: Yeah the cycle detection only was for the directed edges. I am not sure if we even have cycles in bidirected edges.

pgmpy/base/ADMG.py
Spinachboul: Yeah! Fixed that

pgmpy/base/ADMG.py
Spinachboul: Oh yeah! done that, thanks.

Spinachboul:
I have new proposition for finding the ancestors and descendants for ADMG.
Since we also have bidirected edges, can create a temporary DAG graph from it, only containing the directed edges.
I thought this would make more sense since we have existing logic in DAG for getting the ancestors & descendants.

Spinachboul:
> Is there a reason to not add the bidirected edges to the graph and maintain a separate dict for it? While this might work, I think it makes the implementation a bit unintuitive.

Yeah! It does feel unintuitive creating a different structure just for handling the bidirected edgs.
I have this idea of adding on another arg called as "type" which would take values either directed/bidirected.
Changes in the approach would happen specially in the add_bidirected_edges func when we create 2 keys **key_uv** and **key_vu**, they will be considered distinct directed edges from the perspective of networkx

ankurankan:
> Changes in the approach would happen specially in the add_bidirected_edges func when we create 2 keys key_uv and key_vu, they will be considered distinct directed edges from the perspective of networkx

Why don't we represent bidirected edges with two edges? So, for an u <-> v, we would add the edges u -> v and u <- v in the network. We will need to implement some extra functionality to handle to things like paths, but I still think it would be the cleanest solution. What do you think?

Spinachboul:
> > Changes in the approach would happen specially in the add_bidirected_edges func when we create 2 keys key_uv and key_vu, they will be considered distinct directed edges from the perspective of networkx
> 
> Why don't we represent bidirected edges with two edges? So, for an u <-> v, we would add the edges u -> v and u <- v in the network. We will need to implement some extra functionality to handle to things like paths, but I still think it would be the cleanest solution. What do you think?

Yeah exactly, I was hitting on the same point. Also we would need to have a **type** attr which would maintain the semantic meaning of the edges within the ADMG. Otherwise I feel like it would be a simple multigraph of multiple directed edges, which would be spurious and also would affect other methods

Spinachboul:
@ankurankan 
admg_plot

This is how we will visualize the ADMG structure.
I am still working on the ADMG tests, would be updating them soon

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+    def add_bidirected_edges(self, ebunch):
+        for u, v in ebunch:
+            self._add_bidirected_edge(u, v)
+
+    def add_edge(self, u, v, **attr):
+        raise NotImplementedError(
+            "Use add_directed_edge or add_bidirected_edge to add edges."
+        )
+
+    def _get_parent(self, node):
```
ankurankan: I would suggest renaming this to something like `_get_directed_parents` to make it more obvious.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+
+        parents = set()
+        for pred in self.predecessors(node):
+            data = self.get_edge_data(pred, node)
+            for key in data:
+                if data[key].get("type") == "directed":
+                    parents.add(pred)
+        return parents
+
+    def get_parents(self, nodes):
```
ankurankan: I would suggest splitting this into two methods: 1) get_directed_parents, 2) get_bidirected_parents instead of returning the tuple. In ADMG algorithms, we will usually need this information separately, so having two separate methods would make it cleaner.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+        if bidirected_ebunch:
+            self.add_bidirected_edges(bidirected_ebunch)
+
+    def add_node(self, node):
+        super().add_node(node)
+
+    def add_nodes_from(self, nodes, **attr):
+        return super().add_nodes_from(nodes, **attr)
+
+    def _add_directed_edge(self, u, v):
```
ankurankan: Why are `_add_directed_edge` and `_add_bidirected_edge` private?

pgmpy/base/ADMG.py
ankurankan: I would also avoid creating private methods when the private method is used only once. For example `_get_parent` is called only from `get_parents`. In this case it would be better to have the `_get_parent` logic in the `get_parents` method.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+        return markov_blanket
+
+    def to_dag(self):
+        """
+        Converts the ADMG to a directed acyclic graph (DAG).
+        For each bidirected x<->y in the ADMG, add a latent node {x}_{y}
+        and add two edgws x<----{x}_{y}-->y
+        """
+        dag_edges = []
+        dag_nodes = set()
```
ankurankan: We can get nodes directly using something like: `self.nodes()` without going through the loop.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+                    dag_nodes.update(sorted_pair)
+
+        # Create a new DAG instance
+        dag_instance = pgmpy_DAG()
+        dag_instance.add_nodes_from(dag_nodes)
+        dag_instance.add_edges_from(dag_edges)
+
+        return dag_instance
+
+    def _is_d_connected_internal(
```
ankurankan: Same here. This method seems to be only used by `is_m_separated`.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+        nodes_u,
+        nodes_v,
+        conditional_set=None,
+    ):
+        """
+        Checks if two sets of nodes are m_connected given a conditional set.
+        """
+        return not self.is_m_separated(nodes_u, nodes_v, conditional_set)
+
+    def m_connected_nodes(self, nodes_u, nodes_v, conditional_set=None):
```
ankurankan: A simpler way would be to convert the ADMG to the canonical DAG (`to_dag`), call the `active_trail_nodes` and filter out any of the nodes starting with `L_`. The `active_trail_nodes` implements the Bayes' ball algorithm, so it should also be much faster.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+        to a directed acyclic graph (DAG) using the `to_dag` method.
+        """
+        new_dag = self.to_dag()
+
+        if new_dag.is_dconnected(start, end, observed=observed):
+            return True
+        else:
+            return False
+
+    def is_m_separated(
```
ankurankan: Would rename this to `is_mseparated` to keep it consistent with how we have defined d-connected and d-separated.

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+
+        for u in nodes_u_set:
+            for v in nodes_v_set:
+                # if they are d_connected, they must also be m_connected
+                if self._is_d_connected_internal(u, v, observed=conditional_set):
+                    return False
+
+        return True
+
+    def is_m_connected(
```
ankurankan: `is_m_connected` -> `is_mconnected`

pgmpy/base/ADMG.py
ankurankan: `m_connected_nodes` -> `mconnected_nodes`

pgmpy/base/ADMG.py
```diff
@@ -0,0 +1,410 @@
...
+            super().remove_edge(u, v, key=key)
+            raise ValueError("Adding this edge would create a cycle in the graph.")
+
+    def _add_bidirected_edge(self, u, v):
+        if u is None or v is None:
+            raise ValueError("Can't add since one of~ the nodes is None")
+        if u == v:
+            raise ValueError("Cannot add a bidirected edge from a node to itself.")
+
+        if u not in self.nodes:
```
ankurankan: I don't think we need to include this as the `super().add_edge` would automatically add nodes if they do not exist.

ankurankan (changes requested):
@Spinachboul I have added a few comments but broadly looks good. The docstrings still need work and we will need to expand the tests a bit. I will think of potential edge cases to add to the tests.

ankurankan (approved):
Looks good.

PR #2247, opened 2025-07-03
[ENH] Add a log_likelihood method for DynamicBayesianNetwork

### LINKED ISSUE #2228: [ENH] Extend log_likelihood_score function to DynamicBayesianNetwork model
Yes, my feature request is related to a problem raised during a project ([link](https://github.com/petrello/AnomalyDetectionFSAE)) of mine. The project involved utilising dynamic Bayesian networks to develop an anomaly detection system. Due to this aim, I wanted to utilise the `pgmpy.metrics.metrics.log_likelihood_score(model, data)` function.

The problem was such that - even if in the online documentation ([doc_link](https://pgmpy.org/_modules/pgmpy/metrics/metrics.html#log_likelihood_score)) it is specified that the metric is valid for any `pgmpy.base.DAG` - a `ValueError` is raised when I pass as `model` parameter a dynamic Bayesian network. In particular, I can paste here the exact traceback I got:

```
ValueError                                Traceback (most recent call last)
~\AppData\Local\Temp\ipykernel_14288\2474272222.py in <module>
      1 from pgmpy.metrics import log_likelihood_score
      2 
----> 3 res = log_likelihood_score(dbn_model, test_normal_data)
      4 
      5 res

c:\...\...\anaconda3\lib\site-packages\pgmpy\metrics\metrics.py in log_likelihood_score(model, data)
    171     # Step 1: Check the inputs
    172     if not isinstance(model, BayesianNetwork):
--> 173         raise ValueError(f"Only Bayesian Networks are supported. Got {type(model)}.")
    174     elif not isinstance(data, pd.DataFrame):
    175         raise ValueError(f"data must be a pandas.DataFrame instance. Got {type(data)}")

ValueError: Only Bayesian Networks are supported. Got <class 'pgmpy.models.DynamicBayesianNetwork.DynamicBayesianNetwork'>.
```

As a solution to this problem, I was able to implement a custom function able to compute the log-likelihood score given `pgmpy.models.DynamicBayesianNetwork.DynamicBayesianNetwork`. Here is the function:

```python
from typing import Tuple
from tqdm.notebook import tqdm

def compute_log_likelihood(model, data: pd.DataFrame, show_progress: bool = True) -> Tuple[float, np.ndarray]:
        """
        Compute the log-likelihood scores for data given the trained DBN model.
        
        Calculates how likely the observed data is under the current model parameters.
        Lower log-likelihood values indicate less probable observations (potential anomalies).
        
        Args:
            data: DataFrame where each row contains values for all nodes at time t and t+1
            show_progress: Whether to display a progress bar during computation
            
        Returns:
            A tuple containing:
                - Total log-likelihood score (sum of all individual scores)
                - Array of log-likelihood scores for each data point
                
        Raises:
            Exception: If data format is incorrect or values not found in CPD states
        """
        # Get all the nodes in the network
        nodes = list(model.nodes())
        
        # We're only interested in calculating likelihood for nodes at time slice 1
        # since they depend on values in time slice 0
        nodes_t1 = [node for node in nodes if node[1] == 1]
        
        # Ensure data has the correct format
        if isinstance(data, pd.DataFrame):
            data_values = data.values
        else:
            raise Exception("Data must be in DataFrame format")
        
        # Initialize variables for log-likelihood calculation
        total_log_likelihood = 0
        log_likelihood_scores = []
        
        n_samples = len(data_values)
        
        # Create a mapping from node names to CPD objects for quick lookup
        node_to_cpd = {cpd.variable: cpd for cpd in model.get_cpds()}
        
        # Process each data point
        for i in tqdm(range(n_samples), desc="Computing log-likelihood", disable=not show_progress):
            row_log_likelihood = 0
            
            # Calculate log probability for each node at time t+1
            for node in nodes_t1:
                # Get the CPD for this node
                cpd = node_to_cpd.get(node)
                if cpd is None:
                    continue
                    
                # Get the current variable and its value
                current_var = node[0]
                current_val = data[(current_var, 1)].iloc[i]
                
                # Convert value to state number according to CPD
                # Each node has discrete states identified by indices in the CPD
                if node in cpd.state_names:
                    try:
                        current_no = cpd.state_names[node].index(current_val)
                    except (ValueError, IndexError):
                        raise ValueError(f"Value {current_val} not found in state_names for node {node}")
                    
                # Get the evidence variables (parents of the current node)
                evidence = model.get_parents(node)
                evidence_vals = {}
                
                # Collect values for all evidence variables
                for ev_node in evidence:
                    ev_var = ev_node[0]
                    ev_time = ev_node[1]

                    ev_val = data[(ev_var, ev_time)].iloc[i]
                    
                    # Convert to state number
                    if ev_node in cpd.state_names:
                        try:
                            ev_no = cpd.state_names[ev_node].index(ev_val)
                        except (ValueError, IndexError):
                            raise ValueError(f"Value {ev_val} not found in state_names for evidence {ev_node}")
                    
                    evidence_vals[ev_node] = ev_no
                
                # Get probability from CPD
                if evidence:
                    # Create index for evidence
                    evidence_index = []
                    for ev_node in cpd.variables[1:]:
                        if ev_node in evidence_vals:
                            evidence_index.append(evidence_vals[ev_node])
                        else:
                            # Skip if evidence is missing
                            break

                    # Extract probability from the CPD table using the evidence indices
                    prob = cpd.values[current_no][tuple(evidence_index)]
                else:
                    # No evidence, use marginal probability
                    prob = cpd.values[current_no]
                
                # Add log probability to sum (with a small epsilon to avoid log(0))
                epsilon = 1e-10  # Small constant to prevent log(0)
                prob = max(prob, epsilon)
                node_log_prob = np.log(prob)
                row_log_likelihood += node_log_prob
                
            # Add this row's log-likelihood to the total
            total_log_likelihood += row_log_likelihood
            log_likelihood_scores.append(row_log_likelihood)
        
        return total_log_likelihood, np.array(log_likelihood_scores)


# # Example usage:
# from utilities import DBNModel
# 
# # DBNModel is a custom class defined in the utilities package
# dbn_model = DBNModel(intra_slice_edges, inter_slice_edges)
# dbn_model.train(pd_data_train)
# total, scores = dbn_model.compute_log_likelihood(pd_data_test)
```

Now, this could be a baseline for a future solution to integrate into the **pgmpy** library. However, it must be noticed that the above function has been implemented for my specific scenario, meaning the `compute_log_likelihood` function only works for **discrete** dynamic Bayesian networks and within a Jupyter notebook environment.

I ask to authors and contributors if we can think about a possible solution that can work in any scenario and is not affected by any of the aforementioned limitations.

**Additional context**
The custom function I implemented to compute the log-likelihood score is a class method from a custom class `DBNModel` ([link_to_implementation](https://github.com/petrello/AnomalyDetectionFSAE/blob/master/src/utilities/dbn_model.py)), which is essentially a wrapper of  `pgmpy.models.DynamicBayesianNetwork.DynamicBayesianNetwork`.

petrello:

- Fixes #2228

- Adds `log_likelihood(self, data)` method to `DynamicBayesianNetwork` class
- Adds util method `_log_likelihood_node(self, node, sample_data, cpd)` to `DynamicBayesianNetwork` class
- Adds new class of unit tests `TestLogLikelihood`

### DIFF
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -633,7 +633,9 @@ class DynamicBayesianNetwork(DAG):
 
         # Process each data sample
         for i in tqdm(
-            range(n_samples), desc="Computing log-likelihood", disable=not(show_progress and config.SHOW_PROGRESS)
+            range(n_samples),
+            desc="Computing log-likelihood",
+            disable=not(show_progress and config.SHOW_PROGRESS)
         ):
             sample_log_likelihood = 0.0
 
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -140,6 +140,7 @@ class DynamicBayesianNetwork(DAG):
     add_edges
     add_edges_from
     add_cpds
+    log_likelihood
     initialize_initial_state
     inter_slice
     intra_slice
@@ -532,6 +533,9 @@ class DynamicBayesianNetwork(DAG):
                 raise ValueError("CPD defined on variable not in the model", cpd)
 
         self.cpds.extend(cpds)
+
+    def log_likelihood(self, data):
+        pass
 
     def get_cpds(self, node=None, time_slice=None):
         """
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -645,8 +645,8 @@ class DynamicBayesianNetwork(DAG):
 
                 # Compute log-probability for each node in this time slice
                 for node in nodes_t:
-                    node_log_prob = self._compute_node_log_probability(
-                        node, data.iloc[i], node_to_cpd[node], epsilon
+                    node_log_prob = self._log_likelihood_node(
+                        node, data.iloc[i], node_to_cpd[node]
                     )
                     sample_log_likelihood += node_log_prob
 
@@ -654,6 +654,75 @@ class DynamicBayesianNetwork(DAG):
 
         return total_log_likelihood
 
+    def _log_likelihood_node(self, node, sample_data, cpd):
+        """
+        Compute the log-probability of a specific node given the sample data.
+
+        Parameters
+        ----------
+        node : tuple
+            Node identifier in format (variable_name, time_slice)
+        sample_data : pd.Series
+            Single data sample containing values for all variables
+        cpd : pgmpy.factors.discrete.TabularCPD
+            Conditional Probability Distribution for the node
+
+        Returns
+        -------
+        float
+            Log-probability of the node given the evidence
+        """
+
+        # Get the observed value for this node
+        observed_value = sample_data[node]
+
+        # Convert observed value to state index according to CPD
+        try:
+            observed_state_idx = cpd.state_names[node].index(observed_value)
+        except (ValueError, IndexError):
+            raise ValueError(
+                f"Observed value '{observed_value}' for node {node} "
+                f"not found in CPD state names: {cpd.state_names[node]}"
+            )
+
+        # Get parents (evidence variables) for this node
+        parents = self.get_parents(node)
+
+        if not parents:
+            # No parents - use marginal probability
+            prob = cpd.values[observed_state_idx]
+        else:
+            # Has parents - use conditional probability
+            evidence_indices = []
+
+            # Get evidence values for all parents
+            for parent in parents:
+                if parent in sample_data.index:
+                    parent_value = sample_data[parent]
+
+                    # Convert parent value to state index
+                    try:
+                        parent_state_idx = cpd.state_names[parent].index(parent_value)
+                    except (ValueError, IndexError):
+                        raise ValueError(
+                            f"Parent value '{parent_value}' for node {parent} "
+                            f"not found in CPD state names: {cpd.state_names[parent]}"
+                        )
+
+                    evidence_indices.append(parent_state_idx)
+                else:
+                    raise ValueError(
+                        f"Parent node {parent} not found in sample data. "
+                        "Ensure all parent nodes are included in the dataset."
+                    )
+
+            prob = cpd.values[observed_state_idx, tuple(evidence_indices)]
+
+        # Ensure probability is not zero for numerical stability
+        # (with a small epsilon to avoid log(0))
+        prob = max(prob, 1e-10)
+
+        return np.log(prob)
 
     def get_cpds(self, node=None, time_slice=None):
         """
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -1270,7 +1270,7 @@ class DynamicBayesianNetwork(DAG):
         from pgmpy.sampling import BayesianModelSampling
 
         if show_progress and config.SHOW_PROGRESS:
-            pbar = tqdm(total=n_time_slices * len(self._nodes()))
+            _ = tqdm(total=n_time_slices * len(self._nodes()))
 
         # Step 1: Create some data structures for easily accessing values
         do = {} if do is None else do
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -673,7 +673,7 @@ class DynamicBayesianNetwork(DAG):
         """
 
         # Get the observed value for this node
-        observed_value = sample_data[node]
+        observed_value = sample_data[(node[0], node[1])]
 
         # Convert observed value to state index according to CPD
         try:
@@ -696,8 +696,8 @@ class DynamicBayesianNetwork(DAG):
 
             # Get evidence values for all parents
             for parent in parents:
-                if parent in sample_data.index:
-                    parent_value = sample_data[parent]
+                if (parent[0], parent[1]) in sample_data.index:
+                    parent_value = sample_data[(parent[0], parent[1])]
 
                     # Convert parent value to state index
                     try:
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -535,6 +535,45 @@ class DynamicBayesianNetwork(DAG):
         self.cpds.extend(cpds)
 
     def log_likelihood(self, data):
+        """
+        Computes the log-likelihood of a given dataset.
+
+        The log-likelihood measure can be used to check how well the specified
+        model describes the data. A higher score means a better fit.
+
+        Parameters
+        ----------
+        data: pd.DataFrame instance
+            The dataset against which to score the model.
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> import pandas as pd
+        >>> from pgmpy.models import DynamicBayesianNetwork as DBN
+        >>> model = DBN(
+        ...     [
+        ...         (("A", 0), ("B", 0)),
+        ...         (("A", 0), ("C", 0)),
+        ...         (("B", 0), ("D", 0)),
+        ...         (("C", 0), ("D", 0)),
+        ...         (("A", 0), ("A", 1)),
+        ...         (("B", 0), ("B", 1)),
+        ...         (("C", 0), ("C", 1)),
+        ...         (("D", 0), ("D", 1)),
+        ...     ]
+        ... )
+        >>> data = np.random.randint(low=0, high=2, size=(1000, 20))
+        >>> colnames = []
+        >>> for t in range(5):
+        ...     colnames.extend([("A", t), ("B", t), ("C", t), ("D", t)])
+        ...
+        >>> df = pd.DataFrame(data, columns=colnames)
+        >>> model.fit(df)
+        >>> data = model.simulate(int(1e4))
+        >>> model.log_likelihood(data)
+        -103818.57516969478
+        """
         pass
 
     def get_cpds(self, node=None, time_slice=None):
--- a/pgmpy/models/DynamicBayesianNetwork.py
+++ b/pgmpy/models/DynamicBayesianNetwork.py
@@ -564,50 +564,50 @@ class DynamicBayesianNetwork(DAG):
         float
             The total log-likelihood score for the entire dataset. This is the
             sum of log-likelihood scores for all individual data points.
+            More negative values indicate worse fit to the data.
 
+        Raises
+        ------
+        ValueError
+            If data is not a pandas DataFrame, contains missing values,
+            has incorrect column structure, or contains values not present
+            in the model's CPD state names.
+
+        Examples
+        --------
         Examples
         --------
         >>> import numpy as np
         >>> import pandas as pd
         >>> from pgmpy.models import DynamicBayesianNetwork as DBN
-        >>> model = DBN(
-        ...     [
-        ...         (("A", 0), ("B", 0)),
-        ...         (("A", 0), ("C", 0)),
-        ...         (("B", 0), ("D", 0)),
-        ...         (("C", 0), ("D", 0)),
-        ...         (("A", 0), ("A", 1)),
-        ...         (("B", 0), ("B", 1)),
-        ...         (("C", 0), ("C", 1)),
-        ...         (("D", 0), ("D", 1)),
-        ...     ]
-        ... )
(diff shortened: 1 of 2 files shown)


dbolser:
I think I've fixed the black issues. You just need to merge my PR on your PR @petrello :-)

dbolser:
Thanks

petrello:
@dbolser Sorry for the late reply, and thanks a lot for your PR. I decided not to merge it, as I opted to fix the formatting manually to keep this PR self-contained.

Regarding the Black configuration issue, I believe it would be more appropriate to open a separate issue and then wait for feedback from @ankurankan before proceeding with a fix.

Thanks again!

dbolser:
No problem, are the black CI tests still failing?

petrello:
Not for this PR because I manually formatted the code to match Black's requirements. However, it looks like the underlying Black CI issue is still present in the main pgmpy repository.

ankurankan:
@petrello @dbolser Thanks for the PR and sorry for the super late reply. Is this ready for a review? 

For the formatting issues, we have a pre-commit hook (https://github.com/pgmpy/pgmpy/blob/dev/.pre-commit-config.yaml) that you can install, and it will take care of all formatting fixes automatically. You just need to run the following from the pgmpy base directory:

```
pip install pre-commit
pre-commit install
```

petrello:
Hi @ankurankan, no worries at all. Yes, the PR is ready to be reviewed.

Also, thank you for the tip about the pre-commit hook. I might have missed some steps while configuring my local environment.

dbolser:
Cool, the commit hook sounds good!

You can close this now (I think).

ankurankan:
@dbolser, I think all the tests are passing, so no need to make any changes (or close this) in this PR regarding the hook. And sorry again for the delay in the review, I still haven't gotten a chance to go through it.

@XAheli @Spinachboul @Nimish-4 Would any of you be interested in doing a first round of reviews?

XAheli:
> @dbolser, I think all the tests are passing, so no need to make any changes (or close this) in this PR regarding the hook. And sorry again for the delay in the review, I still haven't gotten a chance to go through it.
> 
> @XAheli @Spinachboul @Nimish-4 Would any of you be interested in doing a first round of reviews?

Surely!!

Spinachboul:
> @dbolser, I think all the tests are passing, so no need to make any changes (or close this) in this PR regarding the hook. And sorry again for the delay in the review, I still haven't gotten a chance to go through it.
> 
> @XAheli @Spinachboul @Nimish-4 Would any of you be interested in doing a first round of reviews?


Yes, will do!

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+
+        # Input validation
+        if not isinstance(data, pd.DataFrame):
+            raise ValueError(
+                f"Data must be a pandas.DataFrame instance. Got: {type(data)}"
+            )
+
+        if min(data.columns, key=lambda x: x[1])[1] != 0:
+            raise ValueError("Data column names must start from time slice 0.")
+
```
XAheli, line 624: What happens if `data.columns` is empty? The `min()` function will raise a `ValueError`. Should we add a check for empty DataFrames before this validation? 

also what if columns are not tuples at all? This will cause a `TypeError` when accessing `x[1]`. Should tuple structure validation happen prior this check?

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+        if set(self.nodes()) - set(data.columns):
+            raise ValueError(
+                f"Missing columns in data. Can't find values for the following variables: "
+                f" {set(self.nodes()) - set(data.columns)}"
+            )
```
XAheli, line 629: Should include a validation that `data.columns` is a subset of or exactly matches `self.nodes()`

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+        if data.isnull().any().any():
+            raise ValueError(
+                "Data contains missing values. Log-likelihood computation "
+                "requires complete observations."
+            )
+
+        self.check_model()
+
+        # Create mapping from node to CPD for efficient lookup
+        node_to_cpd = {cpd.variable: cpd for cpd in self.get_cpds()}
```
XAheli, line 647: `node_to_cpd` is built from `self.get_cpds()` but `get_cpds()` could return empty and fail or raise KeyError. Given that users may call `log_likelihood` before fitting/adding cpds, error should be clear

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+            for t in sorted(time_slices):
+                # Get nodes in current time slice
+                nodes_t = [node for node in nodes if node[1] == t]
+
+                # Compute log-probability for each node in this time slice
+                for node in nodes_t:
+                    node_log_prob = self._log_likelihood_node(
+                        node, data.iloc[i], node_to_cpd[node]
+                    )
+                    sample_log_likelihood += node_log_prob
```
XAheli, line 679: The docstring mentions "Variables in the first time slice (t=0) use their marginal probabilities" but I don't see handling for t=0 vs t>0. Is this handled implicitly by the parent structure?

cc @ankurankan

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+                    raise ValueError(
+                        f"Parent node {parent} not found in sample data. "
+                        "Ensure all parent nodes are included in the dataset."
+                    )
+
+            prob = cpd.values[observed_state_idx][tuple(evidence_indices)]
+
+        # Ensure probability is not zero for numerical stability
+        # (with a small epsilon to avoid log(0))
+        prob = max(prob, 1e-10)
```
XAheli, line 769: Why specifically 1e-10 as the epsilon value? Is this documented anywhere?

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+            )
+
+        # Check for missing values
+        if data.isnull().any().any():
+            raise ValueError(
+                "Data contains missing values. Log-likelihood computation "
+                "requires complete observations."
+            )
+
+        self.check_model()
```
XAheli, line 644: maybe adding a parameter to `log_likelihood` to allow skipping model validation when we know the model is unchanged ?

pgmpy/tests/test_models/test_DynamicBayesianNetwork.py
XAheli, line 1: more edge cases would be great
e.g. 

- `test_log_likelihood_empty_df_raises`
- `test_log_likelihood_invalid_column_names_raises`
- `test_log_likelihood_extra_columns_warns`
- `test_log_likelihood_no_cpds_raises`
- `test_log_likelihood_epsilon_warning`

XAheli (commented):
Added some comments kindly look into it.

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+                f"Observed value '{observed_value}' for node {node} "
+                f"not found in CPD state names: {cpd.state_names[node]}"
+            )
+
+        # Get parents (evidence variables) for this node
+        parents = self.get_parents(node)
+
+        if not parents:
+            # No parents - use marginal probability
+            prob = cpd.values[observed_state_idx]
```
Spinachboul: A shape check here will probably make sense for `cpd_values` for correct indexing else could return `IndexError` or other incompatibility issues

pgmpy/models/DynamicBayesianNetwork.py
```diff
@@ -533,6 +534,215 @@ def add_cpds(self, *cpds):
...
+        Raises
+        ------
+        ValueError
+            If observed value is not found in CPD state names,
+            if parent values are not found in CPD state names,
+            or if required parent nodes are missing from sample data.
+        """
+
+        # Get the observed value for this node
+        observed_value = sample_data[(node[0], node[1])]
```
Spinachboul: Could we encounter `NaN` values here, and handling them gracefully makes sense?
Also if the node key `node` does not exist in the `sample_data.index`, then what happens?

dbolser:
Just to be clear, this is @petrello's PR. I just injected myself for some practice.

pgmpy/models/DynamicBayesianNetwork.py
petrello: Yes, we can add the validation step you mentioned, but it might not be needed. Notice that extra columns in `data.columns` won't be considered since the method will refer only to `self.nodes()`. Therefore, we are only interested in checking if the dataframe `data` contains the nodes.

pgmpy/models/DynamicBayesianNetwork.py
petrello: Ok, we can either validate the result of `self.get_cpds` or the dictionary `node_to_cpd`. For example, if `get_cpds()` returns empty, then `note_to_cpd` will be empty as well.

We can opt for something like:
```suggestion
        # Create mapping from node to CPD for efficient lookup
        node_to_cpd = {cpd.variable: cpd for cpd in self.get_cpds()}

        if not node_to_cpd:
            raise ValueError("No CPDs found in the model.")
```

To manage possible KeyError exceptions, we can define a `cpd` variable and wrap it with a `try-except` statement:
```python
    # Compute log-probability for each node in this time slice
    for node in nodes_t:
        try:
            cpd = node_to_cpd[node]
        except:
            raise KeyError(
                f"CPD for node {node} not found in the model."
            )

        node_log_prob = self._log_likelihood_node(
            node, data.iloc[i], cpd
        )
        sample_log_likelihood += node_log_prob
```
since `node_to_cpd[node]` could raise that kind of exception.

pgmpy/models/DynamicBayesianNetwork.py
petrello: The decision of using marginal probability rather than conditional probability is handled inside the `_log_likelihood_node` method as follows:

```python
# Get parents (evidence variables) for this node

if not parents:
    # No parents - use marginal probability
    ...
else:
    # Has parents - use conditional probability
    ...
```

pgmpy/models/DynamicBayesianNetwork.py
petrello: No, this is actually just a small value to avoid log(0).

pgmpy/models/DynamicBayesianNetwork.py
petrello: I propose to check if `data` is empty or has empty columns (all) using `data.dropna().empty` property. This way, we ensure `data` has at least one non-empty column so that `min()` won't fail. Then, check for each `data.columns` if the column name is a tuple and has a length of 2, and whether the second item of the tuple is a non-negative integer.

```suggestion
        if data.dropna().empty:
            raise ValueError("DataFrame is empty.")

        for col in data.columns:
            if not isinstance(col, tuple) or len(col) != 2:
                raise ValueError(
                    "Data column names must be tuples of the form (node_name, time_slice). "
                    f"Got column type {type(col)} of length {len(col) if isinstance(col, tuple) else 'N/A'}"
                )
            if not isinstance(col[1], int) or col[1] < 0:
                raise ValueError(
                    "Time slice in data column names must be a non-negative integer."
                )
        
        if min(data.columns, key=lambda x: x[1])[1] != 0:
            raise ValueError("Data column names must start from time slice 0.")
```

pgmpy/tests/test_models/test_DynamicBayesianNetwork.py
petrello: Perfect, thank you.

What do you mean by `test_log_likelihood_epsilon_warning`?

dbolser:
I'll leave this to you guys

petrello:
@ankurankan @Spinachboul @XAheli Hi! Just checking back on this PR after the latest updates.

I've addressed the previous points and refined the implementation of `log_likelihood` for `DynamicBayesianNetwork`, aiming to keep it consistent with existing APIs.

I'd really appreciate any feedback when you have time. Happy to make further changes if needed!

PR #2251, opened 2025-07-16
Add CausalGraph

marrov:
- Adds CausalGraph as an extension of DAG through composition that adds roles to nodes
- Also adds some factory functions

### Notes
The first commit from this PR is LLM-assisted. I will clean this up and properly test during the next few days, just wanted to get a prototype out for early review. For context, I generated these requirements that the LLM agents used as a starting point.

---

# CausalGraph Class Requirements

## 1. Core Concept

The `CausalGraph` class is designed to represent a causal graphical model. It encapsulates not just the graph structure (nodes and edges) but also the functional roles of the variables (nodes) within the causal model.

## 2. Implementation Strategy: Composition over Inheritance

- The `CausalGraph` class should **not** inherit from `pgmpy.base.DAG`.
- Instead, it should use **composition**. It will hold an instance of a `pgmpy.base.DAG` object internally.
- The class should "forward" relevant methods from the internal `DAG` instance to its own interface, allowing users to perform standard graph operations.

## 3. Data Structure

The class should contain at least two main attributes:

- **`graph`**: An instance of `pgmpy.base.DAG`, representing the nodes (variables) and directed edges (causal relationships).
- **`roles`**: A dictionary to store the roles of different variables.
  - Keys are role names (e.g., `"exposure"`, `"outcome"`, `"adjustment_set"`).
  - Values are sets of variable names (nodes) assigned to that role.

## 4. Initialization / Constructor

- The constructor should be flexible and allow instantiation from a graph-like object and initial role assignments.
- Example from notes:

  ```python
  causal_graph = CausalGraph(
      graph=[("U", "X"), ("X", "M"), ("M", "Y"), ("U", "Y")],
      exposure="X",
      outcome="Y"
  )
  ```

- This implies the constructor should accept a list of edges for the graph and keyword arguments for initial roles like exposure and outcome.

## 5. Variable Roles

- The class must manage metadata about the roles of variables in the causal system.
- **Required Roles**: The primary roles identified are `exposure` and `outcome` (or `target`). These roles should contain a single variable.
- **Identifiable Roles**: The system is designed to identify other roles through identification algorithms, such as:
  - `adjustment_set` (can contain multiple variables)
  - `instrumental_variable`
  - `mediator`
- **Data-related Roles**: The notes also consider a more granular classification of variable roles with respect to a given dataset:
  1. Exposure
  2. Adjustment set
  3. Target/Outcome
  4. Unobserved confounders (present in the graph but not in the data)
  5. Other observed variables

## 6. Interaction with Other Components

- **Identification Algorithms (`BaseCausalIdentification`)**:
  - Identification algorithms will take a `CausalGraph` instance as input.
  - They will process the graph to determine new roles (e.g., find an adjustment set).
  - They will return a new `CausalGraph` instance containing the newly identified roles.
- **Causal Estimators/Regressors (`BaseRegressor`)**:
  - Causal estimators will accept a `CausalGraph` instance as a `model` parameter.
  - The estimator will use the roles defined in the `CausalGraph` (e.g., `adjustment_set`) to configure its estimation strategy.
  - The design should be simple enough that an estimator can internally run an identification algorithm if the required roles are not already present in the provided `CausalGraph`.

## 7. Special Cases and Simplifications

- The design should accommodate simplified, pre-defined causal structures.
- An example is the `RCT` (Randomized Controlled Trial), which could be a factory function or a specialized subclass that creates a `CausalGraph` with a structure and roles (treatment, outcome) implied by the experimental design.

## 8. Immutability

- `CausalGraph` instances should be immutable.
- Methods that would typically modify a graph (e.g., `add_node`, `add_edge`, or methods that assign new roles) should return a *new* `CausalGraph` instance with the changes applied, rather than modifying the instance in-place.

## 9. Method Forwarding

- A minimal set of methods from the underlying `DAG` object should be exposed. The exact set can be expanded during implementation, but a good starting point would be:
  - `get_nodes`
  - `get_edges`
  - `get_parents`
  - `get_children`
  - `is_active_trail`

## 10. Role Aliasing

- To improve accessibility, the system should support aliasing for common role names (e.g., mapping "treatment" to "exposure" or "target" to "outcome").
- **Suggested Design Pattern**: A separate, module-level dictionary can serve as a canonical mapping of aliases to standard role names. This keeps the aliasing logic decoupled from the `CausalGraph` class itself, making it easy to maintain and reference.

  ```python
  # In a separate config or utils file
  ROLE_ALIASES = {
      "treatment": "exposure",
      "target": "outcome",
      # ... other aliases
  }
  ```

## 11. Constructor and Initialization Details

- **Graph Input Formats**: To maintain flexibility, the constructor could accept graph data in various formats, potentially reusing the logic from `pgmpy.base.DAG`'s constructor. Suggested formats include:
  - A list of edges (as in the example).
  - An existing `networkx` Graph or DiGraph object.
- **Passing Multi-Variable Roles**: For roles that accept multiple variables (like `adjustment_set`), the constructor could accept them in one of the following ways:
  - **Option 1: Keyword argument with a list/set.**
    - `CausalGraph(..., adjustment_set={"Z1", "Z2"})`
    - **Pro**: Explicit and clear.
    - **Con**: Can make the constructor signature long if many roles are passed.
  - **Option 2: A single `roles` dictionary.**
    - `CausalGraph(..., roles={"adjustment_set": {"Z1", "Z2"}})`
    - **Pro**: Keeps the constructor signature clean.
    - **Con**: Slightly more verbose for the user.

## 12. Open Questions

- **Role Overwriting**: What is the desired behavior if an identification algorithm is asked to find a role (e.g., `adjustment_set`) that already exists on the `CausalGraph` instance? Should it be overwritten, merged, or should an error be raised?


fkiraly (changes requested):
Neat!

My main comment revolves around composition using the graph object. I was thinking, should we rely more on standard graph libraries for anything not related to roles, e.g., `networkx`, which is a dep of `pgmpy`?

pgmpy/base/causal_graph.py
```diff
@@ -0,0 +1,245 @@
...
+        role_name = self._resolve_role_alias(role_name)
+
+        new_roles = self._roles.copy()
+        new_roles.pop(role_name, None)
+
+        new_cg = CausalGraph(graph=self._graph)
+        new_cg._roles = new_roles
+        return new_cg
+
+    def __getattr__(self, name):
```
fkiraly, line 126: I am a bit concerned about this because this is basically manual inheritance.

This is what we did in `sktime` with `ForecastingHorizon` in relation to `pandas.Index`... so it is causing me a bit of discomfort.

One thing I wonder about: why not directly inherit from `networkx` `Graph` then - it is meant to be inherited from (if you look at docstring). That way we might deduplicate.

Though if you try that, I would suggest to do it in a different branch, to have options.

pgmpy/base/causal_graph.py
```diff
@@ -0,0 +1,245 @@
...
+
+class CausalGraph:
+    """
+    A causal graphical model combining a directed acyclic graph (DAG)
+    structure with metadata about the roles of variables in the causal
+    system (e.g., exposure, outcome, adjustment set).
+
+    Parameters
+    ----------
+    graph : DAG, list of edges, or any graph-like object
```
fkiraly, line 24: the code assumes that it must have methods `edges` or `nodes`, similar to `networkx` `Graph`

pgmpy/base/causal_graph.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def __repr__(self) -> str:
+        return self.__str__()
+
+    def __eq__(self, other) -> bool:
+        """Check equality between CausalGraph instances."""
+        if not isinstance(other, CausalGraph):
+            return False
+
+        if set(self.nodes()) != set(other.nodes()) or set(self.edges()) != set(
```
fkiraly, line 163: is this better delegated to the equality of `self._graph`?

fkiraly (changes requested):
The main thing I worry about in this design is that it violate sthe "Law of Demeter", that is, it looks like shallow composition, with `graph` a component, but in fact it is closer to inheritance from `graph` in a non-pythonic pattern ( delegating `__getitem__`).

This brings me to question, why do we not look at inheritance in the first place?

`networkx` `Graph` supports data attached to nodes (which is what we need as added feature), and the only "diff" to what we want is a lookup of "which roles are there", if I understand correctly.

So, we should seriously consider the approach where we:

* directly inherit from `networkx` `Graph`
* add a few methods as syntactic sugar

fkiraly:
Alternative design suggestion here: https://github.com/pgmpy/pgmpy/pull/2252
(changes mainly internal structure)

marrov:
Closing due to better alternative design in #2252

PR #2252, opened 2025-07-19
Vertex role feature for `DAG` based causal models

fkiraly:
Alternative design for the `Causalgraph`, also see https://github.com/pgmpy/pgmpy/pull/2251

* inherits directly from `networkx` and attempts to use idiomatic patterns where possible
* design-wise, tries to use the `networkx` data attributes for roles. The rest is just syntactic sugar.
* mixin design, so can be used to extend existing classes `DAG` and `UndirectedGraph` if desired

Compared to #2251, also:

* removes role aliases, I feel that incurs too many "hardcoding" choices and associated maintenance burden
* uses `list` instead of `set`

FYI @marrov, @ankurankan


pgmpy/base/causal_graph.py
```diff
@@ -0,0 +1,214 @@
...
+            The name of the role to check.
+
+        Returns
+        -------
+        bool
+            True if the role exists and has variables assigned, False otherwise.
+        """
+        return role in self.get_roles()
+
+    def with_role(self, role: str, variables, inplace=False) -> "CausalGraph":
```
marrov: Type hint missing? `variables: Iterable[Hashable]`

marrov:
So this is what I get for not reading through the `networkx` docs. Using the built-in attributes makes the most sense, I agree. That being said, a couple of comments:

*Major*
- While not dealing with aliasing adds simplicity, I think this is an important feature based on @ankurankan comments during prototyping. Maybe this should not be part of this class, which is a different discussion, but I do think this should happen somewhere. I also think that this should be easily accessible/readable for users.
- The tests have not been modified so of course they are failing. I could look into this if we settle on this design (I think we should)
- In `CausalGraph` if "roles" not in `attr`, it is never initialised before being used in the for-loop at the end of the method. This will cause an error.

*Minor*
- I'd prefer an inmutable design - I know that the underlying `networkx` graph are mutable but I would not expose a `inplace=False` param.
- Missing type hint on `variables`
- Why lists instead of sets? I think set operations added clarity
- Super-nit, but in `is_valid_causal_structure()` if the output is a `bool` it should be false if not valid, not raise an error, no? There's also a formatting issue in the error message: missing space between "outcome'" and "role".
- Alt implementation for `get_roles()` that also fixes the docstring-output type mismatch:
```python
def get_role_sets(self):
    """Get dict of sets of roles preset in the graph.

    Returns
    -------
    Dict with str keys and values being list of nodes
        keys are roles present in the graph, and sets are nodes with that role
    """
    r_dict = {}
    for n, data in self.nodes(data=True):
        role = data.get("role", None)
        if role is not None:
            if role not in r_dict:
                r_dict[role] = []
            r_dict[role].append(n)
    return r_dict
```

pgmpy/base/causal_graph.py
fkiraly: I do not like long type hints if they cause code to wrap over multiple lines... but I do not mind

fkiraly:
> _Major_
> 
>     * While not dealing with aliasing adds simplicity, I think this is an important feature based on @ankurankan comments during prototyping. Maybe this should not be part of this class, which is a different discussion, but I do think this should happen somewhere. I also think that this should be easily accessible/readable for users.

Yes, though the exact place is not entirely clear to me. I thought in the `CausalModel` rather than in the `CausalGraph`? E.g, `RCT` concrete class.
 
>     * The tests have not been modified so of course they are failing. I could look into this if we settle on this design (I think we should)

Sure! I was just waiting for your review, e..g, whether we want sets or lists, etc.

 >     * In `CausalGraph` if "roles" not in `attr`, it is never initialised before being used in the for-loop at the end of the method. This will cause an error.

I think I fixed it now. Would have probably been caught by tests.

> _Minor_
> 
>     * I'd prefer an inmutable design - I know that the underlying `networkx` graph are mutable but I would not expose a `inplace=False` param.

I agree - I know I thought I had a good reason when I added this, but I no longer am sure what it was or whether it existed.

Maybe it was a compromise, if you look at `networkx` it has `add` and `clear` methods, so `add_roles` and `clear_roles` would mutate the graph if they existed.

>     * Missing type hint on `variables`

:-)

>     * Why lists instead of sets? I think set operations added clarity

Yes, but `list` felt more idiomatic for `networkx`. I do ont have strong opinions though

>     * Super-nit, but in `is_valid_causal_structure()` if the output is a `bool` it should be false if not valid, not raise an error, no? There's also a formatting issue in the error message: missing space between "outcome'" and "role".

You wrote that method, I barely touched it.

>     * Alt implementation for `get_roles()` that also fixes the docstring-output type mismatch:

I see, makes sense.

ankurankan:
Thanks for the implementations, @fkiraly @marrov. I like this mixin design, and in the future, as we expand the variable roles feature to other graph types, such as PDAG and ADMG, we should be able to reuse this for those classes.

One question I have is that the `CausalGraph` class is inheriting from `nx.DiGraph`. However, for features like causal identification (such as identifying backdoor variables), we would need to use some of the methods implemented in the DAG class. So, I am wondering if it might be a better idea to add the mixin to the current `DAG` class instead of defining a new `CasualGraph` class. 

The other problem I see with defining a CausalGraph class is that there is no clear definition of what a causal graph is, for example it could be a directed graph (as is the case with DAGs), or a mix of directed + undirected edges (as with PDAGs), or a mix of directed + bidirected edges (as with ADMGs), and the list goes on.

fkiraly:
> One question I have is that the `CausalGraph` class is inheriting from `nx.DiGraph`. However, for features like causal identification (such as identifying backdoor variables), we would need to use some of the methods implemented in the DAG class. So, I am wondering if it might be a better idea to add the mixin to the current `DAG` class instead of defining a new `CasualGraph` class.

Yes, I agree - after disussion today, that is what I will do, in two steps:

* first, fix tests here
* then, in a separate PR move the mixin to the `DAG` and `PDAG` classes, remove the `CausalGraph`.

> The other problem I see with defining a CausalGraph class is that there is no clear definition of what a causal graph is, for example it could be a directed graph (as is the case with DAGs), or a mix of directed + undirected edges (as with PDAGs), or a mix of directed + bidirected edges (as with ADMGs), and the list goes on.

I would say, it is simply a labelled graph with a special vertex label type "roles".

fkiraly:
superseded by https://github.com/pgmpy/pgmpy/pull/2265

ankurankan:
Closing as feature implemented in #2265

PR #2264, opened 2025-07-30
[ENH] remove unused imports, linting

fkiraly:
This PR removes some unused imports and lints assorted locations in the code base.

It is not systematic, simply a few places I spotted.

I also would suggest to add more stringent linting to the pre-commit, to automate this.


ankurankan (approved):
Thanks for the fixes @fkiraly

PR #2265, opened 2025-07-31
[ENH] causal graph with roles - feature added to `DAG`

fkiraly:
This PR implements a design that adds abstract roles to `DAG`, for use in causal discovery and causal prediction algorithms.

Refines the designs of #2252 and #2251 after discussion.

Makes the following changes:

* adds a `_GraphRolesMixin` that can be used in any graph or related class based on `networkx`
* adds a number of "roles" related methods to `DAG`
* adds `roles` argument to `DAG.__init__` for at-construction specification
* adds tests for aforementioned methods and constructor usage
* fixes some omissions in the docstring of `DAG` unrelated to this PR
* linting


pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
...
+
+        if not valid:
+            raise ValueError(
+                f"CausalGraph must have at least one 'exposure' and one 'outcome'"
+                f"role defined, but {problem_str}."
+            )
+        return True
+
+
+class _CausalGraph(_GraphRolesMixin, nx.DiGraph):
```
marrov: Should this still be here? I'd remove it completely now that the mixin is added to `DAG`

pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
...
+        Returns
+        -------
+        List of str
+            list of all roles defined in the graph.
+        """
+        roles = {d.get("role", None) for _, d in self.nodes(data=True)}
+        roles.discard(None)  # remove "None"
+        return list(roles)
+
+    def get_role_sets(self):
```
marrov: Misleading func name and docstring because it is returning a dict of lists.

pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
...
+            The name of the role to remove, e.g., "exposure", "outcome".
+        variables : str, set, list, or iterable, default = all variables with the role
+            The variables to remove the role from. If not provided,
+            all variables with the specified role will have it removed.
+        inplace=False : bool, optional
+            If True, modifies the current graph in place. Defaults to False.
+
+        Returns
+        -------
+        CausalGraph
```
marrov: Given this is a mixin the return type would depend on the class, no? Not sure if this the right way to indicate the return now that this has been added to `DAG`.

pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env python3
+
+from typing import Hashable
+
+
```
marrov: Nit: extra line

pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
...
+        Parameters
+        ----------
+        role : str
+            The role to match.
+
+        Returns
+        -------
+        List of nodes with the specified role.
+        """
+        G = self
```
marrov, line 21: This is to avoid a line break on L25? It looks a bit strange, why not just use `self`?

pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
...
+            The name of the role to check.
+
+        Returns
+        -------
+        bool
+            True if the role exists and has variables assigned, False otherwise.
+        """
+        return role in self.get_roles()
+
+    def with_role(self, role: str, variables, inplace=False):
```
marrov, line 68: In general (not only for this line) type hinting is inconsistent. For example in `has_roles` we have all func arguments and the return type hinted but in this one we are missing type hints for the return, `variables` and `inplace`.

pgmpy/tests/test_base/test_causal_graph.py
marrov, line 1: Not sure I love the `xfail` - is this just a pointer to raise a future issue to implement this in `DAG`?

marrov, line 1: Coverage is ~80% while target is 90%. Not sure if this is fine @ankurankan. Otherwise, I could look into adding some tests to increase coverage.

marrov (changes requested):
Left some comments, most minor, but I would remove the unused `_CausalGraph`

pgmpy/base/_mixin_roles.py
fkiraly: `self` does not "feel" like a graph, I suppose. I did this for semantic clarity, but we can remove.

pgmpy/base/_mixin_roles.py
fkiraly: I suppose this is because your type hinting was consistent, and I did not care about adding type hints? So we are now in this inconsistent state. Should we add type hints throughout?

pgmpy/tests/test_base/test_causal_graph.py
fkiraly: yes - or, at least, think about design implications. The `networkx` classes themselves do not implement this kind of equality or hash.

pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
+#!/usr/bin/env python3
+
+from typing import Hashable
+
+
+import networkx as nx
```
ankurankan: After `_CausalGraph` class is removed, I don't think we need this import.

pgmpy/base/_mixin_roles.py
```diff
@@ -0,0 +1,228 @@
...
+
+        Returns
+        -------
+        bool
+            True if the role exists and has variables assigned, False otherwise.
+        """
+        return role in self.get_roles()
+
+    def with_role(self, role: str, variables, inplace=False):
+        """Return a new CausalGraph with the specified role assignment.
```
ankurankan: Maybe `CausalGraph` -> `causal graph`? Because the mixin might be used with different types of causal graphs.

ankurankan (commented):
I have added a couple of very minor comments but looks good to me otherwise. 

> Coverage is ~80% while target is 90%.

That's fine with me :)

pgmpy/base/_mixin_roles.py
fkiraly: changed

pgmpy/base/_mixin_roles.py
fkiraly: removed

pgmpy/base/_mixin_roles.py
fkiraly: fixed

pgmpy/base/_mixin_roles.py
fkiraly: imports fixed (removed)

pgmpy/base/_mixin_roles.py
fkiraly: removed

pgmpy/base/_mixin_roles.py
fkiraly: changed to "graph", to be general

pgmpy/tests/test_base/test_causal_graph.py
fkiraly: fixed by removing the class, apparently

marrov (approved):
LGTM

fkiraly:
coverage seemed still at 87% - added a test for `get_role_dict` which was not covered

ankurankan (approved):

Issue #2262, opened 2025-07-30
[ENH] Add methods to convert DAGs to dagitty and lavaan representation


ankurankan:
pgmpy currently has methods `DAG.from_lavaan` and `DAG.from_dagitty` to initialize a DAG object through a model specified in lavaan or dagitty syntax. To improve interoperability, we should also add methods that convert a pgmpy DAG object to Lavaan syntax and Dagitty syntax.

Add methods:
1. `DAG.to_lavaan`: Returns a string representing the equivalent lavaan model
2. `DAG.to_dagitty`: Returns a string representing the equivalent dagitty model

Reference syntax:
Lavaan: https://lavaan.ugent.be/tutorial/sem.html
Dagitty: https://cran.r-project.org/web/packages/dagitty/dagitty.pdf Page 10.

XAheli:
Hi! I'd like to work on this

ankurankan:
Fixed.

PR #2269, opened 2025-08-05
Add `DAG.to_lavaan()` and `DAG.to_dagitty()` conversion methods

XAheli:
I've implemented the conversion methods requested in issue #2262 

Added two new methods to the `DAG` class:
- `to_lavaan()` - Converts DAG to lavaan syntax 
- `to_dagitty()` - Converts DAG to dagitty syntax 

## Key Features

**Edge Case Handling:**
- Empty DAGs return empty representations
- Isolated nodes (nodes without edges) 
- Disconnected graph components 
- Mixed node types (strings, integers, tuples) are supported through string conversion

**Deterministic Output:**
- Consistent ordering using `sorted()` 
- Same DAG structure always produces identical output regardless of edge insertion order

**Testing:**
- Tests for various node name types including unicode characters

## Examples

```python
# Simple case
dag = DAG([('X', 'Y'), ('Z', 'Y')])
print(dag.to_lavaan())  # Y ~ X + Z
print(dag.to_dagitty()) # dag {\nX -> Y\nZ -> Y\n}

# Works with any hashable node names
dag = DAG([(1, 2), ('A', 2)])
print(dag.to_lavaan())  # 2 ~ 1 + A
```

## Implementation Notes

- Node names are converted to strings using `str()` for maximum compatibility
- Lavaan format only includes nodes with parents (standard for regression equations)  
- Dagitty format includes all nodes (even isolated ones)


pgmpy/base/DAG.py
```diff
@@ -1467,6 +1469,155 @@ def to_graphviz(self, plot_edge_strength=False):
...
+        References
+        ----------
+        lavaan syntax: http://lavaan.ugent.be/tutorial/syntax1.html
+        """
+        if not self.edges():
+            return ""
+
+        # Group children by their parents
+        child_to_parents = {}
+        for parent, child in self.edges():
```
ankurankan: I think a simpler logic here could be to iterate over each node in the graph instead. For each non-root node in the graph, we would get one lavaan equation. The pseudo-code would be something like:

```
for node in nodes:
     if node is not root-node:
          equations.append(f"{node} ~ {'+'.join(self.get_parents(node))})
 
```

pgmpy/base/DAG.py
```diff
@@ -1467,6 +1469,155 @@ def to_graphviz(self, plot_edge_strength=False):
...
+            nodes_in_edges.add(parent)
+            nodes_in_edges.add(child)
+
+        isolated_nodes = set(self.nodes()) - nodes_in_edges
+        if isolated_nodes:
+            # Sort by string representation
+            for node in sorted(isolated_nodes, key=str):
+                statements.append(str(node))
+
+        if not statements:
```
ankurankan: Do we need a separate `if` statement for this? If `statements` is empty, line 1618 should automatically return an empty string.

pgmpy/tests/test_base/test_dag_conversion.py
```diff
@@ -0,0 +1,279 @@
...
+to be converted to lavaan and dagitty syntax representations for interoperability
+with R packages.
+"""
+
+import pytest
+
+from pgmpy.base import DAG
+
+
+class TestDAGConversion:
```
ankurankan: I think it would be nice to add a couple of tests with more complicated models. We can construct a DAG using a dagitty model (`DAG.from_dagitty` or `pgmpy.utils.get_example_model`), and then export it back to the dagitty representation (`DAG.to_dagitty`), and compare if our output matches the initial dagitty model. 

For dagitty models, you can copy the model from: https://www.dagitty.net/dags.html#. Go to Example (on top) and copy the model code (right sidebar). Sebastiani et al. 2005 is a decently large network.

pgmpy/tests/test_base/test_dag_conversion.py
```diff
@@ -0,0 +1,279 @@
+#!/usr/bin/env python3
```
ankurankan: I would suggest keeping these tests in the `test_DAG.py` file.

pgmpy/base/DAG.py
```diff
@@ -1467,6 +1469,155 @@ def to_graphviz(self, plot_edge_strength=False):
...
+                self.edges(), key=lambda x: (str(x[0]), str(x[1]))
+            ):
+                # Convert to string
+                parent_str = str(parent)
+                child_str = str(child)
+                edge_statements.append(f"{parent_str} -> {child_str}")
+            statements.extend(edge_statements)
+
+        # Add isolated nodes (nodes with no edges)
+        nodes_in_edges = set()
```
ankurankan: A simpler way to get isolated nodes is to use the networkx function: `nx.isolates`. Ref: https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.isolate.isolates.html

ankurankan (changes requested):
Thanks for the PR. Functionality looks good, I have added a few comments for improving the implementation.

pgmpy/base/DAG.py
XAheli: fixed

pgmpy/tests/test_base/test_dag_conversion.py
XAheli: fixed

pgmpy/tests/test_base/test_dag_conversion.py
XAheli: Added

pgmpy/base/DAG.py
XAheli: fixed

pgmpy/base/DAG.py
```diff
@@ -1467,6 +1469,139 @@ def to_graphviz(self, plot_edge_strength=False):
...
+        ----------
+        lavaan syntax: http://lavaan.ugent.be/tutorial/syntax1.html
+        """
+        lavaan_statements = []
+
+        # Iterate over each node and create equation for non-root nodes
+        for node in sorted(self.nodes(), key=str):
+            parents = self.get_parents(node)
+            if parents:  # If node has parents (i.e., not a root node)
+                # Convert node and parents to strings and sort parents
```
ankurankan: I would suggest removing these comments as they are kind of obvious given the comment on the for loop.

pgmpy/base/DAG.py
```diff
@@ -1467,6 +1469,139 @@ def to_graphviz(self, plot_edge_strength=False):
...
+        statements = []
+
+        # Add edges
+        if self.edges():
+            edge_statements = []
+            # Sort by string representation to handle mixed types
+            for parent, child in sorted(
+                self.edges(), key=lambda x: (str(x[0]), str(x[1]))
+            ):
+                # Convert to string
```
ankurankan: Would remove this

pgmpy/base/DAG.py
```diff
@@ -1467,6 +1469,139 @@ def to_graphviz(self, plot_edge_strength=False):
...
+        - If node names contain spaces or special characters, they will be used as-is.
+        - Users should ensure node names are valid in R/dagitty context if needed.
+
+        References
+        ----------
+        dagitty syntax: https://cran.r-project.org/web/packages/dagitty/dagitty.pdf
+        """
+        statements = []
+
+        # Add edges
```
ankurankan: Would remove this.

ankurankan: Because these methods are relatively straightforward, I would suggest writing a single comment explaining how the method works and removing the minor comments.

ankurankan (changes requested):
Looks good to me. I have added a few very minor comments about the comments. I think we avoid over-commenting the code.

XAheli:
> Looks good to me. I have added a few very minor comments about the comments. I think we avoid over-commenting the code.

fixed

ankurankan:
@XAheli This branch is showing conflicts. Can you either fix them or give permission to maintainers to edit your branch[1]? 

[1] https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

ankurankan (approved):
Looks good to me. Thanks for the PR!

Issue #2112, opened 2025-05-25
[ENH] Add methods for pairwise causal discovery


ankurankan:
The currently causal discovery methods implemented in pgmpy work on a set of variables > 2. Identifying the causal direction between 2 variables is a more difficult problem compared to causal discovery and requires stronger assumptions. We would first like to come up with a list of potential methods that we can implement. Here's a review paper: https://www.jmlr.org/papers/volume17/14-518/14-518.pdf

andoriyaprashant:
Hi @ankurankan I’d like to work on this issue
Could you please assign it to me
Thanks

rahulsjha:
Can you please assign this to me?

rahulsjha:
Can we used methods like  Information Geometric Causal Inference (IGCI), Additive Noise Models (ANMs)

ankurankan:
Yes, those can be good to start with. If you would like to work on any of those, I would suggest creating a new issue for the method that you are working on, describe the method that you would implement there, and then create a PR.

rahulsjha:
Okay.

rahulsjha:
please review I have raised a pr

XAheli:
I'd like to take this up!
FYI @ankurankan

XAheli:
> This issue has been open for 1 month with no activity and is now marked as stale.

Issue wip

PR #2273, opened 2025-08-06
ANM Bivariate Causal Discovery

Manas-7854:
implemented ANM algorithm

todo:
tests for ANM
PNL
tests for PNL

Refs #2112 #2286


pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Reshape data for sklearn
+    x = x.reshape(-1, 1)
+    y = y.reshape(-1, 1)
+
+    # Define the Gaussian Process Regressor kernel
+    kernel = C(1.0, (1e-3, 1e3)) * RBF(10, (1e-2, 1e2))
+
+    # Test for X -> Y
+    gp_xy = GaussianProcessRegressor(kernel=kernel, n_restarts_optimizer=10)
```
ankurankan: I think we should accept any sklearn-compatible estimator as an argument instead.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,67 @@
...
+    p_value_xy = hsic_test(
+        torch.from_numpy(x).float(), torch.from_numpy(res_xy).float()
+    )
+
+    # Test for Y -> X
+    gp_yx = GaussianProcessRegressor(kernel=kernel, n_restarts_optimizer=10)
+    gp_yx.fit(y, x)
+    x_pred = gp_yx.predict(y)
+    res_yx = x - x_pred
+    p_value_yx = hsic_test(
```
ankurankan: This can be any independence test, according to my understanding. We should accept that as an argument as well. You can have a look at `pgmpy.estimators.CITests` for a list of independence tests implemented in pgmpy.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,67 @@
+import torch
+from sklearn.gaussian_process import GaussianProcessRegressor
+from sklearn.gaussian_process.kernels import RBF
+from sklearn.gaussian_process.kernels import ConstantKernel as C
+from torch_hsic import hsic
+
+
+def anm_bivariate(x, y):
```
ankurankan: Could be renamed to be simply `ANM`

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    Performs bivariate causal discovery using the Additive Noise Model (ANM).
+
+    This function tests for a causal relationship between two variables, x and y,
+    in both directions (x -> y and y -> x) based on the ANM principle. It uses
+    a Gaussian Process Regressor to model the functional relationship and the
+    Hilbert-Schmidt Independence Criterion (HSIC) to test for the independence
+    of the residuals.
+
+    Args:
```
ankurankan: Please have a look at other docstrings for the expected format. We use numpydoc sytle.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,67 @@
...
+    # Test for Y -> X
+    gp_yx = GaussianProcessRegressor(kernel=kernel, n_restarts_optimizer=10)
+    gp_yx.fit(y, x)
+    x_pred = gp_yx.predict(y)
+    res_yx = x - x_pred
+    p_value_yx = hsic_test(
+        torch.from_numpy(y).float(), torch.from_numpy(res_yx).float()
+    )
+
+    return {"p_value_xy": p_value_xy, "p_value_yx": p_value_yx}
```
ankurankan: I think we should return the orientation between the edges instead of just the p-values. This will also make the output uniform across different bivariate causal discovery methods.

ankurankan (changes requested):
I would suggest keeping this PR only for ANM and do a separate PR for PNL.

ankurankan (changes requested):
Refer comments

pgmpy/estimators/bivariate.py
Manas-7854: the continuous and mixed tests were working fine as independence tests thus i have removed hsic for now,
let me know if you want me to add that back

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+
+import numpy as np
+import pandas as pd
+from sklearn.base import clone
+
+from pgmpy.estimators import CITests as ci_tests
+
+# A set of allowed independence tests for this ANM implementation.
+# Only tests suitable for continuous or mixed data are included.
+ALLOWED_ANM_CI_TESTS = {ci_tests.pearsonr, ci_tests.gcm, ci_tests.pillai_trace}
```
ankurankan, line 11: I don't think we need to limit the method to only continuous or mixed variables. Categorical variables should also work fine. Am I missing something here?

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+    Parameters
+    ----------
+    x : np.ndarray
+        A 1D NumPy array representing the first variable.
+    y : np.ndarray
+        A 1D NumPy array representing the second variable.
+    regressor : sklearn-compatible estimator
+        An instance of a regressor model (e.g., from scikit-learn) with
+        `.fit()` and `.predict()` methods.
+    independence_test : callable
```
ankurankan, line 40: We should specify how a custom `independence_test` should look and the arguments it should accept.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+    This function tests for a causal relationship between two variables, `x` and `y`,
+    in both directions (`x` -> `y` and `y` -> `x`). It uses a provided regressor
+    and a specific set of allowed pgmpy conditional independence tests.
+
+    The ANM principle assumes that for the true causal direction, the residuals of a
+    regression are independent of the cause variable. This implementation infers the
+    causal direction by comparing the p-values from these independence tests.
+
+    Parameters
+    ----------
```
ankurankan, line 32: Can you please add an empty line between each argument description? It looks cleaner and is a bit easier to read.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+ALLOWED_ANM_CI_TESTS = {ci_tests.pearsonr, ci_tests.gcm, ci_tests.pillai_trace}
+
+
+def ANM(
+    x: np.ndarray,
+    y: np.ndarray,
+    regressor,
+    independence_test: Callable,
+    significance_level: float = 0.05,
+) -> str:
```
ankurankan, line 20: I would suggest returning a tuple or None (if couldn't determine direction). It is harder to work with strings if we want to use the output within some other logic.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+        The inferred causal orientation. One of 'x->y', 'y->x', or 'x--y'
+        (for an undecided relationship).
+
+    Raises
+    ------
+    ValueError
+        If the provided `independence_test` is not in the allowed set of tests.
+    """
+    # --- Validate that the provided independence test is allowed ---
+    if independence_test not in ALLOWED_ANM_CI_TESTS:
```
ankurankan, line 59: I would suggest using the `pgmpy.estimators.CITest.get_callable_ci_test` method. Have a look at examples in the codebase on how we use it. `pgmpy.metrics.metrics.py` Line 98.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+        )
+
+    # --- Internal helper to run the pgmpy test ---
+    def run_independence_test(cause_arr, residual_arr):
+        # For pgmpy tests, create a DataFrame and call with named args
+        df = pd.DataFrame(
+            {"cause": cause_arr.flatten(), "residual": residual_arr.flatten()}
+        )
+        # pgmpy tests return a tuple (statistic, p_value, ...); we need the p_value
+        return independence_test(X="cause", Y="residual", Z=[], data=df, boolean=False)[
```
ankurankan, line 73: Would suggest assigning values for better readability. Something like:
```
_, p_value, _, _ = independence_test( .... 
return p_value
```

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+    str
+        The inferred causal orientation. One of 'x->y', 'y->x', or 'x--y'
+        (for an undecided relationship).
+
+    Raises
+    ------
+    ValueError
+        If the provided `independence_test` is not in the allowed set of tests.
+    """
+    # --- Validate that the provided independence test is allowed ---
```
ankurankan, line 58: We should add checks here for the dimensionality of `x` and `y`. If they are not in the expected dimensions, would suggest converting them at the start of the method. Else throw an error.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+            {"cause": cause_arr.flatten(), "residual": residual_arr.flatten()}
+        )
+        # pgmpy tests return a tuple (statistic, p_value, ...); we need the p_value
+        return independence_test(X="cause", Y="residual", Z=[], data=df, boolean=False)[
+            1
+        ]
+
+    # Reshape data for scikit-learn
+    x_reshaped = x.reshape(-1, 1)
+    y_reshaped = y.reshape(-1, 1)
```
ankurankan, line 79: I don't think we need to reshape y. We are doing a `ravel` or `flatten` before using it anywhere.

pgmpy/estimators/bivariate.py
```diff
@@ -0,0 +1,104 @@
...
+    # --- Test for Y -> X ---
+    reg_yx = clone(regressor)
+    reg_yx.fit(y_reshaped, x_reshaped.ravel())
+    x_pred = reg_yx.predict(y_reshaped)
+    residuals_yx = x_reshaped.flatten() - x_pred.flatten()
+    p_value_yx = run_independence_test(y, residuals_yx)
+
+    # --- Decision Logic ---
+    accept_xy = p_value_xy >= significance_level
+    accept_yx = p_value_yx >= significance_level
```
ankurankan, line 97: I thought the algorithm in the paper compared the two p-values or effect sizes with each other and not with a significance level.

pgmpy/tests/test_estimators/test_Bivariate.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def setUp(self):
+        """Set up common data and models for the tests."""
+        self.sample_size = 200
+        np.random.seed(42)
+
+    def test_linear_gaussian_case(self):
+        """
+        Test ANM on a linear Gaussian model: Y = 2*X + N.
+        ANM should not be able to distinguish the direction.
```
ankurankan, line 21: I think the algorithm should be able to distinguish the direction here. We have a clear direction from X to Y in the data-generating method. Am I missing something here?

pgmpy/tests/test_estimators/test_Bivariate.py
```diff
@@ -0,0 +1,68 @@
+import unittest
+
+import numpy as np
+from sklearn.linear_model import LinearRegression
+
+from pgmpy.estimators import CITests as ci_tests
+from pgmpy.estimators.bivariate import ANM
+
+
+class TestANM(unittest.TestCase):
```
ankurankan, line 10: We also need to add tests where the algorithm is able to find the correct orientation.

pgmpy/tests/test_estimators/test_Bivariate.py
```diff
@@ -0,0 +1,68 @@
...
+        Y = 2 * X + np.random.randn(self.sample_size)
+        estimator = LinearRegression()
+
+        # chi_square is a discrete test and should not be allowed.
+        # The `with self.assertRaises(...)` block checks that a ValueError is correctly raised.
+        with self.assertRaises(ValueError):
+            ANM(X, Y, regressor=estimator, independence_test=ci_tests.chi_square)
+
+
+if __name__ == "__main__":
```
ankurankan, line 67: I am not sure what this does?

pgmpy/tests/test_estimators/test_Bivariate.py
```diff
@@ -0,0 +1,68 @@
...
+from pgmpy.estimators import CITests as ci_tests
+from pgmpy.estimators.bivariate import ANM
+
+
+class TestANM(unittest.TestCase):
+    """Test suite for the ANM causal discovery function."""
+
+    def setUp(self):
+        """Set up common data and models for the tests."""
+        self.sample_size = 200
```
ankurankan, line 15: Would suggest a higher sample size so that sample size doesn't become a cause for variability in the tests.

pgmpy/tests/test_estimators/test_Bivariate.py
ankurankan, line 10: We are trying to move away from unittests, so would suggest writing tests in pytest.

pgmpy/estimators/bivariate.py
ankurankan: Logistic regression should work with discrete data.

ankurankan (changes requested):
@Manas-7854 I have added a few comments. I think there are some

pgmpy/tests/test_estimators/test_Bivariate.py
Manas-7854: this is a linear-gaussian case and in the paper its given that ANM doesn't work for the linear-gaussian case
"Note that this model includes the special case when all the fi are linear and all the pni
 are Gaussian"
 
 i think we should also implement a linear-nonGaussian case for which the ANM should work

here is the paper for ref (refer to section 2): https://papers.nips.cc/paper_files/paper/2008/file/f7664060cc52bc6f3d620bcedc94a4b6-Paper.pdf

pgmpy/estimators/bivariate.py
Manas-7854: i am not sure if ANM works for discrete data asw, i couldn't find it mentioned explicitly that i works/doesn't work for discrete data. although the models for which ANM is a extension of did require data to be continous.

here is the paper for ref (its mentioned in the abstract)
https://papers.nips.cc/paper_files/paper/2008/file/f7664060cc52bc6f3d620bcedc94a4b6-Paper.pdf

Manas-7854: also i tried using discrete CITests and the tests were failing :)

pgmpy/tests/test_estimators/test_Bivariate.py
Manas-7854: are refering to the `test_disallowed_discrete_test`, that was added to not allow user to use discrete CITests
we can remove this if we allow discrete CITests asw

ankurankan:
Closing due to inactivity.

Issue #2258, opened 2025-07-29
[ENH] Implement the MAG class


ankurankan:
Ref #1225

Spinachboul:
@ankurankan Commenting here for getting assigned

PR #2282, opened 2025-08-08
Base class for all the ancestral graphical models

Spinachboul:
Works on the issue https://github.com/pgmpy/pgmpy/issues/2258

- Implements the base class for MAGs and PAGs which need to handle the different edge types.
- Also has specific implementations to the both the classes (Initially they have minimal methods)

FYI @ankurankan


pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,299 @@
+#!/usr/bin/env python
```
ankurankan: Would suggest removing the shebang

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,299 @@
...
+
+import networkx as nx
+
+from pgmpy.base.DAG import DAG
+
+
+class AncestralBase(nx.DiGraph, DAG):
```
ankurankan: Do we need to inherit `DAG` here? Can we not work only with the `DiGraph` class?

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,299 @@
...
+    """
+    Base class for all ancestral graphical models.
+
+    This class extends the pgmpy.base.DAG to handle mixed graphs with
+    directed, bidirected, and partially directed edges. Edges are represented
+    with an attribute that specifies the marks at each endpoint.
+
+    The valid marks are:
+    - 'tail': A tail mark (e.g., in A -> B, A has a tail)
```
ankurankan: I think we should use some simplified marks for , such as:
tail - `-`
arrowhead - '>' or '<' (should be stored as just '>')
circle - 'o'

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,154 @@
...
+    directed, bidirected, and partially directed edges.
+    """
+
+    def __init__(self, ebunch=None):
+        self.graph = nx.DiGraph()
+        if ebunch:
+            self.add_edges_from(ebunch)
+
+    def add_edge(self, u, v, u_mark, v_mark):
+        if u_mark not in {"tail", "arrowhead", "circle"} or v_mark not in {
```
ankurankan: Would suggest using `-`, `>`, and `o` instead of full strings. Would be easier to type out and faster to compare.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+    Base class for all ancestral graphical models.
+
+    Wraps a networkx.DiGraph to handle mixed graphs with
+    directed, bidirected, and partially directed edges.
+    """
+
+    def __init__(self, ebunch=None):
+        self.graph = nx.DiGraph()
+        if ebunch:
+            self.add_edges_from(ebunch)
```
ankurankan: I think it would be better not to define `graph` as an attribute but to represent the graph as the object itself. Similar to how DAG works currently. We would change this to something like:
```python
super().__init__(ebunch)
if ebunch:
     self.add_edges_from(ebunch)
```

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+            "arrowhead",
+            "circle",
+        }:
+            raise ValueError("Marks must be one of 'tail', 'arrowhead', or 'circle'.")
+
+        self.graph.add_edge(u, v, mark=v_mark)
+        self.graph.add_edge(v, u, mark=u_mark)
+
+        # if self._is_symmetric_edge(u_mark, v_mark):
+        #     self.graph.add_edge(v, u, mark=u_mark)
```
ankurankan: I think this can be removed

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+
```
ankurankan, line 198: I would suggest adding a `to_adjacency_matrix` method for getting the equivalent adjacency matrix and a dict of node to index number map. This can simplify implementations of a lot of helper methods. For example, for checking if an edge (u, v) is directed in an adjacency matrix M, we can simply check: `M[index(u), index(v)] == '>' and M[index(v), index[u]] == '-'`. This will also be helpful in implementing dfs and neighbor methods.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,132 @@
+import networkx as nx
+
+
+class AncestralBase:
```
ankurankan: I would suggest adding a `to_adjacency_matrix` method that returns the adjacency matrix representation with the edge type in the matrix and a dict of node to index number. This would be helpful in implementing a lot of edge matching methods. For example, to check if an edge (u, v) is directed, we can simply check: `M[index(u), index(v)] == '>' and M[index(v), index(u)] == '-'`

ankurankan: I would also suggest adding the `get_neighbors` method. Usually, with orientation rules, we are interested in triples of variables with certain types of relationships with the neighbours. For example, we want to find variables that form the triplet of the form: `A o-> B o-> C`. The easiest way to implement these is to iterate over nodes, assume them as `B`, and check for neighboring nodes such that they meet the edge types. In the example above, we would need to find all neighboring nodes of `B` such that they have the edge `o->`. With the `get_neighbors` method, this would be `get_neighbours('B', u_type='>', v_type='o')`. For `dfs` algorithms, we would similarly need methods to find neighbours that satisfy the edge type requirements.

ankurankan (changes requested):

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,132 @@
...
+            self.graph.has_edge(u, v)
+            and self.graph.get_edge_data(u, v).get("mark") == "circle"
+        )
+
+    def has_tail_at(self, u, v):
+        return (
+            self.graph.has_edge(u, v)
+            and self.graph.get_edge_data(u, v).get("mark") == "tail"
+        )
+
```
ankurankan: These should not be needed once we have the `get_neighbors` method.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,132 @@
...
+class AncestralBase:
+    """
+    Base class for all ancestral graphical models.
+
+    Wraps a networkx.DiGraph to handle mixed graphs with
+    directed, bidirected, and partially directed edges.
+    """
+
+    def __init__(self, ebunch=None):
```
ankurankan: We need to accept latent variables as well similar to the DAG class.

pgmpy/base/AncestralBase.py
ankurankan: For the `dfs` method, I would suggest adding a `get_reachable_nodes` method, that does a DFS, and finds all the nodes that are reachable for the given node with a certain type of edge.

ankurankan (changes requested):
Please refer the comments.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+            A square matrix of shape (n_nodes, n_nodes) where M[i, j]
+            is the mark at node j for edge (i, j).
+        node_index : dict
+            Mapping from node label to row/col index.
+        """
+        nodes = list(self.nodes)
+        n = len(nodes)
+        node_index = {node: i for i, node in enumerate(nodes)}
+
+        M = np.full((n, n), "", dtype=object)
```
ankurankan: Having the default as `0`  instead of `""` is better as it is a default convention used everywhere.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+            u_mark, v_mark = data["marks"]
+
+            # Mark from u→v is v_mark (mark at v's end)
+            M[u_idx, v_idx] = v_mark
+            # Mark from v→u is u_mark (mark at u's end)
+            M[v_idx, u_idx] = u_mark
+
+        return M, node_index
+
+    def add_edge(self, u, v, u_mark, v_mark):
```
ankurankan, line 144: Missing docstring

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+        if u == v:
+            raise ValueError("Nodes cannot be the same for an edge.")
+        if u_mark not in {"-", ">", "o"} or v_mark not in {"-", ">", "o"}:
+            raise ValueError("Marks must be one of '-', '>', or 'o'.")
+        super().add_edge(u, v, marks=(u_mark, v_mark))
+
+    def add_edges_from(self, ebunch):
```
ankurankan, line 199: Missing docstring

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+        for u, v, u_mark, v_mark in ebunch:
+            self.add_edge(u, v, u_mark, v_mark)
+
+    def _get_marks(self, u, v):
+        """
+        Return (mark_at_u, mark_at_v) for the edge between u and v.
+        Works regardless of stored edge direction.
+        """
+        data = self.get_edge_data(u, v)
+        if data is not None:  # edge stored as (u, v)
```
ankurankan: Remove the comment

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+            return data["marks"]
+        data = self.get_edge_data(v, u)
+        if data is not None:  # edge stored as (v, u) → reverse marks
```
ankurankan: Same here

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+            u_mark, v_mark = data["marks"]
+            return v_mark, u_mark
+        raise ValueError(f"No edge between {u} and {v}")
+
+    def get_neighbors(self, node, u_type=None, v_type=None):
+        """
+        Return neighbors of a node that satisfy edge mark constraints.
```
ankurankan: docstring

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+        u_type = mark at neighbor's side (when going FROM neighbor TO node)
+        v_type = mark at node's side (when going FROM node TO neighbor)
+        """
+        if node not in self:
+            return set()
+        neighbors = set()
+        for neighbor in nx.all_neighbors(self, node):
+            try:
```
ankurankan: Why do we need a try-except block here?

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+        n = len(nodes)
+        node_index = {node: i for i, node in enumerate(nodes)}
+
+        M = np.full((n, n), "", dtype=object)
+
+        for u, v, data in self.edges(data=True):
+            u_idx, v_idx = node_index[u], node_index[v]
+            u_mark, v_mark = data["marks"]
+
+            # Mark from u→v is v_mark (mark at v's end)
```
ankurankan: Remove comments

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
+from collections import deque
+from typing import Hashable, Iterable, Optional
+
+import networkx as nx
+import numpy as np
+
+
+class AncestralBase(nx.DiGraph):
```
ankurankan: Do we need this to inherit from `DiGraph` or would it be easier with `Graph`? Combined with my comment on storing the marks as dict, this should make methods like `_get_marks` redundant; hence simplifying logic.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+                if (u_type is None or node_mark == u_type) and (
+                    v_type is None or neighbor_mark == v_type
+                ):
+                    neighbors.add(neighbor)
+            except ValueError:
+                continue
+        return neighbors
+
+    def get_parents(self, node):
+        """Get nodes that have '>' pointing TO this node"""
```
ankurankan: docstring

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,152 @@
...
+            M[v_idx, u_idx] = u_mark
+
+        return M, node_index
+
+    def add_edge(self, u, v, u_mark, v_mark):
+        if u == v:
+            raise ValueError("Nodes cannot be the same for an edge.")
+        if u_mark not in {"-", ">", "o"} or v_mark not in {"-", ">", "o"}:
+            raise ValueError("Marks must be one of '-', '>', or 'o'.")
+        super().add_edge(u, v, marks=(u_mark, v_mark))
```
ankurankan: I would suggest storing the marks as a dict of `node: mark` instead of a tuple. It would make it obvious which mark is for which node and would make it independent of the mark storing logic here.

ankurankan (changes requested):
I have added a few comments. Most of the methods still need docstrings.

pgmpy/base/AncestralBase.py
Spinachboul: yeah that makes sense:
does this implementation work?
```python
def add_edge(self, u, v, marks):
        if u == v:
            raise ValueError("Nodes cannot be the same for an edge.")
        
        if not isinstance(marks, dict) or set(marks.keys()) != {u, v}:
            raise ValueError("Marks must be a dict with keys as the two nodes.")
        u_mark = marks[u]
        v_mark = marks[v]
        valid_marks = {">", "-", "o"}
        if u_mark not in valid_marks or v_mark not in valid_marks:
            raise ValueError(f"Marks must be one of {valid_marks}.")
        super().add_edge(u, v, marks=(u_mark, v_mark))
```

ankurankan (changes requested):
@Spinachboul All of the docstrings need an Examples and Returns section. Looks good otherwise.

Spinachboul:
@ankurankan Added the examples and returns in the docstrings, kindly check.
Thanks!

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,211 @@
...
+        graph = AncestralBase()
+        self.assertEqual(len(graph.nodes), 0)
+        self.assertEqual(len(graph.edges), 0)
+
+    def test_init_with_edges(self):
+        """Test initialization with edge list."""
+        edges = [("A", "B", "-", ">"), ("B", "C", ">", "-")]
+        graph = AncestralBase(ebunch=edges)
+        self.assertEqual(len(graph.nodes), 3)
+        self.assertEqual(len(graph.edges), 2)
```
ankurankan, line 38: Could you add checks here that the correct edge marks are also stored here?

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,211 @@
...
+        """Test adding valid edges with different mark combinations."""
+        self.graph.add_edge("A", "B", "-", ">")
+        self.graph.add_edge("B", "C", ">", "-")
+        self.graph.add_edge("C", "D", "o", "o")
+
+        self.assertEqual(len(self.graph.edges), 3)
+        self.assertIn("A", self.graph.nodes)
+        self.assertIn("B", self.graph.nodes)
+        self.assertIn("C", self.graph.nodes)
+        self.assertIn("D", self.graph.nodes)
```
ankurankan: Same here. Add checks for the marks.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,211 @@
...
+        with self.assertRaises(ValueError):
+            self.graph.add_edge("A", "B", "z", "w")
+
+    def test_add_edges_from(self):
+        """Test adding multiple edges from list."""
+        edges = [("A", "B", "-", ">"), ("B", "C", ">", "-"), ("A", "C", "o", "o")]
+        self.graph.add_edges_from(edges)
+
+        self.assertEqual(len(self.graph.edges), 3)
+        self.assertEqual(len(self.graph.nodes), 3)
```
ankurankan: Same here.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,211 @@
...
+        self.assertEqual(neighbors_b, {"A"})
+
+    def test_get_neighbors_nonexistent_node(self):
+        """Test getting neighbors for non-existent node."""
+        neighbors = self.graph.get_neighbors("Z")
+        self.assertEqual(neighbors, set())
+
+    def test_get_parents(self):
+        """Test getting parent nodes."""
+        self.graph.add_edge("A", "B", "-", ">")
```
ankurankan: This example is too simple. Add edges other than only directed ones and check that correct ones are identified.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,211 @@
...
+
+        parents_d = self.graph.get_parents("D")
+        self.assertEqual(parents_d, {"B"})
+
+        parents_a = self.graph.get_parents("A")
+        self.assertEqual(parents_a, set())
+
+    def test_get_children(self):
+        """Test getting child nodes."""
+        self.graph.add_edge("A", "B", "-", ">")
```
ankurankan: Same here.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,211 @@
...
+
+        spouses_c = self.graph.get_spouses("C")
+        self.assertEqual(spouses_c, {"D"})
+
+        spouses_b = self.graph.get_spouses("B")
+        self.assertEqual(spouses_b, {"A"})
+
+    def test_get_ancestors(self):
+        """Test getting all ancestors."""
+        self.graph.add_edge("A", "B", "-", ">")
```
ankurankan: Same here.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,211 @@
...
+        self.assertEqual(ancestors_d, {"A", "B", "C", "D", "E"})
+
+        ancestors_c = self.graph.get_ancestors("C")
+        self.assertEqual(ancestors_c, {"A", "B", "C", "E"})
+
+        ancestors_a = self.graph.get_ancestors("A")
+        self.assertEqual(ancestors_a, set("A"))
+
+    def test_get_descendants(self):
+        """Test getting all descendants."""
```
ankurankan, line 103: Same here.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,440 @@
...
+
+
+class AncestralBase(nx.Graph):
+    def __init__(
+        self,
+        ebunch: Optional[Iterable[tuple[Hashable, Hashable]]] = None,
+        latents: set[Hashable] = set(),
+    ):
+        """
+        Ancestral graph base class.
```
ankurankan, line 15: Can you add a description here about the types of edges that are allowed an how they are represented.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,440 @@
...
+            The other endpoint of the edge.
+
+        u_mark : str
+            Mark at node u for edge (u, v). Must be one of {">", "-", "o"}.
+
+        v_mark : str
+            Mark at node v for edge (u, v). Must be one of {">",
+            "-", "o"}.
+
+        Raises
```
ankurankan: Missing example

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,440 @@
...
+            current = queue.popleft()
+            if current not in visited:
+                visited.add(current)
+                ancestors.add(current)
+                queue.extend(self.get_parents(current))
+        return ancestors
+
+    def get_descendants(self, node):
+        """
+        Get all descendant nodes (children, grandchildren, etc.)
```
ankurankan, line 407: Add that the variable itself is included.

ankurankan (changes requested):
The documentation and tests still need some work

Spinachboul:
@ankurankan In the docstrings, I thought of also adding the reverse directed edge, just to make the idea of edge representation clear to the users.
All the other changes have been made.
Kindly check!

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,235 @@
...
+        self.graph.add_edge("A", "B", "-", ">")
+        self.graph.add_edge("B", "C", "-", ">")
+        self.graph.add_edge("C", "D", "-", ">")
+        self.graph.add_edge("E", "C", "-", ">")
+
+        self.graph.add_edge("X", "B", "o", "o")
+        self.graph.add_edge("Y", "C", ">", ">")
+
+        ancestors_d = self.graph.get_ancestors("D")
+        self.assertEqual(ancestors_d, {"A", "B", "C", "D", "E", "Y"})
```
ankurankan: This is not correct. `Y` is not an ancestor. Every ancestor needs to have a directed path to the node. The path from `Y` to `D` here is `Y <-> C -> D`, which is not a directed path. Have a look at the definition on page 5: https://www.jmlr.org/papers/volume9/zhang08a/zhang08a.pdf. You will need to check both sides of the edges.

ankurankan (changes requested):
Refer comment

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,235 @@
...
+        self.graph.add_edge("A", "B", "-", ">")
+        self.graph.add_edge("C", "B", "-", ">")
+        self.graph.add_edge("B", "D", "-", ">")
+
+        self.graph.add_edge("X", "B", "o", "o")
+        self.graph.add_edge("Y", "B", ">", ">")
+        self.graph.add_edge("Z", "B", ">", "-")
+
+        parents_b = self.graph.get_parents("B")
+        self.assertEqual(parents_b, {"A", "C", "Y"})
```
ankurankan: This is not correct either. `Y` is not a parent. Refer the definition.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,235 @@
...
+        self.assertEqual(parents_a, set())
+
+    def test_get_children(self):
+        """Test getting child nodes."""
+        self.graph.add_edge("A", "B", "-", ">")
+        self.graph.add_edge("A", "C", "-", ">")
+        self.graph.add_edge("B", "D", "-", ">")
+
+        self.graph.add_edge("A", "X", "o", "o")
+        self.graph.add_edge("A", "Z", ">", "-")
```
ankurankan: Would suggest adding a bidirected edge in the graph here as well. Will help in catching issues like the ones we saw above.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,235 @@
...
+        """Test getting spouse nodes (bidirectional arrows)."""
+        self.graph.add_edge("A", "B", ">", ">")
+        self.graph.add_edge("A", "C", "-", ">")
+        self.graph.add_edge("C", "D", ">", ">")
+
+        self.graph.add_edge("B", "X", "-", ">")
+        self.graph.add_edge("C", "Y", "o", "o")
+
+        spouses_a = self.graph.get_spouses("A")
+        self.assertEqual(spouses_a, {"B"})
```
ankurankan: I would suggest being a bit thorough in the tests. For example, if we are testing for spouses of `A`, I would add every possible type of edge at `A`, and then we can make sure that we don't select any of those nodes accidentaly.

ankurankan (changes requested):

pgmpy/tests/test_base/test_AncestralBase.py
Spinachboul: Will this work, I am trying to add every possible edge type at A:
directed, bidirected, undirected
```
self.graph.add_edge("A", "B", ">", ">")
        self.graph.add_edge("A", "C", "-", ">")
        self.graph.add_edge("C", "D", ">", ">")
        self.graph.add_edge("A", "X", "o", "o")

        self.graph.add_edge("B", "X", "-", ">")
        self.graph.add_edge("C", "Y", "o", "o")
```

pgmpy/tests/test_base/test_AncestralBase.py
Spinachboul: Yeah, earlier I thought that solving issues would work but then there is a problem at the `get_parents` function implementation itself, where it only checks if `u_type== ">"` and not the other end.
Have fixed that!

pgmpy/tests/test_base/test_AncestralBase.py
Spinachboul: Same change in `get_children` as the above for `get_parents`

pgmpy/tests/test_base/test_AncestralBase.py
Spinachboul: Yeah that makes sense, will do that

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,486 @@
...
+        ebunch : Iterable[tuple], optional
+            An iterable of edges of the form (u, v, u_mark, v_mark) used to
+            initialize the graph. Each mark must be one of {">", "-", "o"}.
+            Default is None, which initializes an empty graph.
+
+        latents : set, optional
+            Set of latent (unobserved) variables in the graph. Default is
+            an empty set.
+
+        Edge Representation
```
ankurankan: You can't have a separate section. I don't think it will get parsed. The content of this section needs to be at the top.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,486 @@
...
+        # Undirected edge C — E
+        >>> g.add_edge("C", "E", "o", "o")
+        >>> g["C"]["E"]["marks"]
+        {'C': 'o', 'E': 'o'}
+
+
+        Raises
+        ------
+        ValueError
+            If marks are invalid or nodes are the same.
```
ankurankan: Returns section is still missing.

pgmpy/base/AncestralBase.py
```diff
@@ -0,0 +1,486 @@
...
+        super().add_edge(u, v, marks={u: u_mark, v: v_mark})
+
+    def add_edges_from(self, ebunch):
+        """
+        Add multiple edges from an iterable of (u, v, marks) tuples.
+
+        Parameters
+        ----------
+        ebunch : Iterable[tuple]
+            Each tuple should be of the form (u, v, u_mark, v_mark)."""
```
ankurankan: Missing examples section and returns section

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -0,0 +1,239 @@
...
+        self.assertEqual(M.shape, (3, 3))
+        self.assertEqual(len(node_index), 3)
+
+        self.assertIn("A", node_index)
+        self.assertIn("B", node_index)
+        self.assertIn("C", node_index)
+
+        a_idx, b_idx, c_idx = node_index["A"], node_index["B"], node_index["C"]
+
+        self.assertEqual(M[a_idx, b_idx], ">")
```
ankurankan: You can directly compare matrices or 2D list instead of single elements.

ankurankan (changes requested):
Have added a few comments. I would suggest taking a bit of time and going through everything in a bit more detail, as similar things have been missing in multiple iterations of review.

About the tests for methods (like neighbors, parent, children, etc.), we are creating graphs in each of the tests. I think it would be cleaner if you define a complex graph in the setup method with at least one node that has all types of possible edges. This would give you a uniform graph on which you can test every method.

Spinachboul:
@ankurankan Apologies! I might have not been looking at the classes carefully.
So I have defined the edges like this:
```
edges = [
            ("A", "B", "-", ">"),
            ("C", "A", "-", ">"),
            ("A", "D", "o", "o"),
            ("A", "E", ">", ">"),
            ("B", "X", "-", ">"),
            ("Y", "C", "-", ">"),
        ]
```
That potentially means, the graph is going to look something like this:
image



- Question A) **Should I be continuing with this graph itself (which I have made intentionally complex) or some other graph structure would be more suitable (int the `Setup` method)**

- Question B) **For all the relationship methods, should I be testing only for a couple of nodes (including the one which has all the edge types) or all the nodes present in the graph?**

ankurankan:
@Spinachboul 

> Should I be continuing with this graph itself (which I have made intentionally complex) or some other graph structure would be more suitable (int the Setup method)

You are only considering 4 types of edges at `A`. However, in total, there are 9 types of possible edges. The missing types are: 1. Undirected Edge: (A, F, '-', '-'), 2. Only one side is undecided in orientation (A, G, '-', 'o'), (A, H, 'o', '-'), (A, I, '>', 'o'), (A, J, 'o', '>'). There should be edges like this at `A` as well, to make sure that each method is able to correctly filter between all edge types.

> For all the relationship methods, should I be testing only for a couple of nodes (including the one which has all the edge types) or all the nodes present in the graph?

Testing for a couple of them is fine. The important thing is to make sure that each method has to consider all edge types to make sure they are working correctly. So, node `A` should definitely be used for testing in the above graph.

Spinachboul:
@ankurankan So as per your suggestions, I am taking the following graph for testing

here is how I defined the edges:
```
edges = [
            ("A", "B", "-", ">"),
            ("A", "C", ">", "-"),
            ("A", "D", "o", "o"),
            ("A", "E", ">", ">"),
            ("A", "F", "-", "-"),
            ("A", "G", "-", "o"),
            ("A", "H", "o", "-"),
            ("A", "I", "o", ">"),
            ("A", "J", ">", "o"),
        ]
```

Here is the visual representation of the same
image

ankurankan:
@Spinachboul I would still keep some of the paths that you had in the previous graph in addition to the graph above (i.e., some nodes that won't be neighbors of `A`)

Spinachboul:
@ankurankan Oh Ok, I will include that too
Thanks!

ankurankan (approved):
@Spinachboul Looks good. I would, however, suggest using the big graph for adjacency matrix tests as well. We can add it in another PR though.

Spinachboul:
> @Spinachboul Looks good. I would, however, suggest using the big graph for adjacency matrix tests as well. We can add it in another PR though.

Yeah! I would be doing that tomorrow.
Thanks

PR #2288, opened 2025-08-12
Add DoubleMLRegressor 

### LINKED ISSUE #2268: [ENH] DoubleML Regressor in sklearn Compatible Design
Implement a DoubleMl Regressor in sklearn compatible design

- should be a scikit-learn compatible regressor - so fit/predict API
- should pass parametrize_with_checks for regressors
- init parameter is DAG (after merge of "roles PR") with roles defined for exposure, target (can be inferred) and adjustment set
- Reference prototype implementation: https://github.com/pgmpy/pgmpy/pull/2246

**Additional context**

DARHWOLF:

- Fixes #2268 

- 
-
-


marrov:
I think one key aspect is wether we want to re-implement DML or wrap an existing library. The options are:

### Option A: Keep a custom DML implementation in pgmpy (current draft PR)
Pros:
- Tight integration with pgmpy DAG roles and causal workflow.
- Minimal dependencies and smaller install footprint.
- Full control over API and behavior tailored to pgmpy.

Cons:
- Requires implementing and maintaining correct cross-fitting, which is currently the main technical issue in this PR (to be discussed in another comment); harder to match maturity of dedicated libraries.
- New methods require re-implementing rather than extension through a compatibility layer (adapter patter, something like `BaseDMLRegressor`)

### Option B: Wrap doubleml
Pros:
- Actively maintained with statistically robust implementations & established references.
- Lighter dependency than econml
- Easy to adapt pgmpy roles to DoubleML’s design matrices: exposure (X), outcome (Y), and adjustment/covariates (Z).

Cons:
- Adds a dependency (but can be optional via extras).
- API mismatch: DoubleML has its own estimator classes (not strictly sklearn BaseEstimator), though it consumes sklearn-style learners.

### Option C: Wrap econml
Pros:
- Broad method coverage beyond DML

Cons:
- Heavier dependency footprint and larger install size.
- API sometimes more complex for simple PLR DML use cases.
- Overkill if we only needs standard DML; more surface area to maintain compatibility.

Considering the above, my vote is for B. Thoughts @DARHWOLF, @fkiraly, @ankurankan?

marrov:
The main technical issue in this draft PR is that the DML algorithms requires cross-fitting of some kind to justify orthogonality, which is currently not implemented. `n_folds` is provided but not used. I would suggest that you implement this, but maybe the comment above on wrapping vs reimplementing makes this comment obsolete.

pgmpy/tests/test_models/test_DoubleMLRegressor.py
```diff
@@ -0,0 +1,27 @@
+from sklearn.dummy import DummyRegressor
+from sklearn.utils.estimator_checks import parametrize_with_checks
+
+from pgmpy.models.DoubleMLRegressor import DoubleMLRegressor
+
+
+def make_estimator():
+    class DummyDAG:
```
marrov: Update to use `pgmpy`'s DAG class

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+                        "Only single exposure supported in " "skeleton."
+                    )
+                treat = exp_list[0] if exp_list[0] in df.columns else df.columns[0]
+
+        # adjustments priority: explicit -> dag.get_adjustment_set() -> []
+        if self.adjustment_set is not None:
+            adj = [c for c in self.adjustment_set if c in df.columns and c != treat]
+        else:
+            adj = []
+            if hasattr(self.dag, "get_adjustment_set"):
```
marrov: `get_adjustment_set` is not method of the `DAG` (nor the mixin) - use `dag.get_role("adjustment")`

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        return pd.DataFrame(arr, columns=cols)
+
+    def _resolve_treatment_and_adjustments(
+        self, df: pd.DataFrame
+    ) -> Tuple[str, List[str]]:
+        # treatment priority: explicit -> dag.exposure -> first column
+        if self.treatment_col is not None and self.treatment_col in df.columns:
+            treat = self.treatment_col
+        else:
+            exp = getattr(self.dag, "exposure", None)
```
marrov: Similar here, `.exposure` is not an attribute - you should use the `get_role()` method instead.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        arr = np.asarray(X)
+        if arr.ndim == 1:
+            arr = arr.reshape(-1, 1)
+        cols = [f"x{i}" for i in range(arr.shape[1])]
+        return pd.DataFrame(arr, columns=cols)
+
+    def _resolve_treatment_and_adjustments(
+        self, df: pd.DataFrame
+    ) -> Tuple[str, List[str]]:
+        # treatment priority: explicit -> dag.exposure -> first column
```
marrov: Not sure I agree with this treatment of roles - I think we should only support DAG roles. Passing it explicitly overrides the DAG, so they could be mismatched and the first col approach if there is no DAG roles is a potential source of a silent failure. Current fallback `treat = df.columns[0]` is dangerous. IMO, first validate that X is a pandas DataFrame and that all role-referenced nodes exist in X.columns, and throw an error otherwise. This will avoid silent misalignment.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+    def _as_dataframe(self, X) -> pd.DataFrame:
+        if isinstance(X, pd.DataFrame):
+            return X.copy()
+        arr = np.asarray(X)
+        if arr.ndim == 1:
+            arr = arr.reshape(-1, 1)
+        cols = [f"x{i}" for i in range(arr.shape[1])]
+        return pd.DataFrame(arr, columns=cols)
+
+    def _resolve_treatment_and_adjustments(
```
marrov: I think you can use the `is_valid_causal_structure()` method and then `get_roles` and check that against a list of must have roles

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, RegressorMixin, clone
+from sklearn.utils.validation import check_is_fitted, check_X_y
+
+
+class DoubleMLRegressor(BaseEstimator, RegressorMixin):
+    def __init__(
+        self,
+        dag: Any,
```
marrov: I think `dag` should be typed as `DAG` from pgmpy explicitly

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        estimator_g: Any,
+        estimator_m: Optional[Any] = None,
+        adjustment_set: Optional[Sequence[str]] = None,
+        treatment_col: Optional[str] = None,
```
marrov: Remove override, use `DAG` role exclusively

pgmpy/models/__init__.py
```diff
@@ -8,15 +8,16 @@
 from .JunctionTree import JunctionTree
 from .LinearGaussianBayesianNetwork import LinearGaussianBayesianNetwork
 from .MarkovChain import MarkovChain
-from .MarkovNetwork import MarkovNetwork
+
+# from .MarkovNetwork import MarkovNetwork
```
marrov: Remove comment

pgmpy/models/__init__.py
```diff
@@ -8,15 +8,16 @@
...
 from .NaiveBayes import NaiveBayes
 from .SEM import SEM, SEMAlg, SEMGraph
 
 __all__ = [
     "DiscreteBayesianNetwork",
     "BayesianNetwork",
     "DiscreteMarkovNetwork",
-    "MarkovNetwork",
+    "DoubleMLRegressor" "MarkovNetwork",
```
marrov: Add comma to properly separate, this will throw an error. Also, import the DML regressor properly: `from .DoubleMLRegressor import DoubleMLRegressor`.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
+"""flake8: noqa"""
```
marrov: Why are you disabling linting here?

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+
+from typing import Any, List, Optional, Sequence, Tuple
+
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, RegressorMixin, clone
+from sklearn.utils.validation import check_is_fitted, check_X_y
+
+
+class DoubleMLRegressor(BaseEstimator, RegressorMixin):
```
marrov: Missing class docstring

pgmpy/models/DoubleMLRegressor.py
marrov: Furthermore, I think we should limit ourselves to single treatment & single outcome

pgmpy/tests/test_models/test_DoubleMLRegressor.py
marrov, line 1: I would like to see a more comprehensive test with synthetic data and use of the `parametrize_with_checks` decorator for testing similar to how @XAheli has used it on this PR #2295.

fkiraly (changes requested):
Great start!

* I think the `predict` method is wrong - it is not using the fitted estimators at all.
* we should the `pgmpy` `DAG` class and its methods, instead of a custom DAG class.
* the `adjustment_set` and `treatment_col` parameter are redundant with the DAG, this should be present in the roles of the DAG.
    * of course we might like to give an alternative option to pass these, e.g., as a dict in place of the DAG
* should we name the `dag` variable maybe `model`?
* we should try to remove the case treatment for `X_for_g` etc. Not sure if it is possible
* why are we using `predict_proba` sometimes in `fit`? Please explain
* should the regressor have `predict_proba`? That would be easy to obtain, simply dispatch to the `predict_proba` of the underlying estimators.

pgmpy/models/DoubleMLRegressor.py
fkiraly: > Furthermore, I think we should limit ourselves to single treatment & single outcome

Hm, I disagree with this - why, if we can easily have it in a general form?

fkiraly:
> Considering the above, my vote is for B. Thoughts @DARHWOLF, @fkiraly, @ankurankan?

I would prefer option A for the start, then maybe B.

Why A? In order to learn about the integration between the new DAG featuers and the new API design for predictive causal models. Juggling an external dependency may make this phase of explorative development more difficult.

Once we have a more consolidated API, then B might prove to be a good approach, I would re-review it based on what we learnt in A. As a pre-condition, we also need to have some degree of soft dependency isolation, and a working design for that aspect of the library (currently it is suboptimal, e.g., the `torch` dep discussion)

pgmpy/models/DoubleMLRegressor.py
ankurankan: This can be `DAG`, `PDAG`, or `ADMG`. All of them would have will have the same roles as DAG: `exposure`, `outcome`, and `adjustment`

pgmpy/models/DoubleMLRegressor.py
marrov: Just for simplicity of this first PR I thought it would limit the scope but no harm in having this from the get-go.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def _resolve_treatment_and_adjustments(
+        self, df: pd.DataFrame
+    ) -> Tuple[str, List[str]]:
+        # treatment priority: explicit -> dag.exposure -> first column
+        if self.treatment_col is not None and self.treatment_col in df.columns:
+            treat = self.treatment_col
+        else:
+            exp = getattr(self.dag, "exposure", None)
+            if exp is None:
```
ankurankan: Shouldn't use a default value here. Throw an error if exposure is not defined to be explicit.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+                    )
+                treat = exp_list[0] if exp_list[0] in df.columns else df.columns[0]
+
+        # adjustments priority: explicit -> dag.get_adjustment_set() -> []
+        if self.adjustment_set is not None:
+            adj = [c for c in self.adjustment_set if c in df.columns and c != treat]
+        else:
+            adj = []
+            if hasattr(self.dag, "get_adjustment_set"):
+                try:
```
ankurankan: Assume that the adjustment set is already present in the model. If not, throw an error.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        self.adjustment_set = adjustment_set
+        self.treatment_col = treatment_col
+        self.n_folds = n_folds
+        self.random_state = random_state
+
+    def _more_tags(self):
+        return {"requires_y": True, "no_sparse_input": True}
+
+    # ---------- small helpers ----------
+    def _as_dataframe(self, X) -> pd.DataFrame:
```
ankurankan: I am not sure if this method is valid. Given an array, we have no way to know which column corresponds to which variable in the model.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+                raise ValueError(
+                    f"Predict input has fewer columns than required: "
+                    f"need {n_needed}, got {X_df.shape[1]}"
+                )
+            mapped_cols = list(X_df.columns[:n_needed])
+        else:
+            mapped_cols = self._design_columns
+
+        X_design = X_df[mapped_cols].to_numpy(dtype=float)
+        preds = self.intercept_ + X_design.dot(self.coef_)
```
ankurankan: This should be: `preds = self.intercept_ + self.treatment_effect_ * X[exposure_variable] + ml_g.predict(X[adjustment])`

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        # fit m (treatment mechanism)
+        t_vec = df[treat_col].to_numpy(dtype=float)
+        if X_for_g.shape[1] == 0:
+            m_hat = np.repeat(t_vec.mean(), len(t_vec))
+            self.estimator_m_ = None
+        else:
+            ml_m.fit(X_for_g, t_vec)
+            self.estimator_m_ = ml_m
+            # if classifier, prefer predict_proba for propensity
+            if hasattr(ml_m, "predict_proba"):
```
ankurankan: I think we should focus just on the continuous case first to keep it simple.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        self.g_hat_ = g_hat
+        self.m_hat_ = m_hat
+
+        # orthogonalization
+        y_res = y_vec - self.g_hat_
+        t_res = t_vec - self.m_hat_
+
+        # regress y_res on t_res (with intercept)
+        X_res = np.column_stack([np.ones(len(t_res)), t_res])
+        theta_coef, *_ = np.linalg.lstsq(X_res, y_res, rcond=None)
```
ankurankan: Would suggest using the linear regression implementation from statsmodels, it will also give us summary stats for the fit.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+
+        intercept = float(theta_coef[0])
+        theta = float(theta_coef[1])  # single-treatment effect
+
+        # store both a clear treatment_effect_ and a coef_ vector
+        self.treatment_effect_ = theta
+        # create coef_ vector aligned with design_cols
+        # position 0 is treatment, rest correspond to adjustments
+        self.coef_ = np.concatenate(
```
ankurankan: Could you help me understand what is going on here? For a single exposure and single outcome case, we should have only one coefficient.

ankurankan (changes requested):

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        feature_names = list(X.columns)
+
+        self.n_features_in_ = X_arr.shape[1]
+        self.feature_names_in_ = np.array(feature_names, dtype=object)
+
+        df = X.copy()
+        df["outcome"] = np.asarray(y_arr).ravel()
+
+        treat_col, adj_cols = self._resolve_treatment_and_adjustments(df)
+        design_cols = [treat_col] + [c for c in adj_cols if c != treat_col]
```
ankurankan: There should never be any common variables between `treatment` and `adjustment`.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+
+        # clone nuisance learners and fit (full-sample, simple) ----
+        ml_g = clone(self.estimator_g)
+        ml_m = (
+            clone(self.estimator_m)
+            if self.estimator_m is not None
+            else clone(self.estimator_g)
+        )
+
+        adj_cols_for_nuisance = [c for c in design_cols if c != treat_col]
```
ankurankan: Do we need to redefine this variable? This is same as `adj_cols` above. Correct?

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,188 @@
...
+        if len(adj_cols_for_nuisance) == 0:
```
ankurankan: If there is no adjustment variable, you can skip the two residual model training and directly train the linear regression models on the exposure and outcome variables themselves.

pgmpy/models/DoubleMLRegressor.py
DARHWOLF: any suggestions to remedy this?
How do we prevent mis-mapping*? 
i am thinking that we should add an optional `feature_names` param be passed at `init`

fkiraly: what do you mean with "mipmapping" - that is a computer graphics term which I am not sure how it applies
https://en.wikipedia.org/wiki/Mipmap

DARHWOLF: My apologies, i meant "mis-mapping".
I meant mis-mapping, for example selecting the wrong column as the treatment or include/exclude wrong covariates

pgmpy/models/DoubleMLRegressor.py
DARHWOLF: Removed

pgmpy/models/DoubleMLRegressor.py
DARHWOLF: used `get_role()` method

pgmpy/models/DoubleMLRegressor.py
DARHWOLF: Done

pgmpy/models/DoubleMLRegressor.py
DARHWOLF: Done

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,428 @@
...
+        # reject ambiguous arrays
+        raise ValueError(
+            "Ambiguous input: X is a NumPy array but no feature_names were provided. "
+            "Either pass X as a pandas DataFrame with column names matching the DAG roles, "
+            "or construct the estimator with feature_names=[...]."
+        )
+
+    def _read_roles(self) -> Tuple[str, List[str]]:
+        if not (
+            hasattr(self.dag, "get_role")
```
XAheli: Quick suggestion: maybe instead of checking this you can directly implement the `get_role` function! Thoughts @ankurankan @marrov ?

marrov: @DARHWOLF, I agree with @XAheli  here - given that you are already forcing that `dag` should be a `pgmpy` DAG it is ok to directly use `get_role`. This method is essentially an undercorver type validation. Just let it fail loudly if the user does not input a `DAG`.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,428 @@
...
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, RegressorMixin, clone
+from sklearn.model_selection import KFold, StratifiedKFold
+from sklearn.utils.validation import check_is_fitted, check_X_y
+
+from pgmpy.base.DAG import DAG
+
+
+class _SklearnTags:
```
marrov: Not an expert in `sklearn` tag's so would love @fkiraly's thoughts on this approach but I looks quite over-engineered. A quick search on the sklearn guidelines proposes a much simper approach: https://scikit-learn.org/stable/developers/develop.html#estimator-tags - Can you elaborate on why this simpler approach is not enough? 

Note: same as Ankur's comment in Aheli's PR (https://github.com/pgmpy/pgmpy/pull/2295#discussion_r2299144999)

fkiraly (changes requested):
Tests are failing *at collection*, i.e,. not when running.

Can you kindly ensure that tests (e.g., `parametrize_with_checks`) at least run?

To explain, this is a failure likely in some import or constructor *before tests are even run*.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,428 @@
...
+        return self._tags.items()
+
+    def __getitem__(self, key):
+        return self._tags[key]
+
+    def __repr__(self):
+        return f"_SklearnTags({self._tags!r}, input_tags={self.input_tags!r})"
+
+
+class DoubleMLRegressor(BaseEstimator, RegressorMixin):
```
marrov: I think this error is pretty informative so I'll just paste it as-is:
```
RuntimeError: Estimator DoubleMLRegressor seems to be a Regressor, but the `regressor_tags` tag is not set. Either set the tag manually or inherit from the RegressorMixin. Note that the order of inheritance matters, the RegressorMixin should come before BaseEstimator.
```

pgmpy/models/DoubleMLRegressor.py
marrov: As discussed today, in order to maintain compatibility using `sklearn-compat` is probably a good idea: https://github.com/sklearn-compat/sklearn-compat?tab=readme-ov-file#tags-__sklearn_tags__-and-estimator-tags

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,428 @@
...
+    ):
+
+        self.dag = dag
+        self.estimator_g = estimator_g
+        self.estimator_m = estimator_m
+        self.n_folds = int(n_folds)
+        self.random_state = random_state
+
+        # implementing feature names for mapping of NumPy arrays
+        if feature_names is None:
```
marrov: Another error that's self explanatory.
```
RuntimeError: Cannot clone object DoubleMLRegressor(dag=<pgmpy.base.DAG.DAG object>, estimator_g=DummyRegressor(), estimator_m=DummyRegressor(), feature_names=['x0', 'x1', 'x2'], n_folds=3, random_state=0), as the constructor either does not set or modifies parameter feature_names
```
The appropriate [sklearn resource on the topic](https://sklearn.org/stable/developers/develop.html#instantiation) is also quite easy to follow.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+
+    def __init__(
+        self,
+        dag: DAG,
+        estimator_g: Any,
+        estimator_m: Optional[Any] = None,
+        n_folds: int = 5,
+        random_state: Optional[int] = None,
+        feature_names: Optional[Sequence[str]] = None,
```
marrov: We should avoid the `feature_names` param. Check @XAheli's implementation and the comments there for some guidance.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+            object.__setattr__(self, name, container)
+            return container
+        # internal attrs raise
+        if name.startswith("_"):
+            raise AttributeError(name)
+        # default for boolean-like probes
+        return False
+
+    def __repr__(self):
+        return f"_SklearnTagsDict({dict(self)!r}, input_tags={self.input_tags!r})"
```
ankurankan: Do we need these? I don't see it getting used anywhere.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+        Outcome nuisance model prototype (must implement fit/predict).
+    estimator_m : estimator-like or None
+        Treatment nuisance model prototype (if None, estimator_g is used for both).
+    n_folds : int
+        Number of folds for cross-fitting (>=2).
+    random_state : int or None
+        Random seed for folding.
+    allow_array_unnamed : bool
+        If True, allow NumPy arrays without feature_names by automatically naming columns
+        'x0','x1',... (unsafe for real causal work). Default False.
```
ankurankan: Missing `Examples` and `References` sections in the docstring.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+    """
+
+    def __init__(
+        self,
+        dag: DAG,
+        estimator_g: Any,
+        estimator_m: Optional[Any] = None,
+        n_folds: int = 5,
+        random_state: Optional[int] = None,
```
ankurankan: Could you change `random_state` to `seed`? We use `seed` as the default argument name across `pgmpy`, so we should keep it consistent.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+          columns will be set to feature_names (order assumed correct).
+        - If X is an array-like: convert to ndarray, reshape if 1-D, and set columns.
+        - If feature_names is provided, use it; otherwise generate generic names
+          feature_0, feature_1, ...ror instructing the user to pass a DataFrame or feature_names.
+        """
+        if isinstance(X, pd.DataFrame):
+            return X.copy()
+        arr = np.asarray(X)
+        if arr.ndim == 1:
+            arr = arr.reshape(-1, 1)
```
ankurankan: Does this condition ever get triggered? I was under the assumption that sklearn always passes a 2-D array, even if it is a single column.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+            hasattr(self.dag, "get_role")
+            and hasattr(self.dag, "is_valid_causal_structure")
+        ):
+            # if DAG class is available check type; otherwise raise guidance
+            if isinstance(DAG, type) and not isinstance(self.dag, DAG):
+                raise ValueError(
+                    "dag must be an instance of pgmpy's DAG or implement get_role/is_valid_causal_structure."
+                )
+            # else proceed but still require get_role/is_valid_causal_structure
+            if not hasattr(self.dag, "get_role") or not hasattr(
```
ankurankan: I think all the model classes are now guaranteed to have both `get_role` and `is_valid_causal_structure` methods. We do not need these checks.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+            feature_names = [f"feature_{i}" for i in range(arr.shape[1])]
+        return pd.DataFrame(arr, columns=feature_names)
+
+    def _read_roles(self) -> Tuple[str, List[str]]:
+        if not (
+            hasattr(self.dag, "get_role")
+            and hasattr(self.dag, "is_valid_causal_structure")
+        ):
+            # if DAG class is available check type; otherwise raise guidance
+            if isinstance(DAG, type) and not isinstance(self.dag, DAG):
```
ankurankan: I am not sure what exactly `isinstance(DAG, type)` is supposed to check, and when it will fail.

pgmpy/models/DoubleMLRegressor.py
ankurankan: This is also a repeated check. Line 180 has the same checks.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+        try:
+            self.dag.is_valid_causal_structure()
+        except Exception as e:
+            raise ValueError(f"DAG validation failed: {e}")
+
+        exposure_list = self.dag.get_role("exposure") or []
+        if not exposure_list:
+            raise ValueError(
+                "DAG must define an 'exposure' role. Use dag.with_role('exposure', var)."
+            )
```
ankurankan: Doesn't the `is_valid_causal_structure` method already perform this check?

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+            ):
+                raise ValueError(
+                    "dag must implement get_role(role) and is_valid_causal_structure()."
+                )
+
+        # Validate the DAG; is_valid_causal_structure raises useful errors if invalid
+        try:
+            self.dag.is_valid_causal_structure()
+        except Exception as e:
+            raise ValueError(f"DAG validation failed: {e}")
```
ankurankan: I don't think a try-except block is needed here. The `is_valid_causal_structure` should automatically throw an error with a relevant error message.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+                "DAG must define an 'exposure' role. Use dag.with_role('exposure', var)."
+            )
+        if len(exposure_list) != 1:
+            raise NotImplementedError(
+                "This estimator supports exactly one exposure variable."
+            )
+        exposure_col = exposure_list[0]
+
+        # adjustments
+        adj_raw = self.dag.get_role("adjustment") or []
```
ankurankan: `get_role` by default returns `[]` if there are no adjusment variables.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+        if isinstance(adj_raw, (list, tuple, set)):
+            adj_list = list(adj_raw)
+        elif adj_raw is None:
+            adj_list = []
+        else:
+            adj_list = [adj_raw]
```
ankurankan: Not sure if this is needed.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+        missing = set(required_features) - set(X_df.columns)
+        if missing:
+            raise ValueError(
+                f"Missing required columns in input data: {sorted(missing)}. Required columns: {required_features}"
+            )
+
+        # return DataFrame with exact ordering of required features
+        return X_df[required_features].copy()
+
+    # -----Fit ---- Predict ---
```
ankurankan: This comment is not needed.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+        # validate input & set sklearn convention attributes
+        X_arr, y_arr = check_X_y(
+            X, y, accept_sparse=False, ensure_2d=True, force_all_finite=True
+        )
+        self.n_features_in_ = X_arr.shape[1]
+
+        try:
+            n_folds_requested = int(self.n_folds)
+        except Exception:
+            raise ValueError(f"n_folds must be integer-like; got {self.n_folds!r}")
```
ankurankan: Can use `isinstance` for the check here instead of try-except block

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+        if exposure_col not in X_df.columns:
+            raise ValueError(
+                f"Exposure '{exposure_col}' not found in input columns for predict()."
+            )
+
+        # compute g_pred using stored nuisance models or y_mean_ fallback
+        if (
+            hasattr(self, "estimator_g_")
+            and self.estimator_g_ is not None
+            and len(adj_cols) > 0
```
ankurankan: Isn't this check already getting done by `check_is_fitted` in Line 432?

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,467 @@
...
+            ml_m = (
+                clone(self.estimator_m)
+                if self.estimator_m is not None
+                else clone(self.estimator_g)
+            )
+
+            X_train = (
+                X_for_nuisance[train_idx]
+                if X_for_nuisance.size
+                else np.empty((train_idx.shape[0], 0))
```
ankurankan: In what situation would we have size = 0 here?

ankurankan (changes requested):
I have added a few comments. I think there is a lot of scope in improving the readability of code, for example, there are many redundant checks, and we can get rid of some of the methods. Some other suggestions for improving the readability: 
1. Create new methods only when the method logic is self-contained.
2. Try to have all the validation code at the start of each method to keep the core algorithm clean.
3. In pgmpy, we try to add `Step` comments on each block of code/algorithm so the pseudocode of the algorithm is obvious from the comments. It also helps in navigating the algorithm; for example, see the comments in: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/estimators/expert.py.

ankurankan:
@DARHWOLF The error that you are getting is due to the `predict` method returning different predictions when `sample_weight` is specified vs when the samples themself account for the weights. Ideally, both approaches should return equivalent predictions, but that's not the case in the current implementation.

The reason is that the `fit` method is currently ignoring the `sample_weight` argument. A straightforward fix would be to set `n_fold=1` in the test regressor and ensure that `sample_weight` is passed to all nuisance and effect models. For `n_fold > 1`, achieving equivalent predictions is a bit difficult, since the fold based grouping would lead to different fitted models. 

@fkiraly @marrov Thoughts?

pgmpy/models/DoubleMLRegressor.py
ankurankan, line 1: This needs to be under `pgmpy/prediction`. You can also refer @XAheli 's PR for the directory structure.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+    random_state : int or None
+        Random seed for folding.
+    allow_array_unnamed : bool
+        If True, allow NumPy arrays without feature_names by automatically naming columns
+        'x0','x1',... (unsafe for real causal work). Default False.
+    """
+
+    def __init__(
+        self,
+        dag: DAG,
```
ankurankan: `dag` -> `casual_graph`. And this can be either a DAG, PDAG, ADMG, MAG, or PAG.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+from sklearn.linear_model import LinearRegression
+from sklearn.model_selection import KFold
+from sklearn.utils.validation import check_is_fitted, validate_data
+
+from pgmpy.base.DAG import DAG
+
+
+class DoubleMLRegressor(RegressorMixin, BaseEstimator):
+    """
+    Double-ML skeleton (single-exposure) with cross-fitting and explicit feature_names support.
```
ankurankan: I think we should mathematically describe here what exactly the fitting and prediction procedure is. There are a lot of variants of these DoubleML estimators, so it is good to be explicit here.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+
+    Parameters
+    ----------
+    dag : object
+        DAG object
+    estimator_g : estimator-like
+        Outcome nuisance model prototype (must implement fit/predict).
+    estimator_m : estimator-like or None
+        Treatment nuisance model prototype (if None, estimator_g is used for both).
```
ankurankan: I think we should have more descriptive names for these two arguments. Maybe `exposure_model` and `outcome_model` or `exposure_estimator` and `outcome_estimator`? Another idea could be nuisance_estimators that accept either a single estimator or a tuple of estimators. @fkiraly @marrov Thoughts?

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+    n_folds : int
+        Number of folds for cross-fitting (>=2).
```
ankurankan: We should allow any value > 0. `n_folds=1` should imply no cross-fitting.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+    random_state : int or None
+        Random seed for folding.
+    allow_array_unnamed : bool
+        If True, allow NumPy arrays without feature_names by automatically naming columns
```
ankurankan: I don't think we need this argument. By default, we should numpy arrays to pandas dataframe in the `fit` method.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+                X.columns = [f"x{i}" for i in range(X.shape[1])]
+            return X
+
+        else:
+            arr = np.asarray(X)
+            if arr.ndim == 1:
+                arr = arr.reshape(-1, 1)
+            if feature_names is None:
+                feature_names = [f"x{i}" for i in range(arr.shape[1])]
+            return pd.DataFrame(arr, columns=feature_names)
```
ankurankan: I think we need to change the logic of this whole method. If a numpy array is passed, we simply convert it to a data frame with column names `0, 1, 2 ...`. This dataframe would then be of the form that `check_estimators` also passes. @XAheli has implemented this new behavior; you can review that PR as well.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def _read_roles(self) -> Tuple[str, List[str]]:
+        """
+        Read roles from DAG without mutating the original user-supplied DAG.
+        """
+        if not isinstance(self.dag, DAG):
+            raise ValueError("causal_graph must be an instance of pgmpy's DAG class.")
+        dag_copy = copy.deepcopy(self.dag)
```
ankurankan: No need to use `deepcopy` anymore. The problem with changing object has been fixed now.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+        return X_df[required_features].copy()
+
+    def fit(self, X, y, sample_weight: Optional[Any] = None):
+        # Step 0: Validate inputs and arguments.
+        try:
+            n_folds = int(self.n_folds)
+            if n_folds < 1:
+                raise ValueError
+        except (TypeError, ValueError):
+            raise ValueError("n_folds must be an integer >= 1")
```
ankurankan: No need to use try-except here. Check if `self.n_folds` is an integer using the `isinstance` method and if not, throw an error. Then check if it is greater than or equal to 1; if not, throw an error.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+
+        X_arr, y_arr = validate_data(
+            self, X, y, accept_sparse=False, ensure_2d=True, force_all_finite=True
+        )
```
ankurankan: I recently found out that `validate_data` returns both `X_arr` and `y_arr` as numpy arrays even if `X` and `y` were dataframes. I would suggest using `validate_data` to just do the validation and then in the code below use the original `X` and `y` instead of `X_arr` and `y_arr`.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+        # Step 2: Prepare nuisance covariates excluding treatment
+        if len(adj_cols) == 0:
+            # Use an intercept only column to ensure estimators recieve a 2D array when adj_col is empty
+            covariates_df = np.ones((n_samples, 1), dtype=float)
+        else:
+            # use the adjustment columns as the features for nuisance models
+            covariates_df = df[adj_cols].to_numpy(dtype=float)
+
+        target_vec = df["outcome"].to_numpy(dtype=float)
+        exposure_vec = df[exposure_col].to_numpy(dtype=float)
```
ankurankan: Do we need to convert these to numpy arrays? The sklearn fit/predict methods should be able to directly accept dataframe objects.

pgmpy/tests/test_models/test_DoubleMLRegressor.py
```diff
@@ -0,0 +1,137 @@
...
+
+    return G
+
+
+def make_estimator_for_checks():
+    """
+    Return an unfitted DoubleMLRegressor instance configured to accept numpy arrays.
+    Important: passes feature_names so sklearn's tests (which passes ndarrays) can map columns.
+    """
+    G = make_role_dag(
```
ankurankan: No need for the `make_role_dag` method. This can, for example, be: `DAG([('x0', 'y'), ('x0', 'x1'), ('x0', 'x2')], roles={'exposure': 'x0', 'outcome': 'y'})`. Also, with the new default behavior of the `fit` method, these variable names need to be `0`, `1`, `2`, ...

pgmpy/tests/test_models/test_DoubleMLRegressor.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # Construct DataFrame with column names that match DAG roles
+    df = pd.DataFrame(
+        {
+            "x0": D,  # exposure
+            "x1": Z[:, 0],
+            "x2": Z[:, 1],
+        }
+    )
+    y = pd.Series(Y, name="y")
```
ankurankan: An easier way to simulate this is to use the simulate method in pgmpy. Have a look at @XAheli 's PR for examples. Would also suggest using smaller values for `beta` here to keep the variances low, or you can also standardize the data after simulation.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+        X_t_for_ols = t_res.reshape(-1, 1)
+
+        # perform a simple OLS of y_res on t_res (and intercept)
+        estimator = LinearRegression()
+        estimator.fit(X_t_for_ols, y_res, sample_weight=sample_weight)
+        theta_coef = estimator.coef_
+        intercept = estimator.intercept_
+
+        self.treatment_effect_ = theta_coef
+        self.intercept_ = intercept
```
ankurankan: I would suggest exposing the `estimator` instead of the `coef_` and `intercept_`.  This would help in making the exposed attributes uniform across all causal prediction algorithms. For example, in the Naive Adjustment case estimator can be any sklearn model, and it might not have a `coef_` and `intercept_` attribute.

pgmpy/models/DoubleMLRegressor.py
```diff
@@ -0,0 +1,299 @@
...
+        X_arr = validate_data(
+            self, X, reset=False, ensure_2d=True, force_all_finite=True
+        )
+        found = X_arr.shape[1]
+        expected = getattr(self, "n_features_in_", None)
+        if expected is not None and found != expected:
+            raise ValueError(
+                f"Found array with {found} feature(s) (shape[1]={found}) while "
+                f"{self.__class__.__name__} is expecting {expected} features as input."
+            )
```
ankurankan: Isn't `check_is_fitted` in Line 260 already verifying this?

ankurankan (changes requested):

pgmpy/models/__init__.py
ankurankan, line 1: This should go in `pgmpy/predictions/__init__.py`

pgmpy/tests/test_models/test_DoubleMLRegressor.py
ankurankan, line 1: Similarly, this should be in `pgmpy/tests/test_prediction/test_DoubleMLRegressor.py`.

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,368 @@
...
+
+    2. Orthogonalization:
+        - Compute residuals:
+            y_res = Y - g_hat(X)
+            t_res = T - m_hat(X)
+        - These residuals remove variation explained by X, isolating the effect of T on Y.
+
+    3. Final Estimation:
+        - Fit a effect estimator on y_res on t_res:
+            y_res = theta * t_res + E
```
ankurankan: Need to fix for the general case.

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,368 @@
...
+        - The estimated coefficient theta is the causal effect of T on Y, adjusted for confounders X.
+
+    For new data (T_new, X_new), the predicted outcome is:
+        Y_pred = intercept + theta * T_new + g_hat(X_new)
```
ankurankan: Need to fix for the general case.

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,368 @@
...
+           treatment and structural parameters. The Econometrics Journal, 21(1),
+           C1-C68.
+
+    """
+
+    def __init__(
+        self,
+        causal_graph,
+        nuisance_estimators=None,
+        effect_estimator=None,
```
marrov, line 169: Maybe if someone only knows DML PLR they may get confusing about adding themselves the effect estimator. Would it be a good idea to have a `LinearRegressor` as the default?

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,417 @@
...
+                raise ValueError(
+                    "If nuisance_estimators is a tuple, it must have exactly two elements."
+                )
+            treatment_est = clone(self.nuisance_estimators[0])
+            outcome_est = clone(self.nuisance_estimators[1])
+        else:
+            treatment_est = clone(self.nuisance_estimators)
+            outcome_est = clone(self.nuisance_estimators)
+
+        if self.effect_estimator is None:
```
marrov, line 258: I just realized that my comment above is handled here but I think this is not the right approach, this should happen in the `__init__`, no? So in line 168 `self.effect_estimator = LinearRegression() if effect_estimator is None else effect_estimator`.

But more important than the location is that there is a bug here! When `self.effect_estimator` is None, you set `self.effect_estimator = LinearRegression()` but never assign it to `effect_est`. Later, line 369 tries to use `effect_est.fit(...)`, which will fail with a `NameError: name 'effect_est' is not defined.`or something like this. line 250 should be: `effect_est = LinearRegression()`.

Could this be checked by a test?

ankurankan: @marrov 
> I just realized that my comment above is handled here but I think this is not the right approach, this should happen in the __init__, no?

I did try that initially, but sklearn doesn't allow any changes to the attributes corresponding to the arguments in the `__init__` method. Also tried passing `LinearRegression()` as the default argument value, but sklearn doesn't allow that either. The default value can be a class (i.e., `LinearRegression`), but not an instance of the class. So, ended up moving it to the `fit` method. Any ideas to implement this in a better way?

> But more important than the location is that there is a bug here! When self.effect_estimator is None, you set self.effect_estimator = LinearRegression() but never assign it to effect_est. Later, line 369 tries to use effect_est.fit(...), which will fail with a NameError: name 'effect_est' is not defined.or something like this. line 250 should be: effect_est = LinearRegression()

Yeah, that is a bug. I will fix it. Thanks.

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,417 @@
...
+        Returns
+        -------
+        self : object
+            Fitted estimator.
+        """
+        # Step 0: Validate inputs
+
+        # Step 0.1: Check `nuisance_estimators`, `effect_estimator`, and assign variables.
+        if self.nuisance_estimators is None:
+            self.nuisance_estimators = (LinearRegression(), LinearRegression())
```
marrov: This mutates the `__init__` parameter `self.nuisance_estimators`, which according to sklearn conventions, you should never do during fit(). Is there any issue with just doing:
```python
if self.nuisance_estimators is None:
    treatment_est = LinearRegression()
    outcome_est = LinearRegression()
```

marrov:
Maybe this has already been discussed and/or I am mistaken but just wanted to ask a silly question. When doing k folds, there are two options on what you could do with the results, you can either:

1. Fit one effect model per fold on that fold's residuals, then average coefficients
2.  Stack all out-of-sample residuals and fit a single effect model (current approach)

I think this is called DML1 and DML2 by the original authors (https://docs.doubleml.org/stable/guide/algorithms.html#double-machine-learning-algorithms). In general they prefer DML2, which is the one we're implementing. I think, given this we should:

1. Mention this in the doctrings with the appropriate link to the source reference
2. Actually implement both methods, defaulting to DML2 and keep it as an optional param. I would not stop the merge for this, but maybe it can be a good update PR/good first issue if someone wants to implement DML1.

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,417 @@
...
+        if len(self.adjustment_vars_ + self.pretreatment_vars_) == 0:
+            X_new_covariates = pd.DataFrame(
+                {"_intercept": np.ones(X_df.shape[0])}, index=X_df.index
+            )
+        else:
+            X_new_covariates = X_df[self.adjustment_vars_ + self.pretreatment_vars_]
+
+        # Step 2: Compute and return predictions. Average the predictions from
+        #         each fold's nuisance model to compute the outcome prediction.
+        treatment_preds = np.column_stack(
```
marrov: Minor optimisation: When n_folds=1, you have only one estimator in the list, so column_stack and mean add overhead though it's likely quite minimal.
```python
if len(self.treatment_est_) == 1:
    treatment_pred_mean = self.treatment_est_[0].predict(X_new_covariates)
    outcome_pred_mean = self.outcome_est_[0].predict(X_new_covariates)
else:
    treatment_preds = np.column_stack(...)
    treatment_pred_mean = np.mean(treatment_preds, axis=1)

# Same for outcome
```

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,417 @@
...
+        seed: Optional[int] = None,
+    ):
+
+        self.causal_graph = causal_graph
+        self.nuisance_estimators = nuisance_estimators
+        self.effect_estimator = effect_estimator
+        self.n_folds = n_folds
+        self.seed = seed
+
+    def __sklearn_tags__(self):
```
marrov, line 180: When n_folds > 1 with no seed, the estimator is non-deterministic due to random fold splits so it should have the `non_deterministic` tag, no?

marrov:
There may be a couple of edge cases not covered by tests (made with help from my LLM friend):

- Single sample per fold: What if n_samples < n_folds? Should raise informative error.
- High-dimensional confounders: Performance with many adjustment variables
- Perfect collinearity: What if treatment is perfectly predicted by adjustments? (degeneracy check)
- All-constant treatment: If T has zero variance, residualization fails
- Sample weights edge cases: All-zero weights, negative weights (should sklearn handle this?)

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -0,0 +1,417 @@
...
+        # Step 2: Prepare covariate dataframe. If no adjustment or pretreatment
+        #         variables, use intercept only.
+        if len(self.adjustment_vars_ + self.pretreatment_vars_) == 0:
+            covariates_df = pd.DataFrame(
+                {"_intercept": np.ones(self.n_samples_)}, index=df.index
+            )
+        else:
+            covariates_df = df[self.adjustment_vars_ + self.pretreatment_vars_]
+
+        # Step 3: Fit nuisance models
```
marrov, line 323: The fitting logic here is duplicated and prone to issues when updating the code. It could maybe be extracted out to a helper function, e.g.:
```python
def _fit_nuisance_models(self, covariates_df, exposure_df, outcome_df, sample_weight):
    """Fit nuisance models with optional cross-fitting."""
    if self.n_folds == 1:
        return self._fit_no_crossfit(...)
    else:
        return self._fit_with_crossfit(...)
```

marrov:
One final question: the current predict logic is mathematically justified only when `effect_estimator` solves the low-dimensional linear score (so basically when the user choses a linear regressor for the effect model) or is the logic generally valid? Because I personally am not super sure about this. Thoughts @ankurankan @fkiraly ?

pgmpy/prediction/DoubleMLRegressor.py
marrov: Not really, how would you recommend that this be solved best @fkiraly?

pgmpy/prediction/DoubleMLRegressor.py
ankurankan: According to the docs, I think this needs to be True if the estimator is non-deterministic even with a seed.
```
non_deterministic: bool, default=False
Whether the estimator is not deterministic given a fixed random_state.
```

ankurankan:
@marrov 

> I think this is called DML1 and DML2 by the original authors (https://docs.doubleml.org/stable/guide/algorithms.html#double-machine-learning-algorithms). In general they prefer DML2, which is the one we're implementing. I think, given this we should:

I had completely missed this. From what I understand, we are currently implementing DML1. DML2 takes a weighted mean over the estimates of the effect estimator of each fold, and DML1 uses equal weights. So, I think our stacking approach is equivalent to doing an unweighted mean. I will clarify this in the documentation and open a new issue for DML2.

**Update**: Sorry, ignore the above. I looked at the paper again and I was misunderstanding some equations. You are correct, our current implementation is DML2. 

> One final question: the current predict logic is mathematically justified only when effect_estimator solves the low-dimensional linear score (so basically when the user choses a linear regressor for the effect model) or is the logic generally valid?

I think the theory gives guarantees for only a certain class of estimators (that can be represented as Z-estimators; Linear regression and GLM can be formulated as Z-estimators). But in practice, I think this should work for any estimator. I will add this to the documentation as well.

pgmpy/prediction/DoubleMLRegressor.py
ankurankan: I had written it like this to make the code uniform for all the cases. As you mentioned, I also think the overhead would be negligible and it reduces the amount of code.

ankurankan:
> Single sample per fold: What if n_samples < n_folds? Should raise informative error.

Have added a check for this.

> High-dimensional confounders: Performance with many adjustment variables

Added a test for this.

> Perfect collinearity: What if treatment is perfectly predicted by adjustments?

I don't think we can add a check for this without defining a correlation threshold below which we call it perfectly correlated. Most packages allow highly correlated variables in the covariates (e.g., R's `lm` method), and let the user figure out the problem.

> All-constant treatment: If T has zero variance, residualization fails

I think we will need to define a cutoff of the variance here as well. Also, most packages allow this as well for standard estimators.

> Sample weights edge cases: All-zero weights, negative weights (should sklearn handle this?)

sklearn seems to handle this automatically.

Issue #2280, opened 2025-08-07
[ENH] Add variable roles to `ADMG` class


ankurankan:
Ref #2278

daehyun99:
@ankurankan HI! I'd love to work on this issue. Would it be okay if I gave it a try?

Spinachboul:
Hi @daehyun99 
I would request if I work on this issue since I did have some contribution in implementing the `ADMG` class and things become tricky when the graph contains bidirected edges as well as the latent confounders.
Thanks

daehyun99:
Hi @Spinachboul 
I understand.
I've been assigned a different issue, so please handle this one directly.
Thanks.

ankurankan:
I think it's still a good first issue, so I am assigning it to @daehyun99. @Spinachboul, could you instead help us with reviewing the PR?

Spinachboul:
Yeah sure!
@daehyun99 u can take on this one 
Thanks

daehyun99:
@ankurankan @Spinachboul 
I understand. Thanks for the clarification.

I've been assigned this issue, and I'd be happy to take it on.
I'm a first-time contributor to this open-source project, so I was wondering if you could offer some advice.
I'm also looking at issues #2279. Which one do you think would be a better starting point for a newcomer?

Thanks again for your help!

Spinachboul:
@daehyun99 I would say start with the refactoring the PDAG class since it closely resembles the DAG structure and easier to work with.
@ankurankan Your guidance would also be required here.

ankurankan:
@daehyun99 Yeah, you can start with the PDAG class. I would suggest having a look at how the `_GraphRolesMixin`  works in the `DAG` class and try to replicate similar functionality in the `PDAG` class. Feel free to ping us on discord if you have any questions or something isn't clear.

daehyun99:
HI, @ankurankan @Spinachboul 
I was writing test code for the `ADMG` class, and an error occurred during the commit process.

- error message
```sh
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

pgmpy/tests/test_base/test_ADMG.py:1:1: F401 'networkx as nx' imported but unused
pgmpy/tests/test_base/test_ADMG.py:338:9: F841 local variable 'separated' is assigned to but never used
```


The issue is that `nx` and `separated` are unused in the following code sections. How should I handle this?

Thanks.

- pgmpy/tests/test_base/test_ADMG.py
```py
import networkx as nx
```

[...]

PR #2294, opened 2025-08-13
Feat: Add logic and test code for `_GraphRolesMixin` as a parent class of `ADMG`

daehyun99:

- Ref #2280 (partially fixes it)

- Test: Verify `_GraphRolesMixin` as a parent class of `ADMG` 
- Feat: Add `_GraphRolesMixin` as a parent class of `ADMG`

### DIFF
--- a/pgmpy/base/ADMG.py
+++ b/pgmpy/base/ADMG.py
@@ -1,4 +1,5 @@
 import collections
+from typing import Hashable, Iterable, Sequence
 
 import networkx as nx
 from networkx import MultiDiGraph
@@ -57,13 +58,33 @@ class ADMG(_GraphRolesMixin, MultiDiGraph):
         for role, vars in roles.items():
             self.with_role(role=role, variables=vars, inplace=True)
 
-    def add_node(self, node, **kwargs):
+    def add_node(
+        self,
+        node: Hashable,
+        latent: bool = False,
+        **kwargs,
+    ):
         """
         Adds a node to the ADMG from the MultiDiGraph class.
+
+        Parameters
+        ----------
+        node : str, int, or any hashable python object.
+            The node to add to the graph.
+
+        latent: boolean (default: False)
+            Specifies whether the variable is latent or not.
         """
+        if latent:
+            self.latents.add(node)
+
         super().add_node(node, **kwargs)
 
-    def add_nodes_from(self, nodes, **kwargs):
+    def add_nodes_from(
+        self,
+        nodes: Iterable[Hashable],
+        latent: Sequence[bool] | bool = False,
+    ):
         """
         Adds multiple nodes to the graph.
 
@@ -71,8 +92,18 @@ class ADMG(_GraphRolesMixin, MultiDiGraph):
         ----------
         nodes : iterable
             An iterable of nodes to add.
+
+        latent: bool, list, tuple (default=False)
+            A container of boolean. The value at index i tells whether the
+            node at index i is latent or not.
         """
-        return super().add_nodes_from(nodes, **kwargs)
+        nodes = list(nodes)
+
+        if isinstance(latent, bool):
+            latent = [latent] * len(nodes)
+
+        for index in range(len(nodes)):
+            self.add_node(node=nodes[index], latents=latent[index])
 
     def add_directed_edges(self, ebunch):
         """
--- a/pgmpy/base/ADMG.py
+++ b/pgmpy/base/ADMG.py
@@ -57,13 +57,13 @@ class ADMG(_GraphRolesMixin, MultiDiGraph):
         for role, vars in roles.items():
             self.with_role(role=role, variables=vars, inplace=True)
 
-    def add_node(self, node, **attr):
+    def add_node(self, node, **kwargs):
         """
         Adds a node to the ADMG from the MultiDiGraph class.
         """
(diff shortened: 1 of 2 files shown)


pgmpy/base/ADMG.py
```diff
@@ -35,11 +48,20 @@ def __init__(self, directed_ebunch=None, bidirected_ebunch=None, latents=None):
...
+        if roles is None:
+            roles = {}
+        elif not isinstance(roles, dict):
+            raise TypeError("Roles must be provided as a dictionary.")
+
+        # set the roles to the vertices as networkx attributes
+        for role, vars in roles.items():
+            self.with_role(role=role, variables=vars, inplace=True)
+
+    def add_node(self, node, **attr):
```
ankurankan: This change is not required. Correct?

pgmpy/base/ADMG.py
```diff
@@ -35,11 +48,20 @@ def __init__(self, directed_ebunch=None, bidirected_ebunch=None, latents=None):
...
         """
         Adds a node to the ADMG from the MultiDiGraph class.
         """
-        super().add_node(node)
+        super().add_node(node, **attr)
```
ankurankan: Same here.

pgmpy/tests/test_base/test_ADMG.py
```diff
@@ -265,6 +276,8 @@ def test_is_m_separated(self):
 
         # Test with conditional set
         separated = self.admg.is_mseparated("A", "D", conditional_set={"C"})
+        assert separated is True
```
ankurankan: We can simplify this by not assigning the new variable and directly calling the assert on the expression. Similar to line 280

ankurankan (changes requested):
@daehyun99 Thanks for the PR. I have added a couple of minor comments, but looks good to me otherwise.

pgmpy/base/ADMG.py
daehyun99: @ankurankan 
The following code snippet (lines 57-58) is essential for adding `roles` values to the `ADMG` class.

```python
        for role, vars in roles.items():
            self.with_role(role=role, variables=vars, inplace=True)
```

This implementation pattern is consistent across the `DAG` and `PDAG` classes when they also handle `roles` values. From a consistency perspective, it seems best to maintain this approach.

The `self.with_role` method internally uses `add_node`.

```python
def add_node(self, node, **attr):
```

As I understand it, This `add_node` method is originally defined in the `networkx.DiGraph` class, which the `ADMG`, `DAG`, and `PDAG` classes all inherit from.

- [`DiGraph`'s `add_node` method](https://github.com/networkx/networkx/blob/36e8a1ee85ca0ab4195a486451ca7d72153e2e00/networkx/classes/digraph.py#L439)

The problem is that the `DiGraph` class's `add_node` method does not have `roles` defined as a parameter.

Because of this, the `pgmpy` library's `DAG` class's `add_node` method handles `roles` by receiving them as `**kwargs`.

- [`DAG`'s `add_node` method](https://github.com/pgmpy/pgmpy/blob/d25b841e0e92f666a75c1f27d5341f4e3d680321/pgmpy/base/DAG.py#L284)

In contrast, the `PDAG` class doesn't define its own `add_node` method and instead directly inherits the one from `DiGraph`.

This leaves us with two potential options for the `ADMG` class:

1.  **Delete the `add_node` method from the `ADMG` class:** This would make its implementation consistent with the `PDAG` class.
2.  **Add a `**attr` argument to the `ADMG` class's `add_node` method:** This would align it with the `DAG` class's implementation.

My suggestion is to first ask the developer who added the `add_node` method to the `ADMG` class for their reasoning. Understanding their original intent will help us make the best decision. @Spinachboul

Thanks.

pgmpy/base/ADMG.py
daehyun99: The content is included in the reply above.

pgmpy/tests/test_base/test_ADMG.py
daehyun99: I've committed the updates to this part.
```py
        # Test with conditional set
        assert self.admg.is_mseparated("A", "D", conditional_set={"C"}) is True
        assert self.admg.is_mseparated("A", "D", conditional_set=set()) is False
```

pgmpy/base/ADMG.py
Spinachboul: @daehyun99  I feel the `add_node` function which is there is `ADMG` class can still stretch itself to handle different attributes such as `roles` which would make it fully compatible with the `DAG` class.

daehyun99: @ankurankan Once you reply to this, I will proceed as requested.

ankurankan: @daehyun99 @Spinachboul I had missed that the method signature for `ADMG` had deviated from `DAG`. I think we can keep the `attr` but rename it to `kwargs` to make it consistent with rest of the package. I would also suggest adding the `latent` argument ot both `ADMG.add_node` and `ADMG.add_nodes_from` as well similar to `DAG` methods. @daehyun99 Could you please also add the `latent` argument here?

daehyun99: @ankurankan I understand.

ankurankan (approved):
Looks good to me. Thanks!!

PR #2296, opened 2025-08-15
Corrections to LinearGaussianBayesianNetwork

gdreiman1:

- Fixes # [2167](https://github.com/pgmpy/pgmpy/issues/2167)

#### Code changes in `LinearGaussianBayesianNetwork`
- I updated `to_joint_gaussian` to properly use variance instead of std. 
- I updated `fit` to properly provide both `mle` and `unbiased` options for estimating the std. I also updated the doc string.
#### Test changes
- I had to update several tests, mostly this meant squaring defined test values for var that were previously set improperly as std. 
- In several of the `simulate` tests I had to substantially increase the sample rate. This makes sense since previously the stored "variances" were much too small (actually std).


pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -608,39 +611,41 @@ def fit(
...
 
+        if method not in {"mle", "unbiased"}:
+            raise ValueError("method must be one of {'mle', 'unbiased'}")
+
         # Step 2: Estimate the LinearGaussianCPDs
         cpds = []
         for node in self.nodes():
             parents = self.get_parents(node)
 
-            # Step 2.1: If node doesn't have any parents (i.e. root node),
```
ankurankan, line 616: Could you please leave this comment in the code?

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -608,39 +611,41 @@ def fit(
...
                 cpds.append(
                     LinearGaussianCPD(
                         variable=node,
                         beta=[data.loc[:, node].mean()],
-                        std=data.loc[:, node].var(),
+                        std=data.loc[:, node].std(ddof=ddof),
                     )
                 )
-
-            # Step 2.2: Else, fit a linear regression model and take the coefficients and intercept.
```
ankurankan, line 627: This one as well.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -608,39 +611,41 @@ def fit(
...
-            #           Compute error variance using predicted values.
             else:
+                # Regression: use ddof=0 for MLE, ddof=p for unbiased
```
ankurankan: I would suggest removing this comment, as it is very clear when the ddof is defined.

pgmpy/models/LinearGaussianBayesianNetwork.py
```diff
@@ -581,6 +581,9 @@ def fit(
         data: pd.DataFrame
             A pandas DataFrame with the data to which to fit the model
             structure. All variables must be continuous valued.
+        method: str
```
ankurankan: Because the fitting method for the coefficients is still MLE, I think we should not change the `method` argument. Rather, I would suggest adding a separate argument for variance estimate correction. For example, `ddof=None`. When `None`, we do the unbiased estimate, else user can specify a specific value. This would also give the user full control if they want to change it to a specific value. What do you think?

ankurankan (commented):
@gdriman1 Thanks for the fixes. I have added a few comments. 

About the failing tests, I don't remember how I had computed the values. For the `to_joint_gaussian`, we can compute that by hand but I have also verified our new results with the R pacakge dagitty, and it matches. Code to reproduce:
```python
from pgmpy.base import DAG
lgbn = DAG.from_dagitty("dag{X -> Y [beta=0.3] Y -> Z [beta=0.5]}")
df = lgbn.to_joint_gaussian()
```
Equivalent dagitty code:
```R
library(dagitty)
dag <- dagitty("dag{X -> Y [beta=0.3] Y -> Z [beta=0.5]}"
d <- simulateSEM(dag, empirical=T, standardized=F)
cov(d)
```

I am still double checking the prediction method.

ankurankan:
About the prediction, the LLM is correct: there is a mistake in the computation of `cov_aa`. The indexing used there is incorrect and works only when one variable is missing. I think it's cleaner to replace it with `np.ix_`. The LLM is also correct about replacing `np.linalg.inv` with `np.linalg.solve`. When multiplying by an inverse matrix, `solve` is more stable. Do you mind making these changes in this PR?

One way to test whether the predicted values are correct is to simulate data from a model with very low variances, drop some of the rows, and predict it back. Because of the low variance, the predicted mean should be close to the actual value.

ankurankan:
@gdreiman1 Was wondering if there were any updates on this? Thanks.

gdreiman1:
@ankurankan Sorry my day job got really busy, should be able to return to this next week!

pgmpy/models/LinearGaussianBayesianNetwork.py
gdreiman1: How would you feel about renaming `method` to something like `std_estimator`? I agree that the OLS is MLE so it is confusing, so maybe making the clarification will help? 

I don't know that it really makes sense to let the user specify a global `ddof`. If you are not setting it `MLE` (and thus `ddof = 0`) then you likely need to calculate the appropriate `ddof` for each `cpd`. I could imagine weird edge cases where users to could pass a `ddof` that was greater or equal to the number of observations which could either give you divide by zero or negative std (probably just an error but still not great).

I also kinda feel like the `unbiased` std should probably be the default not `MLE`. Thoughts?
@ankurankan

ankurankan: @gdreiman1 
I would suggest keeping the `method` argument (or actually renaming it to `estimator` to make it uniform with other BN classes) as I want to keep it reserved for future addition of other estimation methods, for example, Bayesian or EM based. 

> I don't know that it really makes sense to let the user specify a global ddof.

Yeah, I agree. I am sorry, I don't know what exactly I was thinking, but you are right that we can't have a single  `ddof` value. How about we add the `std_estimator` as an additional argument that accepts a string input: `unbiased` or `mle`, and uses the ddof based on that?

gdreiman1: @ankurankan Yeah that sounds good, will make those changes and push.

gdreiman1:
@ankurankan I made changes suggested. I also replicated the previously failing tests in `R` and updated `true_values` so those now pass. 

 One note: in `test_predict` there is a portion with the `'alarm'` model with parts commented out. I think that this model fundamentally shouldn't be modeled using `LinearGaussianBayesNetwork` since it is a discrete network (see [bnlearn info](https://www.bnlearn.com/documentation/man/alarm.html)). I think we should remove that entire portion of the test. 
 
Possible additions:
- Add test for multiple missing vars to address changes in `predict`?
- Replace `'alarm'` model with something that is fully gaussian but keep same ideas?

ankurankan:
@gdreiman1 Thanks for the fixes.

> Add test for multiple missing vars to address changes in predict?

That would be great.

> Replace 'alarm' model with something that is fully gaussian but keep same ideas?

Yes, that would be nice too. We can load linear gaussian example models from the bnlearn repo as well now, for example: `get_example_model('ecoli70')`.

gdreiman1:
> @gdreiman1 Thanks for the fixes.
> 
> > Add test for multiple missing vars to address changes in predict?
> 
> That would be great.
> 
> > Replace 'alarm' model with something that is fully gaussian but keep same ideas?
> 
> Yes, that would be nice too. We can load linear gaussian example models from the bnlearn repo as well now, for example: `get_example_model('ecoli70')`.

@ankurankan I added a new test using `ecoli70` which I think addresses both of these test ideas.

Also, `utils.get_example_model` had a `std = var` error. I am worried there might be more of these, but seems like a broader code review might be needed.

ankurankan (approved):
Sorry for the delay on the review. Looks good to me now.

> Also, utils.get_example_model had a std = var error. I am worried there might be more of these, but seems like a broader code review might be needed.

Thanks for catching this. I did a quick search if `LinearGaussianCPD` is getting initialized anywhere else in the codebase, and I don't see anywhere else. I will keep an eye on this.

Issue #2306, opened 2025-08-17
[BUG] `ADMG.__init__` fails to add `latent` nodes in `nodes`


daehyun99:
**Describe the bug**
Ref #2303 

**To Reproduce**

```python
from pgmpy.base.ADMG import ADMG

# SetUP admg1
admg1= ADMG(
    directed_ebunch=[("A", "B"), ("B", "C"), ("D", "E")],
    bidirected_ebunch=[("A", "D"), ("B", "E")],
    latents=["F"],
)

# SetUP admg2
admg2 = ADMG()
admg2.add_directed_edges([("A", "B"), ("B", "C"), ("D", "E")])
admg2.add_bidirected_edges([("A", "D"), ("B", "E")])
admg2.add_node("F", latent=True)
```

```python
print(admg1.__eq__(admg2) is True)

print("admg1 nodes: ", admg1.nodes())
print("admg1 latents: ", admg1.latents)

print("admg2 nodes: ", admg2.nodes())
print("admg2 latents: ", admg2.latents)
```
- output
```
False
admg1 nodes:  ['A', 'B', 'C', 'D', 'E']
admg1 latents:  {'F'}
admg2 nodes:  ['A', 'B', 'C', 'D', 'E', 'F']
admg2 latents:  {'F'}
```

**Expected behavior**
- Expected output
```
True
admg1 nodes:  ['A', 'B', 'C', 'D', 'E', 'F']
admg1 latents:  {'F'}
admg2 nodes:  ['A', 'B', 'C', 'D', 'E', 'F']
admg2 latents:  {'F'}
```

**Additional context**

**Versions**
<details>

</details>

daehyun99:
same comment as in [#2304](https://github.com/pgmpy/pgmpy/issues/2304#issuecomment-3241297504)

Since this bug is no longer a concern, I am closing this issue.

ankurankan:
Ref #2322

Issue #2305, opened 2025-08-17
[BUG] `PDAG.__init__` fails to add `latent` nodes in `nodes`


daehyun99:
**Describe the bug**
Ref #2303 

**To Reproduce**

```python
from pgmpy.base.DAG import PDAG

# SetUP pdag1
pdag1= PDAG(
    directed_ebunch=[("A", "C"), ("D", "C")],
    latents=["F"],
)

# SetUP pdag2
pdag2 = PDAG()
pdag2.add_nodes_from(["A", "C", "D"])
pdag2.add_edges_from([("A", "C"), ("D", "C")])
pdag2.add_node("F", latent=True)
```
```python
print(pdag1.__eq__(pdag2) is True)

print("pdag1 nodes: ", pdag1.nodes())
print("pdag1 latents: ", pdag1.latents)

print("pdag2 nodes: ", pdag2.nodes())
print("pdag2 latents: ", pdag2.latents)
```
- output
```
False
pdag1 nodes:  ['D', 'C', 'A']
pdag1 latents:  {'F'}
pdag2 nodes:  ['A', 'C', 'D', 'F']
pdag2 latents:  set()
```

**Expected behavior**
- Expected output
```
True
pdag1 nodes:  ['D', 'C', 'A', 'F']
pdag1 latents:  {'F'}
pdag2 nodes:  ['A', 'C', 'D', 'F']
pdag2 latents:  {'F'}
```

**Additional context**

**Versions**
<details>

</details>

daehyun99:
same comment as in [#2304](https://github.com/pgmpy/pgmpy/issues/2304#issuecomment-3241297504)

Since this bug is no longer a concern, I am closing this issue.

ankurankan:
Ref #2322

Issue #2304, opened 2025-08-17
[BUG] `DAG.__init__` fails to add `latent` nodes in `nodes`


daehyun99:
**Describe the bug**
Ref #2303 

**To Reproduce**

```python
from pgmpy.base.DAG import DAG

# SetUP dag1
dag1= DAG(
    ebunch=[("A", "B"), ("B", "C"), ("D", "E")],
    latents=["F"],
)

# SetUP dag2
dag2 = DAG()
dag2.add_nodes_from(["A","B", "C", "D", "E"])
dag2.add_edges_from([("A", "B"), ("B", "C"), ("D", "E")])
dag2.add_node("F", latent=True)
```
```python
print(dag1.__eq__(dag2) is True)

print("dag1 nodes: ", dag1.nodes())
print("dag1 latents: ", dag1.latents)

print("dag2 nodes: ", dag2.nodes())
print("dag2 latents: ", dag2.latents)
```
- output
```
False
dag1 nodes:  ['A', 'B', 'C', 'D', 'E']
dag1 latents:  {'F'}
dag2 nodes:  ['A', 'B', 'C', 'D', 'E', 'F']
dag2 latents:  {'F'}
```

**Expected behavior**
- Expected output
```
True
dag1 nodes:  ['A', 'B', 'C', 'D', 'E', 'F']
dag1 latents:  {'F'}
dag2 nodes:  ['A', 'B', 'C', 'D', 'E', 'F']
dag2 latents:  {'F'}
```

**Additional context**

**Versions**
<details>

</details>

daehyun99:
With the change to treat `latents` as a role, it's automatically prevented to designate a node as a latent variable if it doesn't already exist in the graph.
Since this bug is no longer a concern, I am closing this issue.

- example1
```py
from pgmpy.base.DAG import DAG

# SetUP dag1
dag1= DAG(
    ebunch=[("A", "B"), ("B", "C"), ("D", "E")],
    roles={"latent": "F"},
)
```
- output1
```
ValueError: Variable 'F' not found in the graph.
```

- example2
```py
# SetUP dag2
dag2 = DAG()
dag2.add_nodes_from(["A","B", "C", "D", "E"])
dag2.add_edges_from([("A", "B"), ("B", "C"), ("D", "E")])
dag2 = dag2.with_role("latents", "F")
```
- output2
```
ValueError: Variable 'F' not found in the graph.
```

ankurankan:
Ref #2322

Issue #2303, opened 2025-08-17
[BUG] `__init__` fails to add `latent` nodes in `nodes`


daehyun99:
**Describe the bug**
- When declaring causal graph classes through the `__init__` method, a problem occurred where the `latents` values were omitted.

**To Reproduce**
- Please refer to the sub-issue for bug reproduction.

**Expected behavior**
- When executing the `nodes()` method of the causal graph class, the `latents` nodes must also be included.
- When using the `__eq__` method, it should return `True` for two instances of the same causal graph class, even if they were declared in different ways.

**Additional context**

**Versions**
<details>

</details>

daehyun99:
same comment as in [#2304](https://github.com/pgmpy/pgmpy/issues/2304#issuecomment-3241297504)

Since this bug is no longer a concern, I am closing this issue.

ankurankan:
Ref #2322

PR #2310, opened 2025-08-20
Feat: Add logic and test code for `add_node`, `add_nodes_from` method of `PDAG`

### LINKED ISSUE #2307: [BUG] unable to handle `latent` values within the `add_node` method in `PDAG`
**Describe the bug**
For the `PDAG` class, the `add_node` method isn't implemented directly; it's inherited from the `networkx` library.
Because of this, it seems unable to handle `latent` values within the `add_node` method.

**To Reproduce**
```python
from pgmpy.base.DAG import PDAG

# SetUP pdag1
pdag1 = PDAG(
    latents=['F']
)

# SetUP pdag2
pdag2 = PDAG()
pdag2.add_node("F", latent=True)

print("pdag1 latents: ", pdag1.latents)
print("pdag2 latents: ", pdag2.latents)
```
- output
```
pdag1 latents:  {'F'}
pdag2 latents:  set()
```
**Expected behavior**
- Expected output
```
pdag1 latents:  {'F'}
pdag2 latents:  {'F'}
```

**Additional context**
This bug seems to be related to issue [#2294](https://github.com/pgmpy/pgmpy/pull/2294#discussion_r2272703659), which we discussed last time.
We need to implement the `add_node`, `add_nodes_from`, ~~`add_edge`, and `add_edges_from`~~ methods in the `PDAG` class.
**Versions**
<details>

</details>

daehyun99:

- Fixes #2307 

- Test: Implement code to test the `add_node`, `add_nodes_from` method of `PDAG`
- Feat: Implement `add_node`, `add_nodes_from` method of `PDAG`
- Feat, Docs: Add `ADMG` to `__init__.py` for public access

### DIFF
--- a/pgmpy/base/ADMG.py
+++ b/pgmpy/base/ADMG.py
@@ -74,6 +74,17 @@ class ADMG(_GraphRolesMixin, MultiDiGraph):
 
         latent: boolean (default: False)
             Specifies whether the variable is latent or not.
+
+        Examples
+        --------
+        >>> from pgmpy.base import ADMG
+        >>> admg = ADMG()
+        >>> admg.add_node(node="A")
+        >>> admg.add_node(node="B", latent=True)
+        >>> sorted(admg.nodes())
+        ['A', 'B']
+        >>> admg.latents
+        {'B'}
         """
         if latent:
             self.latents.add(node)
@@ -96,6 +107,18 @@ class ADMG(_GraphRolesMixin, MultiDiGraph):
         latent: bool, list, tuple (default=False)
             A container of boolean. The value at index i tells whether the
             node at index i is latent or not.
+
+        Examples
+        --------
+        >>> from pgmpy.base import ADMG
+        >>> admg = ADMG()
+        >>> admg.add_nodes_from(nodes=["A", "B"])
+        >>> admg.add_nodes_from(nodes=["C", "D"], latent=True)
+        >>> admg.add_nodes_from(nodes=["E", "F"], latent=[False, True])
+        >>> sorted(admg.nodes())
+        ['A', 'B', 'C', 'D', 'E', 'F']
+        >>> admg.latents
+        {'C', 'D', 'F'}
         """
         nodes = list(nodes)
 
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -2367,3 +2367,75 @@ class PDAG(_GraphRolesMixin, nx.DiGraph):
             and self.latents == other.latents
             and self.get_role_dict() == other.get_role_dict()
         )
+
+    def add_node(
+        self,
+        node: Hashable,
+        latent: bool = False,
+        **kwargs,
+    ):
+        """
+        Adds a single node to the Graph.
+
+        Parameters
+        ----------
+        node: str, int, or any hashable python object.
+            The node to add to the graph.
+
+        latent: boolean (default: False)
+            Specifies whether the variable is latent or not.
+
+        Examples
+        --------
+        >>> from pgmpy.base import PDAG
+        >>> pdag = PDAG()
+        >>> pdag.add_node(node="A")
+        >>> pdag.add_node(node="B", latent=True)
+        >>> sorted(pdag.nodes())
+        ['A', 'B']
+        >>> pdag.latents
+        {'B'}
+        """
+        if latent:
+            self.latents.add(node)
+
+        super().add_node(node, **kwargs)
+
+    def add_nodes_from(
+        self,
+        nodes: Iterable[Hashable],
+        latent: Sequence[bool] | bool = False,
+    ):
+        """
+        Add multiple nodes to the Graph.
+
+        Parameters
+        ----------
+        nodes: iterable container
+            A container (list, dict, set) of nodes (str, int or any hashable python
+            object).
+
+        latent: bool or list of bool (default: False)
+            If bool, specifies if all nodes are latent.
+            If list of bool, specifies for each node whether it is latent or not.
+            The length of the list must be equal to the number of nodes.
+
+        Examples
+        --------
+        >>> from pgmpy.base import PDAG
+        >>> pdag = PDAG()
+        >>> pdag.add_nodes_from(nodes=["A", "B"])
+        >>> pdag.add_nodes_from(nodes=["C", "D"], latent=True)
+        >>> pdag.add_nodes_from(nodes=["E", "F"], latent=[False, True])
+        >>> sorted(pdag.nodes())
+        ['A', 'B', 'C', 'D', 'E', 'F']
+        >>> pdag.latents
+        {'C', 'D', 'F'}
+        """
+        nodes = list(nodes)
+
+        if isinstance(latent, bool):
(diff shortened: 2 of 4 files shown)


fkiraly (changes requested):
Question: a node being latent seem to be a role.

So, stupid question - would it not be best if:

* `latents` becomes a property, which simply returns all nodes that have the `"latent"` role?
* all interaction with `latents` is factored through the roles interface. That way, `add_node` and all the other `networkx` default methods do not need to be overwritten, and latents can be added by default `networkx` syntax.

https://networkx.org/documentation/stable/reference/classes/generated/networkx.Graph.add_node.html

Alternatively, `latents` can simply be an attribute, separate from `roles`.

daehyun99:
@fkiraly First, I've been contributing to the pgmpy library for less than a month. I've also only worked with the `DAG`, `PDAG`, and `ADMG` classes in the `pgmpy/base/` directory. Therefore, please bear in mind that there's a high chance I could be wrong.

To get straight to the point, I think using a role interface to handle `latents` seems like a good approach. 
Although it requires additional work in the short term, the role interface approach seems better for long-term extensibility

Based on my research, the `latents` parameter was added starting with [#1407](https://github.com/pgmpy/pgmpy/pull/1407). 
Since then, it seems the `latents` parameter is being used not only in the `DAG`, `PDAG`, and `ADMG` classes in the `pgmpy/base/` directory, but also in the `Network` classes in the `pgmpy/models/` directory.

### (1) To implement a role interface for handling `latents`, we need to modify the existing test code.
For example, in the test code snippet from `pgmpy/tests/test_base/test_DAG.py` [lines 76-85], I would have to change the parts that are written as `add_node(node_name, latent=True)`.
- [test_add_node_string](https://github.com/pgmpy/pgmpy/blob/cc18b31f627af666da5cb76dcbe567431fb46c21/pgmpy/tests/test_base/test_DAG.py#L83)
```py
# pgmpy/tests/test_base/test_DAG.py [line_number 76 - 85]
    def test_add_node_string(self):
        self.graph = DAG()
        self.graph.add_node("a")
        self.assertListEqual(list(self.graph.nodes()), ["a"])
        self.assertEqual(self.graph.latents, set())

        self.graph = DAG()
        self.graph.add_node("a", latent=True)
        self.assertListEqual(list(self.graph.nodes()), ["a"])
        self.assertEqual(self.graph.latents, set(["a"]))
```

### (2) Extensibility would likely improve.

Some `Network` models have a `latents` parameter but do not inherit from `DAG` or implement an `add_node` method.

```py
# (ex) pgmpy/models/DiscreteMarkovNetwork.py

class DiscreteMarkovNetwork(UndirectedGraph):
```
If we handle `latents` using the approach from that PR, we might have to add an `add_node()` method to these models. 
However, if we use a role interface to handle `latents`, we would only need to have the class inherit from `_GraphRolesMixin`.
- [DiscreteMarkovNetwork](https://github.com/pgmpy/pgmpy/blob/cc18b31f627af666da5cb76dcbe567431fb46c21/pgmpy/models/DiscreteMarkovNetwork.py#L16)

Please let me know if I missed anything or misunderstood something.
Thanks.

ankurankan:
@fkiraly Thanks for the suggestion. I think it makes sense to have `latents` as variable roles. This would require a few changes that I have tried to outline in #2322. @daehyun99 If you would like to work on that, please leave a comment there, and I can assign it to you.

daehyun99:
@ankurankan I think we need to clean up the issues and PRs I've already submitted before I can start working on issue #2322.

daehyun99:
Issue: #2307
PR: #2310

I think we can close this issue and PR.

daehyun99:
main issue: #2303
sub issues: #2304 #2305 #2306
PR: #2315

I think this PR should be closed because we are planning to remove the `add_nodes_from` method.

However, the issues still need to be resolved. I'll revise the issues and the code, then open a new PR.

## example
```py
dag= DAG(
    latents=["A"],
)
print("dag nodes: ", dag.nodes())
```
- output
```
dag nodes:  []
```
- Expected output
```
dag nodes:  ['A']
```

daehyun99:
issue: #2311
PR: #2312

What are your thoughts on this issue and PR (excluding the `add_node` and `add_nodes_from` methods)?

I think it would be good to standardize the method for `ADMG` and edge-related methods.

From what I've checked, it seems that none of the Network models inherit from `PDAG` or `ADMG`.
Also, most of the Network models have an `add_edge` method.

Therefore, I believe that `ADMG` and `PDAG` should have `add_directed_edges`, `add_undirected_edges`(`PDAG`), and `add_bidirected_edges`(`ADMG`) methods.

ankurankan:
@daehyun99 

> I think it would be good to standardize the method for ADMG and edge-related methods.

I think we can keep the `add_edge` methods, but I would suggest having a single method instead of individual methods for directed and (bi/un)directed edges. We can add an additional argument `type` to the `add_edge` method to specify the type of edge. This would reduce the amount of boilerplate code for separate methods while also having the same checks (such as no cycles). What do you think?

daehyun99:
@ankurankan It looks good to use only the `add_edge` method and distinguish by `type`.

This might be a silly question, but what do you think about this additional approach?

---
Untitled

The first image shows the inheritance relationships of classes related to `edge` in pgmpy.
Each base graph has an `add_edge` method, which is inherited by various graph models.
Additionally, it seems that `MAG`, `PAG`, and others will be added as base graphs in pgmpy in the future.
Considering extensibility, what about implementing a `_mixin_edges`?

---
Untitled (1)

`_mixin_edges` explicitly has `u`, `v`, `type`, and `weight` parameters.

The `add_edge` method of `nx.Graph`, `nx.DiGraph`, `nx.MultiDiGraph` is responsible for adding an edge.
The `add_edge` method of `_mixin_edges` manages the basic properties that an edge defined in pgmpy should have (`type`, `weight`).
The `add_edge` method within each graph performs individual logic and exception handling for that specific graph.

All edge-related methods of all graphs will explicitly have `type` and `weight` parameters.

Couldn't this approach, where the role of the `add_edge` method is clearly divided by its location, enhance code consistency and maintainability?

Please let me know if I missed anything or misunderstood something.
Thanks.

ankurankan:
@daehyun99 Handling edge types consistently across classes sounds like a great idea. I’m unsure about the best design, though: most edge-addition logic is class-specific, so we’d still need class-level `add_edge` methods. This also seems like a longer-term change because edge types are stored differently currently - PDAG/ADMG store bidirected edges as two edges, while our MAG/PAG implementations store the type as an edge attribute - and making this uniform would likely require updates to some algorithms. Would you mind opening a new issue outlining your idea (proposed method/class signatures and responsibilities) so we can discuss and iterate on that?

Spinachboul:
@daehyun99 @ankurankan Yeah this looks like a great idea! Would love to be in the thread in the issue

daehyun99:
> Issue: #2303, #2304, #2305, #2306, #2307, #2311
> PR: #2310, #2312, #2315

@ankurankan  I've handled the above issues and created issue #2330 regarding `_mixin_edges`.

Can we closing this issue?

Issue #2254, opened 2025-07-29
[ENH] Implement the FCI algorithm for causal discovery


ankurankan:
The Fast Causal Inference (FCI) algorithm generalizes the PC algorithm to discover causal structures with latent and selection variables. The algorithm outputs a Partial Ancestral Graph (PAG) that is an equivalence class over Maximal Ancestral Graphs (MAG). 

Depends on: #2253 

Reference for implementation: Zhang, Jiji. "On the completeness of orientation rules for causal discovery in the presence of latent confounders and selection bias." Artificial Intelligence 172.16-17 (2008): 1873-1896.

Original paper: Spirtes, Peter L., Christopher Meek, and Thomas S. Richardson. "Causal inference in the presence of latent variables and selection bias." arXiv preprint arXiv:1302.4983 (2013).

Spinachboul:
@ankurankan 
Thanks for raising the issue

PR #2321, opened 2025-08-26
Implementation of Maximal Ancestral Graphs

### LINKED ISSUE #2258: [ENH] Implement the MAG class

Spinachboul:
Solves the issue https://github.com/pgmpy/pgmpy/issues/2258

Currently a minimal implementation, does not take into account the `m-separation` algorithm, which will be implemented later after solving https://github.com/pgmpy/pgmpy/issues/2254

@ankurankan


Spinachboul:
@ankurankan 

Should I be using the following graph for the tests, or something needs to be changed?
Since as done in `AncestralBase` class, I will be using the same graph throughout the test functions!
image

Spinachboul:
@ankurankan  I have also included the changes which are to be made in the docstrings of `AncestralBase` class
Kindly check, thanks!

And also I have tested the docstrings for the MAG class, they all seem to be correct

pgmpy/base/AncestralBase.py
```diff
@@ -19,12 +19,10 @@ def __init__(
...
         - Directed: ("A", "B", "-", ">") is stored as
           ("A", "B", {"marks": {"A": "-", "B": ">"}})
-        - Reverse directed: ("A", "B", ">", "-") is stored as
-          ("A", "B", {"marks": {"A": ">", "B": "-"}})
         - Bidirected: ("A", "B", ">", ">") is stored as
           ("A", "B", {"marks": {"A": ">", "B": ">"}})
-        - Undirected: ("A", "B", "o", "o") is stored as
-          ("A", "B", {"marks": {"A": "o", "B": "o"}})
+        - Undirected: ("A", "B", "-", "-") is stored as
+          ("A", "B", {"marks": {"A": "-", "B": "-"}})
```
ankurankan, line 28: Could add an unknown example with `o`

pgmpy/base/AncestralBase.py
```diff
@@ -39,7 +37,7 @@ def __init__(
 
         Examples
         --------
-        >>> from pgmpy.base import AncestralBase
+        >>> from pgmpy.base.AncestralBase import AncestralBase
```
ankurankan: Add `AncestralBase` to `base/__init__.py` file to simpify this import.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,320 @@
...
+    and undirected (-) edges, and are closed under marginalization and conditioning.
+    """
+
+    def __init__(
+        self,
+        ebunch: Optional[Iterable[tuple[Hashable, Hashable]]] = None,
+        latents: set[Hashable] = set(),
+    ):
+        """
+        Initialize a Maximal Ancestral Graph.
```
ankurankan: Should add a small description of what a MAG is

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,320 @@
...
+        Returns
+        -------
+        MAG
+            A new instance of a Maximal Ancestral Graph.
+
+        Examples
+        --------
+        >>> from pgmpy.base.MAG import MAG
+        >>> mag = MAG(
+        ...     ebunch=[("A", "B", "-", ">"), ("B", "C", "-", ">")], latents={"L"}
```
ankurankan: latent nodes need to be in the graph

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,320 @@
...
+        ... )
+        >>> list(mag.nodes())
+        ['A', 'B', 'C']
+        """
+        super().__init__(ebunch=ebunch, latents=latents)
+
+    def _is_collider(self, u_node, c_node, v_node):
```
ankurankan: We need the argument names to be consistent. `has_inducing_path` methods uses `u`, `v`.

ankurankan (changes requested):

pgmpy/base/AncestralBase.py
Spinachboul: So should I be removing this line `from pgmpy.base.AncestralBase import AncestralBase` from all the examples in the docstrings?
And should I also add `MAG` class in `base/__init__.py` and the do the same?

ankurankan: Yes, both the `AcestralBase` and `MAG` should be in the `base/__init__.py` file. That would simplify the import path from `from pgmpy.base.AncestralBase import AncestralBase` to `from pgmpy.base import AncestralBase`. I think this shorter import path is much cleaner and doesn't require as much typing.

Spinachboul:
@ankurankan  The tests are failing only because of this error `TypeError: 'module' object is not callable`
and there are other classes which are being affected as well.
Could you please guide me here?
Thanks!

ankurankan:
@Spinachboul 

> The tests are failing only because of this error `TypeError: 'module' object is not callable`

You will need to change the import paths in the tests as well to `from pgmpy.base import AncestralGraph` or `from pgmpy.base import MAG`

Spinachboul:
@ankurankan Still getting the same error
Actually this error is occuring  for all the other test files,  under `test_estimators` and `test_base`

Spinachboul:
@ankurankan Could you please guide me on what has been done incorrectly here, since I cant seem to resolve this issue `TypeError: 'module' object is not callable` even after importing correctly in the tests

ankurankan:
@Spinachboul This is a circular import issue. In `DAG.py`, can you move the `pgmpy.base import UndirectedGraph` import inside the `DAG.moralize` method?

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
+from typing import Hashable, Iterable, Optional
+
+import networkx as nx
+
+from pgmpy.base import AncestralBase
+
+
+class MAG(AncestralBase):
+    """
+    Class for representing Maximal Ancestral Graphs (MAGs).
```
ankurankan, line 10: The docstring format is: 1 line short description, an empty line, and then the full description. This didn't throw a pre-commit hook error?

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
+from typing import Hashable, Iterable, Optional
+
+import networkx as nx
+
+from pgmpy.base import AncestralBase
+
+
+class MAG(AncestralBase):
```
ankurankan, line 8: This should also be inheriting `_GraphRolesMixin`. Or might be better to inherit that in `AncestralBase` itself.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
...
+    and selection bias. Every pair of nodes in a MAG is connected in such a way
+    that the graph is "maximal," meaning no additional edges can be added
+    without changing the set of implied conditional independence relations.
+    """
+
+    def __init__(
+        self,
+        ebunch: Optional[Iterable[tuple[Hashable, Hashable]]] = None,
+        latents: set[Hashable] = set(),
+    ):
```
ankurankan, line 30: Will need to add `roles` here, similar to the `DAG` class.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
...
+        Returns
+        -------
+        bool
+            True if `c` is a collider on the path, False otherwise.
+
+        Examples
+        --------
+        >>> from pgmpy.base import MAG
+        >>> mag = MAG()
+        >>> mag.add_edge("X", "Z", "-", ">")  # X -> Z
```
ankurankan: Remove the comments.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
...
+                if not self._is_collider(prev_node, curr_node, next_node):
+                    is_inducing = False
+                    break
+
+                ancestors_uv_vu = self.get_ancestors(u).union(self.get_ancestors(v))
+                if curr_node not in W and curr_node not in ancestors_uv_vu:
+                    is_inducing = False
+                    break
+
+            if is_inducing:
```
ankurankan: Isn't this if condition redundant?

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
...
+        for path in nx.all_simple_paths(self, source=u, target=v):
+            if len(path) <= 2:
+                continue
+
+            is_inducing = True
+            for i in range(1, len(path) - 1):
+                prev_node, curr_node, next_node = path[i - 1], path[i], path[i + 1]
+
+                if not self._is_collider(prev_node, curr_node, next_node):
+                    is_inducing = False
```
ankurankan, line 187: Can directly return the `False` value here.

pgmpy/tests/test_base/test_MAG.py
```diff
@@ -0,0 +1,81 @@
+import unittest
+
+from pgmpy.base import MAG
+
+
+class TestMAG(unittest.TestCase):
```
ankurankan: Could you also add the example from Figure 1 of the Zhang 2008 paper and check that methods are working according to the definitions?

ankurankan: Also please try to write these tests as pytest tests. We are trying to move away from unittest, so all new tests should be in pytest format.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
...
+            marks = self.edges[u, v]["marks"]
+            if u in X and marks.get(u) == "-" and marks.get(v) == ">":
+                edges_to_remove.append((u, v))
+            elif v in X and marks.get(v) == "-" and marks.get(u) == ">":
+                edges_to_remove.append((u, v))
+
+        new_mag.remove_edges_from(edges_to_remove)
+        return new_mag
+
+    def copy(self):
```
ankurankan: Can we move this to `AncestralBase` class? Should work for both MAG and PAG then

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,325 @@
...
+        True
+        """
+        if not self.has_edge(u, v):
+            return False
+
+        graph_without_edge = self.copy()
+        graph_without_edge.remove_edge(u, v)
+        return not graph_without_edge.has_inducing_path(u, v, self.latents)
+
+    def is_invisible_edge(self, u, v):
```
ankurankan: Wouldn't really add both these methods, given that one is the inverse of the other. Can just do a `not is_visible_edge` wherever needed in logic.

ankurankan (changes requested):

pgmpy/base/MAG.py
Spinachboul: Yeah this did not throw any error, but anyways I have changed this in my local

pgmpy/tests/test_base/test_MAG.py
Spinachboul: Ok! Will do that

pgmpy/base/MAG.py
Spinachboul: I guess we need to check if there is an inducing path, so if we return false here , it wont be checking for other paths
Is it really the case? I need to revise the definitions from the paper

pgmpy/base/MAG.py
Spinachboul: Can't we add roles to AncestralBase directly? would be more generalized and used by all the subclasses

pgmpy/tests/test_base/test_MAG.py
Spinachboul: @ankurankan Yeah we can extend that graph and add undirected edges as well

pgmpy/base/AncestralBase.py
```diff
@@ -4,8 +4,10 @@
 import networkx as nx
 import numpy as np
 
+from pgmpy.base._mixin_roles import _GraphRolesMixin
 
-class AncestralBase(nx.Graph):
+
+class AncestralBase(nx.Graph, _GraphRolesMixin):
     def __init__(
```
ankurankan, line 11: Need to add `roles` argument in `__init__`.

pgmpy/base/AncestralBase.py
```diff
@@ -492,3 +490,25 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
...
+
+        new_graph.add_nodes_from(self.nodes())
+
+        for u, v, data in self.edges(data=True):
+            u_mark, v_mark = data["marks"][u], data["marks"][v]
+            new_graph.add_edge(u, v, u_mark, v_mark)
+
+        if hasattr(self, "latents"):
+            new_graph.latents = set(self.latents)
+
```
ankurankan, line 590: Will need to copy roles as well.

pgmpy/base/AncestralBase.py
```diff
@@ -492,3 +490,25 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
...
+    def copy(self):
+        """
+        Return a copy of the graph, preserving edge marks and latents.
+
+        Returns
+        -------
+        MAG
+            Copy of the AncestralBase graph
+        """
+        new_graph = self.__class__()
```
ankurankan: A simpler way might be to simply create the object using the `AncestralBase.__init__` method itself.

pgmpy/base/AncestralBase.py
```diff
@@ -4,8 +4,10 @@
 import networkx as nx
 import numpy as np
 
+from pgmpy.base._mixin_roles import _GraphRolesMixin
 
-class AncestralBase(nx.Graph):
+
+class AncestralBase(nx.Graph, _GraphRolesMixin):
```
ankurankan, line 10: `AncestralBase` tests need to be modified too to consider the roles.

pgmpy/tests/test_base/test_MAG.py
```diff
@@ -0,0 +1,148 @@
+import pytest
```
ankurankan, line 1: I don't see Figure 1 example from the paper. Am I missing something?

ankurankan (changes requested):

pgmpy/tests/test_base/test_MAG.py
Spinachboul: I thought this was the figure you were referring to?

image

ankurankan: Yes, but the example in `mag` is different. Right?

Spinachboul: Ig I only added an extra node "L", otherwise I have taken the same graph or maybe something is wrong?

```
edges = [
        ("A", "B", ">", ">"),
        ("C", "D", ">", ">"),
        ("A", "C", ">", ">"),
        ("B", "D", ">", ">"),
        ("A", "D", "-", ">"),
        ("B", "C", "-", ">"),
        ("A", "L", "-", "-"),
        ("C", "L", "-", "-"),
    ]
```

ankurankan: Can we keep one test on the exact graph? This is to make sure that we are not accidentally misinterpreting any of the definitions. Also, please add a citation for that model, so that we can reference that in future if needed.

Spinachboul:
@ankurankan A few pointers for the latest commit which I made:
- I have used the **exact graph** in the paper for just one of the methods.
- I have used the `__class__` in the `copy` method since creating an instance was throwing errors. I was not able to unpack the values for the the same.
- I have cited the **zhang 2008** paper in the references section in the `AncestralBase` class docstrings in APAC format.

pgmpy/base/AncestralBase.py
```diff
@@ -492,3 +527,85 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
...
+        """
+        new_graph = self.__class__()
+
+        new_graph.add_nodes_from(self.nodes())
+
+        for u, v, data in self.edges(data=True):
+            u_mark, v_mark = data["marks"][u], data["marks"][v]
+            new_graph.add_edge(u, v, u_mark, v_mark)
+
+        if hasattr(self, "latents"):
```
ankurankan: Why the `hasattr` check? Isn't it guaranteed that the object will have a `latents` attribute?

pgmpy/base/AncestralBase.py
```diff
@@ -492,3 +527,85 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
...
+            new_graph.latents = set(self.latents)
+
+        if hasattr(self, "get_role_dict"):
```
ankurankan: Same here.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,295 @@
...
+    independence relations when some variables are latent (unobserved). Unlike
+    simple directed acyclic graphs (DAGs), MAGs allow for special edge types
+    (directed and bidirected) that capture the presence of latent confounding
+    and selection bias. Every pair of nodes in a MAG is connected in such a way
+    that the graph is "maximal," meaning no additional edges can be added
+    without changing the set of implied conditional independence relations.
+
+    References
+    ----------
+        [1] Zhang, J. (2008). Causal Reasoning with Ancestral Graphs. Journal of Machine Learning Research, 9(7).
```
ankurankan: Indentation needs to match; otherwise, it won't render properly on the website.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,295 @@
...
+
+        Vertices of a specific role can be retrieved using ``get_role`` method.
+
+        >>> mag.get_role("exposure")
+        ["A"]
+        >>> mag.get_role("adjustment")
+        ["L", "C"]
+
+        """
+        super().__init__(ebunch=ebunch, latents=latents, roles=roles)
```
ankurankan, line 98: We need to check the edge types here before calling super. Not all edge types are allowed in MAGs.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,295 @@
...
+        Examples
+        --------
+        >>> from pgmpy.base import MAG
+        >>> mag = MAG()
+        >>> mag.add_edge("X", "Z", "-", ">")
+        >>> mag.add_edge("Y", "Z", "-", ">")
+        >>> mag._is_collider("X", "Z", "Y")
+        True
+        """
+        mark_uc_at_c = self.edges[u, c]["marks"][c]
```
ankurankan, line 135: We will need to add checks to see whether the edges exist. Without the checks, the current code would throw a KeyError, which would be hard to debug.

pgmpy/tests/test_base/test_MAG.py
```diff
@@ -0,0 +1,155 @@
...
+    m4 = MAG(
+        ebunch=[("X", "Z", ">", ">"), ("Y", "Z", "-", ">")], latents={"L"}, roles=roles
+    )
+    assert m1 != m4
+
+    m5 = MAG(ebunch=e, latents={"L", "U"}, roles=roles)
+    assert m1 != m5
+
+
+def test_is_collider_fixture(mag):
```
ankurankan: Why do these tests have `fixture` in their names? It is confusing.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,295 @@
...
+            neighbors = list(self.neighbors(node))
+
+            for i in range(len(neighbors)):
+                for j in range(i + 1, len(neighbors)):
+                    u, v = neighbors[i], neighbors[j]
+
+                    if self._is_collider(u, node, v):
+                        new_mag.add_edge(u, v, ">", ">")
+
+        new_mag.remove_nodes_from(X)
```
ankurankan: I don't think any of these manipulations is supposed to remove the node from the graph. Can you double-check?

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,295 @@
...
+        True
+        """
+        if not self.has_edge(u, v):
+            return False
+
+        graph_without_edge = self.copy()
+        graph_without_edge.remove_edge(u, v)
+        return not graph_without_edge.has_inducing_path(u, v, self.latents)
+
+    def lower_manipulation(self, X):
```
ankurankan: This should accept an `inplace` boolean argument. Search for examples in the current codebase.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,295 @@
...
+                for j in range(i + 1, len(neighbors)):
+                    u, v = neighbors[i], neighbors[j]
+
+                    if self._is_collider(u, node, v):
+                        new_mag.add_edge(u, v, ">", ">")
+
+        new_mag.remove_nodes_from(X)
+        return new_mag
+
+    def upper_manipulation(self, X):
```
ankurankan: Same here.

pgmpy/base/MAG.py
```diff
@@ -0,0 +1,295 @@
...
+    def _is_collider(self, u, c, v):
+        """
+        Check if a node is a collider in a path u - c - v.
+
+        A collider is a node with incoming arrowheads on both sides:
+        u -> c <- v.
+
+        Parameters
+        ----------
+        u : Hashable
```
ankurankan, line 109: We need to be consistent with the spacing between the arguments. Leave an empty line between each argument.

pgmpy/tests/test_base/test_AncestralBase.py
```diff
@@ -165,3 +165,48 @@ def test_adjacency_matrix_setter(self):
 
         self.assertEqual(graph["X_0"]["X_1"]["marks"], {"X_0": ">", "X_1": "-"})
         self.assertEqual(graph["X_1"]["X_2"]["marks"], {"X_1": ">", "X_2": "-"})
+
+    def test_init_with_roles(self):
+        """Test initialization with roles assigned at construction."""
```
ankurankan: I don't think these comments are needed.

pgmpy/tests/test_base/test_MAG.py
ankurankan: I still don't see the model from the paper in the tests. Can you point me to it?

pgmpy/tests/test_base/test_MAG.py
```diff
@@ -0,0 +1,155 @@
+import pytest
+
+from pgmpy.base import MAG
+
+
+@pytest.fixture
```
ankurankan, line 7: We need to clean up the tests. Try to keep one test function for each method in the class. This file is unreadable right now.

ankurankan (changes requested):
I am not sure how much help you are taking from LLMs for writing the code, but any generated code needs to be scrutinized very carefully, which I am not seeing in the current PRs. It is also not sustainable to go through this many rounds of review and add comments for relatively simple errors. In the future, can you please spend some more time making sure that everything looks correct before requesting reviews?

Spinachboul:
@ankurankan I did make the changes which we discussed in the meet
My question was : I have taken a relatively simple graph as my second graph and only tested the manipulation functions for it, since I thought those were the trickier methods. Will this be fine or something else also needs to added/modified

Here is the graph which I have taken as `mag2`:
```python
@pytest.fixture
def mag2():
    edges = [
        ("P", "Q", ">", ">"),
        ("Q", "R", "-", ">"),
        ("P", "R", "-", ">"),
        ("P", "L", "-", ">"),
    ]
    return MAG(ebunch=edges, latents={"L"})
```

ankurankan (approved):

PR #2324, opened 2025-08-27
[MNT] make `torch` a soft dependency, isolate soft deps in tests

### LINKED ISSUE #2266: [ENH] Make `torch` a soft dependency
The inclusion of `torch` as a core dependency in `pgmpy` is somewhat problematic, at least because:

1. `torch` wheels for intel-based MacOS laptops have been deprecated.
2. `torch` is a really heavy package so it increases `pgmpy`'s installation time 

Adapt a soft dependency framework and move `torch` from a core dependency to a soft dependency. Identify the parts of `pgmpy` that require `torch` and add a proper error/warning if `torch` is not installed.

fkiraly:
This PR makes `torch` a soft dependency, closes #2266 

* ensures the package can be used, and all modules imported, without `torch` being present
* moves `torch` from the core dep set into the `optional` dep set, and a new "torch only" depset `torch`
* adds a `test-nosoftdeps` CI matrix, to run tests when no optional dependencies are present

This PR also isolates all soft dependencies in the tests, i.e., tests relying on soft dependencies are not executed when the soft dependencies are not present.

Note: this introduces a dependency on `scikit-base` for the dependency checking utilities. The `scikit-base` package has no dependencies itself.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ on: [push, pull_request]
 
 jobs:
   test-nosoftdeps:
-    name: Running tests - OS - ${{ matrix.os }}; Python - ${{ matrix.python-version}}
+    name: test-nosoftdeps (${{ matrix.os }}, python ${{ matrix.python-version}})
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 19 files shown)


ankurankan (approved):
Thanks for the PR @fkiraly

PR #2326, opened 2025-08-28
[MNT] Isolate `torch` and `pyro` as soft dependencies

### LINKED ISSUE #2325: [ENH] Make `pyro` a soft dependency
In a previously closed issue #2266, I argued that `torch` should be made a soft dependency. This was implemented in #2324. However, `pgmpy` also used `torch` indirectly through `pyro` in `FunctionalCPD.py` and `FunctionalBayesianNetwork.py`. Therefore, for the same reasons, I'd like to recommend that `pyro-ppl` is moved as a soft dependency. That being said, not sure if this can be easily done or done at all.

### LINKED ISSUE #2327: [BUG] `FunctionalBayesianNetwork` silently sets backend globally to `torch`
`FunctionalBayesianNetwork`, when constructed, sets the backend globally to `torch`.

This is likely entirely unexpected to a user, and I suspect it also masked a few bugs in tests, see https://github.com/pgmpy/pgmpy/actions/runs/17302360964/job/49116586162?pr=2326 - leading to a number of tests never being executed on the `numpy` backend.

I consider silent switching of a central global variable a serious bug, so this should be remedied.

fkiraly:
Fixes #2325. Follows #2324 and #2266 - as @marrov noticed, `torch` is an indirect dependency through `pyro-ppl`, hence in order to isolate `torch`, we must also isolate `pyro-ppl` as a soft dependency.

This PR carries out isolation of `pyro-ppl` along the lines of #2324, using the `_check_soft_dependencies` and `_safe_import` utilities from `scikit-base`.

It also completes the isolation of `torch` as a soft dependency - due to the indirect dependency on `pyro-ppl`, #2324 was incomplete.

Also fixes https://github.com/pgmpy/pgmpy/issues/2327 (this seems necessary to avoid side effects on the global config, throughout the tests)


marrov:
> Follows #2324 and #2266 - as @marrov noticed, `torch` is an indirect dependency of `pyro-ppl`, hence in order to isolate `torch`, we must also isolate `pyro-ppl` as a soft dependency.
> 
> This PR carries out isolation of `pyro-ppl` along the lines of #2324, using the `_check_soft_dependencies` and `_safe_import` utilities from `scikit-base`.

Thanks! I opened an issue for this, which you can link: #2325

fkiraly:
This is maybe a question for @ankurankan - how much of `pgmpy` actually remains if we remove `pyro-ppl`?

If it is "not much", then maybe we should revert the `torch` isolation PR instead...

fkiraly:
I see, so most models have a torch and a numpy backend. So removing torch only prevents users from using torch based backends, but not the numpy backend.

fkiraly:
I think this is ready now.

The remaining failure is this:
`FAILED pgmpy/tests/test_inference/test_ApproxInference.py::TestApproxInferenceBN::test_query_evidence - AssertionError: False is not true`

I think this is a genuine failure and it was masked previously due to bug https://github.com/pgmpy/pgmpy/issues/2327, which was preventing some estimators from getting tested with the `numpy` backend.

ankurankan (approved):

ankurankan:
Thanks for the PR @fkiraly

PR #2328, opened 2025-08-28
[ENH] fix deprecation warnings

fkiraly:
Fixes a few warnings:

* deprecated default for `observed` in `pandas` `groupby`
* classes starting with `Test` but not being test classes. Objects starting with `Test` should be test classes reserved for `pytest`, if the latter is present.


ankurankan (approved):
Thanks for the PR.

PR #2338, opened 2025-09-03
[ENH] Move `latents` attribute to variable roles

### LINKED ISSUE #2322: [ENH] Move `latents` attribute to variable roles
All the base model classes in pgmpy have the option to specify which variables in the graph are latent. These variables are then stored as an attribute of the class called `latents`. Since we now have the functionality to specify roles to variables, it would make sense to treat `latent` as a role of the variables as well, instead of storing them as a separate attribute.

Changes Required:
1. The `DAG.add_node` specifies a `weight` argument as well. I had a quick look, but I don't see it getting used anywhere. If it's not getting used, we can get rid of it.
2. Use the `GraphRolesMixin` to store the latent variables, similar to how other roles (such as exposure, outcome, etc.) are stored.
3. Add a property called latents to all the base model classes such that users can still do `model.latents` to get all the latent variables in the model.
4. Remove the `add_node` and `add_nodes_from` methods from all classes and update the documentation to show how to add a latent variable using keyword arguments.

daehyun99:

- Fixes #2322

#### 1. `latents`
- [x] Delete `latents` argument and related logic in `DAG.add_node`, `DAG.add_nodes_from`.
- [x] Delete test code for testing `latents` argument.
- [x] Use the `GraphRolesMixin` to store the latent variables.
- [x] Add a property called `latents`.
- [x] Add test code for `latents` using `with_role` method of `DAG`, `PDAG`, `ADMG`, `AncestralBase`, ~~`UndirectedGraph`~~ classes.
- [ ] Update the documentation to show how to add a latent variable using keyword arguments.
#### 2. `weights`
- ~~Delete `weight` argument and related logic in `DAG.add_node`.~~
- ~~Delete test code for testing `weight` argument.~~
- ~~Remove the `add_node` and `add_nodes_from` methods from all classes.~~


ankurankan:
@daehyun99 I guess we found out where the `weight` argument was getting used. I think this should get resolved once you remove the `add_node` and `add_nodes_from` methods as `nx`'s methods should be able to accept the additional argument.

daehyun99:
@ankurankan  I have removed the `add_node` and `add_nodes_from` methods and tested it in my local environment, but the same error occurred.
Also, I found the following comment in the `DAG.add_nodes_from` method, but I need more time to understand what it means.
```py
    def add_nodes_from(
        self,
        nodes: Iterable[Hashable],
        weights: Optional[list[float] | tuple[float]] = None,
    ):
        """
        Add multiple nodes to the Graph.
        **The behviour of adding weights is different than in networkx.
        ...
        """
```
Therefore, I will perform the `latent`-related work first, and then perform the `weight`-related work.

daehyun99:
@ankurankan Hi. Could you please do a code review only for the `latents` work?

While working on `latents`, I modified the code for the `AncestralBase` and `TestAncestralBase` classes. ([commit](https://github.com/pgmpy/pgmpy/pull/2338/commits/57741e1d43d8521d2cda4ed2cc2dd882890227b1))

This may cause conflicts with #2321.

Therefore, it might be confusing to do a code review after the `weights` work is also done.

Thanks.

ankurankan:
@daehyun99 Sorry for the delay. I think it looks good. One issue I see now is that because we have removed the `add_node` and `add_nodes_from` arguments, we can not add latent variables using these methods. Is there a workaround to this or using the `with_roles` method the only way?

ankurankan:
I have also resolved some merge conflicts on your branch, so please make sure to do a `git pull` before adding new commits.

daehyun99:
> Is there a workaround to this or using the `with_roles` method the only way?

@ankurankan I have written the sample code below.

- How to add `latents` values to the declaration of a graph model
```py
# way1
>>> G = DAG(
...     ebunch=[("U", "X"), ("X", "M"), ("M", "Y"), ("U", "Y"), ("V", "M")],
...     latents={"U", "V"},
... )
>>> sorted(G.latents)
['U', 'V']

# way2
>>> G = DAG(
...     ebunch=[("U", "X"), ("X", "M"), ("M", "Y"), ("U", "Y"), ("V", "M")],
...     roles ={"latents": ("U", "V")},
... )
>>> sorted(G.latents)
['U', 'V']
```

- How to add `latents` values to a previously declared graph model
```py
# way1
>>> G.latents = "X"
>>> sorted(G.latents)
['U', 'V', 'X']

# way2
>>> G.with_role("latents", "X", inplace=True)
>>> sorted(G.latents)
['U', 'V', 'X']
```

- How to remove `latents` values from a graph model
```py
# only way
>>> G.without_role("latents", "X", inplace=True)
>>> sorted(G.latents)
['U', 'V']
```

daehyun99:
@ankurankan Ready to merge.

pgmpy/base/_mixin_roles.py
```diff
@@ -149,3 +156,31 @@ def is_valid_causal_structure(self) -> bool:
...
+        {'a'}
+        """
+        if self.has_role("latents"):
+            return set(self.get_role("latents"))
+        else:
+            return set()
+
+    @latents.setter
+    def latents(self, variables):
+        self.with_role(role="latents", variables=variables, inplace=True)
```
ankurankan, line 200: I don't find this behavior for a setter method to be intuitive. I think the setter method should replace the existing latents with the ones provided in `variables`. For the general modification operations, we would still have the `with` and `without`  role method. What do you think?

pgmpy/tests/test_base/test_ADMG.py
```diff
@@ -56,6 +57,50 @@ def test_initialization_with_roles(self):
...
+        admg.with_role(role="latents", variables="C", inplace=True)
+        admg.with_role(role="latents", variables=["D", "E", "F"], inplace=True)
+
+        assert admg.latents == {"A", "B", "C", "D", "E", "F"}
+        assert set(admg.get_role("latents")) == {"A", "B", "C", "D", "E", "F"}
+
+        with pytest.raises(ValueError, match="Variable 'G' not found in the graph."):
+            admg.with_role(role="latents", variables="G", inplace=True)
+
+    def test_latnets_without_role(self):
```
ankurankan: A small typo: `_latnets_` -> `_latents_`

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -73,79 +73,36 @@ def test_add_node_string(self):
...
     def test_add_nodes_from_non_string(self):
         self.graph = DAG()
         self.graph.add_nodes_from([1, 2, 3, 4])
 
-        self.graph = DAG()
-        self.graph.add_nodes_from([1, 2, 3, 4], latent=True)
-
-        self.graph = DAG()
-        self.graph.add_nodes_from([1, 2, 3, 4], latent=[True, False, False, False])
+        # TODO: What needs to be resolved before merging PR #2338
```
ankurankan: Thanks for catching this. Could you add some assert statements here, please?

ankurankan (changes requested):
@daehyun99 Thanks a lot for the PR, it mostly looks good to me. I have added a few very minor comments. Could you please have a look?

pgmpy/base/_mixin_roles.py
daehyun99: @ankurankan 
Yes. I also think the behavior of the setter is not intuitive. 
I've modified the code to replace the existing `latents` value.

```py
    @latents.setter
    def latents(self, variables):
        """
        Replace the `latents` nodes.

        Parameters
        ----------
        variables: set of nodes (default: empty set)
            A set of latent variables in the graph. These are not observed
            variables but are used to represent unobserved confounding or
            other latent structures.
        """
        if self.has_role("latents"):
            self.without_role(
                role="latents", variables=self.get_role("latents"), inplace=True
            )
        self.with_role(role="latents", variables=variables, inplace=True)
```

pgmpy/tests/test_base/test_DAG.py
daehyun99: It has been applied.
Additionally, I added assert statements for some tests.

ankurankan (approved):
@daehyun99 Looks good to me. Thanks a lot for the PR.

Issue #2259, opened 2025-07-29
[ENH] Implement the PAG class


ankurankan:
Ref #1225

Spinachboul:
@ankurankan Commenting here for getting assigned

PR #2346, opened 2025-09-08
Partial Ancestral Graph + FCI Algorithm

Spinachboul:
Solves https://github.com/pgmpy/pgmpy/issues/2259

Also will include the implementation of `FCI` (Fast Causal Inference Algorithm) which produces PAG.


ankurankan:
@daehyun99, since you would be working on the refactor of the `PAG` class as well, would you like to give this a review too?

daehyun99:
@ankurankan Sure, I'll give it a look, though I'm still learning.

Spinachboul:
@ankurankan
In the tests, should I be using smaller graphs (but with all possible edge configurations) so that it is easier to test the functions and debug them.
This is the current graph I am considering which I realised is a little too complex:
```
edges = [
        ("A", "B", "-", ">"),
        ("A", "C", ">", "-"),
        ("B", "C", "-", "-"),
        ("D", "E", "o", "o"),
        ("E", "F", "o", "o"),
        ("U", "W", "o", ">"),
        ("V", "W", ">", "-"),
        ("X", "W", ">", "-"),
        ("U", "V", "-", "-"),
        ("U", "X", "-", "-"),
        ("M", "N", "-", ">"),
        ("N", "O", "-", ">"),
        ("C", "D", "-", "-"),
        ("X", "F", "-", "-"),
        

    ]
    return PAG(ebunch=edges)
```

> Also please pin if there are better options from the related papers.

Spinachboul:
> @daehyun99, since you would be working on the refactor of the `PAG` class as well, would you like to give this a review too?

Yeah that is a good idea, since only the test functions need to be implemented now.
so simultaneously we can start the refactoring process.
@ankurankan Your thoughts?

ankurankan:
@Spinachboul @daehyun99  I would suggest waiting till this is done. If anything goes wrong with this PR, it will end up as additional work to fix it twice.

Spinachboul:
Oh ok, makes sense

Spinachboul:
> @ankurankan In the tests, should I be using smaller graphs (but with all possible edge configurations) so that it is easier to test the functions and debug them. This is the current graph I am considering which I realised is a little too complex:
> 
> ```
> edges = [
>         ("A", "B", "-", ">"),
>         ("A", "C", ">", "-"),
>         ("B", "C", "-", "-"),
>         ("D", "E", "o", "o"),
>         ("E", "F", "o", "o"),
>         ("U", "W", "o", ">"),
>         ("V", "W", ">", "-"),
>         ("X", "W", ">", "-"),
>         ("U", "V", "-", "-"),
>         ("U", "X", "-", "-"),
>         ("M", "N", "-", ">"),
>         ("N", "O", "-", ">"),
>         ("C", "D", "-", "-"),
>         ("X", "F", "-", "-"),
>         
> 
>     ]
>     return PAG(ebunch=edges)
> ```
> 
> > Also please pin if there are better options from the related papers.

  @ankurankan Your thoughts?

ankurankan:
@Spinachboul Yes, it's usually better to use the minimal graph that has all potential cases that we need to handle.

Spinachboul:
> @Spinachboul Yes, it's usually better to use the minimal graph that has all potential cases that we need to handle.

Yeah that makes sense, will make the changes

Spinachboul:
@ankurankan 
I would request you to give this PR a review.
I still haven't completed the tests for rules 3, 4 and 5, would do that soon.
I plan to resolve the current issues and your suggestions (post review) together.

Spinachboul:
@ankurankan 
I am working on the tests of `FCI` and `BaseConstrainedEstimator`
I wonder should I begin workin on them in a separate PR or should I include them in this PR itself?
Separate PR as this might make things more systematic to handle and review,.

pgmpy/estimators/BaseConstraintEstimator.py
```diff
@@ -354,3 +356,69 @@ def _get_potential_sepsets(
...
             combinations(separating_set_v, lim_neighbors),
         )
+
+    @staticmethod
+    def orient_colliders(
+        skeleton: UndirectedGraph,
+        separating_sets: Dict[FrozenSet, Set],
+        temporal_ordering: Dict[Hashable, int] = dict(),
+        graph_cls: Type = PDAG,  # By default
+        # here this will return the list of tuples
```
Spinachboul: @ankurankan 
I am not able to recall why did I write this comment, and there also seems to be a design conflict since the function expects the subclasses to include `directed_ebunch` and `undirected_ebunch` in the constructor of PAG.

Could you maybe give a direction on where should go from here, or maybe explain why this comment was made.

Spinachboul:
@ankurankan 
I have updated the `orient_colliders` method in the **BaseConstraintEstimator**
Could you please have a look [here](https://github.com/pgmpy/pgmpy/pull/2346/changes#diff-94916f165053bc97e160e0e6f42b3c22df5b0f73c3be7e44e42b74e3aea366b9R510) and comment on the same.

Also I would request you to have a look at the `test_FCI` files, and suggest any better approach (if any) for the usage of the BNLearn datasets in the examples.

I would also be fixing the issues regarding the orientation rules in the `test_PAG` file soon.

Thanks!

Spinachboul:
@ankurankan 
Could you please review this PR once when you have time?
Feel free to skip the tests which have been commented since that is intentional.
The CI seems to be passing now.
Thanks!

Spinachboul:
@ankurankan 

This PR seems to have corrupted, closing this and will create a new branch with a new PR

PR #2347, opened 2025-09-10
Dataset loader

### LINKED ISSUE #2323: [ENH] Add data loader functionality to easily import example causal datasets
The repository https://github.com/cmu-phil/example-causal-datasets contains a collection of datasets, some of which include causal ground truth. It would be useful to add a `get_example_dataset` method (similar to `get_example_model` that can read the datasets from the repo into a pandas DataFrame and also load the ground truth model.

Theavinash02:
For [#2323](https://github.com/pgmpy/pgmpy/issues/2323) ,I am working on adding an API for loading example datasets from [cmu-phil/example-causal-datasets](https://github.com/cmu-phil/example-causal-datasets)

Currently, get_example_dataset only supports single-file datasets.
However, some datasets in the example-causal-datasets repo (like Blue-driver, Adult, Dry-bean) have multiple data files in their data/ folder. Users should be able to:
- load the default file,
- select a specific variant, or
- merge all files into a single DataFrame.

Without this, the method can’t handle these common multi-file cases.

Fixes: #2323 

**Progress:**
- Implemented Scenario 1 (single-file datasets):
     - Loads data into a DataFrame.
     - Optionally loads ground truth (ground.truth folder).
     - Verified with datasets like Airfoil-self-noise, Algerian-forest-fires, Auto-mpg.

- Started Scenario 2 (multi-file datasets):
    - Added a registry entry for Blue-driver with default, variants, all_files, and header_file.
    - Implemented `_resolve_files_for_entry` helper to decide which files to load based on variant and merge options.
    - Tested the resolver logic locally — it returns the correct file list for default, variant, and merge cases.

Scenario 2 is in progress. [Need to review and clean the code.]


Theavinash02:
Hi @ankurankan ,
I have come across this error while in CI test. Could you please check mentioned below and kindly advise on the same?

```
All done! ✨ 🍰 ✨
1 file would be reformatted, 177 files would be left unchanged.
Skipping .ipynb files as Jupyter dependencies are not installed.
You can fix this by running pip install "black[jupyter]"
would reformat /home/runner/work/pgmpy/pgmpy/pgmpy/utils/registry_example_dataset.py
```

I have done CI check again after pip install "black[jupyter]", and still facing the same error.

daehyun99:
@Theavinash02 
Hi, Does the same error occur even after executing the following command in the CLI?
```sh
black pgmpy/utils/registry_example_dataset.py
```

Theavinash02:
Hi @daehyun99 ,
Yes, I have tried the command you have mentioned. Still has the same error.

Theavinash02:
Hi @daehyun99 ,
I will try once again and update here!
Thanks
Avinash

ankurankan:
@Theavinash02 Thanks a lot for the PR. This is in a good direction. I am, however, still thinking about the design of this. I will add a comment on the issue with the design that I am thinking.

Theavinash02:
Thank @ankurankan .

pgmpy/utils/registry_example_dataset.py
ankurankan, line 1: I don't think we will need an explicit list of datasets like this. The `dataset_class` decorator that you have defined in `_registry.py` should be able to automatically handle adding datasets to the registry. When you write a class for the dataset and decorate with the `dataset_class`, it should automatically add that dataset to the registry.

pgmpy/datasets/_registry.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+# Exceptions ---------------------------------------------------------
+class DatasetRegistrationError(Exception):
+    """Raised for problems during dataset registration (duplicate names, invalid attrs)."""
+
+
+class DatasetNotFoundError(KeyError):
+    """Raised when requested dataset name is not found in registry."""
+
```
ankurankan: I would avoid defining new custom errors unless absolutely necessary. I think we should be able to handle both these scenarios with just `ValueError` and a good error message.

Theavinash02: Hi Ankur. 
I am reworking the entire code again. I will share the same once completed🤝🏻

pgmpy/utils/registry_example_dataset.py
Theavinash02: Sure, Ankur 👍🏻

ankurankan:
@Theavinash02 Just wanted to check if there were any updates on this? Thanks.

Theavinash02:
Hi @ankurankan ,
I am busy with IIT and my job. 
I haven't progressed much in the issue yet. I will start and get back to you with an update. Thanks

Theavinash02:
Hi @ankurankan ,
Could you please review the code for the new API?
I have added two datasets for testing. Once confirmed, I will add the n number of datasets from the repo as discussed before.
Thanks
Avinash

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+    text = raw.decode("utf-8-sig", errors="ignore")
+
+    sachs_graph = parse_sachs_graph(text)
+    if sachs_graph is not None:
+        return sachs_graph
+
+    tier_graph = parse_tier_graph(text)
+    if tier_graph is not None:
+        return tier_graph
+    return None
```
ankurankan: If these methods are only getting used in `BaseDataset`, I would put these inside the class with non-public method names starting with `_`.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+        raw = cls.load_or_fetch("ground_truth")
+        return parse_knowledge_txt(raw)
+
+
+def load_abalone(
+    variant: Optional[str] = None, load_ground_truth: bool = True
+) -> Tuple[pd.DataFrame, Optional[DAG]]:
+    df = Abalone.load(variant=variant)
+    gt = Abalone.load_ground_truth() if load_ground_truth else None
+    return df, gt
```
ankurankan: I don't think we should implement these individual `load_*` methods. The `load_dataset` method should handle all loading by name.

pgmpy/utils/registry_example_dataset.py
```diff
@@ -0,0 +1,216 @@
+_EXDATA_REGISTRY_SCENARIO1 = {
```
ankurankan: I am not sure what the purpose of this variable is?

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+from pathlib import Path
+from typing import Any, Dict, Optional, Tuple
+
+import networkx as nx
+import pandas as pd
+import requests
+
+from pgmpy.base import DAG
+
+from .registry import DATASETS, dataset_class
```
ankurankan: Please use absolute paths: `from pgmpy.datasets import DATASETS, dataset_class`

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def load_df_from_txt(raw: bytes) -> pd.DataFrame:
+    try:
+        return pd.read_csv(io.BytesIO(raw), sep=None, engine="python")
+    except Exception as e:
+        try:
+            return pd.read_csv(io.BytesIO(raw), sep=",", engine="python")
+        except Exception:
+            raise e
```
ankurankan: Please don't use try-except blocks. Unexpected events should be handled explicitly. Should look into the reason why this could throw an error and that case should be handled explicitly.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+
+
+def parse_sachs_graph(
```
ankurankan: If this is a method specific to the `Sachs` dataset, this method should be in the `Sachs` class.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+                        G.add_edge(u, v)
+                except Exception:
+                    continue
+        return DAG(G)
+
+    except Exception:
+        return None
+
+
+def parse_tier_graph(
```
ankurankan: If this is a generalized method, it should be in `BaseDataset`.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+            ext = ".parquet"
+        elif key.startswith("ground_truth"):
+            ext = ".json"
+        return os.path.join(
+            PGMPY_DATA_HOME, f"{cls.name}_{key.replace(':','-')}_{safe}{ext}"
+        )
+
+    @classmethod
+    def _fetch(cls, key: str) -> bytes:
+        raise NotImplementedError
```
ankurankan: Isn't fetch logic is going to be the same for all datasets? If yes, that should be implemented here.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+    if ds is None:
+        raise ValueError(
+            f"Dataset '{name}' not found. Available datasets: {DATASETS.list_all()}"
+        )
+
+    try:
+        X = ds.load()
+    except TypeError:
+        X = ds.load()
+
```
ankurankan: We should avoid using try-except unless absolutely necessary.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+        missing_in_gt = data_cols - gt_nodes
+        gt.add_nodes_from(missing_in_gt)
+
+        missing_in_data = gt_nodes - data_cols
+        if missing_in_data:
+            import warnings
+
+            warnings.warn(
+                f"Ground truth for '{name}' has nodes not in data: {missing_in_data}"
+            )
```
ankurankan: Use `pgmpy.global_vars.logger` for giving warning messages. You can find examples of how to use it at other places in the repo.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+        "is_simulated": False,
+        "n_variables": 9,
+        "n_samples": 4177,
+        "is_discrete": False,
+        "is_continuous": False,
+        "is_mixed": True,
+        "is_ordinal": False,
+    }
+
+    base_url = "https://raw.githubusercontent.com/cmu-phil/example-causal-datasets/main/real/abalone/"
```
ankurankan: I have created a fork of the original dataset repo in pgmpy: https://github.com/pgmpy/example-causal-datasets. Could you please update the link to our repo's? This would make sure that our code does randomly break with updates to the original repo.

pgmpy/datasets/_base.py
```diff
@@ -0,0 +1,342 @@
...
+    @classmethod
+    def load(cls):
+        raise NotImplementedError
+
+    @classmethod
+    def load_ground_truth(cls):
+        return None
+
+
+def load_dataset(
```
ankurankan: I think we might need to add a `variant` argument here to allow users to choose which variant to load. It can be `None` by default which loads the default variant of the dataset.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,342 @@
...
+    @classmethod
+    def load(cls, variant: Optional[str] = None) -> pd.DataFrame:
+        v = variant or cls.DEFAULT_VARIANT
+        raw = cls.load_or_fetch(f"data:{v}")
+        return load_df_from_txt(raw)
+
+    @classmethod
+    def load_ground_truth(cls) -> Optional[DAG]:
+        raw = cls.load_or_fetch("ground_truth")
+        return parse_knowledge_txt(raw)
```
ankurankan: I see that both the `load` and `load_ground_truth` methods are exactly the same in both dataset classes. Why not implement this in the `BaseDataset` class? The idea then would be that if the default `load` and `load_ground_truth` doesn't work of some specific dataset, we can implement these methods for just those specific classes.

ankurankan (changes requested):
@Theavinash02 I think this is going in the correct direction but still needs a bit of work. I have added a few comments; could you please also write the docstrings for the methods?

Theavinash02:
Hi @ankurankan ,
I have addressed the review comments. Could you please review the code once and kindly share your feedback on the same?
Thanks

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+
+import networkx as nx
+import pandas as pd
+import requests
+
+from pgmpy.base import DAG
+from pgmpy.datasets.registry import DATASETS, dataset_class
+from pgmpy.global_vars import logger
+
+PGMPY_DATA_HOME = os.environ.get(
```
ankurankan: I think this should go into the `Config` class in `pgmpy.global_vars`.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+    "PGMPY_DATA_HOME", os.path.join(Path.home(), ".pgmpy", "data")
+)
+
+
+class BaseDataset:
+    name: str = ""
+    description: str = ""
+    tags: Dict[str, Any] = {}
```
ankurankan: Do we need to specify these in the `BaseDataset`? I think its fine to only define these in the child classes.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+
+    VARIANT_URLS: Dict[str, str] = {}
+    GROUND_TRUTH_URL: str = ""
+    DEFAULT_VARIANT: str = ""
```
ankurankan: Same for these.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+
+    @staticmethod
+    def _load_df_from_txt(raw: bytes) -> pd.DataFrame:
+        return pd.read_csv(io.BytesIO(raw), sep="\t")
+
+    @classmethod
+    def cache_path(cls, key: str) -> str:
+        cls._ensure_dir(PGMPY_DATA_HOME)
+        safe = hashlib.sha256(f"{cls.name}:{key}".encode()).hexdigest()[:16]
+        ext = ".bin"
```
ankurankan: Why do we need to have a default value? Shouldn't we throw an error if neither of the keys below match?

pgmpy/datasets/registry.py
```diff
@@ -0,0 +1,28 @@
+class Registry:
```
ankurankan: I think `Registry` is a bit vague, maybe rename it to `DatasetRegistry`?

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+        raise ValueError(
+            f"Dataset '{name}' not found. Available datasets: {DATASETS.list_all()}"
+        )
+
+    X = ds.load(variant=variant)
+
+    gt = ds.load_ground_truth() if load_ground_truth else None
+
+    if gt is None:
+        gt = DAG()
```
ankurankan: I would suggest keeping the return type based on the `load_ground_truth` value. If True, return both data and graph. If False, return only the data.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+
+    @classmethod
+    def _parse_ground_truth(cls, raw: bytes) -> Optional[DAG]:
+        raise NotImplementedError
+
+
+def load_dataset(
+    name: str, variant: Optional[str] = None, load_ground_truth: bool = True
+) -> Tuple[pd.DataFrame, Optional[DAG]]:
+    ds = DATASETS.get(name)
```
ankurankan: Would suggest using more descriptive names than `ds`. Could call it `dataset`, or `data`.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+    if ds is None:
+        raise ValueError(
+            f"Dataset '{name}' not found. Available datasets: {DATASETS.list_all()}"
+        )
+
+    X = ds.load(variant=variant)
+
+    gt = ds.load_ground_truth() if load_ground_truth else None
```
ankurankan: Same here, would suggest using the full name `ground_truth`.

pgmpy/datasets/_base.py
```diff
@@ -0,0 +1,265 @@
...
+
+    if data_cols != gt_nodes:
+        missing_in_gt = data_cols - gt_nodes
+        gt.add_nodes_from(missing_in_gt)
+
+        missing_in_data = gt_nodes - data_cols
+        if missing_in_data:
+            logger.warning(
+                f"Ground truth for '{name}' has nodes not in data: {missing_in_data}"
+            )
```
ankurankan: This doesn't seem to be adding any edges to the graph?

pgmpy/datasets/_base.py
```diff
@@ -0,0 +1,265 @@
...
+
+    gt = ds.load_ground_truth() if load_ground_truth else None
+
+    if gt is None:
+        gt = DAG()
+
+    data_cols = set(X.columns)
+    gt_nodes = set(gt.nodes())
+
+    if data_cols != gt_nodes:
```
ankurankan: I don't think this condition would ever be true. If this happens, that would mean that the dataset is incorrect.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+        elif key.startswith("ground_truth"):
+            ext = ".txt"
+        return os.path.join(
+            PGMPY_DATA_HOME, f"{cls.name}_{key.replace(':','-')}_{safe}{ext}"
+        )
+
+    @classmethod
+    def _fetch(cls, url: str) -> bytes:
+        resp = requests.get(url, timeout=60)
+        resp.raise_for_status()
```
ankurankan: What does this `raise_for_status` do here?

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+        if key.startswith("data"):
+            ext = ".csv"
+        elif key.startswith("ground_truth"):
+            ext = ".txt"
+        return os.path.join(
+            PGMPY_DATA_HOME, f"{cls.name}_{key.replace(':','-')}_{safe}{ext}"
+        )
+
+    @classmethod
+    def _fetch(cls, url: str) -> bytes:
```
ankurankan: Since this is a 2 line method and is used only within `load_or_fetch`, I would suggest removing this method and moving the logic within `load_or_fetch`.

pgmpy/datasets/datasets.py
```diff
@@ -0,0 +1,265 @@
...
+    def load_ground_truth(cls) -> Optional[DAG]:
+        if not cls.GROUND_TRUTH_URL:
+            return None
+
+        raw = cls.load_or_fetch("ground_truth", cls.GROUND_TRUTH_URL)
+        return cls._parse_ground_truth(raw)
+
+    @classmethod
+    def _parse_ground_truth(cls, raw: bytes) -> Optional[DAG]:
+        raise NotImplementedError
```
ankurankan: Is this not implemented?

ankurankan (changes requested):
@Theavinash02 I have added a few comments. Could you please write a pytest test file first so that we can test that the implemented methods are working as expected?

Theavinash02:
Hi @ankurankan ,
I will check and address all the comments and get back to you.
Thanks
Avinash

Theavinash02:
Hi Ankur,

I have addressed all comments and updated the code. 
Could you please review the updated code and kindly share your feedback? 
For Sachs, as I mentioned before, the sachs dataset (specifically the 'jittered_experimental' variant). That data file contains 20 columns (11 proteins + 9 interventions), but the ground truth graph file provided by the repository only contains nodes for the 11 proteins. The loader detects this mismatch and adds the 9 intervention variables as isolated nodes to the graph. If we can conclude this is wrong, i can remove the same from the API.

Thanks
Avinash 

@ankurankan

ankurankan:
@Theavinash02 I think this should be ready now. Would be great if you could give it a review.

Theavinash02:
Hi Ankur,
Sure. I will review the code once and get back to you. 
Thanks 🤝🏻

@ankurankan

Theavinash02:
Hi Ankur,
I have reviewed the code, and everything is solid 🗿🤝🏻. Could we please get this merged as discussed in the community meeting ?
Once merged, I will start working on adding the remaining datasets as well !
Thanks
Avinash

@ankurankan

PR #2358, opened 2025-09-19
[MNT] migrate linting to `ruff`

fkiraly:
migrates the linting to `ruff`.

Policies (and much of the workflow) are copied from `sktime`, but of course feel free to change as you like, if it works.


fkiraly:
Tested in https://github.com/pgmpy/pgmpy/pull/2359

PR #2361, opened 2025-09-21
Refactor PC to be sklearn-compatible

### LINKED ISSUE #2313: [ENH] Refactor causal discovery algorithms to be sklearn compatible
As pointed out in #2024, having a sklearn-compatible estimator can have many benefits. Casual Discovery algorithms seem like one of the best candidates to start with this refactor.  The current general structure of causal discovery algorithms in pgmpy is:

```python
est = CasualDiscoveryAlgo(data_or_independences)
est.estimate(algo_parameters) # Returns either DAG, PDAG, MAG, or PAG.
```

The proposed sklearn-style API would be:
```python
est = CasualDiscoveryAlgo(algo_parameters)
est.fit(data/independences)
# The fit method here should set the following attributes:
# - graph_: The learned graph object: DAG, PDAG, MAG, or PAG.
# - adjacency_: A pandas Dataframe representing the adjacency matrix.
# - The standard sklearn attributes: feature_names_in_, n_features_in_, etc.
```

Changes required:
1. Implement a `BaseCasualDiscovery` class inheriting sklearn's `BaseEstimator`.
     - This should define a `fit` method that calls `_fit` implemented in each of the causal discovery algorithm classes.
2. Add depreciation warnings to the current causal discovery algorithms and implement the refactored methods under `pgmpy/causal_discovery` directory. This will resolve #1867 as well.
3. Refactor current causal discovery algorithms to inherit `BaseCausalDiscovery` and follow the proposed API.

@fkiraly @marrov Would love to hear any feedback on this.

Nimish-4:
- Fixes #2313

- Added a base estimator and a few checks to make it `sklearn` - compatible.

### Usage

```python3
from pgmpy.causal_discovery import PC

child = get_example_model("child")
child_samples = child.simulate(n_samples=1000, seed=42)

est_model = PC().fit(child_samples)
print(est_model.graph_.edges)
```


pgmpy/causal_discovery/PC.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def _fit(
+        self,
+        X: pd.DataFrame,
+        y=None,
+        independencies=None,
+    ):
+        # Handle cases like complex data, sparse arrays etc. first
+
+        X_arr = check_array(  # noqa: F841
```
ankurankan: I think we don't need to assign to `X_arr` as it is not getting used anywhere. flake8 should have complained about this.

ankurankan: I think using `validate_data` is the standard for new estimators: https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator. It should also automatically set the `n_features_in_` and `feature_names_in_` attributes.

pgmpy/causal_discovery/PC.py
```diff
@@ -0,0 +1,241 @@
...
+            )
+        if n_samples < 2:
+            raise ValueError(f"n_samples = {n_samples}, at least 2 are required.")
+
+        self.feature_names_in_ = list(X.columns)
+
+        # CI test
+        ci_test = get_callable_ci_test(self.ci_test, data=X)
+
+        expert_knowledge = self.expert_knowledge or ExpertKnowledge()
```
ankurankan: Would suggest writing an explicit if-else here to make it obvious.

pgmpy/tests/test_causal_discovery/test_PC.py
```diff
@@ -0,0 +1,18 @@
...
+    def setUp(self):
+        return super().setUp()
+
+    def test_sklearn_compatibiltity(self):
+        pc_estimator = PC()
+
+        try:
+            check_estimator(pc_estimator)
+        except Exception as e:
+            self.fail(f"PC estimator is not an sklearn-type Estimator: {e}")
```
ankurankan: I think we will also need to rewite the existing tests to work with the new API.

pgmpy/causal_discovery/PC.py
```diff
@@ -0,0 +1,241 @@
...
+
+from pgmpy.base import PDAG, UndirectedGraph
+from pgmpy.causal_discovery.base import BaseCausalEstimator
+from pgmpy.estimators import ExpertKnowledge
+from pgmpy.estimators.CITests import get_callable_ci_test
+
+
+class PC(BaseCausalEstimator):
+    """
+    Constraint-based estimation of DAGs using the PC algorithm.
```
ankurankan: We will need to expand on the documentation of the class.

pgmpy/causal_discovery/base.py
```diff
@@ -0,0 +1,356 @@
...
+
+from pgmpy import config
+from pgmpy.base import UndirectedGraph
+from pgmpy.estimators import ExpertKnowledge
+from pgmpy.estimators.CITests import get_callable_ci_test
+from pgmpy.global_vars import logger
+from pgmpy.independencies import Independencies
+
+
+class BaseCausalEstimator(BaseEstimator):
```
ankurankan: I think a better name would be something like `BaseConstraintCausalDiscovery`.

pgmpy/causal_discovery/_base.py
```diff
@@ -0,0 +1,356 @@
...
+            if show_progress and config.SHOW_PROGRESS:
+                pbar.update(1)
+                pbar.set_description(
+                    f"Working for n conditional variables: {lim_neighbors}"
+                )
+
+        if show_progress and config.SHOW_PROGRESS:
+            pbar.update(max_cond_vars - lim_neighbors)
+            pbar.close()
+        return graph, separating_sets
```
ankurankan, line 347: I think it might be a good idea to expose graph and separating_sets, as: `self.skeleton_=graph`, and `self.separating_sets_ = separating_sets`.

ankurankan (changes requested):
Mostly looks good. I have added a few minor comments. Since we are doing a refactor of all the causal discovery algorithms, it might be a good opportunity to also improve our documentation. So, I would suggest expanding the `PC` docstring by a lot, explaning the algorithm, various usecases with the arguments, etc.

**Update:**
We might need to specify the `categorical` and `string` tags to be True. I am not sure what exactly does the `string` tag checks for. Could you please have a look?

pgmpy/tests/test_causal_discovery/test_PC.py
```diff
@@ -0,0 +1,18 @@
+import unittest
+
+from sklearn.utils.estimator_checks import check_estimator
+
+from pgmpy.causal_discovery import PC
+
+
+class TestPCEstimatorClass(unittest.TestCase):
```
ankurankan: We are also trying to move to pytest, so it might be better to update the tests to use pytest instead of unittest.

pgmpy/tests/test_causal_discovery/test_PC.py
ankurankan: Could you please also replace this test with something like:
```python
def make_estimator():
    return PC()


@parametrize_with_checks([make_estimator()])
def test_pc_compatibility(estimator, check):
    check(estimator)
```
I think it runs the same checks but shows a better logging message, which might be helpful for debugging failures.

Nimish-4:
> **Update:** We might need to specify the `categorical` and `string` tags to be True. I am not sure what exactly does the `string` tag checks for. Could you please have a look?

From [the docs](https://scikit-learn.org/stable/modules/generated/sklearn.utils.InputTags.html#sklearn.utils.InputTags), the string tag seems to be for 'an array-like of strings', which I don't think we need to accept here. I did go ahead with the `categorical` tag as we accept categorical data in the df, but  currently the estimator checks are failing, seemingly due to NaNs in the data.

pgmpy/causal_discovery/PC.py
```diff
@@ -0,0 +1,261 @@
...
+
+    enforce_expert_knowledge : bool, default=False
+        Whether to enforce expert knowledge during search.
+
+    n_jobs : int, default=-1
+        Number of parallel jobs.
+
+    show_progress : bool, default=True
+        Whether to show progress bar.
+
```
ankurankan, line 129: I think we will need an `Attributes` section here that would list all the exposure variables/attributes of the class.

pgmpy/causal_discovery/PC.py
```diff
@@ -0,0 +1,261 @@
...
+        Fit data (`X`) and independence relations (optional) to a causal graph. The method
+        builds an initial skeleton graph (undirected) based on conditional independence tests.
+        Then, the v-structures are oriented based on the separating sets between non-adjacent
+        nodes. Finally, Meek's rules are applied to orient as many remaining edges as possible.
+
+    orient_colliders()
+        Orients the edges that form v-structures in a graph skeleton based on
+        the `separating_sets` to form a PDAG. For each pair of non adjacent
+        nodes `u`, `v` , if a common neighbor `z` is not in the separating set of `u` and `v`;
+        then the v-structure is oriented as `u`->`z` , `v`->`z`.
```
ankurankan: I think we should remove this.

pgmpy/causal_discovery/PC.py
```diff
@@ -0,0 +1,261 @@
...
+    """
```
ankurankan, line 186: We should add an examples section here.

pgmpy/causal_discovery/PC.py
```diff
@@ -0,0 +1,261 @@
...
+            self.graph_ = pdag.to_dag()
+        else:
+            raise ValueError(
+                f"return_type must be one of: dag, pdag, cpdag, skeleton. Got: {self.return_type}"
+            )
+
+        return self
+
+    @staticmethod
+    def orient_colliders(
```
ankurankan: I think we should rename this to `_orient_colliders` to make it a private method.

pgmpy/causal_discovery/base.py
```diff
@@ -0,0 +1,391 @@
...
+
+    def _fit(
+        self,
+        X: pd.DataFrame,
+        y=None,
+        independencies: Independencies = None,
+    ):
+        raise NotImplementedError()
+
+    def build_skeleton(
```
ankurankan: I think this should also be a private method, so `_build_skeleton`.

ankurankan (changes requested):
Looks good to me. I have added a few very minor comments. Only the tests still need to be ported.

fkiraly (changes requested):
Interesting!

Some remarks below.

### Regarding software engineering

* I can see that the design does not, and will not perfectly align with `scikit-learn`, e.g., the `y` in `fit`. It is also unclear whether it is meant to be used to define base APIs that do not align with one in `scikit-learn` directly - do you have a reference for that?
* an alternative is `scikit-base` `BaseObject` or `BaseEstimtaor`, this allows you to freely define signatures and tags, at the cost of having to write your own tests (these can be templated and I could add them if desired)

### Regarding the API design

* I would strongly recommend, for a consequential design like this, it is crucial to do the "three examples" exercise: come up with three substantially different algorithms that need to fit the unified API. That typically highlights overgeneralization or overspecialization
* I think the design is too overspecialized. I mostly agree with `fit`, but I think `build_skeleton` is far too specific - it has 10 arguments and kwargs, and it implements a very specific algorithm. Is this going to be exactly the same for all causal discovery method? I kind of doubt it (but @ankurankan can correct me)
* looking at the design, it feels like there may actually be two or three abstract scitypes in play: conditional independence tests that produce independence relations, and algorithms that take independence relations and produce a graph. Then there could be a third type of abstraction that just produces a graph given data.
    * having said that, I think it is crucial to go through the "three examples" exercise. For instance, off the top of my head, I can already name multiple approaches for conditional independence testing, with sub-cases where variables are fixed and other algorithms that do it for many such combinations of variables simultaneously - so this is definitely a variable part.
    * in particular, the selection of `ci_test` in `build_skeleton` looks like it should be an instance of a specific base class instead, and not a selector argument that is then dealt with by massive case distinction
    * naively, I thought causal discovery eats data and produces a graph - is this not true?

I would also like to ask for an explanation of your mental model. What does `fit` do? What does it do, exactly, in the "three examples" you have thought of (or may be thinking of in the future)?

Nimish-4:
@fkiraly Thanks for the review. 
> naively, I thought causal discovery eats data and produces a graph - is this not true?

Yeah, that's pretty much it for causal discovery - start with a dataset and end up with a causal graph.
> in particular, the selection of ci_test in build_skeleton looks like it should be an instance of a specific base class instead, and not a selector argument that is then dealt with by massive case distinction

Yeah, `ci_test` does make sense as a similar base class.

>I would also like to ask for an explanation of your mental model. What does fit do? What does it do, exactly, in the "three examples" you have thought of (or may be thinking of in the future)?

The fit method is just supposed to take the dataset and and find out the underlying causal graph in a single step - and in the process, specifying attributes like the `skeleton` graph, `n_features` etc.

One issue against more abstraction is that there would probably be only 2 algorithms that would use this base class - PC and the PCMCI estimators. Other causal discovery algorithms won't have much use for this base estimator (since they're not constraint based), so more base estimators would probably be need to be defined for them (as is the case right now, they're just not in the `sklearn` format).

The one thing we were thinking of with this PR was to potentially enable this to be used for an end to end `sklearn` pipeline, hence the `BaseEstimator` from `sklearn`. But yeah, the idea is not fleshed out completely, just a starting point. 
@ankurankan Thoughts?

ankurankan:
@fkiraly Thanks a lot for the review.

> I can see that the design does not, and will not perfectly align with scikit-learn, e.g., the y in fit. It is also unclear whether it is meant to be used to define base APIs that do not align with one in scikit-learn directly - do you have a reference for that?

I thought this still aligns with the sklearn unsupervised learning algorithms: https://scikit-learn.org/stable/unsupervised_learning.html. The pattern I see there is that `y` is always assumed to be None.

> an alternative is scikit-base BaseObject or BaseEstimtaor, this allows you to freely define signatures and tags, at the cost of having to write your own tests (these can be templated and I could add them if desired)

The current implementation does inherit from the `BaseEstimator` class. I am not sure if you mean something else here?

> I think the design is too overspecialized. I mostly agree with fit, but I think build_skeleton is far too specific - it has 10 arguments and kwargs, and it implements a very specific algorithm. Is this going to be exactly the same for all causal discovery method? 

The `build_skeleton` method would be common for all the constraint-based causal discovery algorithms.

> I would also like to ask for an explanation of your mental model. What does fit do? What does it do, exactly, in the "three examples" you have thought of

With every causal discovery algorithm, we take a dataset and learn a causal graph. This causal graph can be an instance of one of the following classes: `DAG`, `PDAG`, `MAG`, or `PAG`. The `fit` method in each of the causal discovery algorithms would assign the `causal_graph_` attribute with the learned graph.

My current mental model of the overall class structure is something like this:
```python
class BaseCausalDiscovery(BaseEstimator):
    # defines the sklearn tags
    def score(self, method='shd', ground_truth=None):
         # Uses self.causal_graph_ to compute a score. Certain methods can work without ground truth.

class BaseConstraintCausalDiscovery(BaseCausalDiscovery):
    def _build_skeleton(...):
    def _orient_colliders(...):

class PC(BaseConstraintCausalDiscovery): # Similarly other algorithms such as FCI, PCMCI, etc.
    def fit(...):

class BaseScoreCausalDiscovery(BaseCausalDiscovery):
    def _compute_score_delta(...):
    def _legal_operations(...):

class HillClimbSearch(BaseScoreCausalDiscovery): # Similarly other algorithms like GES
    def fit(...):

class BaseOptimizationCausalDiscovery(BaseCausalDiscovery):
    def _run_optimization(...):

class NOTEARS(BaseOptimizationCausalDiscovery): # Similarly other algorithms like DAGMA
    def fit(...):

class BaseOrderingCausalDiscovery(BaseCasualDiscovery):
    # Not sure of the common methods here yet.

class LinGAM(BaseOrderingCausalDiscovery):
    def fit(...):
```
The goal with the base classes would be to just separate out methods that are common for each "class" of causal discovery algorithms.

fkiraly:
> I thought this still aligns with the sklearn unsupervised learning algorithms: https://scikit-learn.org/stable/unsupervised_learning.html. The pattern I see there is that `y` is always assumed to be None.

I think you are correct about this - this would still fit the unsupervised tabular API (if we are handling tabular data)

> The `build_skeleton` method would be common for all the constraint-based causal discovery algorithms.

Ok, got it - I think the refined design where it is only in the intermediate base class makes more sense then (or a mixin).

> My current mental model of the overall class structure is something like this:

Neat - I like this a lot! Btw, if there is no common methods for multiple estimators, you can directly estimate from the base base.

fkiraly:
`def score(self, method='shd', ground_truth=None):`

This is nice. Though instead of having `method="shd"` I would do it like sktime:

* implement a default score in the class
* other scores become metrics (or metric classes like in `sktime`)

So you would have a vignette like this:

```python
from pgmpy.metrics import MyMetric
from pgmpy.discoery import MyDiscovery
from pgmpy.datasets import load_my_dataset

X = load_my_dataset()
algo = MyDiscovery(stuff=42)
algo.fit(X)

graph = algo.get_causal_graph()  # or similar

MyMetric().evaluate(graph, ground_truth=some_ground_truth)
```

so, different metrics are different functions, or different classes, instead of being addressed by a selector in a single place.

Of course, `MyDiscovery` can have a `score` which calls a sensible default metric.

ankurankan:
@Nimish-4 I think the failing test is because of an older version of xgboost on Python 3.10. We can ignore it for now; I will push some fix for it.

PR #2373, opened 2025-10-05
Permutation-Test file and DAG method added

D03xen:

- Fixes #

-  Adds PermutationTest file based on https://ojs.aaai.org/index.php/AAAI/article/view/34881/37036
-  Adds method `_get_non_descendants` in DAG.py file for computing non_descendants of a node
-


pgmpy/metrics/PermutationTest.py
```diff
@@ -0,0 +1,160 @@
...
+    --------
+    violations: integer
+        The number of LMC violations found in the model.
+    """
+    violations = 0
+    nodes = list(model.nodes())
+    for node in nodes:
+        parents = list(model.get_parents(node))
+        non_descendants = model._get_non_descendants(node)
+        test_nodes = [(nd for nd in non_descendants if nd not in parents)]
```
ankurankan, line 47: This wouldn't work in the case when there are latent variables in the model. Have a look at `metrics.implied_cis` method for an example of how `minimal_dseparators` can be used to generate the CIs.

pgmpy/metrics/PermutationTest.py
```diff
@@ -0,0 +1,160 @@
...
+        for nd in test_nodes:
+            try:
+                p_value = ci_test_func(node, nd, parents, data, significance_level)
+                if not p_value:
+                    violations += 1
+            except Exception as e:
+                logger.debug(f"CI test failed for {node} ⊥ {nd} | {parents}: {e}")
+                continue
```
ankurankan, line 55: Why is there a try-except here? Do we expect this to fail in some case?

pgmpy/metrics/PermutationTest.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    new_edges = []
+    for edge in model.edges():
+        new_source = perm_mapping[edge[0]]
+        new_target = perm_mapping[edge[1]]
+        new_edges.append((new_source, new_target))
+
+    permuted_model = DAG()
+    permuted_model.add_edges_from(new_edges)
+    return permuted_model
```
ankurankan, line 85: We don't really need to create the permuted graph. The list of implied CIs remains the same in all the permuted graphs, just the variable name changes. 

For example, for a given graph `X -> Y -> Z` we have 1 implied CI: `X _|_ Z | Y`. If we consider a permutation mapping of `{X: Y, Y: X}`, the new graph would be: `Y -> X -> Z`, and our implied CI would be `Y _|_ Z | X`. So, instead of computing the permuted graph and then computing the implied CIs, we can apply the mapping to the implied CIs directly and it would give us the implied CIs of the permuted graph. This will be computationally much faster as well.

pgmpy/metrics/PermutationTest.py
```diff
@@ -0,0 +1,160 @@
...
+from pgmpy.base import DAG
+from pgmpy.estimators.CITests import get_callable_ci_test
+from pgmpy.global_vars import logger
+from pgmpy.metrics import implied_cis
+from pgmpy.models import DiscreteBayesianNetwork
+
+
+def lmc_violations(model, ci_test_func: Callable, data, significance_level=0.05):
+    """
+    Counts LMC (Local Markov Chain) violations for a given DAG. For all ordered
```
ankurankan, line 17: `Local Markov Chain` -> `Local Markov Condition`.

ankurankan (changes requested):

PR #2378, opened 2025-10-09
[ENH] Add Naive IV Regressor 

### LINKED ISSUE #2375: [ENH] Naive Instrumental variables sklearn regressor

**Additional context**

DARHWOLF:

- Fixes  #2375 

- 
-
-


pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+import pandas as pd
+from sklearn.base import BaseEstimator, RegressorMixin, clone
+from sklearn.utils.validation import check_is_fitted, validate_data
+
+
+class NaiveIVRegressor(RegressorMixin, BaseEstimator):
+    """
+    Naive Instrumental Variable (IV) regressor (single exposure, single instrument).
+    Closed-form estimator for scalar X and scalar instrument Z:
+        beta_hat = Cov(Z, Y) / Cov(Z, X)
```
ankurankan: This equation is not correct. This would only work for linear models, and we are trying to generalize the implementation to any model.

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+    Intercept is estimated as: intercept = mean(Y) - beta_hat * mean(X)
+    TO : DO
+
+    Parameters
+    ----------
+    causal_graph : optional
+        If provided, used to get roles 'exposure', 'outcome', and 'instrument' via
+        causal_graph.get_role(role_name). If not provided, `instrument` must be set.
```
ankurankan: This can't be optional.

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+    ):
+        self.causal_graph = causal_graph
+        self.stage1_estimator = stage1_estimator
+        self.stage2_estimator = stage2_estimator
+
+    def __sklearn_tags__(self):
+        tags = super().__sklearn_tags__()
+        return tags
+
+    def _prepare_feature_df(self, X) -> pd.DataFrame:
```
ankurankan: I am expecting this method to be common in all of our causal prediction algorithms. It might be a good idea to create a `BaseCausalPrediction` class that is inherited by all prediction classes and move this method there. sklearn tags could also possibly be moved.

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+        required_cols = (
+            [self.exposure_var_]
+            + list(self.instrument_vars_)
+            + list(self.pretreatment_vars_)
+        )
+        missing = [v for v in required_cols if v not in X_df.columns]
+        if missing:
+            raise ValueError(
+                f"Missing required columns in X for NaiveIVRegressor: {missing}"
+            )
```
ankurankan: This is already getting checked in the `_prepare_feature_df` method

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+        X_df = self._prepare_feature_df(X)
+
+        # ensure needed columns present
+        missing = [
+            v
+            for v in [self.exposure_var_] + self.instrument_vars_ + self.control_vars_
+            if v not in X_df.columns
+        ]
+        if missing:
+            raise ValueError(f"Missing required columns in X for prediction: {missing}")
```
ankurankan: Same

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+
+        z_df = X_df[self.instrument_vars_].reset_index(drop=True)
```
ankurankan: Why do we need to reset index?

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+        if len(self.control_vars_) == 0:
+            pretreatment_vars = pd.DataFrame(
+                {"_intercept": np.ones(X_df.shape[0])}, index=X_df.index
+            ).reset_index(drop=True)
+        else:
+            pretreatment_vars = X_df[self.control_vars_].reset_index(drop=True)
```
ankurankan: What is `control_vars_`? It's not defined anywhere.

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+        stage1_X_df = pd.concat(
+            [z_df.reset_index(drop=True), pretreatment_df.reset_index(drop=True)],
+            axis=1,
+        )
+
+        # fit stage1: X ~ Z + W
+        if sample_weight is None:
+            stage1_estimator.fit(stage1_X_df, x_arr)
+        else:
+            stage1_estimator.fit(stage1_X_df, x_arr, sample_weight=sample_weight)
```
ankurankan: You can pass `sample_weight=None` to the estimators; we don't need to do if-else here.

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+
+        x_hat = stage1_estimator.predict(stage1_X_df)
+
+        stage2_X_df = pd.concat(
+            [
+                pd.Series(x_hat, name=self.exposure_var_).reset_index(drop=True),
+                pretreatment_vars.reset_index(drop=True),
```
ankurankan: Why are we dropping indexes? All these dataframes are created from the same dataframe, the indexes would match automatically.

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,205 @@
...
+        # fit stage2: Y ~ X_hat
+        if sample_weight is None:
+            stage2_estimator.fit(stage2_X_df, y_arr)
+        else:
+            stage2_estimator.fit(stage2_X_df, y_arr, sample_weight=sample_weight)
+
+        # store
+        self.stage1_est_ = stage1_estimator
+        self.stage2_est_ = stage2_estimator
+        self.coef_ = np.asarray(stage2_estimator.coef_).ravel()
```
ankurankan: Not all estimators have `coef_`

ankurankan:
@DARHWOLF Could you also add the pseudo-code to the issue?

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,163 @@
...
+            raise ValueError(
+                f"NaiveIVRegressor requires exactly one exposure; got {len(exposure_vars)}"
+            )
+        if len(outcome_vars) != 1:
+            raise ValueError(
+                f"NaiveIVRegressor requires exactly one outcome; got {len(outcome_vars)}"
+            )
+        if len(instrument_vars) < 1:
+            raise ValueError(
+                f"NaiveIVRegressor requires at least one instrument; got {len(instrument_vars)}"
```
ankurankan: I don't think we need the `got ... ` part here. Less than 1 would be 0.

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+
+        self.exposure_var_ = exposure_vars[0]
+        self.outcome_var_ = outcome_vars[0]
+        self.instrument_vars_ = instrument_vars
+        self.pretreatment_vars_ = list(self.causal_graph.get_role("pretreatment"))
```
ankurankan: Why do we need to do `list(self.causal_graph.get_role("pretreatment"))`? Doesn't `get_role` always return a list?

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,163 @@
...
+        self.feature_columns_ = (
+            [self.exposure_var_] + self.instrument_vars_ + self.pretreatment_vars_
+        )
+
+        # Handles sklearn n_features_in_ check by raising appropriate msg
+        X_arr = np.array(X)
+        n_features = 1 if X_arr.ndim == 1 else X_arr.shape[1]
+        self.n_features_in_ = n_features
```
ankurankan: Isn't this already set by the `validate_data` call in Line 88?

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,163 @@
...
+        # Handle sklearn single-feature check
+        if n_features == 1 and len(self.feature_columns_) > 1:
+            raise ValueError("1 feature(s) (n_features = 1)")
```
ankurankan: We can also pass `ensure_min_features` argument to `validate_data` to automatically handle this scenario: https://github.com/scikit-learn/scikit-learn/blob/c60dae2060/sklearn/utils/validation.py#L849

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,163 @@
...
+
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, RegressorMixin, clone
+from sklearn.utils.validation import check_is_fitted, validate_data
+
+
+class NaiveIVRegressor(RegressorMixin, BaseEstimator):
+    """
+    Naive Instrumental Variable (IV) regressor (single exposure, single instrument).
```
ankurankan: I think the current code works for more than one instrument. Right?

pgmpy/prediction/NaivelVRegressor.py
```diff
@@ -0,0 +1,163 @@
...
+
+        X_df = self._prepare_feature_df(X)
+
+        instrument_df = X_df[self.instrument_vars_]
+        pre_treatment = X_df[self.pretreatment_vars_]
+
+        T_hat = self.stage1_est_.predict(instrument_df)
+        t_hat_2d = pd.DataFrame(T_hat.reshape(-1, 1))
+        covariates_df = pd.concat([t_hat_2d, pre_treatment], axis=1)
+        y_pred = self.stage2_est_.predict(covariates_df)
```
ankurankan: I don't think the prediction algorithm is correct here. We only need to use the stage 2 estimator for making predictions. It would simply be `self.stage2_est_.predict(pd.concat([X, pre_treatment]))`. This is what was giving you issues with the `_prepare_features_df` method as the required columns for the `fit` and `predict` method are different. You could add a `required_columns` argument to `_prepare_feature_df` that is passed when the method is called and hence allows for different required columns in predict and fit. I am not sure if that would lead to some sklearn `check_estimator` checks to fail.

pgmpy/tests/test_prediction/test_NaiveIVRegressor.py
ankurankan, line 1: The tests would need to be a lot more thorough. Similar to `NaiveAdjustmentRegressor` or `DoubleMLRegressor` we should simulate some data and check if we are able to recover the true parameters of the model or not.

ankurankan (changes requested):

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,162 @@
...
+            ensure_2d=True,
+            ensure_min_features=2,
+            dtype="numeric",
+        )
+        stage1_estimator = clone(self.stage1_estimator)
+        stage2_estimator = clone(self.stage2_estimator)
+
+        exposure_vars = list(self.causal_graph.get_role("exposure"))
+        outcome_vars = list(self.causal_graph.get_role("outcome"))
+        instrument_vars = list(self.causal_graph.get_role("instrument"))
```
ankurankan: Don't think we need to convert the returned values to a list. The `get_role` method always returns a list.

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,162 @@
...
+            raise ValueError(
+                f"NaiveIVRegressor requires exactly one outcome; got {len(outcome_vars)}"
+            )
+        if len(instrument_vars) < 1:
+            raise ValueError("NaiveIVRegressor requires at least one instrument.")
+
+        self.exposure_var_ = exposure_vars[0]
+        self.outcome_var_ = outcome_vars[0]
+        self.instrument_vars_ = instrument_vars
+        self.pretreatment_vars_ = self.causal_graph.get_role("pretreatment")
```
ankurankan, line 278: Why not directly assign these attributes in lines 100-102?

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,162 @@
...
+        feature_columns_fit_ = (
```
ankurankan: Doesn't this need to be `self.feature_columns_fit_`? The `_prepare_feature_df` method uses `self.feature_columns_fit_`.

pgmpy/tests/test_prediction/test_NaiveIVRegressor.py
```diff
@@ -0,0 +1,82 @@
...
+    assert exposure_vars and outcome_vars
+    assert instrument_vars_ == ["Z"]
+    assert pretreat_vars == ["P"]
+
+    # Now fit with matching DataFrame and verify feature columns include pretreatment
+    rng = np.random.RandomState(2)
+    P = rng.normal(size=50)
+    Z = rng.normal(size=50)
+    E = 0.4 * Z + rng.normal(scale=0.2, size=50)
+    Y = 1.2 * E + 0.3 * P + rng.normal(scale=0.2, size=50)
```
ankurankan, line 206: You can define a LinearGaussianBN and then use the `simulate` method to simulate data from it. Have a look at the tests in Naive Adjustment Regressor or Double ML Regressor for examples.

ankurankan (changes requested):

pgmpy/tests/test_prediction/test_NaiveIVRegressor.py
DARHWOLF: Sure, I'll use the simulate method here. 
I see that the DML regressor does not use the simulate method for this particular test. I was trying to mirror it in order to maintain uniformity. Should this change be implemented there as well? 

https://github.com/pgmpy/pgmpy/pull/2288/files#diff-4f611de6112a0819c48a918878624240f70f1791c1b96dc3e40be44f0ed655fcR258

pgmpy/prediction/NaiveIVRegressor.py
DARHWOLF: I thought this would be cleaner.  Shall I directly assign these attributes?

ankurankan: I don't really see why this is cleaner. Could you explain a bit? To me, we are assigning some variables in lines 100-102 and then again assigning them to other variables. So a simpler way is just to assign the final variables at the beginning.

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,212 @@
...
+
+        # Step 1.1: Get roles from the causal graph and assign to attributes.
+        exposure_vars = self.causal_graph.get_role("exposure")
+        outcome_vars = self.causal_graph.get_role("outcome")
+        instrument_vars = self.causal_graph.get_role("instrument")
+
+        # Step 1.2: Validate that exactly one exposure, one outcome and atleast one instrument are specified.
+        if len(exposure_vars) != 1:
+            raise ValueError(
+                f"NaiveIVRegressor requires exactly one exposure; got {len(exposure_vars)}"
```
ankurankan: The error message should be along the lines of: the current implementation only works for a single exposure and outcome.

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+        if len(outcome_vars) != 1:
+            raise ValueError(
+                f"NaiveIVRegressor requires exactly one outcome; got {len(outcome_vars)}"
```
ankurankan: Same here.

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,212 @@
...
+    
+    stage2_est_ : estimator
+        Fitted second-stage estimator.
+    
+    coef_ : array-like
+        Coefficients from the fitted `stage2_estimator` (if available).
+    
+    Examples
+    --------
+    TO : DO
```
ankurankan: Examples are missing.

ankurankan (changes requested):
Looks good to me. I have added a couple of very minor comments. Could you please have a look?

pgmpy/prediction/NaiveIVRegressor.py
DARHWOLF: @ankurankan should there be `References` as well?

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,310 @@
...
+    >>>
+    >>> # Fit the model and make predictions
+    >>> _ = model.fit(X_array, y_array)
+    >>> preds = model.predict(X_array)
+    >>> preds.shape[0]
+    50
+
+    References
+    ----------
+    [1] “Instrumental Variables Estimation.” Wikipedia: https://en.wikipedia.org/wiki/Instrumental_variables_estimation
```
ankurankan: Needs to be `.. [1]`. Have a look at the numpydoc style: https://numpydoc.readthedocs.io/en/latest/format.html

pgmpy/prediction/NaiveIVRegressor.py
```diff
@@ -0,0 +1,310 @@
...
+
+        return X_df[required_features]
+
+    def fit(self, X, y, sample_weight: Optional[Any] = None):
+        """
+        This method performs two-stage least squares regression using the specified causal graph.
+        It first fits the stage 1 estimator to predict the exposure variable from the instrument,
+        then fits the stage 2 estimator to predict the outcome variable from the predicted exposure
+        and pretreatment variables.
+        """
```
ankurankan, line 235: The documentation will need to be improved here.

ankurankan (changes requested):
@DARHWOLF I have added a few comments. In this case, I think in the case of IVs even the `fit` method doesn't work for estimators other than linear regression: https://www.sciencedirect.com/science/article/pii/0304407674900335. If you could address the comments, we can wrap up this PR. And later make an overall decision on what to do with the prediction methods.

ankurankan (approved):
Looks good to me.

PR #2382, opened 2025-10-15
Noisy family

### LINKED ISSUE #2372: [ENH]Proposal: Implement Canonical Probabilistic Models (Noisy-OR, Noisy-MAX, Noisy-AND, Simple Models)
Hello,
I would like to contribute to the pgmpy library through the angle of independence of causal influence
(ICI) models, for theoretical pleasure as well as a pragmatical necessity.
Indeed, I was working on the reasoning_core project, for generating probabilistic/causal problems in order to train LLMs, using another library, pyagrum to name it. It worked pretty well, but as the size of the problem scaled (size of the DAG or the possible values that each node can take), the exponential growth of the CPT in the basic setting is not manageable. 
Naturally, using the Noisy-OR assumption was a way out of this issues, and is moreover a reasonable and interesting framework to train LLMs at. Even thought this theoretical framework is rather old (more than 30 years), not many "bayesian network" library have them implemented, and pgmpy is one of the few to have an implementation of it (even though it's quite fresh). In that optic, I would like to be part of the further development of the library in the direction of ICI models, that would integrate in the smoothest way within the current pgmpy framework.

Yes. Currently, building Bayesian Networks in `pgmpy` requires users to specify the full Conditional Probability Table (CPT) for each node using `TabularCPD`. The size of this table grows exponentially with the number of parent nodes (e.g., `2^n` parameters for a binary node with `n` binary parents).

This presents a major practical problem for knowledge engineering:
1.  **It is extremely difficult** for domain experts to accurately provide this many parameters.
2.  **It is tedious and highly error-prone** to manually calculate and input the CPT values.
3.  **It is a significant bottleneck** that discourages the creation of larger, more realistic models where nodes often have more than 2-3 parents.

This issue makes it hard for users to apply well-established techniques that simplify model construction.

I propose the implementation of several **Canonical Probabilistic Models** as specialized CPD classes. These classes would provide a higher-level, more intuitive API for model construction by taking a small, linear number of parameters and automatically generating the full `TabularCPD` internally.

This approach aligns with `pgmpy`'s existing class-based design (`TabularCPD`, `BayesianNetwork`, etc.).

**Implementation Plan & API Sketch:**

1.  **Introduce new CPD classes** that store canonical parameters directly, making the model's structure more transparent and easier to maintain.
2.  The `BayesianNetwork.add_cpds` method would be updated to recognize these new classes and internally convert them to `TabularCPD` objects for use by the inference engines.

Here is a sketch of the proposed classes:

*   **`NoisyORCPD` (Highest Priority):** For modeling independent causes.
    ```python
    # API sketch for a leaky Noisy-OR
    cpd_y = NoisyORCPD(variable='Y',
                       variable_card=2,
                       evidence=['X1', 'X2', 'X3'],
                       evidence_card=[2, 2, 2],
                       causal_strengths=[0.8, 0.6, 0.3], # The c_i parameters
                       leak_probability=0.01) # The c_L parameter
    ```

*   **`NoisyMAXCPD`:** A generalization for ordinal variables (e.g., disease severity).
    ```python
    # API sketch for a causal Leaky-MAX
    # 'causal_dists' maps a parent state to a distribution over the child's states
    cpd_severity = NoisyMAXCPD(variable='Severity',
                               variable_card=4, # e.g., none, mild, moderate, severe
                               evidence=['DiseaseA', 'DiseaseB'],
                               evidence_card=[2, 2],
                               causal_dists={
                                   'DiseaseA': [0.1, 0.6, 0.3, 0.0], # P(Z_A | +disease_a)
                                   'DiseaseB': [0.2, 0.7, 0.1, 0.0]  # P(Z_B | +disease_b)
                               },
                               leak_dist=[0.95, 0.03, 0.02, 0.0]) # P(Z_L)
    ```

*   **`NoisyANDCPD`:** For modeling necessary conditions.
    ```python
    # API sketch for a leaky Noisy-AND
    cpd_p = NoisyANDCPD(variable='P',
                       variable_card=2,
                       evidence=['S', 'I', 'C'],
                       evidence_card=[2, 2, 2],
                       inhibitor_probs=[0.9, 0.95, 0.1], # The c_i params
                       substitute_probs=[0.0, 0.0, 0.9], # The s_i params
                       leak_probability=0.05) # The q_L (global inhibitor) param
    ```

Each of these classes would contain a `to_tabular_cpd()` method to perform the conversion for the backend.

1.  **Status Quo:** The user continues to manually calculate and create `TabularCPD`s. This is the existing problem we are trying to solve and is not a scalable solution.

2.  **Helper Functions:** One could create standalone helper functions (e.g., `create_noisy_or_cpd(...)`) that simply return a `TabularCPD` object.
    *   **Drawback:** This is a less desirable solution because the original, intuitive canonical parameters (`c_i`, `c_L`, etc.) are lost once the `TabularCPD` is created. Using classes preserves these parameters, making the model object itself inspectable, more readable, and far easier to modify or maintain. The class-based approach provides a much richer and more object-oriented representation of the model's semantics.

**Additional context**

The utility and structure of these models are extensively documented in the literature. A primary reference for this proposal is the paper: **"Canonical Probabilistic Models for Knowledge Engineering" by F. J. Díez and M. J. Druzdzel (May 2007)**.

Implementing these models would significantly lower the barrier to entry for building complex, real-world Bayesian Networks in `pgmpy`, directly addressing one of the most common and difficult challenges in the knowledge engineering process. It would be a major enhancement to the library's usability and power.

vlcmb:

- Fixes #2372 

- Add of pgmpy/factors/discrete/CI.py that encode the canonical noisy class
- Tiny modification of pgmpy/factors/discrete/__init__.py to add those new class within discrete factors


ankurankan (changes requested):
@vlcmb Thanks a lot for the PR. Design wise, I have a couple of suggestions:
1. I would suggest getting rid of `ABC`. I think using ABC has more drawbacks than benefits, and we have tried to avoid them in pgmpy. For example, using ABC couples the abstract class with any new interface/subclass that we are trying to implement, making it a bit too strict. Plus, using an ABC can make it a bit difficult to evolve as making any changes in the ABC class requires changing all the subclasses. I think `typing.Protocol` is usually a better alternative, but I don't think we really need either of them here. 
2. I would suggest inheriting from `TabularCPD` and make a `super` call in the `__init__` methods, similar to the current NoisyOr implementation: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/factors/discrete/NoisyOR.py#L68. Essentially, this would mean moving the `to_cpd` methods to the `__init__` method itself. This would allow us to use these new classes with the rest of the package without making any other changes.

ankurankan:
The linting tests seem to be failing. I would suggest installing our `pre-commit` hook: https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md#pre-commit-hooks. This will automatically take care of all code formatting changes with each commit.

vlcmb:
Hello @ankurankan ,

I made the required changes as you suggested, and it works well, as the corresponding `test_CI.py` file was added in order to test the new module.

However, two integration checks are not pass yet, as before the Linting one, and another one that concern some statistical tests (that have been recently implemented).

Concerning the Linting one, the error that couldn't be corrected seem to be the following:
```
flake8...................................................................Failed
- hook id: flake8
- exit code: 1

pgmpy/factors/discrete/__init__.py:4:1: F401 '.JointProbabilityDistribution.JointProbabilityDistribution' imported but unused
pgmpy/factors/discrete/__init__.py:5:1: F401 '.NoisyOR.NoisyORCPD' imported but unused
```
However, it seems that the deadlock come from other modules, and I am not very aware of the procedure to deal with it.
Similarly, the second check seems to fail due another module added recently `test_CITests`. At first, I thought it was coming from the 'CI.py`module that I created, but in a funny manner it was referring to something else.

I think we are close to meet the requirements, however I would need a bit of your guidance to conclude the pull request, as I am still in the learning process.

Thanks in advance for you help !

ankurankan:
@vlcmb Sorry for the delay, I was away last couple of weeks. I have fixed the linting issues in this PR and will try to add comments as soon as possible.

pgmpy/factors/discrete/CI.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+
+    def __init__(
+        self,
+        variable,
+        evidence,
+        activation_magnitude,
+        mode="OR",
+        leak=None,
+        isboolean_style=False,
```
ankurankan: Is there a use case in your mind where using True/False as state names wouldn't be sufficient? If not, I think it would be better to simplify the implementation and remove `isboolean_style` argument with the default being True/False.

vlcmb:
Welcome back @ankurankan ! 
Thanks for the fix, and I am waiting for your comments in order to finally merge.
Best,

Issue #2371, opened 2025-10-01
[ENH] Add `PAG.to_dagitty` and `PAG.from_dagitty` methods


ankurankan:
Ref #2352

PR #2387, opened 2025-10-18
[ENH] Implemented `to_dagitty` and `from_dagitty` methods for MAG

### LINKED ISSUE #2370: [ENH] Add `MAG.to_dagitty` and `MAG.from_dagitty` methods
Ref #2352

JATAYU000:

- Fixes #2370

- Methods `to_dagitty` and `from_dagitty` for `MAG` having similar method structure as implemented in `DAG`
- Added Examples for each method
- Update the parser implementation.


ankurankan (changes requested):
Would it be possible to move both of these methods to the `AncestralBase` class? As both `MAG` and `PAG` inherit `AncestralBase`, if we can generalize the implementation, we should be able to reduce duplicated code. That will also fix both #2370 and #2371.

The `from_dagitty` method will need to be generalized. It makes certain assumptions at the moment, for example, each edge is on a single line. Instead of writing the method from scratch, I would suggest modifying our dagitty parser (`pgmpy.utils.parse_dagitty`) instead to handle `MAG` and `PAG` as well.

JATAYU000:
Understood, Will work on it.

JATAYU000:
@ankurankan is `PAG` already implemented in base? Can you point me where it is implemented. I can only see `MAG` which is being inherited by `AncestralBase`.

ankurankan:
@JATAYU000 Sorry, I forgot to mention earlier that it is still in the works: https://github.com/pgmpy/pgmpy/pull/2346/files#diff-f58eaac198e83d52f2277f52b857f97decc963b9bad79300db7bd3f5cccf9aa2R9. The main difference between `MAG` and `PAG` would be that `PAG` would allow for a new edge type where one of the ends of the edge could be unknown, represented using edge type `o`.

JATAYU000:
Makes sense, Thank you

JATAYU000:
@ankurankan I have migrated the methods to `AncestralBase`
Now for the `parse_dagitty` utility I am planning on adding a new argument `target_type` which can be 'DAG', 'MAG' etc.

ankurankan:
@JATAYU000 I have fixed the failing test in the `dev` branch. If you update your branch, it should get fixed.

JATAYU000:
@ankurankan I will start Modifying the `parse_dagitty` utility once #2395 is merged.

JATAYU000:
@ankurankan Current status of this PR,
- Added `target_type` argument to `parse_dagitty` function
- Instead of returning `latent` it now returns a dict `roles`
- Updated the `DAG.from_dagitty` method to use the new `roles`
- Similarly in `Ancestralbase` class as well.

in `DAG` when there are `coefs` it uses `LinearGaussianBayesianNetwork` and it does not take roles, it has latent alone, I assume it only needs latent. I want your input regarding that part

pgmpy/base/AncestralBase.py
```diff
@@ -528,6 +529,124 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
                 queue.extend(self.get_neighbors(current, u_type=u_type, v_type=v_type))
         return reachable
 
+    def to_dagitty(self) -> str:
+        """
+        Convert the MAG to a Dagitty string representation.
+        Returns
```
ankurankan, line 551: There needs to be an empty line before this; otherwise, `sphinx` won't be able to parse the section.

pgmpy/base/AncestralBase.py
```diff
@@ -528,6 +529,124 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
...
+        L [latents]
+        Y [outcome]
+        X [exposure]
+        }
+
+        Notes
+        -----
+        - Node names are converted to string representations.
+        - If node names contain spaces or special characters, they will be used as-is.
+        - Users should ensure node names are valid in R/dagitty context if needed.
```
ankurankan: I would suggest keeping the first statement in the main method description in line 534. I think the rest of the points are a bit obvious, and we can remove the `Notes` section.

pgmpy/base/AncestralBase.py
```diff
@@ -528,6 +529,124 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
...
+
+        References
+        ----------
+        dagitty syntax: https://cran.r-project.org/web/packages/dagitty/dagitty.pdf
+        """
+        dagitty_str = "dag {\n"
```
ankurankan: I think the `dag` in the string will need to be dependent on the class from which the `to_dagitty` method is being called. This can be `mag` or `pag`.

pgmpy/base/AncestralBase.py
```diff
@@ -528,6 +529,124 @@ def get_reachable_nodes(self, node, u_type=None, v_type=None):
...
+        for u, v in self.edges:
+            marks = self.edges[u, v]["marks"]
+            if marks[u] == "-" and marks[v] == ">":
+                dagitty_str += f"{u} -> {v}\n"
+            elif marks[u] == ">" and marks[v] == "-":
+                dagitty_str += f"{v} -> {u}\n"
+            elif marks[u] == ">" and marks[v] == ">":
+                dagitty_str += f"{u} <-> {v}\n"
+            elif marks[u] == "-" and marks[v] == "-":
+                dagitty_str += f"{u} -- {v}\n"
```
ankurankan: I think this currently doesn't consider the edge marker `o`. dagitty represents the `o` marker as a `@`.

pgmpy/utils/parser.py
```diff
@@ -67,30 +67,33 @@ def parse_lavaan(lines):
     return ebunch, latents, err_corr, err_var
 
 
-def parse_dagitty(lines):
+def parse_dagitty(lines, target_type="DAG"):
```
ankurankan: I don't think we would need a `target_type` here. The dagitty syntax already defines the type of model. The dagitty string starts with either `dag`, `mag`, or `pag`. We can use that to determine the class that we want to return.

ankurankan (changes requested):

pgmpy/base/AncestralBase.py
JATAYU000: Okay so daggity represents it as '@' and we should map it to 'o'

JATAYU000:
@ankurankan , I have started resolving the changes requested, Ideally we want the parser to be able to parse all the examples in pages 10 and 11 [dagitty pdf](https://cran.r-project.org/web/packages/dagitty/dagitty.pdf) right?
and apart from the requested changes these points below are accepted?

>  Instead of returning latent it now returns a dict roles
>  Updated the DAG.from_dagitty method to use the new roles
>  Similarly in Ancestralbase class as well.

ankurankan:
@JATAYU000 Sorry for the delay in the reply. I think we will need to write a few of our own examples as the examples shown in the pdf do not cover all the scenarios. 

Also, the tests seem to be failing. Could you please have a look?

JATAYU000:
> I think we will need to write a few of our own examples as the examples shown in the pdf do not cover all the scenarios.

@ankurankan right now the parser works for all the examples except for group operation, It would be helpful if you could point me to some documentation on use of the other scenarios which are not mentioned in the pdf.
Also figured out why the test suites failed, working on it.

ankurankan:
@JATAYU000 

>  It would be helpful if you could point me to some documentation on use of the other scenarios which are not mentioned in the pdf.

I don't know of any resources that have a database of MAGs. What I would suggest is that we write test cases that at least checks that all the different edge types are parsed correctly. One edge case to consider is when a node is disconnected from the graph. You can have a look at the DAG parser tests for some examples: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/tests/test_base/test_DAG.py#L789 

Let me know if you need any help in creating example models to test using, and I can try to write some.

JATAYU000:
Update on the PR:
- Parser is updated for most of the edge cases which I am aware of including the group operations and disconnected graphs
- Added Tests for MAG with some examples which cover most of the edge cases
- I Did not notice before that  `with_roles(`) only supports `exposure` and `outcome` and not `latents` even though there exists a `test test_latents_with_role` which checks if `with_roles()` adds `latent` role when it does not. This issue exists only in DAG which fails to add `latents` through `roles` dict, unlike MAG which passes the `latents` as params to `super()` . I have decided to fix it by appending the latents provided through the roles as well which is what the `test_latents_with_role` expects.

pgmpy/base/DAG.py
```diff
@@ -182,6 +182,8 @@ def __init__(
             roles = {}
         elif not isinstance(roles, dict):
             raise TypeError("Roles must be provided as a dictionary.")
+        else:
+            self.latents = self.latents.union(roles.get("latents", set()))
```
ankurankan: Why is this required? Isn't the latents role being set in Line 177?

pgmpy/base/DAG.py
```diff
@@ -287,9 +289,13 @@ def from_dagitty(cls, string=None, filename=None) -> "DAG":
         else:
             raise ValueError("Either `filename` or `string` need to be specified")
 
-        ebunch, latents, coefs, nodes = parse_dagitty(dagitty_str)
+        ebunch, roles, coefs, nodes = parse_dagitty(dagitty_str)
+        latents = roles["latents"]
```
ankurankan: I don't think we will need to extract `latents`. If it is present in the `roles`, it should get set automatically.

ankurankan (changes requested):

pgmpy/base/DAG.py
JATAYU000: No, I was wrong about how `with_roles()` handled latents at that time , It is clear now.

pgmpy/base/DAG.py
JATAYU000: @ankurankan That was for `LinearGaussianBayesianNetwork` ,sure I can move that inside its block.

JATAYU000:
@ankurankan Sorry for the delay but I have resolved the issues, I had some changes which did not set latents variable, I have removed it since it is not required. Is there anything else to be done in this pr? PAG tests with the parser will be added in [This PR](https://github.com/pgmpy/pgmpy/pull/2346) right?

pgmpy/base/DAG.py
```diff
@@ -287,16 +287,21 @@ def from_dagitty(cls, string=None, filename=None) -> "DAG":
...
             raise ValueError("Either `filename` or `string` need to be specified")
 
-        ebunch, latents, coefs, nodes = parse_dagitty(dagitty_str)
+        ebunch, roles, coefs, nodes = parse_dagitty(dagitty_str)
         if len(coefs) == 0:
-            dag = cls(ebunch=ebunch, latents=latents)
+            if cls.__name__ == "DAG":
+                dag = cls(ebunch=ebunch, roles=roles)
+            else:
+                dag = cls(ebunch=ebunch, latents=roles["latents"])
```
ankurankan: Why do we need the if-else here? Shouldn't we always pass all the roles?

pgmpy/base/DAG.py
```diff
@@ -287,16 +287,21 @@ def from_dagitty(cls, string=None, filename=None) -> "DAG":
...
             dag.add_nodes_from(nodes)
             return dag
         else:
             from pgmpy.factors.continuous import LinearGaussianCPD
             from pgmpy.models import LinearGaussianBayesianNetwork
 
-            lgbn = LinearGaussianBayesianNetwork(ebunch=ebunch, latents=latents)
+            lgbn = LinearGaussianBayesianNetwork(
+                ebunch=ebunch, latents=roles["latents"]
+            )
```
ankurankan: Shouldn't we pass the roles here as well?

pgmpy/tests/test_base/test_MAG.py
```diff
@@ -161,3 +163,43 @@ def test_graph_properties(self, mag):
...
+            f.write(model_str)
+        model_from_file = MAG.from_dagitty(filename="test_model.dagitty")
+        os.remove("test_model.dagitty")
+
+        expected_edges = {("A", "B,"), ("A", "E"), ("A", "J"), ("A", "M")}
+        expected_roles = {"outcome": ["J"], "latents": ["E"], "exposure": ["A"]}
+
+        # Handles order of nodes
+        def normalize_edge(edge):
+            return tuple(sorted(edge))
```
ankurankan: I think this is incorrect. For a directed edge, it can potentially flip the direction.

pgmpy/tests/test_base/test_MAG.py
```diff
@@ -161,3 +163,43 @@ def test_graph_properties(self, mag):
...
+
+    def test_from_dagitty(self):
+        model_str = "mag { E [latent] A [e] J [o] {B, E} -> A; A -- J ; A -- M}"
+        model_from_str = MAG.from_dagitty(model_str)
+        with open("test_model.dagitty", "w") as f:
+            f.write(model_str)
+        model_from_file = MAG.from_dagitty(filename="test_model.dagitty")
+        os.remove("test_model.dagitty")
+
+        expected_edges = {("A", "B,"), ("A", "E"), ("A", "J"), ("A", "M")}
```
ankurankan: I think there is a mistake in the expected edges here. In the first edge `("A", "B,")`, the comma is inside the quotes.

ankurankan (changes requested):
@MAVRICK-1 Sorry for the delay in the review, it somehow fell off my list of PRs to review.

pgmpy/base/DAG.py
JATAYU000: This was left behind before MAG implementation , Forgot to remove it thank you for pointing it out.

JATAYU000:
> I think there is a mistake in the expected edges here. In the first edge ("A", "B,"), the comma is inside the quotes.

That was a mistake in the group item parsing which was responsible for the normalize edge in the tests, I have fixed it and updated the tests

> Shouldn't we pass the roles here as well?
```py
class LinearGaussianBayesianNetwork(DAG):
    def __init__(
        self,
        ebunch: Optional[List[Tuple[Hashable, Hashable]]] = None,
        latents: Set[Hashable] = set(),
        lavaan_str: Optional[str] = None,
        dagitty_str: Optional[str] = None,
    ) -> None:
        super(LinearGaussianBayesianNetwork, self).__init__(
            ebunch=ebunch,
            latents=latents,
        )
```
@ankurankan  Why does it not call super with `roles` and just the `latents` then? Is it because it does not benefit from other roles? Should I update here to pass `roles` as well?

Also idk why these doctest errors are there now , The doctest are failing in dev as well. Are you aware of this Issue?

pgmpy/base/AncestralBase.py
ankurankan: @JATAYU000 Since this is the `AncestralBase` class, I think this should handle the `o` edge marker too. Yes, dagitty represents `o` as `@`

pgmpy/base/DAG.py
ankurankan: I have changed the API a bit, and now `LinearGaussianBayesianNetwork` accepts the same arguments that `DAG` does. Ref #2455

ankurankan (changes requested):
@JATAYU000 I have added a couple of minor comments; looks good otherwise.

ankurankan:
@JATAYU000 I think tests for `to_dagitty` method are missing because of which codecov is failing. Could you please add tests for it?

JATAYU000:
@ankurankan I have added tests for the coverage, should it be 100% because there few lines which are handling exceptions, are they supposed to be covered in the tests as well?

ankurankan (approved):
@JATAYU000  No, that's fine. Looks good to me now, thanks for the PR.

PR #2400, opened 2025-10-23
Add exposures and outcomes properties to graph models

MAVRICK-1:
Changes:
- Add exposures and outcomes properties to _GraphRolesMixin with getter/setter methods
- Extend AncestralBase.__init__ to accept exposures and outcomes parameters
- Update MAG.__init__ to support the new parameters
- Add comprehensive tests for the new functionality
- Maintain backward compatibility with existing code

The implementation allows users to easily assign and manage exposure and outcome roles in causal graphs using an intuitive property-based approach, making it easier to define frequently used node roles in causal inference workflows.

Fixes #2379

PR #2402, opened 2025-10-24
[Refactor] Implement a CoreGraph class according to the design discussion. (2)

### LINKED ISSUE #2383: [Refactor] Implement a `CoreGraph` class according to the design discussion.
Sub-issue 1 of #2376

I will develop based on the meeting contents from 2025-10-13 and the discussion in #2376.

### Tasks:
1. Create a classified list of methods in `AncestralBase` and sort them into `CoreGraph` and `_GraphAlgorithmMixin` based on their purpose.
2. Create the paths `pgmpy/core` and `pgmpy/tests/test_core` and proceed with development according to the classification in Task 1.
3. Update the comments and receive a draft code review from the mentors.
4. Develop the methods according to the contents of the 2025-10-13 meeting and the discussion in #2376.
- 4.1. Modify `CoreGraph` to inherit from `nx.MultiGraph`. (Currently `nx.Graph`)
- 4.2. Develop the methods of `CoreGraph` according to the documentation.
- 4.3. Update the comments and receive a code review from the mentors.

---

### 1. The legacy inheritance structure of `pgmpy`'s base graph classes.

Image

The base graph classes in the existing `pgmpy` library individually inherited from the `networkx` library, leading to the following architectural challenges:

#### 1-1. Inconsistent User APIs

Common graph APIs, such as `add_edge()`, had to be implemented independently within each base graph class.
This resulted in an inconsistent API surface and reduced development productivity due to the redundant implementation of overlapping features.

```py
from pgmpy.base import DAG, ADMG, AncestralBase

G1 = DAG()
G1.add_edge("A", "B")

G2 = ADMG()
G2.add_directed_edges([("A", "B")])

G3 = AncestralBase()
G3.add_edge("A", "B", "-", ">")
```

#### 1-2. Dependency Management Risks with `networkx`.
Direct and scattered inheritance creates a rigid coupling with the `networkx` library.
Any breaking changes or major updates in `networkx` would require manual, individual updates across all base graph classes, increasing maintenance overhead and technical debt.

### 1. The Improved Inheritance Structure of `pgmpy` via `_CoreGraph`.

Image

`_CoreGraph` serves as a common base class for all graph types.
This architecture allows shared graph functionalities to be developed and managed centrally within `_CoreGraph` and `_GraphAlgorithmMixin`.
This approach provides users with a "unified API surface" and eliminates the need for redundant development of overlapping features. Furthermore, it enables "centralized management" of dependencies on the `networkx` library.

### 3. An abstract layer separating internal edge storage from user API representation.

Image

A limitation was identified in the `networkx` graph inherited by `_CoreGraph`, where a single graph instance could not simultaneously represent `directed`, `undirected`, and `bidirectional` edges.
To resolve this, I decoupled the internal data storage format from the external user-facing representation.

```py
# Graph init
edges = [("A", "B", "->"), ("A", "B", "<>"), ("B", "C", "->")]
G = _CoreGraph(ebunch=edges)

# Users view method(`get_edges()`) 
G.get_edges(keys=True, data=True)
[('A', 'B', 0, '->'), ('A', 'B', 1, '<>'), ('B', 'C', 0, '->')]

# pgmpy's data store form
G.edges(keys=True, data=True)
MultiEdgeDataView([('A', 'B', 0, {'A': '-', 'B': '>'}), ('A', 'B', 1, {'A': '>', 'B': '>'}), ('B', 'C', 0, {'B': '-', 'C': '>'})])
```

---
The method list for CoreGraph is updated and managed in the documentation at here.
## 1. CoreGraph's_method_list(Updated)
| method | arguments | description | Reference |
| --- | --- | --- | --- |
| `__init__` | `ebunch: Iterable[str]`, <br>`exposures: set[Hashable] = set()`, <br>`outcomes: set[Hashable] = set()`, <br>`latents: set[Hashable] = set()`, <br>`roles=None` | This should be applicable to all graph models in `pgmpy`. <br> We need to define the allowed edges for each model. | [`DAG.__init__` in pgmpy](https://github.com/pgmpy/pgmpy/blob/7b52e562b4e1e79eb6a2769a939ca060bca0e9f0/pgmpy/base/DAG.py#L151) <br> [#2330 in pgmpy (comment)](https://github.com/pgmpy/pgmpy/issues/2330#issuecomment-3279325571)|
| `add_edge` | `u: Hashable`, <br>`v: Hashable`, <br>`type: str`, <br>`key=None`, <br>`**kwargs` | The `add_edge` method in `pgmpy` should have a `type` argument. | [#2330 in pgmpy (comment)](https://github.com/pgmpy/pgmpy/issues/2330#issuecomment-3268981766) |
| `add_edges_from` | `ebunch: Iterable[Union[tuple[Hashable, Hashable, Hashable],tuple[Hashable, Hashable, Hashable, Hashable]]]` | Refer to the description for `add_edge` | Refer to the `add_edge`'s Ref. |
| `remove_edge` | `u: Hashable`, <br>`v: Hashable`, <br>`type: str`, <br>`key=None`, <br>`**kwargs` | `networkx.MultiGraph.remove_edge` deletes an edge based on the `key` argument. This should be modified to delete based on the `type` argument. | [`MultiGraph.remove_edge` in networkx](https://github.com/networkx/networkx/blob/36d7ba36a46e88dec16fca0a2640d479d2c4f833/networkx/classes/multigraph.py#L624) |
| `remove_edges_from` | `ebunch: Iterable[Union[tuple[Hashable, Hashable, Hashable],tuple[Hashable, Hashable, Hashable, Hashable]]]` | Refer to the description for `remove_edge` | Refer to the `remove_edge`'s Ref. |
| `get_neighbors` | `self`, `node`, `u_type`, `v_type` | Get neighbors of a node with optional edge mark constraints. |
| `get_parents` | `self`, `node` | Get nodes that point to this node with '>'. |
| `get_children` | `self`, `node` | Get nodes that this node points to with '>'. |
| `get_spouses` | `self`, `node` | Get nodes connected by bidirectional '>' edges (spouses). |
| `get_ancestors` | `self`, `node` | Get all ancestor nodes of the given node. |
| `get_descendants` | `self`, `node` | Get all descendant nodes (children, grandchildren, etc.). |
| `get_reachable_nodes` | `self`, `node`, `u_type`, `v_type` | Get all nodes reachable from a given node following specified edge marks. |
| `copy` | do not have arguments | This should be applicable to all graph models in `pgmpy`. | [`MultiGraph.copy` in networkx](https://github.com/networkx/networkx/blob/36d7ba36a46e88dec16fca0a2640d479d2c4f833/networkx/classes/multigraph.py#L1023), <br>[`DAG.copy` in pgmpy](https://github.com/pgmpy/pgmpy/blob/7b52e562b4e1e79eb6a2769a939ca060bca0e9f0/pgmpy/base/DAG.py#L1596) |
| `get_edges` | `type: bool`, `key: bool` | Returns a list of edges in the graph. | --- |
| `__eq__` | `other` | - | [`DAG.__eq__` in pgmpy](https://github.com/pgmpy/pgmpy/blob/7b52e562b4e1e79eb6a2769a939ca060bca0e9f0/pgmpy/base/DAG.py#L1606) |
| `_validating_and_formatting_edges_value` | `ebunch: Iterable[Union[tuple[Hashable, Hashable, Hashable], tuple[Hashable, Hashable, Hashable, Hashable]]]` | Helper method for `add_edges_from()`, `remove_edges_from()` | --- |
| `_get_edges_type_key` | do not have arguments | Returns the edge's `type` and `key` value connecting the two nodes as a list of tuples. | --- |
| `_get_key` | `u: Hashable`, `v: Hashable`, `type: str`, `key: Optional[Hashable]` | Return the key value using u, v, and type values. | --- |
| ~~`_validating_edges_value`~~ | `ebunch: Iterable[Union[tuple[Hashable, Hashable, Hashable], tuple[Hashable, Hashable, Hashable, Hashable]]]` | Unify with `_validating_and_formatting_edges_value()`.<br>~~Helper method for `add_edge()`, `remove_edge()`~~ | --- |
| ~~`is_multigraph`~~ | do not have arguments | We decided not to develop this method in Pull Request #2402. <br> ~~This should be modified to be applicable to all graph models in `pgmpy`.<br>`@abstractmethod` decorator is required.~~ | [`MultiGraph.is_multigraph` in networkx](https://github.com/networkx/networkx/blob/36d7ba36a46e88dec16fca0a2640d479d2c4f833/networkx/classes/multigraph.py#L1015) |
| ~~`is_directed`~~ | do not have arguments | We decided not to develop this method in Pull Request #2402. <br> ~~This should be modified to be applicable to all graph models in `pgmpy`. <br>`@abstractmethod` decorator is required.~~| [`MultiGraph.is_directed` in networkx](https://github.com/networkx/networkx/blob/36d7ba36a46e88dec16fca0a2640d479d2c4f833/networkx/classes/multigraph.py#L1019) |

daehyun99:

- Fixes #2383

- Implement a `TestCoreGraph` class.
- Implement a `CoreGraph` class. (`nx.MultiDiGraph`)


pgmpy/base/base.py
```diff
@@ -0,0 +1,248 @@
+import inspect
+from abc import ABC, abstractmethod
+from typing import Hashable, Iterable, Optional
+
+import networkx as nx
+
+from pgmpy.base._mixin_roles import _GraphRolesMixin
+
+
+class _CoreGraphABC(ABC):
```
ankurankan: Wouldn't suggest using `ABC`. It is too strict and creates dependency between the ABC class and the class inheriting it.

pgmpy/base/base.py
```diff
@@ -0,0 +1,248 @@
...
+
+    >>> [Example_Code]
+
+    """
+
+    # The Constants should be removed once Issue #2383 is merged.
+    # And this Constants must be implemented in each graph model that inherits from `CoreGraph`.
+    ALLOWED_EDGES = {"temp"}
+    IS_DIRECTED = False
+    IS_MULTIGRAPH = True
```
ankurankan: I think a simpler way would be to define a single attribute here, say: `SUPPORTED_EDGE_TYPES`. Each inheriting class should define this attribute, and because we don't expect users to use `CoreGraph` directly, we do not need to define attributes like `IS_DIRECTED` or `IS_MULTIGRAPH` because it would be obvious from the object type (i.e., for each of the inheriting classes like `DAG`, `PDAG` we know what kind of graph it would be)

pgmpy/base/base.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def __init__(
+        self,
+        ebunch: Optional[Iterable[tuple[Hashable, Hashable]]] = None,
+        exposures: set[Hashable] = set(),
+        outcomes: set[Hashable] = set(),
```
ankurankan: I thought we had decided to not move these arguments out of roles in this PR?

pgmpy/base/base.py
```diff
@@ -0,0 +1,248 @@
...
+        """
+        return "CoreGraph"
+
+    def is_directed(self):
+        """Returns True if graph is directed, False otherwise."""
+        return self.IS_DIRECTED
+
+    def is_multigraph(self):
+        """Returns True if graph is a multigraph, False otherwise."""
+        return self.IS_MULTIGRAPH
```
ankurankan: I don't think we will need any of these methods.

pgmpy/base/base.py
ankurankan: I don't think we need this class at all.

pgmpy/base/base.py
```diff
@@ -0,0 +1,248 @@
...
+    @abstractmethod
+    def _graph_type_marker(self):
+        """
+        This is a pure virtual method to prevent this class (and CoreGraph) from being instantiated directly.
+        Concrete classes like DAG, PDAG, etc., must implement it.
+        """
+        pass
+
+
+class CoreGraph(_CoreGraphABC, nx.MultiGraph, _GraphRolesMixin):
```
ankurankan: Would suggest renaming this to `_CoreGraph` to signify that it is an internal/private class.

pgmpy/base/base.py
daehyun99: that work is doing in Pull Request #2400.

I thought #2400 would be merged soon, so I thought it would be good to apply it to `CoreGraph` as well.

What should we do?

pgmpy/base/base.py
```diff
@@ -0,0 +1,191 @@
...
+
+    >>> [Example_Code]
+
+    **Roles:**
+
+    >>> [Example_Code]
+
+    """
+
+    SUPPORTED_EDGE_TYPES = ["--", "-*", "*-", "->", "<-", "*>", "<*", "<>", "**"]
```
ankurankan: The `*` needs to be `o`.

pgmpy/tests/test_base/test_base.py
```diff
@@ -0,0 +1,312 @@
...
+
+import unittest
+
+import pytest
+
+import pgmpy.tests.help_functions as hf
+from pgmpy.base import _CoreGraph
+
+
+class TestCoreGraph(unittest.TestCase):
```
ankurankan: We are trying to move to pytest from unittest. Could you please modify the tests to not use `unittest`?

pgmpy/tests/help_functions.py
```diff
@@ -3,3 +3,21 @@ def recursive_sorted(li):
     for i in range(len(li)):
         li[i] = sorted(li[i])
     return sorted(li)
+
+
+def check_graph_status(
```
ankurankan: Would suggest keeping this method in the `test_base.py` file itself.

daehyun99:
For your information, I expect the test to pass when the #2400 Pull Request is merged.

daehyun99:
Hi, @ankurankan @Spinachboul 
I think all the necessary features and test code for CoreGraph have been developed.
I would appreciate it if you could review the code when you have time.

---

- I defined `IS_MULTIGRAPH` to make the `add_edge()` logic different depending on whether it is a multi-graph.
- If inheriting from `nx.MultiGraph`, edges from u to v and from v to u are added simultaneously. Therefore, I modified it to inherit from `nx.MultiDiGraph`.
- I added the `get_edges()` method to represent both unidirectional edges and bidirectional edges simultaneously.

FYI: https://github.com/pgmpy/pgmpy/pull/2402#issuecomment-3471805178

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+            Each edge given in the container will be added to the
+            graph. The edges can be:
+
+                - 3-tuples (u, v, type)
+                - 4-tuples (u, v, type, key) for an edge with data and key
+
+    latents : set of nodes (default: empty `set()`)
+    exposures : set of nodes (default: empty `set()`)
+    outcomes : set of nodes (default: empty `set()`)
+    roles : dict, optional (default: `None`)
```
ankurankan: This needs to numpydoc style: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+
+    See Also
+    --------
+
+    Notes
+    -----
+
+    Examples
+    --------
+"""
```
ankurankan: This is not needed.

pgmpy/base/__init__.py
```diff
@@ -1,7 +1,17 @@
...
 from .UndirectedGraph import UndirectedGraph
 
-__all__ = ["UndirectedGraph", "DAG", "PDAG", "AncestralBase", "MAG"]
+__all__ = [
+    "ADMG",
+    "_CoreGraph",
+    "UndirectedGraph",
+    "DAG",
+    "PDAG",
+    "AncestralBase",
```
ankurankan, line 14: We wouldn't need `AncestralBase` after `_CoreGraph`. Right?

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+from typing import Hashable, Iterable, Optional, Union
+
+import networkx as nx
+
+from pgmpy.base._mixin_roles import _GraphRolesMixin
+
+
+class _CoreGraph(nx.MultiDiGraph, _GraphRolesMixin):
+    """
+    Base graph class.
```
ankurankan: I would suggest adding a bit more detail here covering the following points:
1. Generalized representation of all graphs in pgmpy.
2. All graph classes inherit and define their own `SUPPORTED_EDGE_TYPES` to restrict.
3. Provides generalized algorithms that work for all graphs.

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+    Returns
+    -------
+    None
+
+    See Also
+    --------
+
+    Notes
+    -----
+
```
ankurankan: If any of the docstring sections are empty, you can remove them.

pgmpy/base/_base.py
```diff
@@ -0,0 +1,766 @@
...
+    -------
+    None
+
+    See Also
+    --------
+
+    Notes
+    -----
+
+    Examples
```
ankurankan, line 44: I would suggest keeping the `Examples` section a bit shorter. The idea would be to show some of the main functionalities and then the docstrings in the individual methods can show more functionality/options.

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+    >>> G.with_role("Custom_role", "A", inplace=True)
+    >>> G.with_role("latents", "D", inplace=True)
+    >>> G.get_role_dict()
+    {'latents': ['D'], 'Custom_role': ['A']}
+    >>> G.without_role("Custom_role", "A", inplace=True)
+    >>> G.get_role_dict()
+    {'latents': ['D']}
+    """
+
+    IS_MULTIGRAPH = True
```
ankurankan: What does `IS_MULTIGRAPH` represent? Which of the graph classes among `DAG`, `PDAG`, `MAG`, `PAG`, `ADMG` would it be True for?

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+    SUPPORTED_EDGE_TYPES = ["--", "-o", "o-", "->", "<-", "o>", "<o", "<>", "oo"]
```
ankurankan: Would be better to store it as a set instead of a list. Would help in faster lookups.

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+        outcomes: Optional[set[Hashable]] = None,
+        latents: Optional[set[Hashable]] = None,
+        roles=None,
+    ):
+        """
+        Notes
+        --------
+        - Sub-graph classes must implement `SUPPORTED_EDGE_TYPES`
+        - Sub-graph classes must implement `IS_MULTIGRAPH`
+        """
```
ankurankan: I would move this to the `_CoreGraph` docstring instead of here.

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+        Notes
+        -----
+        - This method is expected to be usable without being implemented in a subclass of the graph class.
+
+        Examples
+        --------
+        >>> from pgmpy.base import _CoreGraph
+        >>> G = _CoreGraph()
+        >>> G.SUPPORTED_EDGE_TYPES  # check the available edge types
+        ['--', '-o', 'o-', '->', '<-', 'o>', '<o', '<>', 'oo']
```
ankurankan: I don't think users would directly need to interact with `SUPPORTED_EDGE_TYPES`, so I would suggest avoiding using it in the examples.

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+            if (u is None) or (v is None):
+                raise ValueError("Nodes cannot be None.")
+            if u == v:
+                raise ValueError("Nodes cannot be the same for an edge.")
+            if (type is None) or (type not in self.SUPPORTED_EDGE_TYPES):
+                raise ValueError(f"Types must be one of {self.SUPPORTED_EDGE_TYPES}.")
+            result.append((u, v, type, key))
+        return result
+
+    def _get_edges_type_key(self):
```
ankurankan: What does `key` represent here?

ankurankan (changes requested):

pgmpy/base/base.py
daehyun99: @ankurankan 
> 1. What does `IS_MULTIGRAPH` represent?

I think `_CoreGraph` should be the most general graph from which all graphs in `pgmpy` can inherit.
So I think it should inherit from `networkx.MultiDiGraph`, and the graph's multi-edge capability can be set with `IS_MULTIGRAPH: bool`.

I expect that if a new graph class is developed in `pgmpy` in the future, the class development time can be shortened with `IS_MULTIGRAPH`.

> 2. Which of the graph classes among `DAG`, `PDAG`, `MAG`, `PAG`, `ADMG` would it be True for?

I think they are all `False`.
When writing the method, I also considered the possibility of multi-edges since `ADMG` inherits from `networkx.MultiDiGraph`.

https://github.com/pgmpy/pgmpy/blob/2d10e05dd8f4615edbc7fbe24e8c51a3fe8705d0/pgmpy/base/ADMG.py#L10

Perhaps it inherits from `MultiDiGraph`, considering the situation where `directed edges` and `bidirected edges` exist simultaneously between two nodes.

How about setting the multi-edge status with `IS_MULTIGRAPH` and considering the situation where multiple edges of the same type exist between two nodes with a `new variable`?

pgmpy/base/base.py
daehyun99: @ankurankan 
The [`networkx.MultiDiGraph`](https://github.com/networkx/networkx/blob/d302ea14cb75a80513052ba3a9010f14a203a587/networkx/classes/multidigraph.py#L428)'s `add_edge()` and `remove_edge()` methods have a `key` parameter.

The `key` parameter distinguishes edges between two nodes.

```py
from networkx import MultiDiGraph

G = MultiDiGraph()

G.add_edge("A", "B")
G.add_edge("A", "B")
G.add_edge("A", "B")

G.edges(keys=True)
# OutPut : OutMultiEdgeView([('A', 'B', 0), ('A', 'B', 1), ('A', 'B', 2)])
```

In addition to this, `_CoreGraph` sets the `type` of the edge.

Therefore, when using the `_CoreGraph`'s `add_edge()` and `remove_edge()` methods, `key` allows users to correctly add or remove the desired `type` of edge.

pgmpy/base/base.py
ankurankan: In causal graphs, there isn't a scenario where multiple edges of the same type are present between a pair of nodes. Between a given pair of variables, there can be at max 1 edge of 1 type. Hence, I don't think `IS_MULTIGRAPH` would be a useful attribute as it will always be False. I would suggest removing it.

pgmpy/base/base.py
ankurankan: As I mentioned above, there won't be multiple edges of the same type between a pair of variables, so we can remove any functionality related to specifying `key` and let networkx assign default values.

pgmpy/base/base.py
daehyun99: Yes, I understand.
I will apply the review and remove `IS_MULTIGRAPH`.

pgmpy/base/base.py
daehyun99: @ankurankan 
I think the methods related to the `key` value is necessary.

In the `add_edge()` method of `nx.MultiDiGraph`, it's not a problem even if there is no `key` value.
However, in the `remove_edge()` method of [`nx.MultiDiGraph`](https://github.com/networkx/networkx/blob/d302ea14cb75a80513052ba3a9010f14a203a587/networkx/classes/multidigraph.py#L526), it deletes a specific edge using the `u`, `v`, and `key` values.
With the `type` value we defined, we cannot select the edge to delete.

Therefore, I think a method to find the `key` of the edge to be deleted is necessary.

---
Additionally, I have one question regarding the user use case.

There are two graphs where only the edge `key` values are different, and everything else is the same.
In this situation, should `G1.__eq__(G2)` return `True`?
Currently, it is returning `False`.

```py
from pgmpy.base import _CoreGraph

G1 = _CoreGraph()
G1.add_edge("A", "B", "->")
G1.add_edge("A", "B", "<>")
print("G1's edges: ", G1.get_edges(type=True, key=True))
# output: G1's edges:  [('A', 'B', {'type': '->'}, 0), ('A', 'B', {'type': '<>'}, 1)]

G2 = _CoreGraph()
G2.add_edge("A", "B", "<>")
G2.add_edge("A", "B", "->")
print("G2's edges: ", G2.get_edges(type=True, key=True))
# output: G2's edges:  [('A', 'B', {'type': '<>'}, 0), ('A', 'B', {'type': '->'}, 1)]

print(G1.__eq__(G2))
# output: False

```

---
Please let me know if I missed or misunderstood anything.
Thank you.

pgmpy/base/base.py
```diff
@@ -0,0 +1,766 @@
...
+        """
+        return sorted(
+            [
+                (u, v, data.get("type"), key)
+                for u, v, key, data in self.edges(data=True, keys=True)
+            ],
+            key=lambda x: (x[0], x[1]),
+        )
+
+    def _get_key(
```
Spinachboul: Is this function really necessary?
Doesn't networkx handle this implicitly?
`G.edges(keys=True, data=True)`
This should easily return the edges and the keys and the edge's data.
And I dont find much use cases of this helper function elsewhere.
@ankurankan Your thoughts??

Spinachboul:
@daehyun99 I would suggest testing the test functions locally before pushing on to the remote branch. This saves us time from getting the same errors while reviewing the PR

pgmpy/base/base.py
ankurankan: @daehyun99 Got it, I had missed that `key` would be required for the `remove_edge` method. I am wondering if it would be possible to remove edges based just on the `type` instead of `key`? The only case where we would have multiple edges is in the case of `ADMG` where one would be a directed edge and the other a bidirected edge. If the user provides the type of edge, we can use that to match which edge needs to be removed. What do you think? I feel having both `key` and `type` as arguments in `remove_edge` makes it a bit complex and gets a bit confusing on what each of the arguments is supposed to do.

daehyun99: >  I am wondering if it would be possible to remove edges based just on the `type` instead of `key`?

Yes, it's possible.
Below is the test code in `test_base.py` that deletes a specific edge only by its `type` value.

```py
    def test_remove_directed_edge(self):
        """Test removing the direct edge of a `_CoreGraph`."""
        edges = [("A", "B", "->"), ("B", "C", "<-")]
        graph = _CoreGraph(ebunch=edges)

        graph.remove_edge("A", "B", "->")
        graph.remove_edge("B", "C", "<-")

        assert not graph.has_edge("A", "B")
        assert not graph.has_edge("B", "C")

        assert graph.get_edges(type=True) == []

        check_graph_status(graph, 3, 0, set(), set(), set(), {})
```

---

> I feel having both `key` and `type` as arguments in `remove_edge` makes it a bit complex and gets a bit confusing on what each of the arguments is supposed to do.

I will modify `remove_edge()` so that the user cannot use the `key` argument.
The `key` value will only be used within the logic.

Thanks.

daehyun99:
> @daehyun99 I would suggest testing the test functions locally before pushing on to the remote branch. This saves us time from getting the same errors while reviewing the PR

Yes, I understand.

---
@ankurankan @Spinachboul 
I'd like to hear your opinion regarding the implementation of `get_*` methods in `_CoreGraph`.

I'm planning to implement `get_*` (e.g., `get_parents`, `get_children`, ...).
Between the two ways below, which one do you think is better?

---
1. way 1: Finding specific nodes with separate methods (e.g., `get_parents`, `get_children`, `get_spouses`).
I think this is similar to the current implementation of the `AncestralBase` class.
- example
```py
G.get_parents("A") # return parent nodes.
G.get_children("A") # return children nodes.
```

2. way 2: Finding specific nodes in a single method (`get_nodes()`) using a `relation` argument.
- example
```py
G.get_nodes("A", relation="parents") # return parent nodes.
G.get_nodes("A", relation="childrens") # return children nodes.
```

I think way 2(`get_nodes()`) would be better from a user perspective, but I'm worried the logic of a single method might become too complex. 
What do you think?

Thanks.

ankurankan:
@daehyun99 I would go with option 1, where the methods are separate. I think that it would be nicer to keep the logic in separate methods; otherwise, we would end up with a long list of if-elses in a single method.

Spinachboul:
> @daehyun99 I would go with option 1, where the methods are separate. I think that it would be nicer to keep the logic in separate methods; otherwise, we would end up with a long list of if-elses in a single method.

I agree here since it would be easy and clearer for the users to understand what the function is and what does it do

pgmpy/base/base.py
daehyun99: FYI: 

> @daehyun99 Got it, I had missed that `key` would be required for the `remove_edge` method. I am wondering if it would be possible to remove edges based just on the `type` instead of `key`? The only case where we would have multiple edges is in the case of `ADMG` where one would be a directed edge and the other a bidirected edge. If the user provides the type of edge, we can use that to match which edge needs to be removed. What do you think? I feel having both `key` and `type` as arguments in `remove_edge` makes it a bit complex and gets a bit confusing on what each of the arguments is supposed to do.

I've modified the code logic to reflect your comments. ([commit](https://github.com/pgmpy/pgmpy/pull/2402/commits/ad0ab4977ce04383e8ba9435b58fe63c8dc9101c))
Now, the `key` value will be declared as the user-input `type` value, instead of an integer.
This means that a maximum of one edge of a specific `type` can exist between two nodes.

Once I finish developing the `get_*` methods, I'll request a review.

Thanks.

daehyun99:
Resolved. ~~This PR's test code depends on #2415.~~

pgmpy/base/base.py
```diff
@@ -0,0 +1,557 @@
...
+        >>> G2.__class__
+        pgmpy.base.base._CoreGraph
+        """
+        ebunch = self._get_edges_type_key()
+
+        graph_copy = self.__class__()
+        graph_copy.add_edges_from(ebunch=ebunch)
+        graph_copy.exposures = self.exposures.copy()
+        graph_copy.outcomes = self.outcomes.copy()
+        graph_copy.latents = self.latents.copy()
```
ankurankan: Do we need to manually set these? As all of these attributes are stored as `roles` internally, they should automatically get copied in the next 2 lines?

ankurankan (changes requested):

pgmpy/base/base.py
daehyun99: > Do we need to manually set these? As all of these attributes are stored as `roles` internally, they should automatically get copied in the next 2 lines?

Yes. I overlooked that.
I think you're right.

pgmpy/base/base.py
```diff
@@ -0,0 +1,888 @@
...
+        elif type in ["--", "-o", "o-"]:
+            reverse_type = f"{type[1]}{type[0]}"
+            super().add_edge(u, v, key=type, **kwargs)
+            super().add_edge(v, u, key=reverse_type, **kwargs)
+        elif type in ["<>", "oo"]:
+            super().add_edge(u, v, key=type, **kwargs)
+            super().add_edge(v, u, key=type, **kwargs)
+        else:
+            raise AssertionError(
+                "This should never happen."
```
ankurankan: A better way to phrase the error would be something like: `f"Unknown edge type. _CoreGraph currently supports the following edge types: .... Got: {type}")`

daehyun99:
@ankurankan @Spinachboul 
I think that `_CoreGraph` class is ready for merged.
Please review.

Thanks.

pgmpy/base/base.py
```diff
@@ -0,0 +1,948 @@
...
+
+    # ----------------------------------------------------------------------
+    # Public API (or Public Methods)
+    # ----------------------------------------------------------------------
+
+    def add_edge(
+        self,
+        u: Hashable,
+        v: Hashable,
+        type: str = None,
```
ankurankan: I think we should rename `type` to `edge_type` as `type` is a keyword in Python.

pgmpy/base/base.py
```diff
@@ -0,0 +1,948 @@
...
+        # Adding edge base on type value.
+        if type in ["<-", "<o"]:
+            u, v = v, u
+            type = f"{type[1]}>"
+        super().add_edge(u, v, key=type, **kwargs)
+
+        # Further adding edge based on the edge's type.
+        if type in ["--", "-o", "o-"]:
+            reverse_type = f"{type[1]}{type[0]}"
+            super().add_edge(v, u, key=reverse_type, **kwargs)
```
ankurankan: Why do we need to add reverse edges?

pgmpy/base/base.py
```diff
@@ -0,0 +1,948 @@
...
+        Examples
+        --------
+        >>> from pgmpy.base import _CoreGraph
+        >>> edges = [("A", "B", "->"), ("B", "C", "->")]
+        >>> G = _CoreGraph()
+        >>> G.add_edges_from(ebunch=edges)
+        >>> G.get_edges(type=True)
+        [('A', 'B', {'type': '->'}), ('B', 'C', {'type': '->'})]
+        """
+        self._validating_edges_value(ebunch=ebunch)
```
ankurankan: I don't think we need to call the validation method here. The edges are going to get validated in the `add_edge` method again in line 263.

pgmpy/base/base.py
```diff
@@ -0,0 +1,948 @@
...
+        >>> G2.__class__
+        pgmpy.base.base._CoreGraph
+        """
+        ebunch = self._get_edges_type_key()
+
+        graph_copy = self.__class__()
+        graph_copy.add_edges_from(ebunch=ebunch)
+        graph_copy.exposures = self.exposures
+        graph_copy.outcomes = self.outcomes
+        graph_copy.latents = self.latents
```
ankurankan: Don't think we will need to add these lines as in 401 we are copying the roles which will already have exposure, outcomes, and latents.

pgmpy/base/_base.py
```diff
@@ -0,0 +1,948 @@
...
+        The type value is matched with the key value of networkx.
+        Helper method for
+            `copy()`,
+            `__eq__()`,
+            `get_edges()`,
+        """
+        return sorted(
+            [(u, v, type) for u, v, type in self.edges(keys=True)],
+            key=lambda x: (x[0], x[1]),
+        )
```
ankurankan: I don't think sorting is required for anything other than `__eq__` method. We can also simply use `graphs_equal` method from networkx instead of implementing our own sorting here.

pgmpy/base/_base.py
```diff
@@ -0,0 +1,948 @@
...
+            graph_copy.with_role(role=role, variables=vars, inplace=True)
+
+        return graph_copy
+
+    def get_edges(self, type: bool = False):
+        """
+        Returns a list of edges in the graph.
+
+        For undirected and bidirected edges, which are stored as two directed
+        edges internally, this method returns only one of them.
```
ankurankan: I don't think we need to store them as two edges anymore as we are storing the type of edge along with them. This was required earlier in PDAG and ADMG classes as we didn't used to store the edge types.

pgmpy/base/base.py
```diff
@@ -0,0 +1,948 @@
...
+
+        self_ebunch = self._get_edges_type_key()
+        other_ebunch = other._get_edges_type_key()
+
+        return (
+            set(self.nodes()) == set(other.nodes())
+            and set(self_ebunch) == set(other_ebunch)
+            and self.exposures == other.exposures
+            and self.outcomes == other.outcomes
+            and self.latents == other.latents
```
ankurankan: Checking for roles dictionary (line 855) should be sufficient, right? These roles would be stored in the roles dictionary as well.

pgmpy/base/base.py
```diff
@@ -0,0 +1,948 @@
...
+            and self.get_role_dict() == other.get_role_dict()
+        )
+
+    def _validating_edges_value(
```
ankurankan: Would suggest renaming this to `_validate_edges`.

pgmpy/base/base.py
```diff
@@ -0,0 +1,948 @@
...
+            `copy()`,
+            `__eq__()`,
+            `get_edges()`,
+        """
+        return sorted(
+            [(u, v, type) for u, v, type in self.edges(keys=True)],
+            key=lambda x: (x[0], x[1]),
+        )
+
+    def _validating_nodes_value(self, node, type):
```
ankurankan: Similary, would suggest renaming this to `_validate_nodes`

ankurankan (changes requested):
@daehyun99 I have added a few comments. I think it mostly looks good except the part where we are storing multiple edges for undirected and bidirected edges. As we anyways store the edge types now, I don't think we should need to create multiple edges for them. Let me know if I am missing something here.

pgmpy/base/_base.py
ankurankan, line 1: Please also rename this file to `_base.py` as this implements non-public class.

pgmpy/base/base.py
daehyun99: > Why do we need to add reverse edges?

@ankurankan 
I will explain this using the `networkx` library code as an example.

In `networkx`, edge information for a graph is stored in `_adj`.

In `pgmpy`, `nx.Graph` is used to represent undirected edges(`UndirectedGraph`), while `nx.DiGraph` is used to represent directed edges(`DAG`).

### In `networkx`
```python
import networkx as nx

graph1 = nx.Graph()
graph1.add_edge("A", "B")
graph1._adj
# output
# {'A': {'B': {}}, 'B': {'A': {}}}
```

The output of `graph1._adj` indicates that edges exist from `A` to `B` and from `B` to `A`.
This represents an undirected edge. (I believe this is how `networkx` internally represents undirected edges.)

```python
import networkx as nx

graph2 = nx.DiGraph()
graph2.add_edge("A", "B")
graph2._adj
# output
# {'A': {'B': {}}, 'B': {}}
```

The output of `graph2._adj` indicates that a directed edge exists from `A` to `B`.
This represents a directed edge.

### In `pgmpy`
```python
from pgmpy.base.base import _CoreGraph

graph3 = _CoreGraph()
graph3.add_edge("A", "B", "->")
graph3._adj
# output
# {'A': {'B': {'->': {}}}, 'B': {}}
```

`_CoreGraph` inherits from `nx.MultiDiGraph`.
It needs to be capable of representing both directed and undirected edges simultaneously.
Therefore, to represent edges that are undirected (`--`) or have unknown directionality (`-o`, `o-`, `oo`), I implemented them using reverse edges.

```python
from pgmpy.base.base import _CoreGraph

graph4 = _CoreGraph()
graph4.add_edge("A", "B", "--")
graph4._adj
# output
# {'A': {'B': {'--': {}}}, 'B': {'A': {'--': {}}}}
```
```python
from pgmpy.base.base import _CoreGraph

graph5 = _CoreGraph()
graph5.add_edge("A", "B", "-o")
graph5._adj
# output
# {'A': {'B': {'-o': {}}}, 'B': {'A': {'o-': {}}}}
```

Please let me know if I have misunderstood anything. Thank you.

pgmpy/base/base.py
daehyun99: > I don't think we need to call the validation method here. The edges are going to get validated in the `add_edge` method again in line 263.

@ankurankan  I agree that the double validation is redundant.

The reason I added line 260 was to ensure that no edges are added to the graph if an error case (like the one below) exists. I thought this all-or-nothing approach would be more user-friendly.

Do you think we should remove line 260?

```python
edges = [("A", "B", "->"), (None, "A", "->"), ("B", "C", "->")]
graph.add_edges_from(ebunch=edges)
```

pgmpy/base/base.py
ankurankan: @daehyun99 The behavior of the methods that you have implemented is correct. I just think that using two edges to represent certain types of edges complicates the implementation a bit. So, my question is if we only use a single edge for every edge type, can we not make the rest of the methods work? Does that lead to more complicated logic in other methods? Or, could we switch to inheriting from `nx.MultiGraph` instead of `nx.MultiDiGraph` to simplify this?

pgmpy/base/base.py
ankurankan: Yeah, that makes sense. I had missed that.

ankurankan:
@daehyun99 I have also fixed some merge conflits on this branch, please make sure to do a `git pull` before pushing anything.

pgmpy/base/base.py
daehyun99: @ankurankan 
I'm not entirely sure, but I think inheriting from `nx.MultiGraph` would lead to more complex changes during the refactoring process.

https://github.com/pgmpy/pgmpy/blob/a1a5f2d9767e36e9f83d1f96ae631191d5b5a37d/pgmpy/base/DAG.py#L190-L200

This is because `nx.MultiGraph` doesn't support edge directionality, whereas `DAG`s strictly prohibit cycles. So, I suspect we might not be able to use `nx.find_cycle()`, which is currently utilized in the `DAG._check_cycles()` method.

---

expect_final

Additionally, I haven't personally worked on the graph classes under `pgmpy/models/`, so I’m not familiar with the implementation details.

However, I suspect there is a high chance of issues arising in classes that inherit from `DAG`, such as `DynamicBayesianNetwork`, `LinearGaussianBayesianNetwork`, and `SEMGraph`.

Therefore, I believe inheriting from `nx.MultiDiGraph` would allow us to handle issues more flexibly during the refactoring process.

---

Furthermore, if the edge handling behavior of `_CoreGraph` and `UndirectedGraph` is consistent within `networkx`, I believe it would lower the complexity of implementing `_GraphAlgorithmMixin`.

What do you think?
Thanks.

pgmpy/base/_base.py
daehyun99: Yes, I understand.
I removed the sorting logic. 

However, since it's needed for testing, I added a function to `test_base.py` that returns sorted values.(`def get_sorted_edges()`)

Thanks.

daehyun99:
Hi @ankurankan 
I’ve incorporated the feedback from our last meeting.
Please review it at your convenience.

Thanks.

---

I add `_convert_edge_type` method. (`_preprocess_edge` -> `_convert_edge_type` -> `_from_api_edge_type`)
this method translate user's `edge_type` input to store form.
it is replacing `_preprocess_edge`

daehyun99:
pgmpy edges (16)

I add `_from_api_edge_type` and `_to_api_edge_type` to convert edge type's representation. (User API<> pgmpy)

I add `get_edges()` method.
because I think that expression of `get_edges()` is more user friendly than `edges()`.

```py
edges = [("A", "B", "->"), ("A", "B", "<>"), ("B", "C", "->")]
G = _CoreGraph(ebunch=edges)

# get_edges
G.get_edges(keys=True, data=True)
[('A', 'B', 0, '->'), ('A', 'B', 1, '<>'), ('B', 'C', 0, '->')]

# edges
G.edges(keys=True, data=True)
MultiEdgeDataView([('A', 'B', 0, {'A': '-', 'B': '>'}), ('A', 'B', 1, {'A': '>', 'B': '>'}), ('B', 'C', 0, {'B': '-', 'C': '>'})])
```

Please review it at your convenience.

Thanks.

ankurankan:
@daehyun99 I had a quick look, and this looks good to me on a high level. @Spinachboul If you have time, could you please review this?

ankurankan (approved):
Looks good to me. Thanks for the massive PR @daehyun99

Spinachboul:
@daehyun99 Thanks for PR. Really excited to see how this would turn out.
Hoping for the best!

PR #2418, opened 2025-11-14
Adds BaseCausalPrediction class

### LINKED ISSUE #2408: [ENH] Implement `BaseCausalPrediction` class
> I am expecting this method to be common in all of our causal prediction algorithms. It might be a good idea to create a `BaseCausalPrediction` class that is inherited by all prediction classes and move this method there. sklearn tags could also possibly be moved.

_Originally posted by @ankurankan in https://github.com/pgmpy/pgmpy/pull/2378#discussion_r2419803599_

DARHWOLF:

- Fixes #2408 

- Adds base class for the `prediction` module for sharing common utilities


pgmpy/prediction/NaiveAdjustmentRegressor.py
```diff
@@ -142,37 +143,6 @@ def __sklearn_tags__(self):
         tags.regressor_tags.poor_score = True
         return tags
```
DARHWOLF: adjustment regressor uses these additional tags... I am not sure if we can move them. 
DML and NaiveIV have the same tags though.

ankurankan:
@XAheli FYI.

pgmpy/prediction/_base.py
ankurankan, line 1: I would suggest changing this file to be simply `_base.py`. This would be more inline with the sklearn naming style as we do not expect users to directly interact with this file. And would also be consistent with pgmpy as we have a `base.py` file in each of the modules.

pgmpy/prediction/BaseCausalPrediction.py
```diff
@@ -0,0 +1,59 @@
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator
+
+
+class BaseCausalPrediction(BaseEstimator):
```
ankurankan: Would suggest renaming it to `_BaseCasualPrediction` as we don't expect users to interact with this class.

pgmpy/prediction/BaseCausalPrediction.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    Base class for causal prediction models in pgmpy. Provides common
```
ankurankan: causal prediction models -> causal prediction algorithms.

pgmpy/prediction/BaseCausalPrediction.py
```diff
@@ -0,0 +1,59 @@
...
+        range index column names (0, 1, ..., n_features-1).
+
+        Parameters
+        ----------
+        X : array-like or DataFrame
+            Input features.
+
+        required_features : list[int] or None
+            Column indices expected from the DAG. If None, uses
+            `self.feature_columns_fit_`.
```
ankurankan: Let's make it required, as that would be much more explicit. In the current form, it would be very difficult to guess what the default is without looking at this docstring.

pgmpy/prediction/NaiveAdjustmentRegressor.py
```diff
@@ -225,12 +195,14 @@ def fit(
         self.outcome_var_ = outcome_vars[0]
         self.adjustment_vars_ = adjustment_vars
         self.pretreatment_vars_ = pretreatment_vars
-        self.feature_columns_ = (
+        self.feature_columns_fit_ = (
```
ankurankan, line 190: I think we will need to modify the docstring for this as well.

pgmpy/tests/test_prediction/test_DoubleMLRegressor.py
```diff
@@ -267,7 +267,7 @@ def test_dag_roles_validation_and_pretreatment_support():
     Y = 1.2 * D + 0.3 * Z + rng.normal(scale=0.2, size=50)
     df = pd.DataFrame({"D": D, "Z": Z, "P": P})
     _ = model.fit(df, pd.Series(Y, name="Y"))
-    assert set(model.feature_columns_) >= {"D", "Z", "P"}
+    assert set(model.feature_columns_fit_) >= {"D", "Z", "P"}
```
ankurankan: Why is this not `==`?

pgmpy/prediction/DoubleMLRegressor.py
```diff
@@ -295,12 +263,12 @@ def fit(self, X, y, sample_weight: Optional[Any] = None):
         self.outcome_var_ = outcome_vars[0]
         self.adjustment_vars_ = self.causal_graph.get_role("adjustment")
         self.pretreatment_vars_ = self.causal_graph.get_role("pretreatment")
-        self.feature_columns_ = (
+        self.feature_columns_fit_ = (
```
ankurankan, line 261: Need to change this in the docstring as well.

ankurankan (changes requested):
@DARHWOLF Thanks for the PR. I have added a few comments. About the tags, I think the additional ones in `NaiveAdjustmentRegressor` are redundant. The `target_tags.required` is True by default when inheriting `RegressorMixin`, the `input_tags.allow_nan` is False by default. So, I think we can just remove them and move the `__sklearn_tags__` to `_BaseCausalPrediction` as well.

ankurankan:
@DARHWOLF Could you please include the changes needed in `NaiveIVRegressor` in this PR as well.

pgmpy/prediction/_base.py
DARHWOLF: Done

pgmpy/prediction/NaiveAdjustmentRegressor.py
DARHWOLF: Modified

pgmpy/prediction/DoubleMLRegressor.py
DARHWOLF: Changed

pgmpy/tests/test_prediction/test_DoubleMLRegressor.py
DARHWOLF: This should be `==` 
I have corrected it here in DML and also in the same test for the iv regressor.

ankurankan (approved):
@DARHWOLF Looks good to me. Thanks for the PR.

PR #2430, opened 2025-11-26
[ENH] Move the orientations argument in ExpertInLoop into ExpertKnowledge

### LINKED ISSUE #2429: [ENH] Move the `orientations` argument in `ExpertInLoop` into `ExpertKnowledge`
The `pgmpy.estimators.ExpertInLoop` algorithm provides the option for users to specify orientations of edges that are used over the orientation returned by the `orientation_fn`. This `orientation` argument should be moved to the `ExpertKnowledge` class, which would allow users to specify the preferred orientation of the edges if the edge exists in the graph.

This might also allow us to remove the `enforce_expert_knowledge` argument in the `PC` algorithm, which allows the users to choose between preferring expert knowledge vs enforcing the expert knowledge.

Aniketsy:
Fix #2429

### DIFF
--- a/pgmpy/estimators/BaseConstraintEstimator.py
+++ b/pgmpy/estimators/BaseConstraintEstimator.py
@@ -61,7 +61,6 @@ class BaseConstraintEstimator(StructureEstimator):
         significance_level: float = 0.01,
         max_cond_vars: int = 5,
         expert_knowledge: Optional[ExpertKnowledge] = None,
-        enforce_expert_knowledge: bool = False,
         n_jobs: int = -1,
         show_progress: bool = True,
         **kwargs,
@@ -185,11 +184,14 @@ class BaseConstraintEstimator(StructureEstimator):
             pbar.set_description("Working for n conditional variables: 0")
 
         # Step 1: Initialize a fully connected undirected graph
-        graph = nx.complete_graph(n=self.variables, create_using=nx.Graph)
-        temporal_ordering = expert_knowledge.temporal_ordering
-        if enforce_expert_knowledge:
+        if expert_knowledge and expert_knowledge.search_space:
+            graph = UndirectedGraph(expert_knowledge.search_space)
+            graph.add_nodes_from(self.variables)
+        else:
+            graph = nx.complete_graph(n=self.variables, create_using=nx.Graph)
+        if expert_knowledge and expert_knowledge.forbidden_edges:
             graph.remove_edges_from(expert_knowledge.forbidden_edges)
-
+        temporal_ordering = expert_knowledge.temporal_ordering
         # Exit condition: 1. If all the nodes in graph has less than `lim_neighbors` neighbors.
         #             or  2. `lim_neighbors` is greater than `max_conditional_variables`.
         while not all(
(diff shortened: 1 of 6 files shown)


ankurankan:
@Aniketsy Thanks for the PR. The linting checks are failing, could you have a look? You can also check our contributing guide for setting up the pre-commit hook, which should automatically take care of linting.

Aniketsy:
@ankurankan sure, I will fix this .

Aniketsy:
```
C:\Users\Aniket.DESKTOP-074O80J\pgmpy\pgmpy>pre-commit run
check yaml...........................................(no files to check)Skipped
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
isort....................................................................Passed
black....................................................................Passed
flake8...................................................................Passed
blackdoc.................................................................Passed
```
@ankurankan  I'm still seeing a linting error on this.

```
pgmpy/estimators/ExpertKnowledge.py:121:14: E721 do not compare types, for exact checks use `is` / `is not`, for instance checks use `isinstance()`
pgmpy/tests/test_estimators/test_PC.py:515:9: F841 local variable 'dag' is assigned to but never used
pgmpy/tests/test_estimators/test_PC.py:526:13: F841 local variable 'dag' is assigned to but never used
```

ankurankan:
@Aniketsy Since you made the commit before pre-commit was installed, you will have to manually fix the errors that you are seeing. For any new commits, pre-commit will complain.

ankurankan:
@Nimish-4 What do you think of this idea of removing the `enforce_expert_knowledge`? We can change the way we define  `required_edges` and `forbidden_edges` in `ExpertKnowledge`. The new definitions would imply that the `required_edges` and `forbidden_edges` do not consider the orientation rather just the pair of variables. And this new `required_orientation` argument would define the edge orientation (essentially meaning that this orientation will be enforced). I find this definition a bit cleaner compared to our current one.

@Aniketsy we will need to make changes to the other causal discovery algorithms to accommodate the `orientation` argument. But let's wait a bit till we have clarity on the API.

Aniketsy:
@ankurankan sure, I’ll wait for the API clarification before making further changes.

Nimish-4:
@ankurankan Yep definition wise it would be better to handle that in the expert knowledge class itself, would look cleaner.

Aniketsy:
Please let me know if I need to make any further changes or if there are any updates regarding this.

ankurankan:
@Aniketsy Sorry for the delay on this; I am still trying to think through how this change would impact the rest of the causal discovery/structure learning classes. 

I think one of the problems is how to define the expert knowledge in a consistent way for each of the causal discovery classes. The PC algorithm sometimes isn't able to respect the forbidden and required edges because they can be inconsistent with what the algorithm learns, and hence, we needed the `enforce_expert_knowledge` argument to override the algorithm using expert knowledge. I don't think this was a problem with some of the other algorithms, like the `HillClimbSearch` where the algorithm can always respect the expert knowledge. So, I haven't been able to think of a way in which we can consistently define the expert knowledge so that the definition holds in every algorithm.

Aniketsy:
@ankurankan Thanks for the explanation, I’ll look into this as well and share my thoughts if I find something useful.

PR #2433, opened 2025-11-27
ports IV identification method

### LINKED ISSUE #2289: [ENH] Port identification methods from `inference.CausalInference` to `identification`
The current CausalInference class in pgmpy implements several identification methods:

1. Backdoor Adjustment
2. Frontdoor Adjustment
3. Instrumental Variables
4. Conditional Instrumental Variables

Recently, we introduced two key changes to improve the implementation of causal inference methods:
1. Enhanced the DAG and PDAG classes to allow specifying variable roles (e.g., exposure, outcome, adjustment).
2. Added a BaseIdentification class to serve as the common base for all identification methods.
With these changes in place, we can now migrate the identification algorithms in CausalInference to their own dedicated classes.

Changes required:
1. Create separate classes for each identification method, inheriting from BaseIdentification.
2. Update implementations to read exposure and outcome variables directly from the causal graph, rather than accepting them as method arguments.
3. Port the existing algorithm logic into the new classes.

DARHWOLF:

- Fixes #2289

- 
-
-


DARHWOLF:
hey @ankurankan, can I please get a preliminary review on this

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,118 @@
+from pgmpy.base import DAG
+from pgmpy.global_vars import logger
+from pgmpy.identification import BaseIdentification
+
+
+class InstrumentVariables(BaseIdentification):
```
ankurankan: `InstrumentVariables` -> `InstrumentalVariables`

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,118 @@
...
+        full_graph_ = causal_graph.copy()
+
+        exposures = full_graph_.get_role("exposures")
+        latent_variables = full_graph_.get_role("latents")
+        all_nodes = set(causal_graph.nodes)
+        observed = list(all_nodes - set(latent_variables))
+        if self.scaling_indicators is None:
+            scaling_indicators = self._get_scaling_indicators(causal_graph)
+        else:
+            scaling_indicators = self.scaling_indicators
```
ankurankan: I think it might be cleaner to handle this is-else within the `_get_scaling_indicators` method.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,118 @@
...
+            raise ValueError(f"The edge from {X} -> {Y} does not exist in the graph")
+
+        if (X in exposures) and (Y in observed):
+            if full_graph_.has_edge(X, Y):
+                full_graph_.remove_edge(X, Y)
+            dependent_var = Y
+
+        elif Y in latent_variables:
+            full_graph_.add_edge("." + Y, scaling_indicators[Y])
+            dependent_var = scaling_indicators[Y]
```
ankurankan, line 118: Isn't `X` and `Y` here are exposure and outcomes respectively?

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,118 @@
...
+    def _identify(self, causal_graph):
+        exposure = causal_graph.get_role("exposures")[0]
+        outcome = causal_graph.get_role("outcomes")[0]
+
+        latent_variables = causal_graph.get_role("latents")
+
+        if self.scaling_indicators is None:
+            scaling_indicators = self._get_scaling_indicators(causal_graph)
+        else:
+            scaling_indicators = self.scaling_indicators
```
ankurankan: Same as before.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,118 @@
...
+        if (exposure in scaling_indicators.keys()) and (
+            scaling_indicators[exposure] == outcome
+        ):
+            logger.warning(
+                f"{outcome} is the scaling indicator of {exposure}. Please specify `scaling_indicators`"
+            )
+
+        transformed_graph, dependent_var = self._iv_transformations(
+            exposure, outcome, causal_graph, scaling_indicators=scaling_indicators
+        )
```
ankurankan, line 157: Because we already handle the `scaling_indicator=None` case in this method, we don't need to set a default value for `scaling_indicators` in `_iv_transformations` method.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,118 @@
+from pgmpy.base import DAG
+from pgmpy.global_vars import logger
+from pgmpy.identification import BaseIdentification
+
+
+class InstrumentVariables(BaseIdentification):
+
+    def __init__(self, variant=None, scaling_indicators=None):
```
ankurankan: I think the conditional variant is still missing.

ankurankan (changes requested):
Looks good design-wise till now. A couple of things:
1. `_validate` method is missing.
2. We will need to think about how should we define roles for conditional instruments. In case of conditional instruments, in additon to specifying the role for a variable, we will need to also specify the conditional variables for it.
3. I think we should be able to define the SEM graphs using our ADMG class. I think the implementation should automatically work as it also has a method for active trails.

pgmpy/identification/instrument.py
DARHWOLF: changed

pgmpy/identification/instrument.py
DARHWOLF: done

pgmpy/identification/instrument.py
DARHWOLF: done

pgmpy/identification/instrument.py
DARHWOLF: added

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,148 @@
...
+
+from pgmpy.base import DAG
+from pgmpy.global_vars import logger
+from pgmpy.identification import BaseIdentification
+from pgmpy.inference.CausalInference import CausalInference
+
+
+class InstrumentalVariables(BaseIdentification):
+
+    def __init__(self, variant=None, scaling_indicators=None):
```
ankurankan: Should we name the normal version as `non-conditional`? That would give a clear distinction between the variants. Having the default as `None` doesn't tell us what it will do.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,148 @@
...
+        self.supported_graph_types = DAG
+        self.variant = variant
+        self.scaling_indicators = scaling_indicators
+
+    def _get_scaling_indicators(self, causal_graph):
+        latent_variables = causal_graph.get_role("latents")
+        observed_nodes = causal_graph.get_role("observed")
+        scaling_indicators = {}
+
+        if self.scaling_indicators is not None:
```
ankurankan: I think we should check here if the scaling indicator is defined for every latent variable. If there are any missing, we should find and assign for them.

pgmpy/identification/instrument.py
ankurankan: Can we add type hinting here?

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,148 @@
...
+            if full_graph_.has_edge(X, Y):
+                full_graph_.remove_edge(X, Y)
+            if parent_y in latent_variables:
+                full_graph_.add_edge(scaling_indicators[parent_y], dependent_var)
+
+        return full_graph_, dependent_var
+
+    def _identify(self, causal_graph):
+        exposure = causal_graph.get_role("exposures")[0]
+        outcome = causal_graph.get_role("outcomes")[0]
```
ankurankan: Can we add checks similar to other identification or prediction algorithms, where we throw an error if there is more than one exposure or outcome variable? With the current implementation, the implementation would give incorrect results when there is more than one exposure or outcome without notifying the user.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,148 @@
...
+class InstrumentalVariables(BaseIdentification):
+
+    def __init__(self, variant=None, scaling_indicators=None):
+        self.supported_graph_types = DAG
+        self.variant = variant
+        self.scaling_indicators = scaling_indicators
+
+    def _get_scaling_indicators(self, causal_graph):
+        latent_variables = causal_graph.get_role("latents")
+        observed_nodes = causal_graph.get_role("observed")
```
ankurankan: Can we still use `nodes() - latents` as observed? We do not, by default, add the observed role to variables. We will either need to change this or assume that only latent roles are specified, and the rest are observed.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,148 @@
...
+                full_graph_.remove_edge(X, Y)
+            if parent_y in latent_variables:
+                full_graph_.add_edge(scaling_indicators[parent_y], dependent_var)
+
+        return full_graph_, dependent_var
+
+    def _identify(self, causal_graph):
+        exposure = causal_graph.get_role("exposures")[0]
+        outcome = causal_graph.get_role("outcomes")[0]
+        observed = causal_graph.get_role("observed")
```
ankurankan: I don't think we currently set the observed role by default. With multi-role support, handling both observed and latent roles might be a bit complicated. So, it might be better that we simply assume variables to be observed if the latent role is not set.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,148 @@
...
+
+        latent_variables = causal_graph.get_role("latents")
+        scaling_indicators = self._get_scaling_indicators(causal_graph)
+
+        if (exposure in scaling_indicators.keys()) and (
+            scaling_indicators[exposure] == outcome
+        ):
+            logger.warning(
+                f"{outcome} is the scaling indicator of {exposure}. Please specify `scaling_indicators`"
+            )
```
ankurankan: Let's throw an error here asking user to specify the correct `scaling_indicators`.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def _get_scaling_indicators(self, causal_graph):
+        latent_variables = causal_graph.get_role("latents")
+        observed_nodes = causal_graph.get_role("observed")
+        scaling_indicators = {}
+
+        if self.scaling_indicators is not None:
+            return self.scaling_indicators
+
+        for node in latent_variables:
```
ankurankan: We might need to handle the edge case here that if `node` is also the `exposure` variable, we shouldn't accidentally assign the `outcome` variable as its scaling indicator. Basically, avoid the condition that we check in lines 75-80.

ankurankan (changes requested):

pgmpy/identification/instrument.py
DARHWOLF: added

pgmpy/identification/instrument.py
DARHWOLF: added line 29

pgmpy/identification/instrument.py
DARHWOLF: added

pgmpy/identification/instrument.py
DARHWOLF: added

pgmpy/identification/instrument.py
DARHWOLF: Done

pgmpy/identification/instrument.py
DARHWOLF: Done

DARHWOLF:
@ankurankan this should be ready for review, please have a look when time permits

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
+from networkx.algorithms.dag import descendants
+
+from pgmpy.base import DAG
+from pgmpy.identification import BaseIdentification
+from pgmpy.inference.CausalInference import CausalInference
+
+
+class InstrumentalVariables(BaseIdentification):
```
ankurankan, line 8: Needs docstring here.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def __init__(self, variant=None, scaling_indicators=None) -> None:
+        self.supported_graph_types = DAG
```
ankurankan: This needs to be a tuple similar to the `Frontdoor` class.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def _get_scaling_indicators(self, causal_graph):
+        exposure = causal_graph.get_role("exposures")
+        outcome = causal_graph.get_role("outcomes")
+        latent_variables = causal_graph.get_role("latents")
+        all_nodes = causal_graph.nodes()
+        observed_nodes = all_nodes - latent_variables
+        scaling_indicators = {}
+
+        if self.scaling_indicators is not None and len(self.scaling_indicators) == len(
```
ankurankan: It might be better to match exactly if there are scaling indicators for each latent variable, rather than just the lengths.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
...
+            latent_variables
+        ):
+            return self.scaling_indicators
+
+        # checks for missing scaling indicators and assigns for the missing nodes (if any)
+        if self.scaling_indicators is not None and len(self.scaling_indicators) < len(
+            latent_variables
+        ):
```
ankurankan: This is the same condition that we check above. Might be better to do either an if-else or since the if condition has a return statement above, we don't need the checks here.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
...
+                            break
+            return self.scaling_indicators
+
+        for node in latent_variables:
+            for neighbour in causal_graph.neighbors(node):
+                if neighbour in observed_nodes:
+                    if not (node in exposure and neighbour != outcome):
+                        scaling_indicators[node] = neighbour
+                        scaling_indicators
+                        break
```
ankurankan: I don't think we need this for loop. Isn't this case already handled by the loop above (line 32-40)?

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
...
+        if len(exposure) != 1:
+            raise ValueError(
+                f"The current implementation suppports only one exposure. Got: {len(exposure)}"
+            )
+        if len(outcome) != 1:
+            raise ValueError(
+                f"The current implementation suppports only one outcome. Got: {len(outcome)}"
+            )
+        all_nodes = causal_graph.nodes()
+        observed = all_nodes - causal_graph.get_role("latents")
```
ankurankan: Since we are using this `observed` variable in multiple methods, can we define a `self.observed` variable in the `__init__` method?

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
...
+                elif exposure in (causal_graph.active_trail_nodes([Z], observed=W))[Z]:
+                    instruments.append(Z)
+                    conditionals.append(W)
+                    # instruments.extend(list(W))
+                else:
+                    continue
+            if bool(instruments):
+                for i, j in enumerate(instruments):
+                    causal_graph.with_role("instrument", j, inplace=True)
+                    causal_graph.with_role("conditional", conditionals[i], inplace=True)
```
ankurankan: We do not need to loop here. We can assign a role to multiple variables in a single `with_role` call. Have a look at the docstring of the method.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,247 @@
...
+                    continue
+
+                # Condition to check if X d-connected to I after conditioning on W.
+                elif exposure in (causal_graph.active_trail_nodes([Z], observed=W))[Z]:
+                    instruments.append(Z)
+                    conditionals.append(W)
+                    # instruments.extend(list(W))
+                else:
+                    continue
+            if bool(instruments):
```
ankurankan: I would suggest using `len` instead of `bool` here. It's better to be explicit about what we are checking.

pgmpy/tests/test_identification/test_instrument.py
ankurankan, line 1: Does this already include the examples from the paper?

ankurankan (changes requested):

pgmpy/identification/instrument.py
DARHWOLF: @ankurankan this is to check for the edge case 
> We might need to handle the edge case here that if node is also the exposure variable, we shouldn't accidentally assign the outcome variable as its scaling indicator. Basically, avoid the condition that we check in lines 75-80.

pgmpy/identification/instrument.py
DARHWOLF: Thanks I have updated this. 
Since this is a useful feature, would it be a good idea to have examples in the docstrings about this, so that users don't miss out

ankurankan (changes requested):
@DARHWOLF I had a look at the implementation (partially). I think a few things are incorrect:
1. The scaling indicator is only required when either the exposure or outcome variables are latent. I have fixed the `_get_scaling_indicator` method to have the correct behaviour. 
2. In the tests for scaling indicators, observed variables were being used as exposure and outcome. I have changed these to use the latent variables.
3. I think the `_iv_transformations` method is incorrect currently. For example, there should be no need to add the edge in line 32. I don't see any edges getting added when the exposure variable (X) is latent. Could you have another look at the algorithm in the paper and double check the implementation?
4. I would also suggest installing a spelling checker in your IDE. I am seeing some spelling mistakes in the error messages and docstrings.

pgmpy/identification/instrument.py
```diff
@@ -0,0 +1,271 @@
...
+        variant="non-conditional",
+        scaling_indicators=None,
+    ) -> None:
+        self.supported_graph_types = (DAG,)
+        self.variant = variant.lower()
+        self.scaling_indicators = (
+            dict() if scaling_indicators is None else scaling_indicators
+        )
+
+    def _get_scaling_indicators(self, causal_graph):
```
DARHWOLF, line 72: comment

PR #2434, opened 2025-11-27
[ENH] Implement a SimpleCausalModel class

### LINKED ISSUE #2426: [ENH] Implement a `SimpleCausalModel` class
Currently, the graphical representation of the causal model is handled through the `DAG` class, along with variable roles (such as exposure, outcome, and adjustment). However, in the case of causal estimation/prediction, users usually do not care about the full causal structure but are rather only interested in the simple causal model case, i.e., a causal model with an edge b/w exposure and outcome, and edges between adjustment and exposure/outcome variables.

In the current implementation of the `DAG` class, users have to specify each of these edges, which can be cumbersome. A possible way to simplify the specification of such models could be to create a new `SimpleCausalModel` class which is a wrapper around the `DAG` class and provides methods for easy specification of these models. A potential API could be like:

```python
class SimpleCausalModel(DAG):
     def __init__(exposure: int or str or iterable,
                          outcome: same^,
                          adjustment: same^,
                          instruement: same^
                          ...)
           # Initialze a DAG model using the variables (if int generate random variable names) and the fixed structure. 
```

Aniketsy:
Fixes #2426 

Please let me know if my approach or fix needs any improvements . I’m open to feedback and happy to make changes based on suggestions.
Thankyou !


pgmpy/base/DAG.py
ankurankan, line 1: Would suggest putting this class in a separate file.

pgmpy/base/DAG.py
```diff
@@ -1925,3 +1925,46 @@ def __hash__(self):
                 ),
             )
         )
+
+
+class SimpleCausalModel(DAG):
+    """
+    A specialized DAG class for simple causal models.
```
ankurankan: The docstring needs to be expanded quite a bit. It should also have Parameters and Examples sections. Refer other classes for examples.

pgmpy/base/DAG.py
```diff
@@ -1925,3 +1925,46 @@ def __hash__(self):
...
+
+    """
+
+    def __init__(
+        self, exposure, outcome, adjustment=None, instrument=None, latents=None
```
ankurankan: `covariates` seems like a more appropriate name instead of `adjustment`.

pgmpy/base/DAG.py
```diff
@@ -1925,3 +1925,46 @@ def __hash__(self):
...
+    ):
+        super(SimpleCausalModel, self).__init__(latents=latents)
+
+        def to_list(var):
+            if isinstance(var, str):
+                return [var]
+            elif isinstance(var, int):
+                return [f"Var_{var}"]
```
ankurankan: This behavior should be explained in the docstring.

pgmpy/base/DAG.py
```diff
@@ -1925,3 +1925,46 @@ def __hash__(self):
...
+            elif var is None:
+                return []
+            return list(var)
+
```
ankurankan: We should not define functions inside methods unless absolutely necessary.

pgmpy/base/DAG.py
ankurankan: We should also add a `mediators` argument here. If mediators are present, the edges would be: `exposure -> mediator` and `mediator -> outcome`

ankurankan: Would suggest using plural term for each of them as we are allowing multiple variables for each.

pgmpy/base/DAG.py
```diff
@@ -1925,3 +1925,46 @@ def __hash__(self):
...
+        for adj in self.adjustment:
+            for exp in self.exposure:
+                self.add_edge(adj, exp)
+            for out in self.outcome:
+                self.add_edge(adj, out)
+
+        # Add edges from instrument variables to exposure variables
+        for inst in self.instrument:
+            for exp in self.exposure:
+                self.add_edge(inst, exp)
```
ankurankan: In addition to adding the edges, we need to define the variable roles as well. Have a look at the `__init__` method of the `DAG` class for examples. I would suggest restructuring the method to first construct the list of edges that would be present in the DAG, and then, at the end, do a `super` call using the constructed edge list and variable roles.

pgmpy/base/DAG.py
ankurankan: Can we add type hinting here?

pgmpy/tests/test_base/test_DAG.py
```diff
@@ -2216,3 +2216,73 @@ def test_sebastiani_round_trip_comparison(self):
             self.assertIn(
                 expected_edge_str, dagitty_output, f"Missing edge: {expected_edge_str}"
             )
+
+
+class TestSimpleCausalModel(unittest.TestCase):
```
ankurankan: Please put this in a new test file. Also, try to use pytest style tests, as we are trying to move from unit tests and are slowly refactoring the tests.

ankurankan: The tests will also need to check that the roles for the variables are correctly set.

ankurankan (changes requested):
@Aniketsy Thanks for the PR. I have added a few comments, please have a look.

Aniketsy:
@ankurankan Thanks for the review. I will update the PR as per your suggestions.

Aniketsy:
@ankurankan  could you please review these changes? I’m getting one test failure and would appreciate your guidance.
Also, I’m facing some difficulty running the tests locally. Could you please guide me on the correct setup or point me to the relevant documentation?

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+class SimpleCausalModel(DAG):
+    """
+    A specialized DAG class for simple causal models.
+
+    This class simplifies the creation of causal graphs commonly used in causal inference,
+    where the structure consists of exposures, outcomes, covariates (confounders), mediators, and instruments.
+    It automatically adds the standard edges:
+        - Exposures -> Outcomes
```
ankurankan: This is not true; this edge is only added if there are no mediators. 

Is it possible to show an ASCII figure here showing the graph that would be created?

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+        - Covariates -> Exposures
+        - Covariates -> Outcomes
+        - Instruments -> Exposures
+        - Exposures -> Mediators
+        - Mediators -> Outcomes
+
+    Parameters
+    ----------
+    exposures: str, int, or iterable
+        The exposure variable(s). If an int is provided, the variable name will be generated as "Var_<int>".
```
ankurankan: Would suggest encoding the variable names depending on their types. Exposures can be "E_<int>", outcomes can be "O_<int>", covariates can be "X_<int>", and so on. It would make it easier to identify what is the role of each node.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+        The instrumental variable(s). If an int is provided, the variable name will be generated as "Var_<int>".
+
+    latents: iterable or None (default: None)
+        List of latent variables.
+
+    Notes
+    -----
+    If any of the `exposures`, `outcomes`, `covariates`, `mediators`, or `instruments` arguments are
+    provided as integers, they will be automatically converted to variable names in the format
+    "Var_<int>".
```
ankurankan: Not required; already explained above.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Examples
+    --------
+    >>> from pgmpy.base.simple_causal_model import SimpleCausalModel
```
ankurankan: Add `SimpleCausalModel` to `pgmpy.base.__init__` and simplify the import path.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+    >>> model = SimpleCausalModel(
+    ...     exposures="X", outcomes="Y", covariates="Z", mediators="M", instruments="I"
+    ... )
+    >>> model.edges()
+    OutEdgeView([('X', 'Y'), ('Z', 'X'), ('Z', 'Y'), ('I', 'X'), ('X', 'M'), ('M', 'Y')])
+        If any of the `exposures`, `outcomes`, `covariates`, `mediators`, or `instruments` arguments are
```
ankurankan: Any text inside code blocks requires one blank line above and below. Refer to numpydoc style documentation for details: https://numpydoc.readthedocs.io/en/latest/format.html

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+        outcomes = self._to_list(outcomes)
+        covariates = self._to_list(covariates)
+        mediators = self._to_list(mediators)
+        instruments = self._to_list(instruments)
+        latents = list(latents) if latents is not None else []
+
+        edges = []
+
+        # Add edges from exposures to outcomes
+        edges += [(exp, out) for exp in exposures for out in outcomes]
```
ankurankan: This should only be done if there are no mediators. If there are mediators, it should be exposure -> mediator -> outcome.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+        edges += [(cov, out) for cov in covariates for out in outcomes]
+
+        # Add edges from instruments to exposures
+        edges += [(inst, exp) for inst in instruments for exp in exposures]
+
+        # Add edges from exposures to mediators and mediators to outcomes
+        edges += [(exp, med) for exp in exposures for med in mediators]
+        edges += [(med, out) for med in mediators for out in outcomes]
+
+        super().__init__(edges)
```
ankurankan: You can pass latents and variable roles here itself. Have a look at the signature of `DAG.__init__`.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,140 @@
...
+        ----------
+        role: str
+            The role to query (e.g., "exposure", "outcome", "covariate", "mediator", "instrument").
+
+        Returns
+        -------
+        set
+            A set of variable names associated with the role.
+        """
+        return self.variable_roles.get(role, set())
```
ankurankan: This class inherits `DAG` class which inherits `_GraphRolesMixin`. Since `_GraphRolesMixin` already has `get_role` method, it should automatically work with this class as well.

pgmpy/base/SimpleCausalModel.py
```diff
@@ -0,0 +1,140 @@
...
+            ('Var_5', 'Var_1'), ('Var_1', 'Var_4'), ('Var_4', 'Var_2')])
+    """
+
+    @staticmethod
+    def _to_list(var):
+        if var is None:
+            return []
+        elif isinstance(var, str):
+            return [var]
+        elif isinstance(var, int):
```
ankurankan, line 106: I don't think the behavior here is correct. When an integer is passed, we should create that many variables with that role.

ankurankan (changes requested):

Aniketsy:
Thanks for the detailed review, @ankurankan , I hope I'm not creating extra work for you. I’ve applied the changes as per your suggestions. Please let me know if anything needs further improvement.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,163 @@
...
+    It automatically adds the standard edges:
+        - Exposures -> Outcomes (only if there are no mediators)
+        - Covariates -> Exposures
+        - Covariates -> Outcomes
+        - Instruments -> Exposures
+        - Exposures -> Mediators
+        - Mediators -> Outcomes
+
+    Diagram
+    -------
```
ankurankan: I am not sure but I don't think we can have `Diagram` section. I don't know how Sphinx would parse it.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,163 @@
...
+    If mediators are present:
+        Covariates   Instruments
+           |             |
+           v             v
+        Exposures ---> Mediators
+           |              |
+           v              v
+        Outcomes <--------
```
ankurankan: This doesn't look correct: Exposures to outcomes edges should not be present. A more standard way to draw these models is to place the instrument, exposure, mediator, and outcome in a single line.

pgmpy/base/SimpleCausalModel.py
```diff
@@ -0,0 +1,163 @@
...
+
+    This class simplifies the creation of causal graphs commonly used in causal inference,
+    where the structure consists of exposures, outcomes, covariates (confounders), mediators, and instruments.
+    It automatically adds the standard edges:
+        - Exposures -> Outcomes (only if there are no mediators)
+        - Covariates -> Exposures
+        - Covariates -> Outcomes
+        - Instruments -> Exposures
+        - Exposures -> Mediators
+        - Mediators -> Outcomes
```
ankurankan, line 18: I would suggest adding a line here saying that if you want to have more control on the model structure, use the `DAG` class.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,163 @@
...
+    If no mediators:
+        Covariates   Instruments
+           |             |
+           v             v
+        Exposures ------> Outcomes
+
+    Parameters
+    ----------
+    exposures: str, int, or iterable
+        The exposure variable(s). If an int ``n`` is provided, ``n`` variables
```
ankurankan: Shouldn't the `n` be within single quote? What does using double quotes do?

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,163 @@
...
+            }.get(role, "Var_")
+            return [f"{prefix}{i}" for i in range(var)]
+        elif isinstance(var, Iterable):
+            prefix = {
+                "exposures": "E_",
+                "outcomes": "O_",
+                "covariates": "X_",
+                "mediators": "M_",
+                "instruments": "I_",
+            }.get(role, "Var_")
```
ankurankan: If the variables are an iterable (i.e., the names of the variables are provided), we shouldn't add prefix.

pgmpy/base/simple_causal_model.py
```diff
@@ -0,0 +1,163 @@
...
+            "covariates": set(covariates),
+            "mediators": set(mediators),
+            "instruments": set(instruments),
+        }
+        latents_set = set(latents) if latents else set()
+        super().__init__(edges, latents=latents_set, roles=roles)
+
+        if latents_set:
+            self.add_nodes_from(latents_set)
+            self.latents = latents_set
```
ankurankan: Why is this needed? Aren't the latents getting added in line 158?

pgmpy/tests/test_base/test_simple_causal_model.py
```diff
@@ -0,0 +1,124 @@
...
+    assert set(model.get_role("exposures")) == {"X1", "X2"}
+    assert set(model.get_role("outcomes")) == {"Y1", "Y2"}
+    assert set(model.get_role("covariates")) == {"Z"}
+    assert set(model.get_role("mediators")) == {"M"}
+    assert set(model.get_role("instruments")) == {"I"}
+
+
+def test_integer_variables():
+    model = SimpleCausalModel(
+        exposures=1, outcomes=2, covariates=3, mediators=4, instruments=5
```
ankurankan: Can you add some latents here?

ankurankan (changes requested):
Mostly looks good now. I have added a few minor comments.

pgmpy/base/simple_causal_model.py
Aniketsy: I added this because the DAG class requires all latents to be present as nodes, otherwise, it raises a `ValueError` during initialization. please correct me if i'm wrong.

Aniketsy:
@ankurankan  Thanks for the review! I've updated the diagram section as you mentioned. Please take a look and let me know if anything else needs improvement. all checks are green now.

ankurankan:
@Aniketsy Looks good to me. I have added a commit with the following changes:
1. Renames the files as we use CamelCase for file names.
2. Adds the SimpleCausalGraph to `base/__init__.py` file to simplify import paths.
3. Modifies the behaviour of how latents are handled. Usually, existing nodes in the graph are specified as latent variables. So, the change now throws an error if the latent variable doesn't already exist in the graph instead of adding them.

ankurankan (dismissed):

Aniketsy:
@ankurankan Thanks! for the changes and improvements.

PR #2444, opened 2025-12-05
[ENH] Support node-specific priors in EM (Dirichlet/ BDeu) (#2432)

### LINKED ISSUE #2432: [ENH] Support node-specific priors in EM (Dirichlet/ BDeu)
I have a feature request regarding ExpectationMaximization. When using apply_smoothing=True, the method internally calls BayesianEstimator.estimate_cpd().

Is there currently a way to pass node-specific Dirichlet pseudo-counts through EM.get_parameters()? I’d like to use different pseudo_counts per node.

Right now I’m using the BDeu prior with equivalent_sample_size = 1, and I locally modified the library so that

   _alpha = float(equivalent_sample_size) / node_cardinality_

instead of using a global value. 

  - adding support for node-specific priors or a mapping-based ESS (e.g. per-node ESS) in the estimator
  - after discussing this with Ankur, he suggested to modify the equivalent_sample_size argument to accept a dictionary of the form 
  
      _{node: equivalent_sample_size}_

Haley-1617:

- Fixes #2432 

- Allow parameter - equivalent_sample_size in BayesianEstimator().estimate_cpd() method to take in dict type, supporting node-specific prior in EM
- Added tests in test_ BayesianEstimator() and test_EM()
-

### DIFF
--- a/pgmpy/estimators/BayesianEstimator.py
+++ b/pgmpy/estimators/BayesianEstimator.py
@@ -149,7 +149,9 @@ class BayesianEstimator(ParameterEstimator):
         node: Hashable,
         prior_type: str = "BDeu",
         pseudo_counts: Union[List[List[float]], np.ndarray, float, int] = [],
-        equivalent_sample_size: Union[int, float] = 5,
+        equivalent_sample_size: Union[
+            int, float, Dict[Hashable, Union[int, float]]
+        ] = 5,
         weighted: bool = False,
     ) -> TabularCPD:
         """
@@ -232,7 +234,12 @@ class BayesianEstimator(ParameterEstimator):
         if prior_type == "k2":
             pseudo_counts = np.ones(cpd_shape, dtype=int)
         elif prior_type == "bdeu":
-            alpha = float(equivalent_sample_size) / (
+            equivalent_sample_size_val = (
+                equivalent_sample_size.get(node, 0)
+                if isinstance(equivalent_sample_size, dict)
+                else equivalent_sample_size
+            )
+            alpha = float(equivalent_sample_size_val) / (
                 node_cardinality * np.prod(parents_cardinalities)
             )
             pseudo_counts = np.ones(cpd_shape, dtype=float) * alpha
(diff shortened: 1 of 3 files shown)


ankurankan (changes requested):
@Haley-1617 Thanks for the PR. I think we should not change the `BayesianEstimator.estimate_cpd` directly. As users usually interact with the `*Estimator.get_parameters` method, I think the option to specify the dictionary of `equivalent_sample_size` should go there. We should keep the `estimate_cpd` method as it is, and the logic for passing the correct `equivalent_sample_size` to `estimate_cpd` should be handled in the `get_parameters` method. I would suggest the following changes:

1. Modify `BayesianEstimator.get_parameters` method to accept a dictionary as the `equivalent_sample_size`. Add logic in the `get_parameters` such that it select the correct value from `equivalent_sample_size` if it is a dict before calling `estimate_cpd`.
2. Similary modify `ExpectationMaximization.get_parameters` to accept a dict and add logic to pass the correct `equivalent_sample_size` value in the `estimate_cpd` call.

Haley-1617:
@ankurankan Thank you for the feedback. I really appreciate it! I think I missed the fact that users usually interact with `get_parameter` rather than `estimate_cpd`, so I should change `get_parameter` instead. 
One more question I have is: do we also need to add `prior_type` and `pseudo_counts` to `EM.get_parameter` as well? Since we also want to support node-specific priors for Dirichlet, which is handled by `pseudo_counts`.
Thank you so much!

ankurankan:
@Haley-1617 
> One more question I have is: do we also need to add prior_type and pseudo_counts to EM.get_parameter as well? 

I think the current `EM.get_parameters` should automatically handle those arguments because we pass a `**kwargs` argument when calling `estimate_cpd` (line 397) ?

I also just noticed that the `BayesianEstimator.get_parameters` already accepts a dict for `equivalent_sample_size`. I think there are two possible ways to allow dicts for `equivalent_sample_size` in `EM.get_parameters`.
1. We either handle the logic in `EM.get_parameters` such that we pass an int when calling `estimate_cpd`.
2. We start accepting dict as `equivalent_sample_size` in `BayesianEstimator.estimate_cpd`. This is what you have implemented in this PR.

I think the second option is better because it doesn't create dependency between the two classes. 

Could you please clarify this a bit in the `estimate_cpd` docstring (maybe add a Notes section)? I think the rest of the PR looks good.

Haley-1617:
Got it, option 2 sounds good to me. I have updated the doc string accordingly. Thank you!

ankurankan (approved):
@Haley-1617 Looks good to me. Thanks for the PR.

Issue #2442, opened 2025-12-04
[ENH] Add doctests in the pre-commit hook


ankurankan:
Check if it is possible to create a pre-commit hook that checks whether the doctests of the files that are being committed pass.

ankurankan:
Fixed in #2446

PR #2446, opened 2025-12-05
[ENH] Add doctest pre-commit hook and implementation script

whyvineet:
# Add doctest validation to pre-commit  
Fixes **#2442**

## Summary
Adds a doctest validation step to pre-commit to ensure all doctest examples remain correct before commits.

## Key Changes
- Added `run_doctests.py` to run doctests on staged Python files.  
- Updated `.pre-commit-config.yaml` with a new doctest hook.

## How it works
Runs automatically on commit, checks doctests in staged `.py` files, and blocks the commit if any fail.

## Testing
```bash
pre-commit run doctest --all-files  
pre-commit run doctest

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,3 +23,12 @@ repos:
     hooks:
     - id: blackdoc
       additional_dependencies: ["black==24.8.0"]
+  - repo: local
+    hooks:
+      - id: doctest
+        name: Run doctests
+        entry: python run_doctests.py
+        language: system
+        files: \.py$
+        pass_filenames: true
+        stages: [commit]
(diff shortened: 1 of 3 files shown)


ankurankan (changes requested):
@whyvineet Thanks for the PR. It seems to work well. A couple of changes I would suggest:
1. Rename `run_doctests.py` to something like `_doctests_precommit.py`. This will make it a bit more obvious what the file is.
2. Expand the documentation in the `run_doctests.py` file a bit to mention how the file/script is getting used.

whyvineet:
@ankurankan Implemented all the suggested changes. Please let me know if anything else needs refinement.

ankurankan (approved):
@whyvineet Looks good. Thanks for the PR :)

Issue #2386, opened 2025-10-17
[BUG] Expectation Maximization (EM) estimator fails with an IndexError


nihilee:
**Describe the bug**
The Expectation Maximization (EM) estimator fails with an `IndexError` when the dataset contains missing values represented by `numpy.nan`, despite the documentation explicitly stating that "If some values in the data are missing the data cells should be set to `numpy.nan`".

The error occurs because when pandas DataFrames contain `np.nan` in integer columns, pandas automatically converts the entire column to `float64` (e.g., `0` becomes `0.0`). During the EM algorithm's E-step, these float values are used as indices to access CPD arrays, but NumPy arrays only accept integer indices, causing an `IndexError`.

**To Reproduce**

```python
import numpy as np
import pandas as pd
from pgmpy.models import DiscreteBayesianNetwork
from pgmpy.estimators import ExpectationMaximization as EM

# Set seed for reproducibility
np.random.seed(42)

# Create dataset with integer values
data = pd.DataFrame(
    np.random.randint(low=0, high=2, size=(1000, 3)),
    columns=["A", "C", "D"],
)

# Add missing values (10% of data) using np.nan as documented
missing_rate = 0.1
mask = np.random.random(data.shape) < missing_rate
data = data.mask(mask)

# Check data types - notice all columns are now float64 due to np.nan
print(f"Data types after adding np.nan: {data.dtypes.to_dict()}")
# Output: {'A': dtype('float64'), 'C': dtype('float64'), 'D': dtype('float64')}

print(f"Missing values:
{data.isnull().sum()}")
print(f"Sample values in column A: {data['A'].head(10).tolist()}")
# Output shows values like [0.0, 1.0, nan, 1.0, ...] instead of [0, 1, nan, 1, ...]

# Define Bayesian Network with latent variable
model = DiscreteBayesianNetwork(
    [("A", "B"), ("C", "B"), ("C", "D")], 
    latents={"B"}
)

# Run EM - this fails with IndexError
estimator = EM(model, data)
params = estimator.get_parameters(latent_card={"B": 3})
```

**Error output**
IndexError: only integers, slices (`:`), ellipsis (`...`), None and long or byte Variables are valid indices (got float)

**Versions**
1.0.0

Aniketsy:
I’d love to look into this. If you’re okay with it, please let me know.

ankurankan:
[...]

PR #2447, opened 2025-12-05
[ENH] Add Support for Randomly Missing Values in EM Algorithm

### LINKED ISSUE #2445: [ENH] Add Support for Randomly Missing Values in EM Algorithm
**Description**
The current EM algorithm in `pgmpy` does not support datasets that contain randomly missing values (sparse NaN entries across rows). Instead, it is designed to handle only cases where entire variables/columns are missing.

**Enhancement**

- Detect missing variables for each data point dynamically.
- Support datasets with sparse, randomly distributed missing values.
- Generate latent combinations and corresponding weights based on model likelihood to properly handle partial missing data.

Aniketsy:
fix #2445 #2386


pgmpy/tests/test_estimators/test_EM.py
```diff
@@ -347,3 +348,25 @@ def tearDown(self):
         get_reusable_executor().shutdown(wait=True)
 
         config.set_backend("numpy")
+
+
+class TestExpectationMaximization(unittest.TestCase):
```
ankurankan, line 421: Ideally, the test should start by simulating some data using known CPDs, then introduce some random NaN values in the data, and test whether the implementation is able to recover the original values (up to some tolerance limits). Please have a look at the other tests for example.

pgmpy/estimators/EM.py
```diff
@@ -346,12 +364,12 @@ def get_parameters(
             estimator = BayesianEstimator.__new__(BayesianEstimator)
```
ankurankan, line 371: I think the logic in Step 3.1 of the algorithm will need to be modified, as it currently fixes all CPDs that do not involve latent variables. However, with randomly missing values, non-latent variables can also have missing values, and we should not be fixing the CPDs for them.

pgmpy/estimators/EM.py
```diff
@@ -137,19 +136,27 @@ def _parallel_compute_weights(
...
-            latent_combinations = np.array(v, dtype=int)
-            df = data_unique.iloc[[i] * latent_combinations.shape[0]].reset_index(
-                drop=True
-            )
-            for index, latent_var in enumerate(latent_card.keys()):
-                df[latent_var] = latent_combinations[:, index]
+            missing_vars = [
+                var
+                for var in latent_card.keys()
+                if var not in data_unique.columns
```
ankurankan: I don't think we should check whether var is not in `data_unique.columns`. This line is going to get executed `n_samples * max_iter` times. Its better to do this check in the `__init__` method if it doesn't already exists.

pgmpy/estimators/EM.py
```diff
@@ -113,13 +103,22 @@ def _get_log_likelihood(self, datapoint: Dict[str, Any]) -> float:
...
                         **{
-                            key: value
+                            key: (
+                                int(value)
+                                if not pd.isna(value)
+                                and isinstance(
+                                    value, (int, float, np.integer, np.floating)
+                                )
+                                else value
+                            )
```
ankurankan: Why is this change required? Doesn't the logic in `_parallel_compute_weights` already make sure that there are no Nan values before the `_get_log_likelihood` call?

pgmpy/tests/test_estimators/test_EM.py
```diff
@@ -192,7 +193,7 @@ def test_em_init_missing_data_handling(self):
             est = EM(self.model1, df)
 
         # Data shape and column removal
-        self.assertEqual(est.data.shape, (2, 3))
+        self.assertEqual(est.data.shape, (3, 3))
```
ankurankan, line 196: Why did this value change?

ankurankan (changes requested):
@Aniketsy Thanks for the PR, I think there are some mistakes in the implementation. Please have a look at my comments. I also feel that the PR is generated using some help from an AI (I am seeing some unnecessary changes and checks that are not required; AI tools typically add those). If that is the case, I would suggest to be a bit more careful with the changes, and try to understand the algorithm a bit more deeply so that we don't end up merging something that outputs incorrect results.

Aniketsy:
@ankurankan Thanks for pointing it out. I made some changes using an llm, and while testing locally I ran into two failing tests. I think those changes were unnecessary. I’ll go through the algorithm again in depth and correct the implementation.

pgmpy/tests/test_estimators/test_EM.py
Aniketsy: Previously, any row containing a `NaN` was removed, now only columns where all values are NaN are dropped. Please correct me if i’m misunderstanding anything.

ankurankan:
@Aniketsy, is this ready for review?

pgmpy/estimators/EM.py
```diff
@@ -113,11 +106,19 @@ def _get_log_likelihood(self, datapoint: Dict[str, Any]) -> float:
         likelihood = 0
         for cpd in self.model_copy.cpds:
             scope = set(cpd.scope())
+            if any(pd.isna(datapoint.get(var)) for var in scope):
+                continue
```
ankurankan: Don't think this is correct. There shouldn't be any missing values in this method. The missing values should have been filled in before being passed here.

pgmpy/estimators/EM.py
```diff
@@ -137,19 +138,26 @@ def _parallel_compute_weights(
...
-            df = data_unique.iloc[[i] * latent_combinations.shape[0]].reset_index(
-                drop=True
-            )
-            for index, latent_var in enumerate(latent_card.keys()):
-                df[latent_var] = latent_combinations[:, index]
+            missing_vars = [
+                var
+                for var in latent_card.keys()
+                if pd.isna(data_unique.iloc[i].get(var))
+            ]
```
ankurankan: This is incorrect, `missing_vars` should also consider the variables that have missing values, not just the ones that are specified in `latent_card`, unless you preprocess the `latent_card` to contain those as well.

ankurankan (changes requested):
@Aniketsy I think the algorithm implementation is incorrect. It currently drops the (partial) row whenever it encounters Nan values. The EM algorithm shouldn't do that; instead it should iterate over all possible states of the missing values and compute a weighted dataset using it. 

If you are using an LLM to generate the code, please stop doing it if you don't fully understand the algorithm. It puts the burden of verifying the implementation on the reviewer and it takes a lot of time to review such PRs when the mistakes are cleverly hidden by the LLM. Please first try to understand what the algorithm is supposed to do. I would be happy to explain the algorithm in any of our community meetings.

Aniketsy:
@ankurankan I was planning to do in weekend, this was not ready, and i apologize I thought this was in draft, and I was removing changes which  I had done previously as i mentioned. I took the help of  llm earlier to fix two failures in which I was stuck.
sure I will ask to explain the algorithm in which part I have doubt or not understand.

Aniketsy:
Also i would like to apologize for creating extra work and causing any inconvenience.

Aniketsy:
@ankurankan I've reverted the unnecessary changes I introduced earlier and spent some time reviewing the implementation again. I would also like to understand the algorithm in more depth in any meeting, I’ve already joined the discord server, please let me know when it might be possible.

> I think the logic in Step 3.1 of the algorithm will need to be modified, as it currently fixes all CPDs that do not involve latent variables. However, with randomly missing values, non-latent variables can also have missing values, and we should not be fixing the CPDs for them.

I haven’t covered this part yet. I want to understand the algorithm more clearly before making changes there. I’ll explore this next, and if I’m confident about implementing it correctly, I’ll update the PR.

Aniketsy:
> I think the logic in Step 3.1 of the algorithm will need to be modified, as it currently fixes all CPDs that do not involve latent variables. However, with randomly missing values, non-latent variables can also have missing values, and we should not be fixing the CPDs for them.

@ankurankan for this i came up with this, please let me know if this needs improvement.
identify all variables in the data that have any missing values and collect them in `vars_with_missing`.
and then while building `fixed_cpd_vars`, we subtract both
- all variables with missing values
- all children of those variables

ankurankan:
@Aniketsy 

>  I would also like to understand the algorithm in more depth in any meeting, I’ve already joined the discord server, please let me know when it might be possible.

We have dev meetings on Mondays and community meetings on Thursdays. You would see them in the events (top left) on discord.

> identify all variables in the data that have any missing values and collect them in vars_with_missing.
and then while building fixed_cpd_vars, we subtract both

This would be the first step. Additionally, in the `_parallel_compute_weights` method, we expand the dataframe by adding rows corresponding to each possible value of the missing variable (line 142). Since now we want to support randomly missing values, we will have to check for each row here which values are missing and based on that expand the dataframe. I also think that checking for every row might slow down the algorithm, so it might be a good idea to precompute a boolean array that has True corresponding to each missing value, and this might give us a faster way to look up the missing variables in each row.

Aniketsy:
@ankurankan I’ve applied the changes based on your suggestions. could you please take a look and let me know if I’m heading in the right direction or if there’s anything that needs improvement.

ankurankan:
Tests are failing. Converting this to draft.

Aniketsy:
I'll revisit this shortly and fix the issues. Thanks!

PR #2458, opened 2025-12-13
[ENH] Add LRU caching to CI tests

### LINKED ISSUE #2435: [ENH] Cache conditional independence (CI) test results
The `pgmpy.estimators.CITests.py` implements a few conditional independence tests. In some scenarios (for example, permutation tests #2119), the same function call is made multiple times. Implementing a caching mechanism can help make a significant improvement in the runtime.

A simple way to implement this could be to decorate each of the CI tests using the `functools.lru_cache` decorator. Changes required:

1. `lru_cache` requires arguments to be hashable. One of the arguments of CI tests is a dataframe that is not hashable. See if we can come up with a workaround for this.
2. Add the `lru_cache` decorator to CI tests.
3. Define a variable in `pgmpy.global_vars.py` to define the size of the cache.

kajal-jotwani:

- Fixes #2435 

- Implemented `_HashableDataFrame` wrapper to allow DataFrames to be hashed by object ID.
- Added `_cached_ci_test` decorator to handle unhashable arguments (lists, dicts) automatically.
-


ankurankan:
@kajal-jotwani The tests seem to be failing. Could you have a look please?

kajal-jotwani:
The cached function was previously defined locally, which caused pickling issues with joblib. I’ve fixed this now.

ankurankan:
@kajal-jotwani Sorry for the delay in reviewing this. I think the hashing of dataframes makes the current implementation quite complicated and will also be quite slow, considering that causal discovery algorithms call CI test functions thousands of times in each run. Sorry, I wasn't able to foresee this issue coming up.

I think a better solution might be to refactor the CI tests into a class structure where the data is passed to the class' '__init__` method along with any hyperparameter. This would allow us to only cache the actual CI test call. The API would be something like this:

```python
class SomeCITest:
      def __init__(self, data, param1, param2, ...):
             self.data = data
             self.param1 = param1
            .....

     @lru_cache(MAX_CACHE_SIZE)
     def test(X, Y, Z):
             # Run the test logic.
```

What do you think?

kajal-jotwani:
@ankurankan thanks for the feedback, this makes sense. I had tried the caching approach earlier, but it didn’t lead to any significant runtime improvement, and I couldn’t find a clean or efficient way to hash the DataFrame without adding extra overhead. I agree that refactoring the CI tests into a class-based structure would be better, since we wouldn’t be caching the DataFrame and could focus only on caching the actual (X, Y, Z) test calls. I’m happy to implement this approach.

ankurankan:
@kajal-jotwani Thanks. Could you please create a new issue to refactor the CI tests? I think we should follow a similar approach to what we did to our metrics: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/metrics/. Essentially, the following steps:

1. Create a new folder for ci tests. Create a `_base.py` and define a `_BaseCITest` class in it that inherits `skbase.BaseObject`. This would automatically provide the registry functionality that we have implemented using `CITestRegistry`. 
2. Separate each CI test in its own file. Make each CI test a class that inherits `_BaseCITest`. The init method should take all the parameters for the tests (including data), and then the actual test method would be uniform for each CI test. 
3. Add some tags to each CI test class. This would require us to move the current registry arguments to tags.

> . I had tried the caching approach earlier, but it didn’t lead to any significant runtime improvement, 

In the case of causal discovery algorithms, we would see much improvments as each CI test gets called only once. But for some metrics, we do repeat the calls, so it would be beneficial in that case.

kajal-jotwani:
> @kajal-jotwani Thanks. Could you please create a new issue to refactor the CI tests? I think we should follow a similar approach to what we did to our metrics: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/metrics/. Essentially, the following steps:

I’ve created the new [issue](https://github.com/pgmpy/pgmpy/issues/2547) for the CI test refactor as suggested.  Just to confirm, should the caching be implemented after the refactor, or can I work on it in this PR?

ankurankan:
@kajal-jotwani Thanks. I would suggest doing the refactor first and then implementing the caching to minimize the amount of changes in a single PR.

kajal-jotwani:
Okay, I’ll work on the refactor first. Thanks!

ankurankan:
@kajal-jotwani I think this is not relevant anymore? Will need to have the caching for the new refactored CI tests.  Closing

Issue #2459, opened 2025-12-15
[ENH] Add datasets


ankurankan:
We recently added a dataset module that supports loading datasets (and their metadata) from https://github.com/pgmpy/example-causal-datasets/tree/main/real
. Each new dataset currently requires a separate file defining its dataset class, so we haven’t added all of them yet. The extension template for adding new datasets is at: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/extension_templates/_dataset.py

This issue is intended to track which datasets still need to be added, split the work across contributors, and prevent duplicate effort. Before you start working on a dataset, please check it off in the list below.

List of the remaining datasets:

- [x] apple-watch-fitbit
- [x] auto-mpg #2462 
- [x] blue-driver
- [x] boston-housing: #2460 
- [x] cities
- [x] college-plans
- [x] contraceptive-method
- [x] covertype
- [x] credit-approval
- [x] cystic-fibrosis
- [x] depression-coping
- [x] dropouts
- [x] dry-bean
- [ ] fmri.9var: https://github.com/pgmpy/pgmpy/pull/2485
- [x] galton-stature
- [x] goldberg
- [x] hitters
- [x] htru2
- [ ] hungary-chickenpox: https://github.com/pgmpy/pgmpy/pull/2475
- [x] iq-brain-size
- [x] lead
- [x] myocardial-infarction-complications
- [x] pima-diabetes
- [x] pittsburgh-bridges
- [x] residential-building
- [x] seoul-bike
- [x] south-german-credit
- [x] spartina
- [x] student-performance
- [x] superconductivity
- [x] uscrime
- [x] wine-quality: #2463 
- [x] yatch-hydrodynamics

MAVRICK-1:
- apple-watch-fitbit https://github.com/pgmpy/pgmpy/pull/2464
- auto-mpg https://github.com/pgmpy/pgmpy/pull/2464
- blue-driver  https://github.com/pgmpy/pgmpy/pull/2464

MAVRICK-1:
- dropouts:  https://github.com/pgmpy/pgmpy/pull/2465
- dry_bean : https://github.com/pgmpy/pgmpy/pull/2465

MAVRICK-1:
-  galton-stature  :  https://github.com/pgmpy/pgmpy/pull/2466

MAVRICK-1:
- goldberg: https://github.com/pgmpy/pgmpy/pull/2467

[...]

PR #2477, opened 2025-12-17
feat: Add lead dataset

sabasiddique1:
- Refs #2459

## List of changes to the codebase in this pull request
- Added Lead dataset class (_lead.py)
- 7 variables, 221 samples
- Continuous data type (covariance matrix format)
- Custom load_dataframe method to parse covariance matrix
- No expert knowledge
- Updated __init__.py to import Lead
- Added lead to test file ALL_DATASETS list

## Description
What was added:

- New dataset file: `pgmpy/datasets/_lead.py`
- Custom implementation to handle covariance matrix format (not standard tabular data)
- Dataset registration in __init__.py
- Test coverage in test_datasets.py
- Special parsing for lower-triangular covariance matrix format


ankurankan:
@sabasiddique1 This is also a covariance matrix, will need to be handled separately.

ankurankan:
Note: The covariance matrix file is not `\t` seprated leading to an error. We should modify the datafile itself to make the space character uniform.

PR #2485, opened 2025-12-17
feat: Add fmri_9var dataset

sabasiddique1:
Refs #2459

## List of changes
- Added FMRI 9-Variable dataset class (_fmri_9var.py)
- 9 variables, 1440 samples
- Continuous data type
- No ground truth or expert knowledge
- Updated __init__.py to import FMRI9Var
- Added fmri_9var to test file ALL_DATASETS list

### What was added:

- New dataset file: `pgmpy/datasets/_fmri_9var.py`
- Dataset registration in `__init__.py`
- Test coverage in `test_datasets.py`


ankurankan:
@sabasiddique1 I think we need to first figure out what the different files in the dataset are, and based on that decide how to implement this. I don't see any references for the dataset in the repo.

PR #2499, opened 2025-12-18
Add cities dataset

MAVRICK-1:
This pull request introduces a new dataset class for "Cities" into the `pgmpy.datasets` module. The main changes involve registering and implementing the new dataset, making it available for use alongside existing datasets.

**New dataset addition:**

* Added a new file `pgmpy/datasets/_cities.py` which defines the `Cities` dataset class, including dataset metadata and URLs for data and ground truth.
* Registered the `Cities` dataset in the `pgmpy/datasets/__init__.py` file to make it importable from the datasets module.

refs #2459

### DIFF
--- a/pgmpy/datasets/_cities.py
+++ b/pgmpy/datasets/_cities.py
@@ -1,0 +1,30 @@
+from pgmpy.datasets import register_dataset_class
+from pgmpy.datasets._base import _BaseDataset, _CovarianceMixin
+
+
+@register_dataset_class
+class Cities(_CovarianceMixin, _BaseDataset):
+    name = "cities"
+
+    tags = {
+        "n_variables": 7,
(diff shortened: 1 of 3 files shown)


pgmpy/datasets/_cities.py
```diff
@@ -0,0 +1,30 @@
...
+from pgmpy.datasets._base import _BaseDataset, _CovarianceMixin
+
+
+@register_dataset_class
+class Cities(_CovarianceMixin, _BaseDataset):
+    name = "cities"
+
+    tags = {
+        "n_variables": 7,
+        "n_samples": 100,  # Typical sample size for covariance matrix datasets
```
ankurankan: This is not correct. The first line of the data file has the sample size; in this case, it is 164.

PR #2500, opened 2025-12-18
Add depression-coping dataset

MAVRICK-1:
This pull request adds support for a new dataset, "Depression Coping," to the `pgmpy.datasets` module. The main changes involve registering the new dataset and providing its metadata and download URL.

**New dataset integration:**

* Registered the new `DepressionCoping` dataset class in `pgmpy.datasets`, making it available for use like other datasets.
* Implemented the `DepressionCoping` class in `_depression_coping.py`, including dataset metadata (number of variables, samples, data type, etc.) and the data file URL.

refs #2459


pgmpy/datasets/_depression_coping.py
```diff
@@ -0,0 +1,30 @@
+from pgmpy.datasets import register_dataset_class
+from pgmpy.datasets._base import _BaseDataset
+
+
+@register_dataset_class
+class DepressionCoping(_BaseDataset):
+    name = "depression_coping"
+
+    tags = {
+        "n_variables": 70,  # Based on observation of many columns including STR1-21, DEP1-20, COP1-20, etc.
```
ankurankan: I would suggest removing this comment here. Have you checked this manually, or is this AI-generated? If AI generated, I would suggest loading the dataset and verifying the number of columns in the dataframe.

ankurankan (changes requested):
Tests are missing in this PR as well; could you please add them?

pgmpy/datasets/_depression_coping.py
MAVRICK-1: forgot to remove it , IT was copilot suggestion

MAVRICK-1:
@ankurankan done

PR #2503, opened 2025-12-21
[MNT] python 3.14 support

fkiraly:
Widens python support to 3.14:

* `pyproject` dependency range extended
* CI tests for 3.14

All dependencies are available under 3.14, and tests seem to pass.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -14,4 +14,4 @@ comment:
   require_changes: true
 codecov:
   notify:
-    after_n_builds: 6
+    after_n_builds: 9
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: ["3.10", "3.13"]
+        python-version: ["3.10", "3.13", "3.14"]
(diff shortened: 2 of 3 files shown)


daehyun99:
The `after_n_builds` setting in `codecov.yml` ensures that the Codecov report is generated only after the specified number of full dependency tests have completed.

You can enable full dependency tests for Python 3.14 by modifying line 63 of the `ci.yml` file as follows:
https://github.com/pgmpy/pgmpy/blob/197e1e0444c77c00581a4c32763811e5b03f8503/.github/workflows/ci.yml#L58-L63

```yaml
       python-version: ["3.10", "3.13", "3.14"]  # Line 63
```

Thanks.

fkiraly:
> You can enable full dependency tests for Python 3.14 by modifying line 63 of the ci.yml file as follows:

Not all soft dependencies might be compatible? Maybe let´s do this in a separate Pr?

I also updated the `after_n_builds`.

daehyun99:
> Not all soft dependencies might be compatible? Maybe let´s do this in a separate Pr?

Ah, I missed that part. I agree with your suggestion. Thank you.

ankurankan (approved):

ankurankan:
Thanks @fkiraly

PR #2504, opened 2025-12-25
[MNT] make `pyparsing` a softdependency

fkiraly:
`pyparsing` is only required for the `readwrite` module and can therefore be isolated as a soft dependency.

This PR carries out the isolation of `pyparsing` as an soft dependency, moving it to the `optional` dependency set.

There were already dependency checks in the code which seemed to assume it is a soft dependency, but the current `main` treats `pyparsing` as a core dependency.
The partial logic is replaced by full isolation using `scikit-base`.

### DIFF
--- a/pgmpy/readwrite/BIF.py
+++ b/pgmpy/readwrite/BIF.py
@@ -4,26 +4,8 @@ from itertools import product
 from string import Template
 
 import numpy as np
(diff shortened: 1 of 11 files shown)


fkiraly:
hm, this does not seem like a good idea - too much depends on this?

PR #2509, opened 2025-12-25
[MNT] replace `pgmpy` `_safe_import` with `scikit-base` instance of the same utility

fkiraly:
This PR replaces the `pgmpy` `_safe_import` with `scikit-base` instance of the same utility.

Since `pgmpy` depends on `scikit-base` the function (which was moved there recently)  can be imported from there; since no release of `pgmpy` has happened since the `_safe_import` utility was introduced, this switch can be made without needing to worry about user code or deprecation.


ankurankan (approved):
Thanks @fkiraly

Issue #2506, opened 2025-12-25
[ENH] Review and suggested changes regarding `extension_template` directory and design


fkiraly:
I came across the `extension_template` directory in `pgmpy`.

I would suggest:

* move it outside `pgmpy`, in root of the repo - this is not a functional class, and interesting only for developers, so it might sense to distribute it with the repo but not with the package.
* I recommed to use `BaseObject` from `scikit-base` which comes with its own tag system, this way you do not need to invent (or maintain) your own tag system
* there are a variety of parameters in the class that look like they could be tags? `name`, `data-url` etc. What is the rationale to not make them tags?

FYI @ankurankan, @Theavinash02

daehyun99:
> move it outside pgmpy, in root of the repo - this is not a functional class, and interesting only for developers, so it might sense to distribute it with the repo but not with the package.

Additionally suggest:
- what do you think about placing the utility files for library development in the `tools/` directory at the project root?

- example : `pgmpy/tools`
```sh
pgmpy/
├── tools/
│   ├── README.md
│   ├── benchmarks/
│   │   └── test_benchmark_joblib.py
│   └── extension_template/
│   │   └── _dataset.py.py
```

https://github.com/pgmpy/pgmpy/pull/2490#issuecomment-3677456981

Thank you.

Theavinash02:
Hi @fkiraly ,
Regarding the point `there are a variety of parameters in the class that look like they could be tags? name, data-url etc. What is the rationale to not make them tags?` ,
We have made the attributes like name and the *_url fields as class attributes because they define the dataset’s identity and how it is loaded. Here, tags are meant only for dataset properties used to filter and discover datasets. Since name and URLs directly affect the registry and data loading logic, keeping them separate from tags makes the design clearer.

Hi @daehyun99 ,
I agree that it would be better to move `extension_template` as suggested to `tools/` directory, as suggested.
Thank you.

Hi @ankurankan,
Could you please share your thoughts on how we could integrate the `Baseobject` from `scikit-base` in a way that aligns well with the current dataset API?

fkiraly:
> what do you think about placing the utility files for library development in the `tools/`

Good idea! Though the specific term "tools" feels a bit non-descriptive. How about `devtools`?

fkiraly:
@Theavinash02, I have opened a showcase PR with how a refactor could look like: https://github.com/pgmpy/pgmpy/pull/2512

More details in there, quick summary:

[...]

PR #2512, opened 2025-12-27
[ENH] showcase to refactor `BaseDataset` to `scikit-base` `BaseObject`

fkiraly:
Showcase to illustrate discussion in #2506 on refactoring `BaseDataset` to `BaseObject`.

The `BaseDataset` and its registry duplicate a lot of features in `scikit-base`.

* pretty printing is default for `BaseObject`
* the tag system is already managed via `_tags` attribute and `get_class_tags` and `get_class_tag` accessors.
* A registry is not necessary, since using `all_objects` from `scikit-base` acts like a registry that can be used to retrieve all datasets from the package. It only needs to be pointed to a directory to search.
    * this is not just for datasets but for any class. If/once we continue refactoring the estimators to `scikit-base`, it works for all types of classes in `pgmpy`.

This is not a complete refactor, but only illustrates the general architectural structure. Concretely, suggested actions:

* move `BaseDataset` to inherit from `BaseObject`
* move tag system to `_tags` and do not directly access the attribute, but use `get_class_tags` etc.
* move `name` attribute to tags, leave all other self-attributes as they are
* optional: replace registry with `all_objects` retrieval

### DIFF
--- a/pgmpy/datasets/_abalone.py
+++ b/pgmpy/datasets/_abalone.py
@@ -2,10 +2,11 @@ from pgmpy.datasets import register_dataset_class
 from pgmpy.datasets._base import _BaseDataset
 
 
-@register_dataset_class
+
 class AbaloneContinuous(_BaseDataset):
-    name = "abalone_continuous"
-    tags = {
+
+    _tags = {
+        "name": "abalone_continuous",
         "n_variables": 8,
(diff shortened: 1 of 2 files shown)


ankurankan:
Closing as #2515 is merged.

PR #2515, opened 2025-12-28
Updated the Dataset API according to #2506 and #2512

Theavinash02:
Hi @fkiraly and @ankurankan ,

I have updated the code accordingly to our discussion in #2506 and #2512 .
Please find the changes below:
- ` _BaseDataset` now inherits from` skbase.base.Baseobject` and updated `_base.py` to use `get_clase_tag` for caching and loading.
- Updated `skbase.lookup.all_objects` to discover datasets instead of custom `DatasetResgistry`
- Replaced the `tags `dictionary with the standard `_tags `class attribute
- Renamed datasets files (eg: _abalone.py to abalone.py) to ensure it is discoverable by `all_objects`.
- Moved and updated the extension template to a new location as discussed and updated the `test_datasets.py` according to the new changes.

Could you please review the changes and share your feedback and comments on the same?

Thanks
Avinash

@daehyun99


fkiraly (changes requested):
Looks great!

Though one thing that gets removed, not replaced, is the registry lookup, if I understand this correctly - is this intentional (e.g., will be added later), or an oversight?

Theavinash02:
Hi @fkiraly ,

Thanks for the review.
Regrading the registry lookup, i have removed the custom `DatasetRegistry` class as we replaced the manual registry mechanism with `skbase.lookup.all_objects`, which dynamically discovers the dataset classes. I have created a new `list_datasets` function in `_base.py` that wraps `all_objects` to list available datasets(variants as well) to ensure the functionality remains the same. And `load_dataset` now uses `all_objects` internally to find the correct class by name. 

Thanks
Avinash

@ankurankan

fkiraly (approved):
Ah, I overlooked `list_datasets`. Great!

Theavinash02:
Hi @ankurankan ,
Could you please review and kindly provide your feedback ?
Thanks

Theavinash02:
Thanks @ankurankan

PR #2518, opened 2026-01-09
Add score method to causal discovery estimator

### LINKED ISSUE #2412: [ENH] Add `score` method to Casual Discovery algorithms
pgmpy currently implements a few scoring metrics for causal models (https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/metrics/metrics.py), and some new ones are in the works (#2054 and #2115 ). Some of these metrics require ground truth data, and others can use the training data itself to give a score for how well the causal model explains the data.

Since we are also refactoring the causal discovery algorithms to be unsupervised sklearn-learn estimators (#2361 and #2410), these metrics can be added as the `score` method for the causal discovery estimators to make the interface compatible with sklearn. This should also enable the users to use other sklearn functionality, such as Grid Search.

Nimish-4:

- Fixes #2412


pgmpy/causal_discovery/_base.py
```diff
@@ -76,6 +78,57 @@ def _check_fit_data(self, X):
 
         return X
 
+    def score(
+        self,
+        X,
+        y=None,
+        method_true_model=SHD,
+        method_test_data=correlation_score,
```
ankurankan: 1. Remove `y`. 
2. rename arguemnt to `scoring_method=None`. Is there a standard name that sklearn uses?
3. Add `ground_truth=None`.

Nimish-4: @ankurankan 
2 - Doesn't seem like there is a standard name in sklearn. A particular metric is used by default in the base classes without naming. [Regressor](https://scikit-learn.org/stable/modules/generated/sklearn.base.RegressorMixin.html#sklearn.base.RegressorMixin.score) [Classifier](https://scikit-learn.org/stable/modules/generated/sklearn.base.ClassifierMixin.html#sklearn.base.ClassifierMixin.score)

3 - If we add a separate argument for `ground_truth`, what should `X` be in case we do have `ground_truth`? Test data wouldn't really be needed since we are directly testing the actual model, so I can't see a way to work with it.

ankurankan: @Nimish-4 

2. Yeah, I think sklearn always uses a default scoring metric and doesn't allow users to specify which one to use. I think we can not do that for causal discovery as there are no **standard** metrics.
3. I think X can be None in that case. I think we will need to keep the default value of X as None.

I had modified some code from this PR for a presentation and I was thinking of an interface like this. What do you think?:
causal_discovery_score

Nimish-4: Yep the `X` part was the only issue, this looks good. I'll make the change.

Nimish-4: @ankurankan Seems like `y` cannot be removed as an argument ... the estimator checks are expecting `y` as an argument. Looks fine otherwise.

ankurankan:
@Nimish-4 Any idea what the failing test is?

Nimish-4:
@ankurankan Looking into it. It is passing locally, could be due to the python version maybe.

ankurankan (dismissed):

ankurankan (approved):

PR #2559, opened 2026-01-30
Fix bugs due to pandas 3.0 string dtype

### LINKED ISSUE #2539: [BUG] Failures due to Pandas 3.0
**Describe the bug**

With the latest `pandas 3.0`, there is support for string data type. This is causing failures in cases where DataFrame values are being reassigned with other types (previously being `object` type). [For example](https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/readwrite/BIF.py#L166) - 

```python
state_df.iloc[:, idx] = state_df.iloc[:, idx].map(
state_maps[parent]
)
.
.
.

TypeError: Invalid value '0    0    1    1      Name: 0, dtype: int64' for dtype 'str'
```

Other parts of code with similar issues can be highlighted here.

**Versions**
Pandas - 3.0
Python - 3.14

Nimish-4:
- Fixes #2539 

- Added string dtype handling in `utils`
- Added object handling in `BIF` readwrite

Ref: [Migration guide](https://pandas.pydata.org/docs/user_guide/migration-3-strings.html#setitem-operations-will-now-raise-an-error-for-non-string-data)


pgmpy/utils/utils.py
```diff
@@ -437,10 +437,16 @@ def preprocess_data(df):
             else:
                 dtypes[col] = "C"
         else:
-            raise ValueError(
-                f"Couldn't infer datatype of column: {col} from data. "
-                "Try specifying the appropriate datatype to the column."
-            )
+            try:
+                if isinstance(df[col].dtype, pd.StringDtype):
+                    dtypes[col] = "C"
```
ankurankan: Can we move this check to line 431 with an `or` operator? Something like:
`elif pd.api.types.is_object_dtype(df[col]) or pd.api.types.is_string_dtype(df[col])`

ankurankan (changes requested):

Nimish-4:
@ankurankan The `ExpertInLoop` tests are failing due to the API key I think. I had made some changes to it because they were failing before this API issue too. Any idea?

PR #2566, opened 2026-02-04
get_ancestors is technically wrong 

NNEngine:
**Issue**: get_ancestors is technically wrong
```
return set(obs_nodes_list) | set(self.dependent)
```

Problems:
- set(self.dependent) splits a string into characters
- Ancestors ≠ observed nodes
- This ignores actual graph semantics

Fix:
```
def get_ancestors(self, nodes):
    if not nodes:
        return set()
    return {self.dependent}
```

Why this is correct:

- In Naive Bayes, only the class variable is an ancestor
- Observed nodes are not ancestors by definition


ankurankan:
@NNEngine The tests are failing.

ankurankan (approved):

Issue #2551, opened 2026-01-29
[ENH] Refactor example model loader


ankurankan:
pgmpy allows users to load example model from the bnlearn repository (https://www.bnlearn.com/bnrepository/) using the `pgmpy.utils.get_example_model` method. These example models are currently also packaged with pgmpy (https://github.com/pgmpy/pgmpy/tree/dev/pgmpy/utils/example_models). As we are in the process of expanding supported example models in pgmpy (#2550), we should refactor this `get_example_model` method into a separate module. Changes:

1. Move all the stored example models into a separate repository.
2. Create a base example model class similar to `_BaseDataset` that defines some tags and the provides the core functionality for loading an example model.
3. Separate out each example model into a class and a separate file. (Again similar to how `datasets` module is implemented).

Updates:
- Example models have been moved to: https://github.com/pgmpy/example_models

ankurankan:
@Theavinash02 @DARHWOLF @XAheli Would any of you be interested in taking this up? Would be nice if we could do this quickly, as this blocks #2550.

Theavinash02:
Hi @ankurankan 
I would like to work on this.

PR #2571, opened 2026-02-05
Example model loader #2551

Theavinash02:
This PR is for #2551. Please find the relevant updates below:

**List of changes to the codebase in this pull request**

- Added class file for three example model types under pgmpy/example_models:
- DAG example (Acid_1996) loaded via DAGitty text.
- Continuous example (arth150) loaded from JSON into a LinearGaussianBayesianNetwork.
- Discrete example (alarm) loaded from a compressed .bif.gz file.
- Added and restructure _base.py example model loader to handle DAG, continuous, and discrete formats.
- Updated `__init__.py` to explicitly expose all example models.

Manually validated the following behaviors:
- list_models() correctly discovers all example models.
- load_model() successfully loads DAG, discrete, and continuous models.
- Loaded models match expected node/edge/CPD counts.

Please review the code and kindly share your review comments. 

Thanks
Avinash

@ankurankan


Theavinash02:
Thank you, Ankur.
Shall I work on adding all the models that we are currently having and create a PR for the same?

@ankurankan

Issue #2612, opened 2026-02-10
[ENH] Design and documentation for bnRep example model support.


shivaansh0610-LUFFY:
### Description

This issue proposes a scoped first step towards supporting example models from the bnRep repository
(https://github.com/manueleleonelli/bnRep) in pgmpy.

The aim is to begin with **documentation and design**, before moving on to the actual implementation of
model conversion and inclusion of converted models. No changes to the existing model-loading logic
are proposed.

This is intended as a follow-up to #2550.

### Proposed scope

- Describe what bnRep offers and how it differs from existing bnlearn example models
- Outline the expected data formats for converted models:
- BIF for discrete (categorical) models
- JSON for continuous models
- Add developer-facing notes under `pgmpy/devtools` describing:
- the intended conversion workflow from bnRep models
- where converted model files should live (i.e., the `pgmpy/example_models` repository)
- Clarify that existing pgmpy model loaders can already consume these formats and do not require
  modification
- (Optional) Add a minimal placeholder or skeleton for a conversion script, without implementing
  the conversion logic itself

No changes to the public API or loader behavior are proposed at this stage.

### Motivation

Starting with documentation and design allows alignment on structure and expectations before
introducing larger code changes. It also provides a clear reference for future contributors who
may work on model conversion or dataset inclusion.

Feedback regarding scope and direction is much appreciated.

ankurankan:
@shivaansh0610-LUFFY Thanks for creating the issue. Once the script for converting the bnrep model files is done, we will be placing the files in our `example_models` repo: https://github.com/pgmpy/example_models. I don't think we will need to make any changes to our model loading functionality as the file formats would be the same as what our current model loader expects. We will simply need to add new model files under `example_models` corresponding to each of the models in the bnrep repository.

shivaansh0610-LUFFY:
> [@shivaansh0610-LUFFY](https://github.com/shivaansh0610-LUFFY) Thanks for creating the issue. Once the script for converting the bnrep model files is done, we will be placing the files in our `example_models` repo: https://github.com/pgmpy/example_models. I don't think we will need to make any changes to our model loading functionality, as the file formats would be the same as what our current model loader expects. We will simply need to add new model files under `example_models` corresponding to each of the models in the bnrep repository.

Thanks for the clarification!
I’ll update the design/documentation to reflect that converted bnRep models would live in the pgmpy/example_models repository and that no changes to the model loader are required.
[...]

Issue #2550, opened 2026-01-29
[ENH] Add example models from bnRep repository


ankurankan:
pgmpy has the option to load example models from the bnlearn repository (https://www.bnlearn.com/bnrepository/). This functionality is provided throught the `pgmpy.utils.get_example_model` method. 

The `bnRep` repository (https://github.com/manueleleonelli/bnRep) also has a collection of Bayesian network models from literature.  We should allow users to directly load these models as well. Changes required:

1. Write an R script that can read in the models from `bnrep` repository and output a `BIF` format file (if categorical) or in JSON format (if continuous). Add this script under `pgmpy/devtools`.
2. Add the option to load these models using the `get_example_model` method.

Eshaan-byte:
Hi @ankurankan  can I work on this issue? Thank you!

ankurankan:
@Eshaan-byte This issue is currently blocked by #2551. What you can do in the meantime is to work on point 1 of this issue while #2551 is resolved. If you would like to work on the first part, please create a new issue and tag me; I will assign that to you.

ankurankan:
Update: The issue is not blocked anymore.

shivaansh0610-LUFFY:
Hi @ankurankan ,
I’m interested in contributing to this issue. I was thinking of starting with a scoped part, such as documentation and design for loading bnRep example models (and/or a minimal conversion script skeleton), before moving to full integration.

Would that be a reasonable place to start? Happy to open a separate issue if preferred.

ankurankan:
@shivaansh0610-LUFFY , I would suggest creating a separate issue for it, and we can discuss it there.

shivaansh0610-LUFFY:
I’ve created a separate issue for the scoped documentation/design work as suggested: 
Issue #2612 
Happy to continue the discussion there.

PR #2613, opened 2026-02-10
Add design and documentation for bnRep example model support

shivaansh0610-LUFFY:

- [ ] Are all the GitHub Actions checks passing? If not, they will need to be fixed before review. You can reference actions logs for the failing check to identify the issue.

- [ ] Please include a short description of the changes you have made. This doesn't need to be a polished description, but it **must** be written manually (**not by an AI model**) by you.
- [ ] Have you **manually** verified that the changes are doing exactly what is expected?

- Related to #2612 and #2550

- Added developer-facing documentation under `pgmpy/devtools/bnrep/README.md.`
- Documented the agreed workflow for supporting bnRep example models
- Clarified that converted models live in `pgmpy/example_models` and no loader or API changes are required

### DIFF
--- a/devtools/bnrep/README.md
+++ b/devtools/bnrep/README.md
@@ -1,0 +1,45 @@
+# bnRep Example Models — Design Notes
+
+## Background
+
+The bnRep repository (https://github.com/manueleleonelli/bnRep) provides a collection of Bayesian
+network models from the literature. These models are useful for benchmarking, education, and
+research, but are not directly loadable in pgmpy in their current form.
+
+This document records the agreed documentation and workflow for supporting bnRep example models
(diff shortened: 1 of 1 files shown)


ankurankan:
@shivaansh0610-LUFFY I don't think we need a design doc like this. I am not sure who this would be helpful to. I would rather suggest starting to add model files based on the models in bnRep repo assuming that those files existed in our `example_models` repo. Once the model files are added, we should then be able to load these models in pgmpy.

shivaansh0610-LUFFY:
Thanks for the clarification, that makes sense.

I’ll follow your suggestion and focus next on adding converted model files from the bnRep repository to the `example_models` repo, assuming the expected formats already align with the existing loaders.

I’ll update the original issue accordingly once I have something concrete to propose.

PR #2622, opened 2026-02-14
Refactor structure score class

### LINKED ISSUE #2536: [MNT] Refactor structure scores
All relevant code is in `pgmpy/estimators/StructureScore.py`.
 
Changes:
1. Create a new directory under `causal_discovery` named `structure_score`.
2. Replace the `StructureScore` class with `BaseStructureScore`, which inherits from `skbase.BaseObject`. Add this to `structure_score/_base.py`.
3. Separate out each scoring method into a separate file.
4. Add `_tags` to each of the classes: `name`, `supported_datatype`,  `is_parameteric`, `is_default`. 
5. Modify `ScoreCache` implementation to be a mixin class or check if simply `lru_cache` can be used instead.
6. Modify score based causal discovery algorithms to use the refactored implementation.

Nimish-4:
The following checklist was written by @ankurankan 

Codex and Claude were used for implementing parts of the PR. I started with precisely specifying how the new design needed to be. I wanted to move the scoring classes into separate files in a new module `structure_score`, move the hyperparameters of each of the scoring classes into their initialization, and unify the signature of the `local_score` method. I also wanted to redesign the score caching and kept it out of the scope for the initial refactor. Claude was given a detailed design instruction and asked to refactor the code based on that. I also instructed claude to add a future warning to the current modules. Once the refactor was done, I asked Claude to implement a new `get_structure_score` method and update its use in `pgmpy/causal_discovery` and `pgmpy/metrics` modules. I found the signature and logic of `get_structure_score` too convoluted, and simplified its scope to work for only `BaseStructureScore` or str objects, removing the kwargs argument. Once this was done, I asked codex to add `lru_cache` decorator to the `local_score` methods in each of the classes, which led to a problem with caching whole objects. I then asked it to implement a `_local_score` method in the Base class, which gets cached in the init method. This led to a cleaner implementation and object instance-level caching. Lastly, I asked codex to port all the tests from `test_StructureScore.py` to a new directory `test_structure_score` and create files for each scoring method. I asked both codex and claude to verify that the new tests match exactly with the old tests. Lastly, I asked for porting documentation from the `StructureScore.py` to the new modules, this did lead to some issues in the documentation getting modified. I then manually went through the docstrings and improved them to a satisfactory level.

Yes, I have redesigned the structure score module. The new design is very similar to the CI test module. The idea is to make the structure score classes composable and split the hyperparameters between the causal discovery methods and the structure score method itself. I have also improved the documentation of each of the classes. The logic of the code remains exactly the same.

I have made sure that the values output from the new implementation exactly match the previous one. I have used tests to check if the integration of the new `structure_score` module still works with the rest of the codebase. I have tried to manually run causal discovery algorithms to see if the new API feels intuitive. This led to the discovery of coupling between causal discovery constructors and `fit` methods #3144

No.

Yes. I prompted the LLM to exactly copy the existing tests into the new module design. I have checked that the values being tested are exactly the same and the comments on comparing the values to other packages are intact.

- Fixes #2536 

- Created a new module `structure_score`.
- Each structure score method moved to a separate file.
- Each structure score class initialization method now takes the method's hyperparameters as arguments. `local_score` method is unified across all methods.
- Scores always get cached now. No option to specify whether to cache or not. Much simpler design for score cache.
- `causal_discovery` module updated to use the new `structure_score` module.


pgmpy/causal_discovery/structure_score/_base.py
```diff
@@ -0,0 +1,311 @@
...
+        >>> from pgmpy.estimators import K2
+        >>> op = ("add", "A", "B")  # Example operation
+        >>> score = K2(data)
+        >>> ratio = score.structure_prior_ratio(op)
+        >>> print(ratio)
+        0
+        """
+        return 0
+
+    def state_counts(
```
ankurankan: I think this is duplicated code. Is it a problem to inherit from both `BaseObject` and `BaseEstimator` so that we don't have to keep this `state_counts` method in both classes?

pgmpy/causal_discovery/structure_score/dirichlet_priors.py
ankurankan, line 1: I think it might be better to split them all into individual files. We won't be able to classify newer structure score methods into any of these.

pgmpy/causal_discovery/structure_score/log_likelihood.py
ankurankan, line 1: Same here.

pgmpy/causal_discovery/structure_score/log_likelihood_gauss.py
ankurankan, line 1: Same here.

ankurankan (changes requested):
@Nimish-4 Thanks for the PR. I think we will still need to implement the `get_scoring_method` equivalent in `_base.py` and update how it is called in the codebase.

Nimish-4:
> @Nimish-4 Thanks for the PR. I think we will still need to implement the `get_scoring_method` equivalent in `_base.py` and update how it is called in the codebase.

Implementing it in `_base.py` results in a circular import error, due to importing all the different scores (which depend upon `BaseStructureScore` itself). Any particular thing you have in mind?

pgmpy/causal_discovery/structure_score/_base.py
Nimish-4: Yep you are right, I was thinking of a standalone file so copy pasted it.

Issue #2419, opened 2025-11-15
[ENH] List of causal datasets


ankurankan:
As we are working on adding data loader methods in pgmpy, this issue serves as a list of publicly available datasets that we can add in the package:

1. Causal estimation + causal discovery: https://github.com/rguo12/awesome-causality-data
2. Pairwise causal discovery: https://webdav.tuebingen.mpg.de/cause-effect/

Theavinash02:
Hi @ankurankan ,
I would like to work on this.  Could you please assign this to me?
Thanks

Mk-D-01:
Hello @ankurankan,
I would like to work on this. Could you please assign it to me?

ankurankan:
@Mk-D-01 This is an umbrella issue for a lot of datasets, please refer individual issues and comment there for assignment.

NAME-ASHWANIYADAV:
@ankurankan is this issue open as i can see @Theavinash02 mentioned a pr and it is merged but this is still open so i would just like to confirm is it still open to work as i am looking for some opportunties

PR #2627, opened 2026-02-15
Pairwise casual discovery(Tubingen dataset)

Theavinash02:
- Created a new mixin class (`_TubingenBenchmarkMixin`) to load tubingen dataset in _base.py under pgmpy/datasets.
- modified the Dataset dataclass to support the Dict types and updated the `__str__` method accordingly in _base.py.
- added metadata from the website to retrieve the weights of each pairs,

This PR is related to #2419 .
- Pairwise causal discovery: https://webdav.tuebingen.mpg.de/cause-effect/

Hi @ankurankan ,
Could you please review this code below and kindly share your feedback?

Thanks
Avinash


pgmpy/datasets/_base.py
```diff
@@ -22,15 +23,18 @@
 @dataclass
 class Dataset:
     name: str
-    data: pd.DataFrame
+    data: Union[pd.DataFrame, Dict[int, pd.DataFrame]]
```
ankurankan: I don't think this is correct. When will this be an `int`; I think this should always be a DataFrame.

pgmpy/datasets/_base.py
```diff
@@ -22,15 +23,18 @@
...
     expert_knowledge: Optional[ExpertKnowledge] = None
-    ground_truth: Optional[DAG] = None
-
+    ground_truth: Union[Optional[DAG], Dict[int, Dict[str, Any]]] = None
```
ankurankan: Again, I think this is wrong.

pgmpy/datasets/_base.py
```diff
@@ -240,6 +244,75 @@ def load_dataframe(cls) -> pd.DataFrame:
...
+                continue
+            pair_id = int(parts[0])
+            weight = float(parts[-1])
+            weights[pair_id] = weight
+        zip_content = cls._get_raw_data("data", cls.data_url)
+        gt_dict = {}
+        with zipfile.ZipFile(io.BytesIO(zip_content)) as z:
+            files = [f for f in z.namelist() if re.match(r"pair\d+_des\.txt", f)]
+            for file in files:
+                try:
```
ankurankan: There shouldn't be any try-except blocks.

pgmpy/datasets/_base.py
```diff
@@ -240,6 +244,75 @@ def load_dataframe(cls) -> pd.DataFrame:
...
+    """
+
+    @classmethod
+    def load_dataframe(cls) -> Dict[int, pd.DataFrame]:
+        zip_content = cls._get_raw_data("data", cls.data_url)
+        pairs_dict = {}
+        with zipfile.ZipFile(io.BytesIO(zip_content)) as z:
+            files = [f for f in z.namelist() if re.match(r"pair\d+\.txt", f)]
+            for file in files:
+                try:
```
ankurankan: Shouldn't do try-except blocks. I am guessing this was generated by an AI.

ankurankan (changes requested):
Tests are missing. I am not sure how this is supposed to be used. I see a single `tubingen` dataset being added; does this load all the datasets in this repo?  If that is the case, I don't think that would be a good idea. As these are different datasets, we should have different names for them.

pgmpy/datasets/_base.py
Theavinash02: Hi Ankur,
I added the try-except block because some of the tubingen pair description files don’t always contain direction or weighting information. In a few cases, the ground truth is missing or written differently, which causes the loader to fail so for now, I added the try-except to prevent a single pair from breaking dataset loading while I confirm how these edge cases should be handled. I was planning to clarify this in yesterday’s meeting as well. I know that catching exceptions isn’t ideal, and I can update this with proper validation once we decide the expected behavior.

Could you please let me know how to proceed with such cases?

pgmpy/datasets/_base.py
Theavinash02: I have included this as I have used dict to store the 108 pairs and users can select the pair according with the pair number and load the same.
Shall I change the approach for this as well?

pgmpy/datasets/_base.py
Theavinash02: Hi Ankur,,

each pair is stored as a separate file (pair1.txt … pair108.txt), so the loader returns a dictionary of DataFrames keyed by pair id instead of a single DataFrame.

Could you please confirm if this approach is okay? 

I need to update the test file as well. I will share the same soon.

Thanks

Theavinash02:
Hi @ankurankan 
Could you please check the comments and please share your thoughts?

Thanks
Avinash

ankurankan:
I think we do not have the same API in mind for how this would work. I expect this to work as follows:

```python
# This only loads the pairs_1 dataset.
pairs_1 = load_dataset('tubingen/1')

# This only load the pairs_12 dataset
pairs_1 = load_dataset('tubingen/12')
```

This is a cleaner API  for `load_dataset` than loading all the datasets at the same time. This also makes the API single-purpose (i.e. doesn't have different behavior depending on the argument).

pgmpy/datasets/_base.py
ankurankan: If that is the case, it will need to be handled explicitly. Ideally, it should be an if-else condition where the if statement checks if the weighting information is available or not. If this if-else approach is not possible, we can do a try-except block, but the except block should only catch the specific error that is thrown rather than doing `except Exception`.

Theavinash02:
Thank you for the approach, Ankur. 
I will work on the same and get back to you.
Thanks
Avinash

@ankurankan

pgmpy/datasets/_base.py
Theavinash02: Sure,Ankur

Theavinash02:
Hi @ankurankan ,
Could you please review the code and kindly share your review comments?
Thanks

ankurankan (changes requested):
Looks good mostly. I think we should unzip the files and put it in our example_datasets folder. Unzipping all the files for each load would be quite expensive.

Theavinash02:
Hi @ankurankan , 
Sure. I will make the changes on the code and the repo as well.
Thanks 😄

Theavinash02:
Hi @ankurankan ,

could you please review the PR below ?
https://github.com/pgmpy/example_datasets/pull/6
Once the above PR is merged, I will start updating the code in this PR.

Thanks,
Avinash

ankurankan (changes requested):
Data PR merged

Theavinash02:
Thanks @ankurankan

Theavinash02:
Hi @ankurankan ,
Could you please review this PR and kindly share your feedback?
Thanks
Avinash

Theavinash02:
Hi @ankurankan ,

Could you please review the updated `class TubingenBenchmarkMixin` and kindly share your comments on the same?

Thanks,
Avinash

pgmpy/datasets/tubingen.py
```diff
@@ -0,0 +1,24 @@
...
+
+class Tubingen(_TubingenBenchmarkMixin, _BaseDataset):
+    """
+    Tubingen Cause-Effect Pairs Dataset.
+    A benchmark collection of independent cause-effect pairs.
+    """
+
+    _tags = {
+        "name": "tubingen",
+        "n_variables": None,
```
ankurankan: This need to be `2`

pgmpy/datasets/tubingen.py
```diff
@@ -0,0 +1,24 @@
...
+        "n_samples": None,
```
ankurankan, line 13: We need to figure out a way to dynamically set this tag when `load_dataset` is called.

pgmpy/datasets/_base.py
```diff
@@ -256,25 +280,49 @@ def load_dataset(name: str) -> Dataset:
...
+            if pair_id is not None
+            else target_cls.load_dataframe()
+        ),
         expert_knowledge=target_cls.load_expert_knowledge(),
-        ground_truth=target_cls.load_ground_truth(),
+        ground_truth=(
+            target_cls.load_ground_truth(pair_id)
+            if pair_id is not None
+            else target_cls.load_ground_truth()
+        ),
```
ankurankan: Shouldn't do if-else here. It complicates code. The `if name.startswith("tubingen")` block should create and return the `Dataset` object directly.

pgmpy/datasets/_base.py
```diff
@@ -256,25 +280,49 @@ def load_dataset(name: str) -> Dataset:
     >>> df = dataset.data
     >>> ground_truth = dataset.ground_truth
     """
+    pair_id = None
+    search_name = name
+    if name == "tubingen" and pair_id is None:
+        raise ValueError(
+            "Tubingen dataset requires a pair id. Use 'tubingen/<pair_id>'."
+        )
```
ankurankan: Remove this. Let the error handling in line 310 catch this.

pgmpy/datasets/_base.py
```diff
@@ -256,25 +280,49 @@ def load_dataset(name: str) -> Dataset:
...
+    if name.startswith("tubingen/"):
+        search_name = "tubingen"
+        name_parts = name.split("/")
+        if len(name_parts) == 2 and name_parts[1].isdigit():
```
ankurankan, line 288: Also check if the `name_parts[1]` is outside available dataset number.

ankurankan (changes requested):
The `load_dataset` method will need to be simplified; otherwise looks good.

pgmpy/datasets/tubingen.py
Theavinash02: Hi @ankurankan ,
I think we can do the same by loading the data first and then updating the n_sample and n_variable with the resulting dataframe's shape before returning the dataset object.
I have updated the code according to this approach.
Please check and let me know.

Theavinash02:
Hi @ankurankan ,
Could you please review the code and  kindly share your feedback?
Thanks
Avinash

Theavinash02:
Hi @ankurankan ,
I have updated the code as per our discussion in dev meeting. Could you please review and kindly let me know?
Thanks

ankurankan (approved):

Issue #2545, opened 2026-01-27
[Refactor] Change to use `pytest` from `unittest`


daehyun99:
`pgmpy` has decided to adopt `pytest` over `unittest`.
Therefore, we need to refactor our existing test code to follow the `pytest` syntax.

### Requirements:
- Replace `unittest`'s `setUp` and `tearDown` with `pytest`'s `fixtures`.
- Please submit your PR following the PR template.

### Example:
Below are examples of test code that are already implemented in `pytest`:
- https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/tests/test_base/test_AncestralBase.py
- https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/tests/test_identification/test_base.py

### Contributing Guide Docs:
- https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md#contributing-to-pgmpy

Thank you.

cc @ankurankan 

---

### 📋 File Changes Tracking

- [x] `test_global_vars`: #2641
- [ ] `test_base/test_DAG`: ⏭️ Skipped (blocked by 2376)
- [x] `test_base/test_PDAG`: #3311
- [x] `test_ci_tests/*.py`:  #3202
- [x] `test_estimators/test_BaseEstimator`: #2631
- [x] `test_estimators/test_BayesianEstimator`: #2755
- [x] `test_estimators/test_CITests`: #2796 (blocked by 2547)
- [ ] `test_estimators/test_EM`: 📝 **Need to Change**
- [x] `test_estimators/test_ExhaustiveSearch`: #2824
- [x] `test_estimators/test_expert`: #2852
- [x] `test_estimators/test_GES`: #2770
- [ ] `test_estimators/test_HillClimbSearch`: ⏭️ Skipped (going to be deprecated)
- [x] `test_estimators/test_MarginalEstimator`: #2765
- [x] `test_estimators/test_MaximumLikelihoodEstimator`: #2819
- [x] `test_estimators/test_MirrorDescentEstimator`: #2816
- [x] `test_estimators/test_MmhcEstimator`: #2773
- [x] `test_estimators/test_ParameterEstimator`: #2779
- [x] `test_estimators/test_PC`: #2750
- [x] `test_estimators/test_ScoreCache`: #2756
- [x] `test_estimators/test_SEMEstimator`: #2768
- [x] `test_estimators/test_StructureScore`: #2910
- [x] `test_estimators/test_TreeSearch`: #2901
[...]

PR #2641, opened 2026-02-19
refactor(tests): convert test_global_vars.py from unittest to pytest

moksha-hub:

- Fixes part of #2545

- Converted `test_global_vars.py` from `unittest` to `pytest` style
- Removed `unittest.TestCase` inheritance from `TestDuplicateFilter`
- Replaced `tearDown` in `TestConfig` with an `autouse` pytest fixture that resets config after each test
- Replaced all `self.assertEqual(a, b)` with plain `assert a == b` statements
- Applied black formatting

### Summary
This PR converts `pgmpy/tests/test_global_vars.py` from `unittest` to `pytest` as part of the broader migration tracked in #2545. The `tearDown` cleanup logic in `TestConfig` is replaced with an `autouse` fixture that resets config state after each test, which is the idiomatic pytest approach. All assertions are converted to plain `assert` statements. I manually verified the tests pass locally.


moksha-hub:
@fabriziov @linzhp @dbolser @sroecker please have a review.
Thank You.

moksha-hub:
@daehyun99 please have a review if changes needed i will be available as quick.

pgmpy/tests/test_global_vars.py
```diff
@@ -10,88 +9,84 @@
...
-        self.assertEqual(config.get_show_progress(), True)
+        assert config.SHOW_PROGRESS is True
+        assert config.get_show_progress() is True
 
     @pytest.mark.skipif(
         not _check_soft_dependencies("torch", severity="none"),
         reason="test only if torch is available",
     )
-    def test_torch_cpu(self):
+    def test_torch_cpu(self):  # pragma: no cover
```
daehyun99: Hi there,

Overall looks good to me.

But I have a quick question.
Is there a specific reason why you added `# pragma: no cover`?
Didn't the coverage test pass even before adding this?

I'm not entirely sure why it's necessary here, so please let me know if I'm missing something.

Thanks.

pgmpy/tests/test_global_vars.py
```diff
@@ -10,88 +9,84 @@
...
+        assert config.SHOW_PROGRESS is True
+        assert config.get_show_progress() is True
 
     @pytest.mark.skipif(
         not _check_soft_dependencies("torch", severity="none")
         or not torch.cuda.is_available(),
         reason="test only if torch and torch.cuda are available",
     )
-    def test_torch_gpu(self):
+    def test_torch_gpu(self):  # pragma: no cover
```
daehyun99, line 58: Here too

moksha-hub:
Hi @daehyun99 , thanks for reviewing!

Yes, the coverage test was passing before, but these test functions (test_torch_cpu and test_torch_gpu) are conditionally skipped when PyTorch is not installed in the CI environment. Without # pragma: no cover, coverage.py would flag them as uncovered lines in the report, even though they're intentionally skipped via @pytest.mark.skipif. Adding # pragma: no cover keeps the coverage report clean and accurate  it prevents misleading "missed" coverage that's actually expected skips, not dead code.

pgmpy/tests/test_global_vars.py
daehyun99: The coverage test results are generated in an environment where `torch` dependencies are installed.

https://github.com/pgmpy/pgmpy/blob/01f33549e6750701e8bd3e2bd56403077a4a4d1a/.github/workflows/ci.yml#L58-L107

In this setup, `test_torch_cpu` runs without being skipped.
However, since the GitHub Actions environment does not support GPU tests, `test_torch_gpu` cannot be executed.

Therefore, I believe `# pragma: no cover` should only be applied to `test_torch_gpu`.
I suggest removing the `# pragma: no cover` tag from `test_torch_cpu` to ensure it is properly tracked in our coverage

```suggestion
    def test_torch_cpu(self): 
```

Thanks.

moksha-hub:
Thanks @daehyun99  for digging into the CI config! You're absolutely right I missed that the test-fulldep job runs with torch installed, meaning test_torch_cpu is properly exercised in CI and should be tracked by coverage. I'll accept your suggestion to remove # pragma: no cover from test_torch_cpu and keep it only on test_torch_gpu since GPU tests genuinely can't run in the GitHub Actions environment.

daehyun99 (commented):
it looks good to me.
I don't see anything else that needs to be done.

Thanks.

cc @ankurankan

ankurankan (dismissed):

ankurankan (approved):

Issue #2567, opened 2026-02-05
[BUG] The dataset template file doesn't ask for `base_url`


ankurankan:
The template files (https://github.com/pgmpy/pgmpy/blob/dev/devtools/extension_templates) are supposed to allow developers to easily add new methods to pgmpy. The `_dataset.py` template is for adding new datasets to the repo. 

The `_dataset.py` template currently has a bug, as it doesn't require the developer to specify `base_url`, but `base_url` is required as it is used by the caching functionality. This should be fixed.

Soham-47:
Hii @ankurankan , can you please assign this issue to me?

MAVRICK-1:
@ankurankan can I work on this?

singhharsh1708:
@ankurankan can you assign me ths?

DogeSamurai23:
Hi @ankurankan  
Is this issue resolved? I would like to contribute to this.

Thanks

NOAH-DCRUZE100:
Hi @ankurankan, I would like to claim this issue. I've identified that the current dataset template is missing a prompt for the base_url, which is necessary for the library to correctly point to the raw data files in the example_datasets repository. Please assign this to me so I can submit the fix.

PR #2679, opened 2026-02-20
Fix dataset template to require base_url

realMelTuc:
## Description

The _dataset.py template had  which caused issues with the caching functionality. Changed it to a placeholder URL that developers should update when creating new datasets.

## Fix

Changed the default value of  from  to a placeholder URL:

This ensures that:
1. The template provides a working placeholder URL
2. Developers will see a proper URL instead of None
3. The caching functionality will work properly since base_url won't be None
4. The example in the comment is now actually used as the default

## Issue

Fixes issue #2567: "The dataset template file doesn't ask for base_url"

## Checklist

- [x] I have read the contributing guidelines
- [x] My changes follow the code style of the project
- [x] I have tested my changes (if applicable)
- [x] My changes do not break existing tests

### DIFF
--- a/devtools/extension_templates/_dataset.py
+++ b/devtools/extension_templates/_dataset.py
@@ -45,9 +45,6 @@ class YourDatasetClass(_BaseDataset):
     # The current parser expects the dataset to be in a tabular form with the first row containing the names of the columns.
     data_url = None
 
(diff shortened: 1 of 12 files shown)


ankurankan:
No checklist. Closing.

PR #2684, opened 2026-02-20
[ENH]: Add get_reference() for programmatic access to dataset/model citations

### LINKED ISSUE #2596: [ENH] Figure out a better way to allow uses to access references for example datasets and example models
Each example dataset and model class currently has the citation for the original source in the class docstring. But since users are expected to interact with these example models/datasets only through the `load_dataset` or `load_model` methods, there is no easy way to access this citation. We need to figure out a way to make it easier for the user to quickly/programmatically get the reference without needing to look into the codebase.

dhairya-pandya:
## ENH: Add get_reference() for programmatic access to dataset/model citations (#2596)
As discussed by @ankurankan in #2596, To solve this issue I have added a `get_reference()` function in both `pgmpy.datasets` and `pgmpy.example_models` that takes dataset or model name and returns the citation by parsing the class docstring at the runtime.

For `Dataset` dataclass, I have added a `reference` field, so it is returned when you call `load_dataset()`. Internally, a helper function scans the docstring for the RST "References" section and extracts everything under it.

This approach means none of the 80+ individual dataset or model files needed any changes — the docstring stays the `single source of truth`. Tests cover the cases where a reference exists (returns a string), doesn't exist (returns `None`), and the name is invalid. throws `ValueError` .

- [] Are all the GitHub Actions checks passing? If not, they will need to be fixed before review. You can reference logs for the failing check to identify the issue.

- [x] Please include a short paragraph (**not bullet points**) describing the changes you have made. This should at least include the problem your PR solves and the implemented solution. This doesn't need to be a polished description, but it **must** be written **manually** by you without any assistance from any AI tools. - The topmost paragraph is to be referenced for this.

- Fixes #2596 

- Added `get_reference(name)` public function to both `pgmpy.datasets` and `pgmpy.example_models` that returns the citation string extracted from the class docstring, or `None` if no references exist.
- Added `reference` field to the `Dataset` dataclass, automatically populated by `load_dataset()`.
- Added `get_reference()` classmethod to `_BaseDataset` and `_BaseExampleModel` base classes.
- Added `_parse_docstring_references(cls)` helper in both `_base.py` files to parse RST-style `References` sections from docstrings at runtime.
- Factored out `_find_dataset_class(name)` and `_find_model_class(name)` helpers to avoid duplicating class-lookup logic between `load_*` and `get_reference`.
- Updated `__init__.py` exports for both packages.
- Added tests: `test_get_reference` (string return for cited, `None` for uncited, `ValueError` for unknown), `test_load_dataset_has_reference` (Dataset.reference field), and `test_get_reference_all_models` (works across all models).

### DIFF
--- a/pgmpy/datasets/__init__.py
+++ b/pgmpy/datasets/__init__.py
@@ -1,8 +1,8 @@
-from ._base import load_dataset, list_datasets, _BaseDataset
-
+from ._base import _BaseDataset, get_reference, list_datasets, load_dataset
 
 __all__ = [
     "_BaseDataset",
-    "load_dataset",
+    "get_reference",
     "list_datasets",
+    "load_dataset",
 ]
--- a/pgmpy/datasets/_base.py
+++ b/pgmpy/datasets/_base.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import hashlib
-import inspect
 import io
 import os
 import re
@@ -18,6 +17,7 @@ from skbase.lookup import all_objects
 from pgmpy.base import DAG
 from pgmpy.estimators import ExpertKnowledge
 from pgmpy.global_vars import PGMPY_DATA_HOME
+from pgmpy.utils import _parse_docstring_references
 
 
 @dataclass
@@ -26,7 +26,7 @@ class Dataset:
     data: pd.DataFrame
     expert_knowledge: Optional[ExpertKnowledge] = None
     ground_truth: Optional[DAG] = None
-    reference: Optional[str] = None
+    reference: str | None = None
 
     tags: Dict[str, Any] = None
 
@@ -187,7 +187,7 @@ class _BaseDataset(BaseObject):
         return DAG.from_dagitty(raw_data)
 
     @classmethod
-    def get_reference(cls) -> Optional[str]:
+    def get_reference(cls) -> str | None:
         """
         Extracts and returns the References section from the class docstring.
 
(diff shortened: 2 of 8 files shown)


dhairya-pandya:
@ankurankan Please review this PR as per your convenience and provide feedback regarding the same! I have discussed the approach used in the PR message

DARHWOLF (changes requested):
Sorry for the delay in reviewing this. 

May I request you explain how this is supposed to work? How would the user be able to  access the citation in this current implementation. 

I would suggest discussing design with @ankurankan before raising a PR. 
Could you please mention the details here (in your own words -- no llm)

I see em dashes in the PR description which indicates LLM usage.

dhairya-pandya:
Sorry for replying late, @DARHWOLF Thanks for reviewing the PR, so this function `get_reference()` helps the user to get references/citations by simply calling it. So there are two ways to call it, getting only reference and other being while loading the model or dataset. Both use simple design, where citations are parsed directly from class docstrings at runtime, no other files needed. 

The `_parse_docstring_reference(cls)` finds the References / strings of `-----` RST Section headers and collects everything below it until the next section or end of docstring.

### Example Usage

```
# Standalone Usage of get_reference()
from pgmpy.datasets import get_reference
from pgmpy.example_models import get_reference

# For datasets
ref = get_reference("abalone_continuous")
print(ref)
# Output: .. [1] Lopez-Paz, D., ...

# For example models
ref = get_reference("bnlearn/alarm")
print(ref)
# Output: ..[1] I. A. Beinlich, ...

# Usage along the Dataset class
from pgmpy.datasets import load_dataset
dataset = load_dataset("abalone_continuous")
print(dataset.reference)
# Output: .. [1] Lopez-Paz, D., ...
print(dataset)
# Shows reference alongside data shape, ground_truth, etc.
```

ankurankan (changes requested):
@dhairya-pandya Sorry for the delay in reviewing this.
1. I don't think we should add a new method for `get_reference`.
2. I would suggest simply adding a `description` field to `Dataset`. I don't have a good solution for example models yet, so I would suggest leaving example datasets for now.
3. Since most of the docstrings only have the reference, I would suggest removing the parsing logic and simply add the `__doc__` attribute to the `description`.
4. Might be better to create a new PR because of the number of merge conflicts.

dhairya-pandya:
Thanks for the review! Ok, I will implement these changes. I found out that this is way simpler. As you told, I am implementing the change of adding `__doc__ ` attribute in description, like this in the load_dataset() function.

```
    return Dataset(
        name=name,
        data=target_cls.load_dataframe(),
        expert_knowledge=target_cls.load_expert_knowledge(),
        ground_truth=target_cls.load_ground_truth(),
        description=target_cls.__doc__,
        tags=target_cls.get_class_tags(),
    )
```

I am also closing this PR and opening a new one due to numerous merge conflicts. Thanks!

PR #2716, opened 2026-02-24
fix: use single RNG instance in MarkovChain.generate_sample() for correct seed behavior

### LINKED ISSUE #2715: MarkovChain.generate_sample() re-seeds RNG on every step, breaking reproducible walks
## Bug Description

`MarkovChain.generate_sample()` accepts a `seed` parameter but misuses it: the seed is passed to `sample_discrete()` inside the per-step loop, which **re-seeds the RNG identically on every iteration**. This means every transition samples from the same initia

moksha-hub:

Yes, to study and know about the codebase and its internal components

Yes. The issue was caused by passing the `seed` into `sample_discrete()` inside the transition loop in `generate_sample()`, which resulted in the RNG being re-seeded at every step. This caused the sampling process to restart from the same state repeatedly instead of progressing.

The fix ensures that NumPy is seeded once before the loop starts, so the random number generator advances correctly across transitions. The rest of the sampling logic remains unchanged and continues to use `sample_discrete()` for transitions.

I traced `MarkovChain.generate_sample()` to confirm that the RNG was being re-seeded on every transition. After updating the seeding behavior, I verified that the random stream progresses correctly between steps.

I ran the `generate_sample` regression tests and the full `test_MarkovChain.py` test file locally, and both passed. I also ran pre-commit checks on the modified files and confirmed that formatting and linting passed, and verified that GitHub Actions completed successfully.

For edge cases, I checked that using the same `start_state` and `seed` produces identical sequences across runs, that the RNG no longer resets at each step, and that sampling still works correctly when `start_state` is not provided (falling back to `random_state()`).

No.

No.

- Fixes #2715

- Updated `MarkovChain.generate_sample()` so that NumPy is seeded once per generated walk instead of passing `seed` into `sample_discrete()` on every transition.
- This fixes the repeated re-seeding bug and allows the RNG stream to advance correctly across steps while keeping `sample_discrete()` as the transition sampling path.
- Expanded the `generate_sample()` docstring to document `start_state`, `size`, and `seed`, and updated the example to show seeded usage.
- Replaced the old mock-based `generate_sample` tests with behavior-based checks in `test_MarkovChain.py`.
- Added a regression test to verify that two runs with the same `start_state` and `seed` produce identical sample sequences.
- Added a regression test to verify that the RNG advances between steps instead of restarting from the same seeded state on each transition.
- Updated the no-`start_state` test to verify the `random_state()` fallback path without depending on mocked transition outputs.
- Removed the unused `sys` import and added a `sample_discrete` import in the test file for the seeded regression check.


DARHWOLF:
Can you update the PR template with the original template. 
It should not be changed in any way or else PR will not be reviewed. 
Converting this to draft till i see the checklist back

ankurankan:
The checklist: https://github.com/pgmpy/pgmpy/blob/dev/.github/PULL_REQUEST_TEMPLATE.md

ankurankan:
Please fill out the checklist. Please follow the instructions. We have also asked you to add the checklist in other PRs, and I see that you keep ignoring the instructions. Please be more careful.

ankurankan (changes requested):
Refer comment

ankurankan:
@moksha-hub I am closing this. You are not following the clear instructions, even after being asked multiple times. We won't be accepting any other contributions from you.

moksha-hub:
@ankurankan apologies for making mistakes i really made changes to the checklist can you please explain what exact i need to follow this is my first time here for contributions

PR #2734, opened 2026-02-27
fix(inference): ApproxInference.query builds state space from samples instead of model

### LINKED ISSUE #1941: Wrong indexing in `ApproxInference._get_factor_from_df()` when giving `state_names`
Providing `state_names` to a `ApproxInference` object doesn't work properly. It gives values `0` always because it can't reindex according to string state names.

## Minimal Reproducible Example

```
import numpy as np
import pandas as pd
from pgmpy.inference import ApproxInference, DBNInference
from pgmpy.models import DynamicBayesianNetwork as DBN

data = np.random.randint(low=0, high=2, size=(1000, 6))
colnames = []
for t in range(2):
    colnames.extend([("A", t), ("B", t), ("C", t)])

df = pd.DataFrame(data, columns=colnames)

model = DBN(
    [
        (("A", 0), ("B", 0)),
        (("A", 0), ("C", 0)),
        (("A", 0), ("A", 1)),
        (("B", 0), ("B", 1)),
        (("C", 0), ("C", 1)),
    ]
)

model.fit(df)
```

If I do:

```
# Query the model
dbn_inf = ApproxInference(model)
print(
    dbn_inf.query(
        [("A", 0)],
        evidence={("B", 0): 1, ("C", 0): 0},
        state_names={("A", 0): ["A = 0", "A = 1"]},
    )
)
```

I wrongly get:

```
+-----------------+-----------------+
| ('A', 0)        |   phi(('A', 0)) |
+=================+=================+
| ('A', 0)(A = 0) |          0.0000 |
+-----------------+-----------------+
| ('A', 0)(A = 1) |          0.0000 |
+-----------------+-----------------+
```

But if I simply comment out `state_names`, I correctly get:

```
+---------------+-----------------+
| ('A', 0)      |   phi(('A', 0)) |
+===============+=================+
| ('A', 0)(0.0) |          0.4800 |
+---------------+-----------------+
| ('A', 0)(1.0) |          0.5200 |
+---------------+-----------------+
```

## **Cause**

If you print `df` and `df_index` inside `ApproxInference._get_factor_from_df()`, you get:

```
# df
(A, 0)
0.0    0.5016
1.0    0.4984
dtype: float64

# df_index 
['A = 0', 'A = 1']
```
None of the values in `df_index` exist in the index of `df`, so by doing `df.reindex(df_index)`, that will be put to `NaN`, i.e., 

```
# df.reindex(df_index)
(A, 0)
A = 0   NaN
A = 1   NaN
dtype: float64
```

## Possible Solution

Use this inside `ApproxInference._get_factor_from_df(df, state_names)`:

```
if len(variables) == 1:
      # Remove this: df_index = state_names[variables[0]]
      df_index = range(0, len(state_names[variables[0]]))
```

Mudassiruddin7:

- Fixes #1941

- `pgmpy/inference/ApproxInference.py`: query() now uses self.model.states as 
  the canonical state list, with fallback to sample-derived states
- `pgmpy/tests/test_inference/test_ApproxInference.py`: test with a heavily skewed 
  distribution where one state has near-zero probability

### DIFF
--- a/pgmpy/inference/ApproxInference.py
+++ b/pgmpy/inference/ApproxInference.py
@@ -190,20 +190,16 @@ class ApproxInference(object):
         # possible states are represented (sampling may miss rare states).
         if state_names is None:
             if hasattr(self.model, "states") and self.model.states:
-                state_names = {
-                    var: self.model.states[var]
-                    for var in variables
-                    if var in self.model.states
-                }
-                # Fall back to sample-derived states for any missing variables
+                state_names = {}
                 for var in variables:
-                    if var not in state_names:
-                        if isinstance(self.model, DynamicBayesianNetwork):
-                            state_names[var] = list(
-                                samples.loc[:, [var]].iloc[:, 0].unique()
-                            )
-                        else:
-                            state_names[var] = list(samples.loc[:, var].unique())
+                    if var in self.model.states:
+                        state_names[var] = list(self.model.states[var])
+                    elif isinstance(self.model, DynamicBayesianNetwork):
+                        state_names[var] = list(
+                            samples.loc[:, [var]].iloc[:, 0].unique()
+                        )
+                    else:
+                        state_names[var] = list(samples.loc[:, var].unique())
             elif isinstance(self.model, DiscreteBayesianNetwork):
                 state_names = {
                     var: list(samples.loc[:, var].unique()) for var in variables
--- a/pgmpy/inference/ApproxInference.py
+++ b/pgmpy/inference/ApproxInference.py
@@ -186,9 +186,25 @@ class ApproxInference(object):
                     seed=seed,
                 )
 
-        # Step 2: If state_names is None, infer it from samples.
+        # Step 2: If state_names is None, read from model to ensure all
+        # possible states are represented (sampling may miss rare states).
         if state_names is None:
-            if isinstance(self.model, DiscreteBayesianNetwork):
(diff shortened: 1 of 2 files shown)


Mudassiruddin7:
query() in ApproxInference was inferring the state space entirely from the sampled data. If a state has very low probability, it may simply never appear in the sample, and then it's missing from the result entirely. The returned factor has the wrong shape, the wrong state index mapping, and is silently wrong in a way that's hard to catch without knowing what the correct answer should be. The fix is to use self.model.states as the canonical list of states for each variable, which is what the model was defined with, and only fall back to sample-derived states when self.model.states isn't available. The sample counts still come from the sampling, but the state ordering and completeness come from the model definition. The new test uses a network with a state that has probability around 0.001 so it almost never appears in any sample, and verifies that the state is still present in the query output.

DARHWOLF:
@Mudassiruddin7 I do not think you have been assigned with this isssue. 

Please close this PR

Mudassiruddin7:
hey @DARHWOLF!! you're right that I wasn't formally assigned before opening this. I commented on issue #1941 when I submitted the PR. I'll leave it to @ankurankan to decide whether to keep this open or close it.

ankurankan:
@Mudassiruddin7 Please do not open PRs without getting the issue assigned to you. As nobody else has commented on the issue, I am assigning the issue to you.

PR #2739, opened 2026-02-28
fix(inference): kwargs not forwarded in query, NaN ATE on empty adjustment set, bare except handlers

### LINKED ISSUE #1616: bug: estimate_ate() will return nan if backdoor is null.
For example, there is a dag looks like: [('T','B'),('B','Y'),('Z','Y'),('Z','T')]
Then if we want to use estimate_ate() to get the ate from T to Y. Then estimate_ate() will return nan since there is no backdoor from T to B.

### LINKED ISSUE #2401: [BUG] "query" function in CausalInference.py ignores **kwargs
**Describe the bug**

The "query" function in CausalInference.py can use VariableElimination or  BeliefPropagation algorithms. Both those algorithms can handle virtual evidence natively. However, passing virtual_evidence as a **kwarg in the "query" function within CausalInference does nothing and is ignored completely.

**To Reproduce**

You can either pass a simple virtual_evidence CPD in a toy network and watch it get ignored or you can check the code and see that **kwargs are not passed to any of the "infer.query" calls in the function (which is where virtual_evidence for example would be passed to VariableElimination or BeliefPropagation algorithms).

**Expected behavior**
CausalInference.query should be able to pass **kwargs to the inference algorithms.

**Versions**

pgmpy==1.0.0
python==3.13.2

Mudassiruddin7:

- Fixes #2401
- Fixes #1616

- `pgmpy/inference/CausalInference.py`: **kwargs forwarded to all 5 infer.query() 
  call sites; frozenset return value corrected; 5 bare except: changed to 
  except Exception:
- `pgmpy/estimators/LinearModel.py`: fitted result stored in self.result instead 
  of overwriting self.estimator
- `pgmpy/inference/LinearModel.py`: same self.result fix
- `pgmpy/inference/ApproxInference.py`: if joint == True changed to if joint
- `pgmpy/tests/test_inference/test_CausalInference.py`: test for kwargs 
  propagation via mock; test for estimate_ate on unconfounded treatment-outcome pair

### DIFF
--- a/pgmpy/estimators/LinearModel.py
+++ b/pgmpy/estimators/LinearModel.py
@@ -1,5 +1,5 @@
 import statsmodels.api as sm
-from statsmodels.api import OLS, GLS, WLS
+from statsmodels.api import GLS, OLS, WLS
 
 
 class LinearEstimator(object):
@@ -22,12 +22,12 @@ class LinearEstimator(object):
         return self.estimator(endog=endog, exog=exog, **kwargs)
 
     def fit(self, X, Y, Z, data, **kwargs):
-        self.estimator = self._model(X, Y, Z, data, **kwargs).fit()
-        self.ate = self.estimator.params[X]
+        self.result = self._model(X, Y, Z, data, **kwargs).fit()
+        self.ate = self.result.params[X]
         return self
 
     def _get_ate(self):
         return self.ate
 
     def summary(self):
-        return self.estimator.summary()
+        return self.result.summary()
--- a/pgmpy/inference/ApproxInference.py
+++ b/pgmpy/inference/ApproxInference.py
@@ -71,7 +71,7 @@ class ApproxInference(object):
         if isinstance(variables, (set, tuple)):
             variables = list(variables)
 
-        if joint == True:
+        if joint:
             return self._get_factor_from_df(
                 samples.groupby(variables, observed=False).size() / samples.shape[0],
                 state_names,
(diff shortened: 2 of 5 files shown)


Mudassiruddin7:
This PR fixes four independent issues I found while working through CausalInference.py. The most user-visible one is that query() accepted **kwargs but never forwarded them to any of the five internal infer.query() calls, so passing elimination_order or show_progress silently did nothing. The estimate_ate NaN issue turned out to be two stacked bugs: get_all_backdoor_adjustment_sets was returning frozenset() when the empty set is a valid adjustment set, but it should return frozenset({frozenset()}), a frozenset containing the empty frozenset. That one-character difference completely broke iteration over the result. On top of that, LinearEstimator.fit() was overwriting self.estimator (the class reference itself) with the fitted result object, so a second call to fit would try to call .fit() on an already-fitted OLS object. Storing the fitted result in self.result instead fixes it. The five bare except: handlers in identification_method() were catching KeyboardInterrupt which made it impossible to Ctrl+C out of a long-running identification call, so I narrowed them to except Exception:. The == True comparison in ApproxInference is a minor PEP 8 issue but worth cleaning up in the same session.

ankurankan:
@Mudassiruddin7 Please do not combine issues in a single PR. You will have to create separate PRs for each of them.

Mudassiruddin7:
> @Mudassiruddin7 Please do not combine issues in a single PR. You will have to create separate PRs for each of them.

understood, apologies for bundling them together. I'll close this PR and open separate ones for each issue. to confirm the split:

- #2401 (kwargs not forwarded) → one PR
- #1616 (estimate_ate NaN / empty adjustment set) → one PR
- the bare except and == True cleanups → I'll fold these into whichever 
  related PR makes most sense, or open a small maintenance PR if you prefer them separate

let me know if that split looks right and I'll get the new PRs up shortly.

ankurankan:
Closing this.

PR #2740, opened 2026-02-28
fix(models): state_names lost on fit, incorrect marginalization in remove_node, BP thread safety in predict

### LINKED ISSUE #2007: Incorrect Marginalization Logic in `DiscreteBayesianNetwork.remove_node`

### LINKED ISSUE #2207: [BUG] model.fit() overwrites user-defined state_names when fitting on integer data
**Describe the bug**

The model.fit() method destructively overwrites CPDs that contain user-defined state_names.
If a user defines a CPD with string state names (e.g., {'A': ['expert_knowledge', 'default']}) and then fits the model on data where 'A' is represented by integers (0, 1), the original state_names are permanently discarded and replaced by the integers.

While a warning about replacing the CPD is shown, it's not obvious that this will also destroy the state name mapping. This leads to two main problems:
--Loss of Information: The model loses the connection between the integers and their string labels.
--Failed Queries: Subsequent inference using the original string labels (e.g., evidence={'A': 'expert_knowledge'}) will fail.

The expected behavior is for .fit() to use the existing state_names to interpret the numerical data, rather than deleting them. This aligns with the common practice in data science libraries of preserving, rather than overwriting, user-defined model information during fitting.
This would prevent users from having to manually pre-process their data to avoid this issue.

**To Reproduce**

```python
import pandas as pd
from pgmpy.estimators import MaximumLikelihoodEstimator

# Define a model with an expert-defined CPD for 'A'
# The states for 'A' are meaningful strings.
model = DiscreteBayesianNetwork([('A', 'B')])
cpd_a = TabularCPD(
    variable='A', variable_card=2,
    values=[[0.7], [0.3]],
    state_names={'A': ['expert_knowledge', 'default']}
)
model.add_cpds(cpd_a)

# Capture the correct, original state names before fitting
original_a_state_names = model.get_cpds('A').state_names['A']

# Create data where 'A' is represented by integers
data = pd.DataFrame(data={
    'A': [1, 0, 1, 1, 0],
    'B': ['x', 'y', 'x', 'y', 'x']
})

# Fit the model. This is where the state names are overwritten.
print("--- Fitting model on integer data ---")
model.fit(data, estimator=MaximumLikelihoodEstimator)
print("--- Fit complete ---\n")

# Get the newly learned CPD for 'B' and show the state name inconsistency
cpd_b_learned = model.get_cpds('B')

print("Original state names for 'A' (before fit):")
print(original_a_state_names)
print("\nState names for 'A' from the new P(B|A)'s perspective (after fit):")
print(cpd_b_learned.state_names['A'])
```

**Expected behavior**
The 'fit()' method should be non-destructive. When it encounters a variable that already has a user-defined CPD with 'state_names', it should:
1.  **Not** replace the existing CPD or its 'state_names'.
2.  Use the existing 'state_names' list to interpret the integer data. For example, it should map the integer '0' in the data to the state 'expert_knowledge' and 1 to 'default'.
3. The newly learned CPD for the child node (P(B|A)) should correctly reference its parent's string-based states.

**Current behavior**

WARNING:pgmpy:Replacing existing CPD for A

Original state names for 'A' (before fit):
['expert_knowledge', 'default']

State names for 'A' from the new P(B|A)'s perspective (after fit):
[0, 1]

**Proposed fix**

The logic within model.fit() and the MaximumLikelihoodEstimator should be modified. If pre-defined state_names exist for a variable, the estimator should attempt to map the data to these names. If the mapping is successful, it should use the mapped states and skip the step of inferring states from the data. If the mapping seems unreasonable(for example, if the data contains values like 0 and 2 when the model was only defined with two states), it can revert to its current behavior.

I am currently working on a fix for this issue and plan to submit a pull request soon.

### LINKED ISSUE #2274: [BUG] Fix flaky test `TestBayesianNetworkFitPredict::test_predict`
The test `pgmpy/tests/test_models/test_DiscreteBayesianNetwork.py::TestBayesianNetworkFitPredict::test_predict` occasionally fails on GitHub Actions. We need to figure out why that is happening and how to fix it. The full traceback:
```bash
_________________ TestBayesianNetworkFitPredict.test_predict __________________
joblib.externals.loky.process_executor._RemoteTraceback: 
"""
  File "C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\_utils.py", line 72, in __call__
    return self.func(**kwargs)
           ~~~~~~~~~^^^^^^^^^^
  File "C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\parallel.py", line 607, in __call__
    return [func(*args, **kwargs) for func, args, kwargs in self.items]
            ~~~~^^^^^^^^^^^^^^^^^
  File "D:\a\pgmpy\pgmpy\pgmpy\inference\ExactInference.py", line 1306, in map_query
    final_distribution = self._query(
        variables=variables,
    ...<3 lines>...
        show_progress=show_progress,
    )
  File "D:\a\pgmpy\pgmpy\pgmpy\inference\ExactInference.py", line 1081, in _query
    clique_potential_list = [self.clique_beliefs[root_node]]
                             ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
KeyError: ('Pclass', 'Survived', 'Sex')
"""

The above exception was the direct cause of the following exception:

self = <pgmpy.tests.test_models.test_DiscreteBayesianNetwork.TestBayesianNetworkFitPredict testMethod=test_predict>

    def test_predict(self):
        titanic = DiscreteBayesianNetwork()
        titanic.add_edges_from([("Sex", "Survived"), ("Pclass", "Survived")])
        titanic.fit(self.titanic_data2[500:])
    
        p1_ve = titanic.predict(self.titanic_data2[["Sex", "Pclass"]][:30])
    
>       p1_bp = titanic.predict(
            self.titanic_data2[["Sex", "Pclass"]][:30], algo=BeliefPropagation
        )

pgmpy\tests\test_models\test_DiscreteBayesianNetwork.py:1113: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pgmpy\models\DiscreteBayesianNetwork.py:799: in predict
    pred_values = Parallel(n_jobs=n_jobs, require="sharedmem")(
C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\parallel.py:2072: in __call__
    return output if self.return_generator else list(output)
                                                ^^^^^^^^^^^^
C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\parallel.py:1682: in _get_outputs
    yield from self._retrieve()
C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\parallel.py:1784: in _retrieve
    self._raise_error_fast()
C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\parallel.py:1859: in _raise_error_fast
    error_job.get_result(self.timeout)
C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\parallel.py:758: in get_result
    return self._return_or_raise()
           ^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <joblib.parallel.BatchCompletionCallBack object at 0x00000254DB8C9FD0>

    def _return_or_raise(self):
        try:
            if self.status == TASK_ERROR:
>               raise self._result
E               KeyError: ('Pclass', 'Survived', 'Sex')

C:\hostedtoolcache\windows\Python\3.13.5\x64\Lib\site-packages\joblib\parallel.py:773: KeyError
```

Mudassiruddin7:

- Fixes #2207
- Fixes #2007
- Fixes #2274

- `pgmpy/models/DiscreteBayesianNetwork.py`: fit() now extracts state_names from 
  existing CPDs when none explicitly passed; remove_node() weights by actual marginal 
  before summing; predict() drops require='sharedmem' from Parallel() call
- `pgmpy/tests/test_models/test_DiscreteBayesianNetwork.py`: three new tests, one 
  per bug

### DIFF
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -2,6 +2,7 @@
 
 import itertools
 from collections import defaultdict
+from copy import deepcopy
 from functools import reduce
 from operator import mul
 from typing import (
@@ -941,8 +942,12 @@ class DiscreteBayesianNetwork(DAG):
             lambda t: t.index.tolist()
         )
         data_unique = data_unique_indexes.index.to_frame()
-        pred_values = Parallel(n_jobs=n_jobs)(
-            delayed(model_inference.query if stochastic else model_inference.map_query)(
+        pred_values = Parallel(n_jobs=n_jobs, require="sharedmem")(
+            delayed(
+                deepcopy(model_inference).query
+                if stochastic
+                else deepcopy(model_inference).map_query
+            )(
                 variables=missing_variables.union(
                     set(data_point.index[data_point.isna()])
                 ),
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -201,11 +201,58 @@ class DiscreteBayesianNetwork(DAG):
          <TabularCPD representing P(D:2) at 0x7f28248e26a0>]
         """
         affected_nodes = [v for u, v in self.edges() if u == node]
+        removed_cpd = self.get_cpds(node=node)
 
         for affected_node in affected_nodes:
-            node_cpd = self.get_cpds(node=affected_node)
-            if node_cpd:
-                node_cpd.marginalize([node], inplace=True)
+            child_cpd = self.get_cpds(node=affected_node)
+            if child_cpd:
+                if removed_cpd:
+                    # Weight by the removed node's distribution before marginalizing
+                    # so that P(child | rest) = sum_node P(child | node, rest) * P(node | ...)
+                    child_factor = child_cpd.to_factor()
+                    removed_factor = removed_cpd.to_factor()
+                    product = child_factor * removed_factor
+                    product.marginalize([node], inplace=True)
+                    product.normalize(inplace=True)
+
+                    # Rebuild the TabularCPD from the marginalized factor
+                    new_evidence = [v for v in child_cpd.variables[1:] if v != node]
+                    new_evidence += [
+                        v
+                        for v in product.variables
+                        if v != affected_node and v not in new_evidence
+                    ]
+                    if new_evidence:
+                        new_evidence_card = [
+                            int(product.get_cardinality([v])[v]) for v in new_evidence
+                        ]
+                    else:
+                        new_evidence_card = None
+
+                    new_cpd = TabularCPD(
+                        variable=affected_node,
+                        variable_card=int(
+                            product.get_cardinality([affected_node])[affected_node]
+                        ),
+                        values=product.values.flatten("C").reshape(
+                            int(
+                                product.get_cardinality([affected_node])[affected_node]
+                            ),
+                            -1,
+                        ),
+                        evidence=new_evidence if new_evidence else None,
+                        evidence_card=new_evidence_card,
+                        state_names={
+                            v: product.state_names[v] for v in product.variables
+                        },
+                    )
+                    # Reorder to match the factor's variable ordering
+                    if new_evidence:
+                        new_cpd.reorder_parents(new_evidence, inplace=True)
+                    self.remove_cpds(affected_node)
+                    self.add_cpds(new_cpd)
+                else:
+                    child_cpd.marginalize([node], inplace=True)
 
         if self.get_cpds(node=node):
             self.remove_cpds(node)
@@ -608,7 +655,7 @@ class DiscreteBayesianNetwork(DAG):
         mm = self.to_markov_model()
         return mm.to_junction_tree()
 
-    def fit(self, data, estimator=None, state_names=[], n_jobs=1, **kwargs) -> "DAG":
+    def fit(self, data, estimator=None, state_names=None, n_jobs=1, **kwargs) -> "DAG":
         """
         Estimates the CPD for each variable based on a given data set.
 
@@ -670,6 +717,13 @@ class DiscreteBayesianNetwork(DAG):
         else:
             if not issubclass(estimator, BaseEstimator):
                 raise TypeError("Estimator object should be a valid pgmpy estimator.")
+
+        # If state_names not explicitly provided, preserve any existing CPD state_names
+        # so that states present in the model but absent from data are not lost.
+        if not state_names and bn.cpds:
+            state_names = {}
+            for cpd in bn.cpds:
+                state_names.update(cpd.state_names)
 
         _estimator = estimator(
             bn,
@@ -883,7 +937,7 @@ class DiscreteBayesianNetwork(DAG):
             lambda t: t.index.tolist()
         )
         data_unique = data_unique_indexes.index.to_frame()
-        pred_values = Parallel(n_jobs=n_jobs, require="sharedmem")(
+        pred_values = Parallel(n_jobs=n_jobs)(
             delayed(model_inference.query if stochastic else model_inference.map_query)(
                 variables=missing_variables.union(
                     set(data_point.index[data_point.isna()])
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -212,16 +212,19 @@ class DiscreteBayesianNetwork(DAG):
                     child_factor = child_cpd.to_factor()
                     removed_factor = removed_cpd.to_factor()
                     product = child_factor * removed_factor
+                    # First marginalize out the removed node itself.
                     product.marginalize([node], inplace=True)
-                    product.normalize(inplace=True)
-
-                    # Rebuild the TabularCPD from the marginalized factor
+                    # Restrict the reconstructed CPD to the child's existing parents only.
+                    # Any extra variables introduced via removed_cpd are marginalized out so that
+                    # the CPD's evidence matches the graph parents, avoiding inconsistency.
                     new_evidence = [v for v in child_cpd.variables[1:] if v != node]
-                    new_evidence += [
+                    extra_vars = [
                         v
                         for v in product.variables
                         if v != affected_node and v not in new_evidence
                     ]
+                    if extra_vars:
+                        product.marginalize(extra_vars, inplace=True)
                     if new_evidence:
                         new_evidence_card = [
                             int(product.get_cardinality([v])[v]) for v in new_evidence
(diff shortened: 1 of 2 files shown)


Mudassiruddin7:
Three correctness bugs in DiscreteBayesianNetwork, all in the same file. The fit() state_names issue came from the default being an empty list that was forwarded to the estimator, which interpreted it as "infer from scratch" and overwrote any names already set on existing CPDs. The fix is to check whether CPDs already have state_names before deciding what to pass to the estimator. The remove_node() marginalization bug was more subtle: the existing code summed uniformly across the removed parent's states, which is only mathematically correct when that parent has a uniform distribution. The proper approach is to weight by the parent's actual marginal CPD before summing, then renormalize. I verified this against hand-computed values for a small asymmetric network. The predict() threading bug was the trickiest to reproduce reliably: passing require='sharedmem' to Parallel() uses threads instead of processes, so all workers shared the same BeliefPropagation instance. Since _query() modifies clique_beliefs in place during calibration, concurrent threads were corrupting each other's state and producing intermittent KeyError on clique lookup. Removing sharedmem switches to loky process-based parallelism where each worker gets its own copy of the inference object via pickling.

ankurankan:
Closing this:

1. Tests are failing and haven't been fixed in a while.
2. One PR for one issue.

PR #2748, opened 2026-02-28
Fix virtual evidence corrupting model state in VE and BP inference

WHOIM1205:
### Bug Summary

`_virtual_evidence()` internally calls `self.__init__(bn)`, which mutates `self.model` by adding auxiliary `__X` nodes.

In `query()` and `map_query()` (both `VariableElimination` and `BeliefPropagation`), virtual evidence was handled using a **recursive call pattern**:

- The outer call saved `orig_model`
- `_virtual_evidence()` modified `self.model`
- A recursive `query()`/`map_query()` was invoked
- The inner call captured the already-mutated model as `orig_model`
- Restoration became a no-op

Result:
- Phantom `__X` nodes accumulated
- Silent model corruption
- Incorrect posteriors in subsequent queries
- BeliefPropagation/VariableElimination objects became unsafe to reuse

---

### Fix Applied

Replaced the recursive pattern with a **linear execution flow** using `try/finally`:

1. Save original model before calling `_virtual_evidence()`
2. Merge generated virtual evidence into `evidence`
3. Fall through to normal inference path
4. Restore original model in a `finally` block

This guarantees restoration:
- Even on exceptions
- Even with early returns
- Without recursive state capture bugs

---

### 🔧 Modified Methods

- `VariableElimination.query()`
- `VariableElimination.map_query()`
- `BeliefPropagation.query()`
- `BeliefPropagation.map_query()`

All now follow:

```python
orig_model = self.model.copy()

try:
    self._virtual_evidence(...)
    evidence = {...}
    result = run_inference(...)
finally:
    self.__init__(orig_model)

### DIFF
--- a/pgmpy/inference/ExactInference.py
+++ b/pgmpy/inference/ExactInference.py
@@ -321,66 +321,72 @@ class VariableElimination(Inference):
             )
 
         # Step 2: If virtual_evidence is provided, modify the network.
+        orig_model = None
         if isinstance(self.model, DiscreteBayesianNetwork) and (
             virtual_evidence is not None
         ):
+            orig_model = self.model.copy()
             self._virtual_evidence(virtual_evidence)
             virt_evidence = {"__" + cpd.variables[0]: 0 for cpd in virtual_evidence}
-            return self.query(
-                variables=variables,
-                evidence={**evidence, **virt_evidence},
-                virtual_evidence=None,
-                elimination_order=elimination_order,
-                joint=joint,
-                show_progress=show_progress,
-            )
+            evidence = {**evidence, **virt_evidence}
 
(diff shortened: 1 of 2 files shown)


WHOIM1205:
@ankurankan 
Fixes model corruption caused by recursive handling of `virtual_evidence` in `query()` and `map_query()`.

Previously, `_virtual_evidence()` mutated `self.model`, and the recursive call pattern captured the already-modified model, making restoration ineffective. This led to accumulation of phantom `__X` nodes and incorrect results in subsequent queries.

Replaced the recursive flow with a linear `try/finally` pattern that saves the original model before modification and guarantees proper restoration after inference.

No API changes. Fully backward compatible.

PR #2772, opened 2026-03-02
ENH: Add Jobs (Lalonde dataset)

### LINKED ISSUE #2620: [ENH] Add causal estimation + causal discovery datasets

- Add  publicly available causal estimation and causal discovery  datasets in the package
 from: https://github.com/rguo12/awesome-causality-data,
 using the extension template: https://github.com/pgmpy/pgmpy/blob/dev/devtools/extension_templates/_dataset.py

- parent issue: https://github.com/pgmpy/pgmpy/issues/2419#issue-3628409669

**Additional context**

KahnSvaer:

- [] Are all the GitHub Actions checks passing? If not, they will need to be fixed before review. You can reference logs for the failing check to identify the issue.

- Fixes #2620

- Added Lalonde (Jobs) dataset specifically the Dehejia-Wahha split and the PSID_control split 
- Added a mixin class on lalonde.py (named so as this seems to be the most common name the dataset seems to go with)
- Added class names to `tests_dataset`

### DIFF
--- a/pgmpy/datasets/lalonde.py
+++ b/pgmpy/datasets/lalonde.py
@@ -1,0 +1,132 @@
+import io
+
+import pandas as pd
+
+from pgmpy.datasets._base import _BaseDataset
+
+
+class _JobsMixin:
+    _common_tags = {
+        "n_variables": 10,
+        "has_ground_truth": False,
+        "has_expert_knowledge": False,
+        "has_missing_data": False,
+        "has_index_col": False,
+        "is_simulated": False,
+        "is_discrete": False,
+        "is_continuous": False,
+        "is_mixed": True,
+        "is_ordinal": False,
+    }
+
+    columns = [
+        "training",
+        "age",
+        "education",
+        "black",
+        "hispanic",
+        "married",
+        "no_degree",
+        "re74",
+        "re75",
+        "re78",
+    ]
+
+    categorical_variables = [
+        "training",
+        "black",
+        "hispanic",
+        "married",
+        "no_degree",
+    ]
+
+    ordinal_variables = dict()
+
+
+class JobsLalonde(_JobsMixin, _BaseDataset):
+    """
+    References
+    ----------
+    .. [1] https://users.nber.org/~rdehejia/nswdata2.html
+    .. [2] Rajeev Dehejia and Sadek Wahba, "Causal Effects in Non-Experimental Studies:
+    Reevaluating the Evaluation of Training Programs," Journal of the American Statistical
+    Association, Vol. 94, No. 448 (December 1999), pp. 1053-1062.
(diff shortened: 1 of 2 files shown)


KahnSvaer:
@ankurankan @DARHWOLF  Please review. This is a dataset from the [`rguo12/awesome-causality-data`](https://github.com/rguo12/awesome-causality-data) repo. Reference of the dataset authoritative path - [`~rdehejia/nswdata2.html`](https://users.nber.org/~rdehejia/nswdata2.html). 

**Description of the dataset**
The dataset comes from a randomized job-training experiment. A group of disadvantaged workers applied for the program, and among these eligible individuals some were randomly assigned to receive training (treated) while others were randomly assigned not to receive it (control). Their later earnings were compared to measure the causal effect of the training.
Because both the treated and control groups come from the same eligible population and were randomized, they are highly comparable and their covariate distributions have strong overlap. This experimental comparison gives the ground-truth treatment effect.
The additional CPS and PSID files are non-experimental comparison groups drawn from general population surveys. These are used to replace the experimental control group in order to study what happens in an observational setting, where the treated group is compared to people from a different population.

For a minimal dataset, we could use the concatenated Dehejia–Wahba sample, since this is the standard experimental benchmark for causal methods. The trade-off is that this would limit us to the Random Control Trial settings, we wouldn’t be able to compare the treated group to CPS/PSID controls, so the Lalonde regression failure case and overlap analysis would not be possible.
For now, I have added a single PSID control file. The others follow the exact same pattern, only the `data_url` in the `JobsPSID` class (`datasets/lalonde.py`) needs to be updated. I can include the remaining files too if that would be preferred.

KahnSvaer:
`test_datasets` is passing, tests seem to be failing on something unrelated.

ankurankan:
@KahnSvaer Thanks for the PR. I would suggest a slightly different approach here. I see that there are two interventional variants to the dataset and six observational ones. I think a good way would be to name all of these datasets as `lalonde/<variant>`. We can also create a directory under `example_datasets` called `lalonde`, where we put all the related dataset classes/files. As the number of datasets are increasing this will also help us in somewhat create groups making it a bit easier to navigate (we have been doing something similar for `example_models`).

Could you please create a new issue and outline this approach there? And then update this PR?

ankurankan (changes requested):
Please refer comment.

KahnSvaer:
@ankurankan Thanks, I’ve added #2780.

I also wanted your input on one thing — for the example models, would you prefer that I keep the files as they are, or should I standardize them to match the existing dataset formats?

Since I’ve already overridden the `load_dataset` method inside the respective classes, either approach should work — just let me know your preference.

ankurankan:
Blocked by: https://github.com/pgmpy/example_datasets/pull/7

PR #2795, opened 2026-03-03
Fix/lgbn do 2714

### LINKED ISSUE #2714: [BUG] LinearGaussianBayesianNetwork.do falls back to DAG.do as LGBN doesn't implement a `do` method.
## Describe the bug
There is a critical architectural gap in the `LinearGaussianBayesianNetwork` (LGBN) class concerning interventional queries (do-calculus). While `DiscreteBayesianNetwork` provides a formal `.do()` method that explicitly severs incoming edges and marginalizes the corresponding Conditional Probability Distributions (CPDs) to produce a valid intervened network, this method is completely missing from `LinearGaussianBayesianNetwork`.

Instead, the logic to handle interventions for continuous variables is inappropriately "leaked" and hardcoded directly inside the `.simulate()` method. In `simulate()`, the model creates a copy, severs incoming edges to intervened nodes, removes their CPDs, and then structurally updates the `beta` (intercept) coefficients of their child nodes. Because this logic is buried inside a sampling function, it is impossible to programmatically manipulate an LGBN's structure for interventional inference, visualization, or exporting without actually running a full simulation loop.

When attempting to call `.do()` on an LGBN, it falls back to the base `DAG.do()` method. This correctly modifies the graph structure (edges are removed), but it blindly copies the continuous parameters (`LinearGaussianCPD` objects) without adjusting them. This results in a corrupted model state where the CPDs still "expect" their original parents as evidence, causing a structural inconsistency that fails `check_model()`.

## To Reproduce

Run the below program 
```python
import sys
from pgmpy.models import LinearGaussianBayesianNetwork
from pgmpy.factors.continuous import LinearGaussianCPD

# 1. Setup a simple Linear Gaussian Bayesian Network: X -> Y -> Z
model = LinearGaussianBayesianNetwork([("X", "Y"), ("Y", "Z")])
cpd_x = LinearGaussianCPD("X", [0], 1)
cpd_y = LinearGaussianCPD("Y", [0, 2], 1, ["X"])
cpd_z = LinearGaussianCPD("Z", [0, 3], 1, ["Y"])
model.add_cpds(cpd_x, cpd_y, cpd_z)

print(f"Original Model Valid? {model.check_model()}")

try:
    # 2. Attempt formal intervention on Y: do(Y)
    # The base DAG.do() removes the incoming edges but blindly copies the Continuous CPDs
    intervened_model = model.do(["Y"])
    
    # 3. Check if the returned object is structurally valid
    # This will raise a ValueError because CPD_Y still references parent 'X' 
    # despite the edge being severed by DAG.do()
    intervened_model.check_model()

except ValueError as e:
    print(f"\n[Structural Validity Error]: {e}")
    print("\nExplanation: DAG.do() severed the structure but didn't update the LinearGaussianCPD.")
    print("The model is now corrupted as the CPDs are out of sync with the graph structure.")

### PR DESCRIPTION
**Pull request by ShyamSathish005**

- [x] `LinearGaussianBayesianNetwork` was basically a second-class citizen for causal interventions — discrete networks had a clean `.do()` method, but the Gaussian version made you run a full simulation just to peek at an intervened structure.
I've fixed this with a proper `.do()` override for LGBNs. It doesn't just cut edges — it pushes the intervened constant through the Beta coefficients to correctly update child node intercepts. I verified this by hand on a simple **X → Y** chain and matched it against the code output. I also made sure calling `.do()` without values does the structural surgery without crashing, and refactored `simulate()` to use the new logic, so the codebase is much DRYer.

---

- Fixes #2714

#### 1. Mathematical Implementation of the Interventional do-operator

In a Linear Gaussian model, variable $Y$ with parents $X_i$ is defined by:
```math
Y \mid X_1, \dots, X_k \sim \mathcal{N}\left(\beta_0 + \sum_{j=1}^{k} \beta_j X_j, \sigma^2\right)
```

When applying \operatorname{do}(X_i = c)$, this PR implements the "Intercept Shift":
```math
\beta_{0_{\text{new}}} = \beta_{0_{\text{old}}} + \beta_{X_i} \cdot c
```

The edge $X_i \to Y$ is then severed, and the evidence list for $Y$ is updated to exclude $X_i$.

#### 2. Architectural Refactoring
* **Decoupling Logic:** I moved the interventional logic out of `simulate()` and into a dedicated `do()` method within the `LinearGaussianBayesianNetwork` class. This ensures that interventions are a first-class structural property of the model rather than just a sampling side-effect.
* **Consistency:** Refactored `simulate()` to internally call `self.do()`, reducing code duplication by about 25 lines and ensuring that any intervention follows the same mathematical path.
* **Numerical Stability:** Intervened nodes are now replaced with a CPD having a tiny variance epsilon ($10^{-10}$) instead of a hard zero to prevent singular covariance matrices during joint Gaussian operations.

#### 3. Manual Verification of Scenarios
I have manually tested the following scenarios on an **Ubuntu 24.04** environment:
* **Intervening on Root vs. Leaf:** Verified that root interventions correctly propagate to children, while leaf interventions only affect the node itself.
* **In-place Mutability:** Confirmed that `inplace=False` (default) creates a deep copy, preserving the original model's CPDs.
* **Multiple Interventions:** Tested cases where two parents are intervened on simultaneously, ensuring the intercept shift sums correctly: $\beta_0 + (\beta_1 \cdot c_1) + (\beta_2 \cdot c_2)$.

---

### GSoC 2026 Context
This PR is a foundational contribution toward the **"Unified BN Interface"** and **"Pairwise Causal Discovery"** project ideas. It ensures that continuous models are no longer "second-class citizens" in the pgmpy interventional API, allowing discovery algorithms (like ANM/IGCI) to query post-intervention states natively.

### DIFF
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -520,6 +520,138 @@ class LinearGaussianBayesianNetwork(DAG):
             model_copy.add_cpds(*[cpd.copy() for cpd in self.cpds])
         return model_copy
 
+    def do(
+        self,
+        nodes: Union[Hashable, List[Hashable]],
+        values: Optional[Dict[Hashable, float]] = None,
+        inplace: bool = False,
+    ) -> "LinearGaussianBayesianNetwork":
+        r"""
+        Applies the do-operator to the Linear Gaussian Bayesian Network.
+
+        The do-operator, :math:`do(X = x)`, has the effect of removing all
+        incoming edges to :math:`X`, fixing :math:`X` to the constant
+        :math:`x`, and propagating that constant through every child's CPD
+        via the **intercept-shift rule**.
+
+        For a child :math:`Y` whose CPD before intervention is:
+
+        .. math::
+
+            Y \mid \mathbf{Pa}(Y) \sim
+            \mathcal{N}\!\bigl(\beta_0 + \beta_X X
+            + {\textstyle\sum_{j \neq X}} \beta_j X_j,\;
+            \sigma^2\bigr)
+
+        the updated intercept after :math:`do(X = c)` is:
+
+        .. math::
+
+            \beta_{0}^{\text{new}} = \beta_{0}^{\text{old}}
+            + \beta_X \cdot c
+
+        and :math:`X` is removed from :math:`Y`'s evidence list.  The
+        variance :math:`\sigma^2` is unchanged because it is independent of
+        the parent values in a Linear Gaussian model.
+
+        When ``values`` is ``None`` the graph surgery is still performed
+        (all edges severed, children updated) but no intercept shift is
+        applied.  This is equivalent to :math:`do(X = 0)`.
+
+        Parameters
+        ----------
+        nodes : hashable or list of hashable
+            The name(s) of the node(s) to intervene on.
+
+        values : dict, optional (default: None)
+            A dictionary mapping each intervened node to its fixed
+            numerical value, e.g. ``{"X": 2.5}``.
+            Keys that are not in ``nodes`` are silently ignored.
+            If ``None``, every intervened node is set to 0.
+
+        inplace : bool, default False
+            If ``True``, modifies the current model in-place;
+            otherwise returns a modified copy.
+
+        Returns
+        -------
+        LinearGaussianBayesianNetwork
+            The mutilated model (new instance, or ``self`` when
+            ``inplace=True``).
+
+        Examples
+        --------
+        >>> from pgmpy.models import LinearGaussianBayesianNetwork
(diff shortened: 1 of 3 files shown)


ankurankan (changes requested):
I think we need to come up with a better solution than setting `std=0`. Setting the std=0 will lead to many method failing.

ShyamSathish005:
> I think we need to come up with a better solution than setting `std=0`. Setting the std=0 will lead to many method failing.

You are crct, instead of trying to "fake" a constant value using a zero-variance CPD, can we or probably treat the intervened node as a fixed constant and structurally remove it from the active probabilistic model after we've updated its children. Does that sound okay?

ankurankan:
It's better to have a proposed solution discussion on the issue itself rather than here. Can you post there? I don't think your proposed solution would be consistent with the current behavior of Discrete BN.

PR #2810, opened 2026-03-04
Fix duplicate variable in temporal ordering filter

### LINKED ISSUE #2800: [BUG] Duplicate variable in min() call in BaseConstraintEstimator
**Describe the bug**

In pgmpy/estimators/BaseConstraintEstimator.py at line 344, max_order is computed as:
`max_order = min(temporal_ordering[u], temporal_ordering[u])`
The second argument should be temporal_ordering[v], not temporal_ordering[u]. As written, min() always returns temporal_ordering[u], ignoring v's ordering entirely.

**To Reproduce**

The bug is visible directly in the source code at line 344. 
[Online reference](https://github.com/pgmpy/pgmpy/blob/6fdd85b2b56052aa02528588f3765fd080a18dbc/pgmpy/estimators/BaseConstraintEstimator.py#L344C13-L344C72)

**Expected behavior**

The line should be:
`max_order = min(temporal_ordering[u], temporal_ordering[v])`
so that neighbors with a temporal order greater than either u or v are correctly discarded from the separating sets.

**Additional context**
pgmpy/estimators/BaseConstraintEstimator.py, line 344.

Iruos8805:

LLM assistance was used to generate tests to catch the bug in the future.  The test basically covers four scenarios. It tests if a node with a higher temporal rank is correctly excluded from the separating sets of a pair with lower ranks.  It makes sure the results are the same even if you swap the order of the two nodes. It uses a mock CI test to confirm that an edge isn't accidentally deleted when its only possible separator is disqualified. It verifies the fix works across both the PC and Base estimator classes.

I have manually verified that if the bug exists in either `BaseConstraintEstimator.py` or `_base.py`, then the tests fail.  Symmertry edge cases, empty constraints edge cases were considered.

I have reviewed the introduced tests and confirmed there are no `try-except` blocks. The error handling relies on pytest assertions. 

The test basically covers four scenarios. It tests if a node with a higher temporal rank is correctly excluded from the separating sets of a pair with lower ranks.  It makes sure the results are the same even if you swap the order of the two nodes. It uses a mock CI test to confirm that an edge isn't accidentally deleted when its only possible separator is disqualified. It verifies the fix works across both the PC and Base estimator classes.

- Fixes #2800 

- `max_order = min(temporal_ordering[u], temporal_ordering[u])` to `max_order = min(temporal_ordering[u], temporal_ordering[v])` so that neighbors with a temporal order greater than either u or v are correctly discarded from the separating sets.
- In pgmpy/estimators/BaseConstraintEstimator.py at line 344, max_order is computed as:
max_order = min(temporal_ordering[u], temporal_ordering[u])
The second argument should be temporal_ordering[v], not temporal_ordering[u]. As written, min() always returns temporal_ordering[u], ignoring v's ordering entirely. 
- The bug is visible directly in the source code at line 344. 
[Online reference](https://github.com/pgmpy/pgmpy/blob/6fdd85b2b56052aa02528588f3765fd080a18dbc/pgmpy/estimators/BaseConstraintEstimator.py#L344C13-L344C72)


Spinachboul:
Seems like a straightforward change here
@DARHWOLF You can merge this one.

ankurankan (changes requested):
@Iruos8805 Thanks for the fix and the PR. Could you please add a unit test to verify this is working properly? This should have been caught earlier.

ankurankan (changes requested):
@Iruos8805 
1. Can you also fix this in the `_ConstrainMixin` class?
1. Can you compress the tests into a single test case and add the tests for some of the algorithms implemented in `causal_discovery` module such as `PC`?

Iruos8805:
@ankurankan I have made the changes suggested. Could you please review it?

ankurankan:
@Iruos8805 Are the tests AI generated? And why are the PR checklist questions not in the description?

Iruos8805:
@ankurankan 
Yes, the `_fake_ci_temporal` and `test_temporal_ordering_sepsets_and_skeleton` tests were generated with AI assistance. To confirm their validity, I manually reintroduced the bug into the source code to verify that these tests correctly flag the error and made necessary edits to tune them according to the requirement. I have now added the LLM checklist. I didn't include it initially because the intial versions of this PR didn't use AI assistance. Also performed a cleanup of my commits to remove the accidental file wide formatting changes from earlier, ensuring that only the relevant additions are included.

ankurankan (approved):
@Iruos8805 Thanks for the PR.

I also think you used AI in your earlier version of the code. For the future, please keep in mind that we have a very strict policy on the PR checklist, and any violations will result in a ban.

PR #2838, opened 2026-03-06
Fix do() method CPD handling and add test case

s2kumar2007:
The current implementation of the do() method marginalizes parent variables from CPDs, which leads to incorrect behavior.

This PR updates the implementation to correctly handle the intervention and adds a test case to verify the behavior.

Changes

1. I have fixed CPD handling in do() method
2.Added test file to verify intervention behavior
3. Tested the file throught pytest module

- the Fix that i suggest to use reduction in place of marginalize function . using reduction function , we can set the value of the parent node to specific value(for example to setting the value to 0) such that it disapper from the CPD.marginalization averages over a variable. i had another idea to create a new CPD with the child node that only contains the child variable after the intervention.
the method

At first glance, it checks whether adj_model has any CPDs.
if it does, it iterates through each node in the nodes list and retrieves its children.
If the CPD exists, it retrieves the parents of the child node from the CPD's variables
(excluding the first variable which is the child itself).
If there are parents, it creates a list of evidence by setting each parent variable to state 0.(do state
0 is needed because after do operation, the variable will only have one state left which is 0)
Then it elimates the parent variables from the CPD by reducing it with the created evidence.
Finally, it removes the old CPD from the model and adds the new reduced CPD to the model.i used pytest to run the test file and add the document regarding the test file in file itsef .

Did you use a Large language model (LLM) to assist you in making changes?

- [x] The PR description is written manually.
- [x] I manually verified that the changes work as expected.
- [x] No unnecessary try-except blocks were added.
- [x] The test cases verify the correct intervention behavior.
-

### DIFF
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1422,7 +1422,6 @@ class DiscreteBayesianNetwork(DAG):
                 new_cpd.normalize()
                 adj_model.remove_cpds(cpd)
                 adj_model.add_cpds(new_cpd)
-        
 
         """  At first glance, it checks whether adj_model has any CPDs.
         if it does, it iterates through each node in the `nodes` list and retrieves its children.
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1422,6 +1422,7 @@ class DiscreteBayesianNetwork(DAG):
                 new_cpd.normalize()
                 adj_model.remove_cpds(cpd)
                 adj_model.add_cpds(new_cpd)
+        
 
         """  At first glance, it checks whether adj_model has any CPDs.
         if it does, it iterates through each node in the `nodes` list and retrieves its children.
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1405,7 +1405,6 @@ class DiscreteBayesianNetwork(DAG):
 
         model = self if inplace else self.copy()
         adj_model = DAG.do(model, nodes, inplace=inplace)
-    
 
         if adj_model.cpds:
             for node in nodes:
@@ -1417,7 +1416,6 @@ class DiscreteBayesianNetwork(DAG):
                     if cpd is None:
                         continue
 
-                    
                     parents = cpd.variables[1:]
 
                     if parents:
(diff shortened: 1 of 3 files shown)


ankurankan:
Why did you remove the checklist from the PR?

s2kumar2007:
Bymistake i deleted the check list . I have now added the checklist back and verified that the GitHub Actions checks are passing.@ankurankan

ankurankan (commented):
@s2kumar2007 This is fully AI generated. I do not think you understand the issue. Can you explain in your words what and why have you made the changes in this PR?

s2kumar2007:
@ankurankan  the do function remove the the parent incoming edge using the marginalize function .  if they is no parent to the nodes  ,calling the marginalize function with empty list  will create a error in that . so i added if statement to check the parent node is present or not .if the parent node is present it will not execute the the marginalize function . this only the safty check now i understood . the do function forcefully removes the parent node such that child node does not depend on parent node. 
I later realized that this was only a safety improvement and not the actual issue.
the fix that i suggest to use reduction in place of marginalize function . using reduction function , we can set the value of the parent node to specific value(for example to setting the value to 0) such that it disapper from the CPD.marginalization averages over a variable. i had another idea to create a new CPD with the child node that only contains the child variable after the intervention.
the method 
1. At first glance, it checks whether adj_model has any CPDs. 
 2. if it does, it iterates through each node in the `nodes` list and retrieves its children.
 3. for each CPD of the children, it checks if the CPD is not None.This is safty check.
 4. If the CPD exists, it retrieves the parents of the child node from the CPD's variables 
 (excluding the first variable which is the child itself).
  5.If there are parents, it creates a list of evidence by setting each parent variable to state 0.(do state
  0 is needed because after do operation, the variable will only have one state left which is 0)
6. Then it elimates the parent variables from the CPD by reducing it with the created evidence.
 Finally, it removes the old CPD from the model and adds the new reduced CPD to the model.i used pytest to run the test file .and add the document regarding the test file in file itsef .

s2kumar2007:
@ankurankan Could tell why this error is coming ? what is reason behind this...
error 
test_do – array shape mismatch

Actual: [[0.25, 0.75], [0.75, 0.25]]

Expected: [0.5, 0.5]

test_simulate_intervention and test_simulate_virtual_intervention –
ValueError: CPD associated with X doesn't have proper parents associated with it.

Could you please clarify the expected behavior for updating child CPDs after intervention?It seems the graph structure changes after do() but the CPDs may not be updated consistently with the new parent structure

ankurankan:
@s2kumar2007 

> if it does, it iterates through each node in the nodes list and retrieves its children.

There are direct methods to retrieve children of a node.

> for each CPD of the children, it checks if the CPD is not None.This is safty check.

Don't need this check. Run `check_model` at the start of the method.

> 5.If there are parents, it creates a list of evidence by setting each parent variable to state 0.(do state
0 is needed because after do operation, the variable will only have one state left which is 0)

If this is a child node, isn't it guaranteed that the do node will exist in the parent?

ankurankan:
@s2kumar2007 Also please add the full checklist back to the PR.

ankurankan (changes requested):
Please refer comments.

s2kumar2007:
@ankurankan Any idea why The OS macOs is always failing .I have updated the checklist.I think the it failing because of the Circular import error . So i have Opened a issue regarding this in #2986.Could you assign it to me.

s2kumar2007:
@ankurankan I see a contradiction between the PR description 
and the existing test:

1. Issue  says child CPDs should be reduced to the do state
2. But test_do expects C's CPD to remain full shape (2,2,2) 
   unchanged after do(["T"])

Could you clarify:
- Should do() accept a specific state value like do({"T": 0})
- Or should it just make T independent with uniform distribution?

ankurankan:
Changes required:

1. The nodes in do need new CPD that have a single state (the do state)
2. For every child of do variables, we reduce them to the do state of the parent variable. 
3. There should be separate do methods for `DAG`, and `DiscreteBN`. For `DAG`, only the structure changes, for `DiscreteBN` TabularCPDs also change.

PR #2843, opened 2026-03-06
create stale.yml file

Spinachboul:
Addition of Stale PR bot tracker.
Runs every midnight
Checks if the PR has been inactive for 14 days and for issues, 30 days
It runs maximum 50 operations per day.

FYI: @ankurankan @Theavinash02

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,10 +2,9 @@ name: "Community Management: Stale Issues & PRs"
 
 on:
   schedule:
-    # command run on notice
-    # runs at 00:30 midnight rush
+    # Runs at 00:30 UTC every day
     - cron: '30 0 * * *'
-    # to trigger action manually
+  # Allows manual triggering
   workflow_dispatch:
 
 permissions: read-all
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,42 @@
...
+          stale-issue-message: 'This issue has been open for 1 month with no activity. It will be closed in 7 days to keep our backlog manageable.'
+          close-issue-message: 'This issue was closed because it has been inactive for 1 month.'
+          days-before-issue-stale: 30
+
+          days-before-close: 7
+          stale-issue-label: 'status: stale'
+          stale-pr-label: 'status: stale'
+
+          exempt-pr-labels: 'pinned,security,in-progress,no-stale'
+          exempt-issue-labels: 'pinned,help-wanted,no-stale'
```
ankurankan: I don't think we need these two.

.github/workflows/stale.yml
```diff
@@ -0,0 +1,42 @@
...
+
+          ascending: true
+
+          operations-per-run: 50
```
ankurankan: Why do we want to limit this?

.github/workflows/stale.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Handle Stale Content
+        uses: actions/stale@v10.2.0
+        with:
+          stale-pr-message: 'This PR has had no activity for 14 days and is now marked as stale. It will be closed in 7 days if no further activity occurs.'
+          close-pr-message: 'This PR was closed due to 3 weeks of inactivity.'
+          days-before-pr-stale: 14
+
+          stale-issue-message: 'This issue has been open for 1 month with no activity. It will be closed in 7 days to keep our backlog manageable.'
+          close-issue-message: 'This issue was closed because it has been inactive for 1 month.'
+          days-before-issue-stale: 30
```
ankurankan: Don't think we should be closing these. Should just add a stale label.

ankurankan (changes requested):

.github/workflows/stale.yml
Spinachboul: Got it! Will add the labels instead.

.github/workflows/stale.yml
Spinachboul: this was set to prevent hitting from github rate limits.
Will remove it.

.github/workflows/stale.yml
Spinachboul: Well the motivation here was that some issues are meant to stay open forever (e.g., a "Welcome" issue or a "Roadmap"). Without these exemptions, the bot would mark them as stale simply because no one has commented on them.

Spinachboul: Removed it in the latest commit

ankurankan (dismissed):

ankurankan (approved):
Thanks @Spinachboul for th PR. I have changed number of days from 14 to 5 for stale PRs.

Issue #2827, opened 2026-03-06
[BUG] The behavior of `DiscreteBayesianNetwork.do` is incorrect


ankurankan:
Currently, the `do` method in `DiscreteBayesianNetwork` marginalizes the specified do state from all the child states. This is incorrect behavior, and instead, the child node CPDs should be reduced to the do state of the parent node.

theadarshcoder:
Hi @ankurankan, I’d be happy to take this up. Could you please assign the issue to me?

ankurankan:
@s2kumar2007 Would you be interested in taking this up? @vbhavh @theadarshcoder Since both of you have 1 issue assigned each, I will try to assign this one to someone else.

s2kumar2007:
Yes , I am interested to take this issue

imshubham22apr-gif:
Hi @ankurankan, I would love to take this up! I was actually just looking into this and have been working on a fix locally.

My Approach:
Currently, the do method applies marginalization to the child CPDs. I will update the logic in DiscreteBayesianNetwork.do so that instead of marginalizing, we apply state reduction to the child node CPDs. By using reduce() (or appropriately slicing the TabularCPD), we can correctly fix the parent node to the specified do state. I'm also adding a focused unit test to ensure the child CPDs reflect this reduction and no longer marginalize the variable.

I can open a PR shortly for you to review if this sounds good!

ankurankan:
@imshubham22apr-gif Please get an issue assigned to you before working on it.

s2kumar2007:
I have submitted a PR for this issue. Please review.@ankurankan

PR #2856, opened 2026-03-07
Fix virtual intervention to use uniform CPD after do operation

### LINKED ISSUE #2864: [BUG]: `simulate()` returns wrong distribution for soft virtual interventions
**Describe the bug**
I was running soft causal interventions using simulate(virtual_intervention=...) and noticed the sampled distribution was way off from what I specified. I passed P(X=0)=0.40 but the samples came back around P(X=0)≈0.79. No error was raised so I almost missed it.
After digging in, I traced it to do() in DiscreteBayesianNetwork. When it severs incoming edges, it marginalizes the CPD over former parents instead of replacing it with a uniform. That marginalized prior then gets multiplied with the virtual evidence likelihood, contaminating the result. Hard interventions don't show this because the delta likelihood zeroes everything out anyway.

**To Reproduce**
I noticed it while working with a simple 3-node network where `X`'s CPD given its parent `D` is asymmetric. Here's the minimal case I narrowed it down to:

```python

# asymmetric P(X|D) is what triggers it — if P(X|D=0) ≈ P(X|D=1),
# the marginal happens to be close to uniform and the bug hides
model = DiscreteBayesianNetwork([('D', 'X'), ('X', 'Y')])
model.add_cpds(
    TabularCPD('D', 2, [[0.5], [0.5]]),
    TabularCPD('X', 2, [[0.9, 0.8], [0.1, 0.2]],
               evidence=['D'], evidence_card=[2]),
    TabularCPD('Y', 2, [[0.3, 0.8], [0.7, 0.2]],
               evidence=['X'], evidence_card=[2]),
)

# requesting a soft intervention — not a hard do(X=0)
samples = model.simulate(
    n_samples=200000,
    virtual_intervention=[TabularCPD('X', 2, [[0.4], [0.6]])],
    show_progress=False, seed=42
)
print(samples['X'].value_counts(normalize=True).sort_index())
# P(X=0) comes out ~0.79 instead of the 0.40 I asked for
```

**Expected behavior**
I expected `P(X=0) ≈ 0.40` and `P(X=1) ≈ 0.60`, matching the distribution I passed into `virtual_intervention.`

**Additional context**
I looked at the existing tests for `virtual_intervention` and they all use hard interventions like `[[1.0], [0.0]`]. That's why this hasn't been caught, with hard interventions the prior doesn't matter since only one state survives. The bug only shows up with soft distributions where multiple states have non-zero probability, which is the whole point of `virtual_intervention.`

**Versions**
<details>
pgmpy 1.0.0
Python 3.11
</details>

Ady0333:

- [X] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side) against our **dev branch** (left side). Please do not request your **dev branch**.
- [X] Have you followed all the steps from our [Contributing Guide](https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md)?
- [X] Make sure that the pull request fully addresses the linked issue and is within its defined scope.

- [X] Please include a short paragraph (**not bullet points**) describing the changes you have made. This should at least include the problem your PR solves and the implemented solution. This doesn't need to be a polished description, but it **must** be written **manually** by you without any assistance from any AI tools.
- [X] Have you **manually** verified that the changes are doing exactly what is expected? Have you considered edge cases?
- [X] Has the LLM added try-except blocks? They will need to be removed; any error handling must be explicit.
- [X] If you used LLM for generating tests, they usually need to be compressed into a smaller set of tests. Please **manually** describe what scenarios the tests are testing for.

---

- Fixes #2864

---

- Replaced `cpd.marginalize(cpd.variables[1:], inplace=True)` in do() with a uniform CPD construction.
- Added test for soft virtual interventions with a non-trivial distribution.

### DIFF
--- a/pgmpy/models/DiscreteBayesianNetwork.py
+++ b/pgmpy/models/DiscreteBayesianNetwork.py
@@ -1409,7 +1409,15 @@ class DiscreteBayesianNetwork(DAG):
         if adj_model.cpds:
             for node in nodes:
                 cpd = adj_model.get_cpds(node=node)
-                cpd.marginalize(cpd.variables[1:], inplace=True)
+                uniform_vals = np.ones((cpd.variable_card, 1)) / cpd.variable_card
+                new_cpd = TabularCPD(
+                    variable=node,
+                    variable_card=cpd.variable_card,
+                    values=uniform_vals,
+                    state_names={node: list(cpd.state_names[node])},
+                )
+                adj_model.remove_cpds(cpd)
+                adj_model.add_cpds(new_cpd)
         return adj_model
 
     def simulate(


Ady0333:
Hello @ankurankan , 
Please review this pr once you have moment. Happy to make changes if needed!!

DARHWOLF:
May I request you to please include the linked issue to the PR description.
The template is not to be edited in any way. 
Converting this to draft till I see am issue linked to this PR

Ady0333:
i will add raise an issue and then link it.

ankurankan:
This is nowhere near the correct solution for the issue. This is also related to #2827 which already has a PR and should fix this as well. Closing.

PR #2862, opened 2026-03-08
fix: return None for inplace operations in role mixins (#2853)

### LINKED ISSUE #2853: [BUG] with_role and without_role return object instead of None when inplace=True
**Describe the bug**

The with_role and without_role methods in the DAG class return the modified graph object (self) even when the inplace parameter is set to True. Following standard Python conventions for inplace operations, these methods should return None when modifying the object in place. The modified object should only be returned when inplace=False.

This behavior was identified during the investigation of doctest failures in PR #2836.

**To Reproduce**

```python
from pgmpy.base import DAG
# Initialize a DAG and add a node
dag = DAG()
dag.add_node("Z")
# Call with_role with inplace=True
result = dag.with_role(role="latents", variables="Z", inplace=True)
# Check the return value
print(f"Return value: {result}")
# Actual: Return value: <pgmpy.base.DAG.DAG object at 0x...>
# Expected: Return value: None
```

**Expected behavior** :
When inplace=True is passed to with_role or without_role, the methods should return None. They should only return the DAG object when inplace=False.

**Additional context** :
This bug was noted by the maintainer @ankurankan during the review of PR #2840. Correcting this will bring these methods in line with standard API practices within pgmpy and the wider Python ecosystem.

Versions

<Details>
Python: 3.8 NetworkX: 3.0+ pgmpy: dev

codervinitjangir:

- [x] Make sure that the pull request fully addresses the linked issue and is within its defined scope.

- In this PR, I have addressed the issue where with_role and without_role methods were incorrectly returning the graph object during inplace operations. I have updated the logic to return None when inplace=True, which aligns with standard Python conventions. I also updated an existing test that was affected by this change and added new test cases to verify the fix.

- Fixes #2853

-
- Modified pgmpy/base/_mixin_roles.py to return None if inplace=True in with_role and without_role.

- Fixed test_with_role_method in pgmpy/tests/test_base/test_AncestralBase.py to stop assigning the return value of an inplace operation.

- Added new tests in pgmpy/tests/test_base/test_mixin_roles.py to verify None is returned for inplace calls.


DARHWOLF (commented):
Seems like a reasonable fix. Thanks for the PR

A few things I want to confirm:
- I see the core graph class' copy method is supposed to return a copy of the dag object. With this change, it will return `None`
- Same for all the other graph classes 
- would this require a fix? 
cc: @ankurankan 

LG when AA approves.

codervinitjangir:
Thanks for the great catch, @DARHWOLF! 🙏

I looked deep into this, and you are absolutely right. While the current copy() implementations in both DAG and _CoreGraph actually rebuild roles from scratch (meaning the sets aren't shared in practice), relying on that makes the with_role/without_role methods fragile to shallow copies.

To make this completely robust, I've pushed a fix with the following updates:

- Defensive Copying: Wrapped the .get("roles", set()) calls with set(...) in both methods. This ensures we always work on a fresh copy of the set, preventing any accidental shared reference mutations.

- New Test: Added test_with_role_inplace_false_does_not_mutate_original to explicitly verify that the original graph's roles remain completely untouched after an inplace=False operation.

- Global Coverage: Regarding the impact on other graph classes — since with_role and without_role live in the _GraphRolesMixin, this single defensive fix automatically covers DAG, BayesianNetwork, and all other core graph classes that inherit from it.

The latest commit is up. Let me know if it looks good to go!

pgmpy/base/_mixin_roles.py
```diff
@@ -105,10 +105,12 @@ def with_role(self, role: str, variables, inplace=False):
...
                     raise ValueError(f"Variable '{var}' not found in the graph.")
                 else:
-                    existing_role = new_graph.nodes[var].get("roles", set())
+                    existing_role = set(new_graph.nodes[var].get("roles", set()))
 
                     existing_role.add(role)
                     new_graph.add_node(var, roles=existing_role)
+        if inplace:
+            return None
         return new_graph
```
ankurankan: Can be simplified as replacing the codeblock with:

```
if not inplace:
     return new_graph
```

pgmpy/base/_mixin_roles.py
```diff
@@ -139,13 +141,16 @@ def without_role(self, role: str, variables=None, inplace=False):
...
                 if isinstance(roles, set) and role in roles:
                     roles.discard(role)
                     if len(roles) == 0:
                         attr.pop("roles")
                     else:
                         attr["roles"] = roles
+
+        if inplace:
+            return None
         return new_graph
```
ankurankan: Same.

pgmpy/base/DAG.py
```diff
@@ -1449,9 +1449,10 @@ def to_graphviz(self, plot_edge_strength=False):
         --------
         >>> from pgmpy.utils import get_example_model
         >>> model = get_example_model("alarm")
-        >>> model.to_graphviz()
-        <AGraph <Swig Object of type 'Agraph_t *' at 0x7fdea4cde040>>
-        >>> model.draw("model.png", prog="neato")
+        >>> agraph = model.to_graphviz()  # doctest: +SKIP
+        >>> agraph  # doctest: +SKIP
+        <AGraph <Swig Object of type 'Agraph_t *' at ...>>
+        >>> agraph.draw("model.png", prog="neato")  # doctest: +SKIP
```
ankurankan: Why is this modification in this PR?

pgmpy/base/_mixin_roles.py
```diff
@@ -96,7 +96,7 @@ def with_role(self, role: str, variables, inplace=False):
                 if var not in new_graph:
                     raise ValueError(f"Variable '{var}' not found in the graph.")
                 else:
-                    existing_role = new_graph.nodes[var].get("roles", set())
+                    existing_role = set(new_graph.nodes[var].get("roles", set()))
```
ankurankan: Why is this required? Shouldn't copying the roles be automatically handled in when `copy` method is called in line 89? This will end up creating a copy even when inplace=False.

pgmpy/base/_mixin_roles.py
```diff
@@ -105,10 +105,12 @@ def with_role(self, role: str, variables, inplace=False):
                 if var not in new_graph.graph:
                     raise ValueError(f"Variable '{var}' not found in the graph.")
                 else:
-                    existing_role = new_graph.nodes[var].get("roles", set())
+                    existing_role = set(new_graph.nodes[var].get("roles", set()))
```
ankurankan: Same

pgmpy/base/_mixin_roles.py
```diff
@@ -139,13 +141,16 @@ def without_role(self, role: str, variables=None, inplace=False):
 
         for v, attr in new_graph.nodes(data=True):
             if variables is None or v in variables:
-                roles = attr.get("roles", set())
+                roles = set(attr.get("roles", set()))
```
ankurankan: Same

pgmpy/tests/test_base/test_mixin_roles.py
```diff
@@ -56,6 +56,26 @@ def test_with_role_inplace_false_returns_new_graph(basic_dag):
...
+    assert result is None
+    assert "roles" not in basic_dag.nodes["X"]
+
+
+def test_with_role_inplace_false_does_not_mutate_original(basic_dag):
+    basic_dag.with_role(role="exposures", variables="X", inplace=True)
+    original_roles = basic_dag.nodes["X"]["roles"].copy()
+    basic_dag.with_role(role="outcomes", variables="X", inplace=False)
+    assert basic_dag.nodes["X"]["roles"] == original_roles
+
```
ankurankan: Can you compress these in a single test?

ankurankan (changes requested):

codervinitjangir:
Hi @ankurankan, thanks for the detailed review! I've updated the PR based on your feedback. Here is the summary of the changes:

1. **Simplified Return Logic**: Updated both with_role and without_role to use if not inplace: return new_graph. This implicitly returns None when inplace=True, which is much cleaner.

2. **Reverted Defensive Copies**: Removed the explicit set() wrappers around role assignments. You are right; since the copy() method already handles fresh role set creation, the manual defensive copies were redundant.

3. **Cleaned Up DAG.py**: Fully reverted pgmpy/base/DAG.py to its state in the dev branch to remove the unrelated doctest modifications that accidentally leaked into this PR.

4. **Compressed Tests**: Merged the three separate tests in test_mixin_roles.py into a single, comprehensive test.

All tests (310 passed) and pre-commit hooks are passing locally. Let me know if this looks good to go!

pgmpy/tests/test_base/test_mixin_roles.py
```diff
@@ -56,6 +56,16 @@ def test_with_role_inplace_false_returns_new_graph(basic_dag):
...
 
 
+def test_inplace_true_returns_none(basic_dag):
+    result = basic_dag.with_role(role="exposures", variables="X", inplace=True)
+    assert result is None
+    assert "exposures" in basic_dag.nodes["X"]["roles"]
+
+    result = basic_dag.without_role(role="exposures", variables="X", inplace=True)
+    assert result is None
+    assert "roles" not in basic_dag.nodes["X"]
```
ankurankan: Can you also add the case when `inplace=False`?

ankurankan (changes requested):

codervinitjangir:
Hi @ankurankan ,

I have added the test case for inplace=False as you requested.

I combined them into a single test test_inplace_argument that checks both True and False cases for with_role and without_role. I added asserts to make sure that inplace=False returns a completely new graph and does not modify the original graph's roles.

I have pushed the commit. Please check and let me know!

ankurankan (approved):
Looks good. Thanks for the PR @codervinitjangir

PR #2866, opened 2026-03-08
fix: handle zero degrees-of-freedom edge case in power_divergence

### LINKED ISSUE #2860: [BUG] power_divergence may produce invalid results when degrees of freedom = 0
## Describe the bug

The power_divergence test in `pgmpy/estimators/CITests.py` can produce invalid results when the `degrees of freedom (dof)` becomes zero.

This occurs when all conditional states are skipped due to insufficient samples.

## To Reproduce

Relevant code:

```
p_value = 1 - stats.chi2.cdf(chi, df=dof)

If dof == 0, the chi-square distribution is not defined, leading to unreliable statistical results.
```

## Expected behavior

The function should detect this edge case and either:

return a well-defined fallback value, or raise a clear exception indicating insufficient data.

## Additional context

This can occur when datasets are very small or when conditioning variables create many sparse state combinations. Handling this case explicitly would make the statistical test more robust.

beingamanforever:
Ensure the statistical test behaves correctly when all conditional strata are skipped and the accumulated degrees of freedom remain zero.

Return a neutral statistic and p-value instead of evaluating the chi-square distribution with `df=0`, which is undefined.

Add a regression test covering sparse conditional strata.

Fixes #2860

* [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side) against our **dev branch** (left side). Please do not request your **dev branch**.
* [x] Have you followed all the steps from our [[Contributing Guide](https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md)](https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md)?
* [x] Make sure that the pull request fully addresses the linked issue and is within its defined scope.
* [x] Are all the GitHub Actions checks passing? If not, they will need to be fixed before review. You can reference logs for the failing check to identify the issue.

* [x] Please include a short paragraph (**not bullet points**) describing the changes you have made. This should at least include the problem your PR solves and the implemented solution. This doesn't need to be a polished description, but it **must** be written **manually** by you without any assistance from any AI tools.
* [x] Have you **manually** verified that the changes are doing exactly what is expected? Have you considered edge cases?
* [x] Has the LLM added try-except blocks? They will need to be removed; any error handling must be explicit.
* [x] If you used LLM for generating tests, they usually need to be compressed into a smaller set of tests. Please **manually** describe what scenarios the tests are testing for.

* Fixes #2860

* Added guard in `power_divergence` to handle the edge case where `dof == 0`.
* Return `(0, 1)` instead of evaluating the chi-square distribution with zero degrees of freedom.
* Added a regression test ensuring correct behavior when all conditional strata are skipped.

### Regression test

```python
def test_power_divergence_zero_dof(self):
    data = pd.DataFrame(
        {
            "X": [0, 0, 1, 1],
            "Y": [0, 1, 0, 1],
            "Z": [0, 1, 0, 1],
        }
    )

    chi, p_value = power_divergence(
        X="X", Y="Y", Z=["Z"], data=data, boolean=False
    )

    self.assertEqual(chi, 0)
    self.assertEqual(p_value, 1)
```

### DIFF
--- a/pgmpy/estimators/CITests.py
+++ b/pgmpy/estimators/CITests.py
@@ -347,7 +347,11 @@ def power_divergence(X, Y, Z, data, boolean=True, lambda_="cressie-read", **kwar
                 c, _, d, _ = stats.chi2_contingency(contingency, lambda_=lambda_)
                 chi += c
                 dof += d
-        p_value = 1 - stats.chi2.cdf(chi, df=dof)
+        if dof == 0:
+            chi = 0
+            p_value = 1
+        else:
+            p_value = 1 - stats.chi2.cdf(chi, df=dof)
 
     # Step 4: Return the values
     if boolean:
--- a/pgmpy/tests/test_estimators/test_CITests.py
+++ b/pgmpy/tests/test_estimators/test_CITests.py
@@ -16,6 +16,7 @@ from pgmpy.estimators.CITests import (
     pearsonr,
     pearsonr_equivalence,
     pillai_trace,
+    power_divergence,
 )
 from pgmpy.factors.continuous import LinearGaussianCPD
 from pgmpy.models import LinearGaussianBayesianNetwork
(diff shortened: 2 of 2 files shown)


ankurankan (changes requested):
I don't think your test case is correct, because we only skip the test if all counts in a column or row are 0, i.e., one of the variables has no data for one of its states. This skipping behavior is not triggered for the test case that you have written in the PR. In the test case, the `dof=0` is due to a different bug (#2886). Can you change the test case to actually trigger the skipping behavior?

ankurankan:
Closing this due to inactivity.

PR #2887, opened 2026-03-10
fix: propagate roles correctly in copy() methods

### LINKED ISSUE #2875: [BUG] Fix copy method in base graph classes
**Describe the bug**
We recently fixed a bug in the `with_role` and `without_role` method wherein the `inplace = True` argument was returning a copy of the graph rather than returning `None`. 

Now that `inplace = True` is being fixed to behave correctly, the `copy()` methods of the graph classes require fixing the line where the with_role() method is being used. 

This needs to be changed to `inplace = False`

For reference: 
https://github.com/pgmpy/pgmpy/pull/2862#pullrequestreview-3911658254

**To Reproduce**

```python
<Paste your code here>
<please include all imports>
<if data is required, try to find a small dummy data set that triggers the problem>
```

**Expected behavior**

**Additional context**

**Versions**
<details>

</details>

P-r-e-m-i-u-m:
Fixes #2875

## Problem
After PR #2862, the `with_role()` method with `inplace=True` no longer modifies the graph in place — it returns a new object. This caused roles to silently not propagate in all `copy()` methods.

## Fix
Changed all 4 `copy()` methods to use `inplace=False` and assign the result back:

- `pgmpy/base/_base.py` — `_CoreGraph.copy()`
- `pgmpy/base/DAG.py` — `DAG.copy()`
- `pgmpy/base/PDAG.py` — `PDAG.copy()`
- `pgmpy/base/AncestralBase.py` — `AncestralBase.copy()`

### DIFF
--- a/pgmpy/base/AncestralBase.py
+++ b/pgmpy/base/AncestralBase.py
@@ -742,6 +742,6 @@ class AncestralBase(nx.Graph, _GraphRolesMixin):
         )
 
         for role, vars in self.get_role_dict().items():
-            ancestral_base.with_role(role=role, variables=vars, inplace=True)
+            ancestral_base = ancestral_base.with_role(role=role, variables=vars, inplace=False)
 
         return ancestral_base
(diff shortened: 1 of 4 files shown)


ankurankan:
@P-r-e-m-i-u-m No checklist. Closing

PR #2889, opened 2026-03-10
BUG: preserve contingency state space in power_divergence

### LINKED ISSUE #2886: [BUG] Incorrect contingency table computation in `power_divergence` test

```python
        data = pd.DataFrame(
            {
                "X": [0, 0, 1, 1],
                "Y": [0, 1, 0, 1],
                "Z": [0, 1, 0, 1],
            }
        )
```

NandiniDhanrale:
Fixes #2886

## Summary

This fixes incorrect contingency table construction in power_divergence when conditioning on Z.

Previously, each Z stratum built the X/Y contingency table using only the states observed in that stratum. That could shrink a table such as 2x2 into 2x1 or 1x2, which changes the effective test and can produce incorrect statistics / degrees of freedom.

This change preserves the full X/Y state space across strata and fills missing combinations with

### DIFF
--- a/doctest_modules.txt
+++ b/doctest_modules.txt
@@ -2,3 +2,4 @@
 # Add a module path here after fixing its doctests to have them checked in CI.
 # Example:
 # pgmpy/base/DAG.py
+pgmpy/base/DAG.py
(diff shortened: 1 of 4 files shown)


ankurankan:
Closing as checklist was removed.

PR #2918, opened 2026-03-12
fix: Improve error message for load_model when model not found

### LINKED ISSUE #2917: [ENH] Improve error message for `load_model` method
Providing an incorrect model name to the `load_model` method currently throws an `IndexError`.

```python
In [9] : from pgmpy.example_models import load_model
In [10]: load_model("bnrep/soilead")
Cell In[10], line 1
----> 1 load_model("bnrep/soilead")

File ~/work/pgmpy/pgmpy/pgmpy/example_models/_base.py:179, in load_model(name)
    174 if target_model is None:
    175     raise ValueError(
    176         f"Model with name '{name}' not found. Please use list_models() to see available datasets."
    177     )
--> 179 return target_model[0].load_model_object()

IndexError: list index out of range
```
In the example above, the `bnrep/soilead` model doesn't exist. It would be better to throw a more informative error saying that the example model doesn't exist and that users can use the `list_models` method to list all the available models.

Alm0stSurely:
## Problem
When calling `load_model()` with a non-existent model name, the function raises an `IndexError` instead of a clear `ValueError`.

Example:
```python
load_model("bnrep/soilead")  # Model doesn't exist
# Before: IndexError: list index out of range
```

## Analysis
The `all_objects()` function returns an empty list `[]` when no models match the filter, not `None`. The existing check `if target_model is None` doesn't catch this case, causing the subsequent `target_model[0]` access to raise an IndexError.

## Solution
Changed the truthiness check from `if target_model is None` to `if not target_model`, which properly handles both `None` and empty list cases.

## Testing
- Added `test_load_model_not_found()` to verify the proper ValueError is raised
- All existing tests pass (4/4 in test_example_models)

## Notes
Fixes #2917

### DIFF
--- a/pgmpy/example_models/_base.py
+++ b/pgmpy/example_models/_base.py
@@ -171,7 +171,7 @@ def load_model(name: str):
         return_names=False,
     )
 
-    if target_model is None:
+    if not target_model:
         raise ValueError(
             f"Model with name '{name}' not found. Please use list_models() to see available datasets."
         )
--- a/pgmpy/tests/test_example_models/test_example_models.py
+++ b/pgmpy/tests/test_example_models/test_example_models.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pytest
 from skbase.lookup import all_objects
 
 from pgmpy.base import DAG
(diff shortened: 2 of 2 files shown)


ankurankan:
Closing as the checklist was removed.

PR #2928, opened 2026-03-12
fix(CITests): use global states for contingency tables in power_divergence

### LINKED ISSUE #2886: [BUG] Incorrect contingency table computation in `power_divergence` test

Dhi13man:
## Summary

Fixes #2886 — the conditional branch of `power_divergence` computed contingency tables using only the unique X/Y values present in each Z-stratum. When a state of X or Y was absent from a stratum, it was silently dropped, producing undersized tables (e.g., `[[1],[1]]` instead of a proper 2×2 table) and incorrect chi-squared statistics.

### Changes

- **`pgmpy/estimators/CITests.py`**: Compute `all_x`/`all_y` from the full dataset before the loop and use `np.searchsorted` to map each stratum's values to global indices, ensuring every contingency table has consistent dimensions matching the full domain of X and Y.

- **`pgmpy/tests/test_estimators/test_CITests.py`**: 
  - Updated expected values for conditional tests that were previously computed with the buggy undersized tables
  - Added regression test `test_contingency_table_includes_all_states` using the exact reproduction case from #2886

### Why the expected test values changed

The fix correctly exposes zero columns/rows that were previously hidden by undersized tables. Strata where a state of X or Y is absent now have the correct table dimensions, causing the existing zero-column skip logic (line 338) to properly identify them as having insufficient samples. This changes which strata contribute to the chi-squared sum, producing different (correct) aggregate statistics.

## Test plan

- [x] All `TestDiscreteTests` tests pass (4/4)
- [x] Regression test for issue #2886 added and passing
- [x] All other CITests pass (pre-existing `test_gcm` failure unrelated to this change)
- [x] Code formatted with black

### DIFF
--- a/pgmpy/estimators/CITests.py
+++ b/pgmpy/estimators/CITests.py
@@ -322,13 +322,18 @@ def power_divergence(X, Y, Z, data, boolean=True, lambda_="cressie-read", **kwar
     else:
         chi = 0
         dof = 0
+        # Use global unique values so every stratum gets a consistently sized table
+        all_x = np.unique(data[X])
+        all_y = np.unique(data[Y])
+        n_x = len(all_x)
+        n_y = len(all_y)
         for z_state, df in data.groupby(Z, observed=True):
-            # Compute the contingency table
-            unique_x, x_inv = np.unique(df[X], return_inverse=True)
-            unique_y, y_inv = np.unique(df[Y], return_inverse=True)
-            contingency = np.bincount(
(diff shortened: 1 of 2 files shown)


ThiefMaster:
AI slop.

Dhi13man:
I had leftover Claude Max and wanted to contribute to backlog clearing in open source projects before it reset, but I see that this could be a nuisance. My apologies.

PR #2950, opened 2026-03-13
Add predict_probability method to LinearGaussianBayesianNetwork

### LINKED ISSUE #2828: [ENH] Add a `predict_probability` method to `LinearGaussianBayesianNetwork`
The `DiscreteBayesianNetwork` class provides two methods for predictions:
1. `predict`: Returns the MAP estimate from the posterior distribution.
2. `predict_probability`: Returns the joint posterior distribution.

However, the `LinearGaussainBayesianNetwork` class only has a `predict` method that returns the posterior distribution. We should add a `predict_probability` method and make the behavior of both `predict` and `predict_probability` similar to `DiscreteBayesianNetwork`.

vbhavh:
feat: add predict_probability method to LinearGaussianBayesianNetwork
- Add predict_probability method that returns conditional distribution (mean, covariance)
- Modify predict method to return DataFrame with MAP estimates (conditional means)
- Update simulate method to use predict_probability internally
- Add comprehensive tests for both methods including error handling
- Update existing tests to use predict_probability where appropriate
- Maintain API consistency with DiscreteBayesianNetwork pattern

Fixes https://github.com/pgmpy/pgmpy/issues/2828

### DIFF
--- a/pgmpy/models/LinearGaussianBayesianNetwork.py
+++ b/pgmpy/models/LinearGaussianBayesianNetwork.py
@@ -694,7 +694,7 @@ class LinearGaussianBayesianNetwork(DAG):
 
         else:
             df_evidence = pd.DataFrame([evidence])
-            missing_vars, mean_cond, cov_cond = model.predict(data=df_evidence)
+            missing_vars, mean_cond, cov_cond = model.predict_probability(data=df_evidence)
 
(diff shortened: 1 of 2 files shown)


vbhavh:
Some tests were not passing, so i closed this and am fixing them.

PR #2957, opened 2026-03-14
[ENH] DAG falsification using permutation-based test

### LINKED ISSUE #2115: [ENH] Method for validating output of causal discovery methods

JATAYU000:

To understand certain code blocks. auto completion
There are no generated code , except for few line suggestions which I have reviwed.
No it has not
No I have not

- Fixes #2115

Note: This pr stacks on abandoned pr #2119 and updated the functions and fixes as mentioned below

- Fix  `DAG._get_non_descendants` to exclude_parents 
- Updated step 2 as per the paper by permuting nodes as in the paper
- As well as implement validate `lmc` and validate `tpa` for Local Markov Condition and Markov Equivalence Class checks
- Not from the paper but made `n_permutation` param similar to how it is handled in the package


JATAYU000:
I have read through the paper and found an package which implemented this [reference py-why/dowhy](https://github.com/py-why/dowhy/blob/5341e8a95fba013b88fde393634a6dabdc0c2a6a/docs/source/example_notebooks/gcm_falsify_dag.ipynb) alse [doc](https://github.com/pgmpy/pgmpy/pull/2957#issuecomment-4060567601) I have taken inspiration from this.
 
There are a few doubts I had for you @ankurankan 
Right now I am wondering what the default `ci_test` should be since I have tried `gcm` , `pearsonr` and there is difference in violations caught etc, I am not aware of how these particular tests work and would take me some time to go through them all, What are your suggestions on it?

Also due to networkx package changes on `d_seperation` the module `py-why/dowhy` handles it with except block is that ok here?

If the updated code is approved will be removed the commented lines as well as update the tests written by the previous pr

ankurankan:
@JATAYU000 

> Right now I am wondering what the default ci_test should be since I have tried gcm , pearsonr and there is difference in violations caught etc, I am not aware of how these particular tests work and would take me some time to go through them all, What are your suggestions on it?

The default can be None. If the user doesn't specify it, our default CI test depends on the type of data. The default values are automatically handled by our `CITestRegistry.get_test` method. You can have a look at the `CorrelationScore` class for examples on how to use it: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/metrics/correlation_score.py

> Also due to networkx package changes on d_seperation the module py-why/dowhy handles it with except block is that ok here?

We have our own implementation of d-separation: `DAG.is_dconnected` and `DAG.active_trail_nodes`

JATAYU000:
> The default values are automatically handled by our `CITestRegistry.get_test`

Makes sense, will go through the examples

> We have our own implementation of d-separation

That is convinient, Will update the logic

The current node permutaion design is good to go?

JATAYU000:
@ankurankan update on this pr, would like to have a review
- have used `DAG.is_dconnected` instead of networkx functions
- updated the tests from previous pr
- Remove old logic parts

ankurankan:
@JATAYU000 Can we compare our implementation with the implementation in dowhy to make sure our implementation is correct: https://www.pywhy.org/dowhy/v0.10/example_notebooks/gcm_falsify_dag.html

JATAYU000:
I had compared the results on the first example, will try the same for others and let you know

ankurankan:
@JATAYU000 The linting checks are failing. If possible, please add those test cases (comparing with dowhy) to our tests file, mentioning that the true values are computed using dowhy.

JATAYU000:
> please add those test cases (comparing with dowhy) to our tests file
Does this mean to have `dowhy` to be added to the tests dependencies and used to compare the results?

@ankurankan  There are a few issues, the Conditional independence test used by `dowhy` is called `kernel_based` which i did not find similar in pgmpy's `ci_registry` also passing on of our `ci_test` to `dowhy` function is also not posssible since it expects
```python
Callable[[np.ndarray, np.ndarray, np.ndarray], float]
```
and our tests follow this pattern
```
ci_test(X, Y, Z, data)
```

The closest in results i Have observed is via `pillai` test, and I am not sure how reliable it is to right tests assuming this.

ankurankan:
@JATAYU000 I think for their last example here: https://www.pywhy.org/dowhy/v0.10/example_notebooks/gcm_falsify_dag.html#Real-Data-(Protein-Network-dataset-by-Sachs-et-al.,-2005), if you use `pillai_trace` the results should be very close. Or if you could share some comparison code I can also try to verify it.

Also, I somehow missed it, but this implementation will need to be compliant with our `UnsupervisedMetric` interface. You can look at examples in the `metrics` directory.

JATAYU000:
> Also, I somehow missed it, but this implementation will need to be compliant with our UnsupervisedMetric interface. You can look at examples in the metrics directory.

Sure, will have a look

**Example 1**
```python
g_true = nx.read_gml("falsify_g_true.gml")
data = pd.read_csv("falsify_data_nonlinear.csv")
result_dowhy = falsify_graph(g_true, data,n_permutations=-1,n_jobs=4)

dag = DAG()
dag.add_nodes_from(g_true.nodes())
dag.add_edges_from(g_true.edges())

result = permutation_test(dag, data, ci_test='pillai', show_progress=True, n_permutations=-1)
```
- **DOWHY results**
```text
The given DAG is informative because 2 [/] 120 of the permutations lie in the Markov                    
equivalence class of the given DAG (p-value: 0.02).                                                   
The given DAG violates 0/11 LMCs and is better than 96.7% of the permuted DAGs (p-value: 0.03).       
Based on the provided significance level (0.05) and because the DAG is informative,                   
we do not reject the DAG.

Graph is falsifiable: True, Graph is falsified: False
```
- **Current implementation**
```text
{'p_value_falsifiable': 0.016666666666666666, 'p_value_falsified': 0.03333333333333333, 'n_lmc_violations': 0, 'n_within_mec': 2, 'ci_lower_falsified': 0.013037551098350586, 'ci_upper_falsified': 0.08258033755910602, 'falsifiable': True, 'falsified': False}

Graph is falsifiable: True, Graph is falsified: False
```
**Example 2**
```python
g_true = nx.read_gml("falsify_g_true.gml")
data = pd.read_csv("falsify_data_nonlinear.csv")
g_given = g_true.copy()
g_given.add_edges_from(([('X4', 'X1')]))  # Add wrong edge from X4 -> X1
g_given.remove_edge('X2', 'X0')  # Remove true edge from X2 -> X0
result_dowhy = falsify_graph(g_given, data,n_permutations=-1,n_jobs=4)

dag = DAG()
dag.add_nodes_from(g_given.nodes())
dag.add_edges_from(g_given.edges())

result = permutation_test(dag, data, ci_test='pillai', show_progress=True, n_permutations=-1)
```
- **DOWHY results**
```text
The given DAG is informative because 1 [/] 120 of the permutations lie in the Markov                    
equivalence class of the given DAG (p-value: 0.01).                                                   
The given DAG violates 2/11 LMCs and is better than 91.7% of the permuted DAGs (p-value: 0.08).       
Based on the provided significance level (0.05) and because the DAG is informative,                   
we reject the DAG.     
                                          
Graph is falsifiable: True, Graph is falsified: True
```
- **Current implementation**
```text
{'p_value_falsifiable': 0.008333333333333333, 'p_value_falsified': 0.1, 'n_lmc_violations': 2, 'n_within_mec': 1, 'ci_lower_falsified': 0.058133592026010775, 'ci_upper_falsified': 0.16668174110895206, 'falsifiable': True, 'falsified': True}

Graph is falsifiable: True, Graph is falsified: True
```

**Example 3 sachs**
```python
data_url = "https://raw.githubusercontent.com/FenTechSolutions/CausalDiscoveryToolbox/master/cdt/data/resources/cyto_full_data.csv"
data_sachs = pd.read_csv(data_url)
g_sachs = nx.read_gml('falsify_sachs.gml')
def create_gradient_boost_regressor(**kwargs) -> SklearnRegressionModel:
    return SklearnRegressionModel(GradientBoostingRegressor(**kwargs))
def gcm(X, Y, Z=None):
    return generalised_cov_based(X, Y, Z=Z, prediction_model_X=create_gradient_boost_regressor,
          prediction_model_Y=create_gradient_boost_regressor)
result_sachs = falsify_graph(g_sachs, data_sachs, n_permutations=100,
    independence_test=gcm,
    conditional_independence_test=gcm,
    plot_histogram=True)

dag = DAG()
dag.add_nodes_from(g_sachs.nodes())
dag.add_edges_from(g_sachs.edges())

result = permutation_test(dag, data_sachs, ci_test='pillai', show_progress=True, n_permutations=100)

```

- **DOWHY results**
```text
The given DAG is informative because 0 [/] 100 of the permutations lie in the Markov                    
equivalence class of the given DAG (p-value: 0.00).                                                   
The given DAG violates 21/49 LMCs and is better than 99.0% of the permuted DAGs (p-value: 0.01).      
Based on the provided significance level (0.05) and because the DAG is informative,                  
we do not reject the DAG.                                                                         

Graph is falsifiable: True, Graph is falsified: False
```
- **Current implementation**
```text
{'p_value_falsifiable': 0.0, 'p_value_falsified': 0.04, 'n_lmc_violations': 25, 'n_within_mec': 0, 'ci_lower_falsified': 0.015663303991547604, 'ci_upper_falsified': 0.09837071435887923, 'falsifiable': True, 'falsified': False,}

Graph is falsifiable: True, Graph is falsified: False

```

JATAYU000:
Have updated the pr to be compliant with the `UnsupervisedMetric`

JATAYU000:
@ankurankan Please have a look at the changes, also the results

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+        Significance level for conditional independence tests. Lower values make
+        the test more conservative for accepting the null hypothesis.
+
+    n_permutations : int, optional
+        Number of random node permutations to generate for the baseline.
+        If None, uses max(20, int(1/significance_level)).
+        If -1, uses all possible permutations (factorial of number of nodes)
+
+    ci_test : {"pillai", "chi_square", "pearsonr", "gcm", "g_sq", "log_likelihood", "freeman_tuckey",
+    "modified_log_likelihood", "neyman", "cressie_read"}
```
ankurankan: I don't think we should list it like this, as this will need to be updated every time we add a new CI test, which would be easy to miss. Should say str or instance of BaseCITest and in the detailed description simply link to the ci tests.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def _evaluate(
+        self,
+        causal_graph: DAG,
+        X: pd.DataFrame,
+        n_permutations: int | None = None,
+        significance_level: float = 0.05,
+        ci_test: str | None = None,
+        return_summary: bool = True,
+        show_progress: bool = True,
```
ankurankan: The `_evaluate` method signature for Unsupervised metric is `_evaluate(self, X, causal_graph)`. All other arguments should be specified in the class initialization. Have a look at our extension template: https://github.com/pgmpy/pgmpy/blob/dev/devtools/extension_templates/_metrics.py#L124

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+    _tags = {
+        "name": "permutation_test",
+        "requires_true_graph": False,
+        "requires_data": True,
+        "lower_is_better": True,
+        "is_symmetric": False,
+        "supported_graph_types": (DAG,),
+    }
+
+    def __init__(self):
```
ankurankan: We should accept a seed value to make the results reproducible.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+        ci_test: str | None = None,
+        return_summary: bool = True,
+        show_progress: bool = True,
+    ):
+
+        # Step 0: Initialize variables and validate inputs.
+        if not isinstance(X, pd.DataFrame):
+            raise TypeError(f"Data should be a pandas DataFrame. Got: {type(X)}")
+
+        if not isinstance(causal_graph, DAG) or isinstance(causal_graph, DynamicBayesianNetwork):
```
ankurankan: This wouldn't work for `DynamicBayesianNetwork`. Will only be `DAG`.

ankurankan: The supported causal graph types should also be defined in the tags. Have a look at the extension template: https://github.com/pgmpy/pgmpy/blob/dev/devtools/extension_templates/_metrics.py#L124

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+            raise TypeError(f"DAG must be a `pgmpy.base.DAG` object. Got: {type(causal_graph)}")
+
+        nodes = list(causal_graph.nodes())
+        data_columns = set(X.columns)
+
+        if not set(nodes).issubset(data_columns):
+            missing_vars = set(nodes) - data_columns
+            raise ValueError(f"Data missing variables present in model: {missing_vars}")
```
ankurankan, line 220: The `_BaseUnsupervisedMetric.evaluate` already does all these checks before calling `_evaluate`.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+        Data to test the graph against. The column names of the DataFrame must match the variable names in the `model`.
+
+    significance_level : float, default=0.05
+        Significance level for conditional independence tests. Lower values make
+        the test more conservative for accepting the null hypothesis.
+
+    n_permutations : int, optional
+        Number of random node permutations to generate for the baseline.
+        If None, uses max(20, int(1/significance_level)).
+        If -1, uses all possible permutations (factorial of number of nodes)
```
ankurankan: I don't think this is true, and I don't think there is a closed-form solution to this: https://oeis.org/A003024 . Do you have a reference for this?

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+
+    ci_test : {"pillai", "chi_square", "pearsonr", "gcm", "g_sq", "log_likelihood", "freeman_tuckey",
+    "modified_log_likelihood", "neyman", "cressie_read"}
+
+        The statistical conditional independence test to use for evaluating the Local Markov Conditions in data.
+        See :class:`pgmpy.estimators.CITests` for more details.
+
+    return_summary : bool, default=True
```
ankurankan: default should be False. We want this to work with causal discovery score methods too which would require a single value to be returned.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+        else:
+            pbar = perm_list
+
+        for permuted_nodes in pbar:
+            perm_mapping = dict(zip(nodes, permuted_nodes))
+            # Relabel nodes in the original DAG
+            nx_permuted_dag = nx.relabel_nodes(causal_graph, perm_mapping, copy=True)
+            permuted_dag = DAG()
+            permuted_dag.add_nodes_from(nx_permuted_dag.nodes())
+            permuted_dag.add_edges_from(nx_permuted_dag.edges())
```
ankurankan: We shouldn't be creating permuted DAGs. For any permuted DAG, the actual independence assertion doesn't change. The only thing that changes is the variable in the assertion. For example, if the DAG A -> B -> C has the independence assertion A _|_ C | B. And if we permute the DAG to be B -> A -> C, this assertion becomes B _|_ C | A. So, the change in the DAG is reflected in the assertion; A and B were exchanged in both. So, instead of permuting and creating new DAGs, we can simply permute on the independence assertions. This would save a lot of compute.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+
+        # Step 2: Generate permutations and compute LMC violations for each to construct null distribution.
+
+        perm_list = self._get_permutation_list(nodes, n_permutations)
+        if show_progress and config.SHOW_PROGRESS:
+            pbar = tqdm(perm_list, desc="Constructing Null Distribution")
+        else:
+            pbar = perm_list
+
+        for permuted_nodes in pbar:
```
ankurankan, line 246: This can be parallelized using joblib

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+                data=data,
+                boolean=False,
+                significance_level=significance_level,
+            )
+            pval = res[1]
+            if pval <= significance_level:
+                n_violations += 1
+        return n_violations, len(triples)
+
+    def _tpa_violations(self, permuted_dag, original_dag):
```
ankurankan: Why not combine this method with `_lmc_violations`? There a lot of repeated computation.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+        tpa_violations = []
+        n_within_mec = 0
+
+        # Step 1: Compute LMC violations for the given DAG.
+
+        n_lmc_violations, _ = self._lmc_violations(causal_graph, X, ci_test, significance_level)
+
+        # Step 2: Generate permutations and compute LMC violations for each to construct null distribution.
+
+        perm_list = self._get_permutation_list(nodes, n_permutations)
```
ankurankan, line 240: Why not simply use itertools.permutations and keep a count variable in the loop? Exit when the count is greater than n_permutations.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+        Evaluate which pairwise parental d-separations (parental triples) in `permuted_dag` are
+        violated assuming `original_dag` is the ground truth DAG.
+        """
+        triples = self._get_parental_triples(permuted_dag)
+        n_violations = 0
+        for node, nd, parents in triples:
+            # Check d-separation in original DAG
+            if original_dag.is_dconnected(node, nd, observed=parents):
+                n_violations += 1
+        return n_violations, len(triples)
```
ankurankan: I am assuming this method is for checking whether `permuted_dag` and the `original_dag` are in the same markov blanket? If that is the case, `DAG.is_iequivalent` should give the same result.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,295 @@
...
+
+    For a given DAG, the test checks whether the DAG has fewer Local Markov Condition (LMC) violations than a baseline.
+    The baseline has the same causal structure as the DAG. This baseline is chosen to be the node permutations of the
+    given DAG. Fewer LMC violations mean that the DAG is more consistent/robust than a 'random' guess.
+
+    The test performs two evaluations:
+    1. Falsifiability: Whether the DAG is informative enough to be falsifiable. This is judged using the fraction of
+    DAGs in the node perumations that are Markov equivalent to given DAG
+    2. Falsification: Whether the given DAG performs significantly better than the node perumations, in terms of fewer
+    LMC violations.
```
ankurankan, line 28: I think it might be better to split these into two metrics.

ankurankan (changes requested):
@JATAYU000 Sorry for the delay in the review. I think the comparison results look good. I have added a few comments on the implementation. Could you also add the dowhy comparison tests into our test suite? Basically, have the same code that you have used for the comparisons, and in the comment, you can give the dowhy code to reproduce and the results that you got from it.

pgmpy/metrics/permutation_test.py
JATAYU000: Yes it raises error if it is `DynamicBayesianNetwork`

pgmpy/metrics/permutation_test.py
JATAYU000: If I used `.is_iequivalent` would lose the ability to report the number of TPA violations in summary, would that not be necessary in the summary?

pgmpy/metrics/permutation_test.py
JATAYU000: I wanted the helpers to be easily understood similar to the paper, Yea It would be better to not repeat.

pgmpy/metrics/permutation_test.py
JATAYU000: Updated

pgmpy/metrics/permutation_test.py
JATAYU000: Updated

pgmpy/metrics/permutation_test.py
JATAYU000: Check is removed as mentioned here
https://github.com/pgmpy/pgmpy/pull/2957#discussion_r3002901016

pgmpy/metrics/permutation_test.py
JATAYU000: Updated

pgmpy/metrics/permutation_test.py
JATAYU000: Updated

pgmpy/metrics/permutation_test.py
JATAYU000: Updated, would this suffice as links?

See :class:`pgmpy.estimators.CITests` for available tests.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,298 @@
...
+    >>> print(f"Wrong model falsifiable: {result_wrong['falsifiable']}")
+    """
+
+    _tags = {
+        "name": "permutation_test",
+        "requires_true_graph": False,
+        "requires_data": True,
+        "lower_is_better": True,
+        "is_symmetric": False,
+        "supported_graph_types": (DAG,),
```
JATAYU000, line 128: supported types tag is defined

pgmpy/metrics/permutation_test.py
JATAYU000: Yes I do have it in the tags https://github.com/pgmpy/pgmpy/pull/2957/changes#r3004954127

pgmpy/metrics/permutation_test.py
JATAYU000: The null distribution is constructed by randomly permuting , we would benefit from deterministic only when n_perms = -1

pgmpy/metrics/permutation_test.py
JATAYU000: How do you expect it to be split as two different classes? wouldn't that be again duplicate computations?
@ankurankan  could you please clarify what you meant?

ankurankan: It will duplicate some computation, but these are two different metrics. I think it is a bit confusing to keep them in a single class just because they share some computation. Separating them would also make the return values a bit simpler, given that we are returning a lot of values for this metric.

pgmpy/metrics/permutation_test.py
ankurankan: Then maybe numpy random permutation? I essentially want to get rid of implementing our own `_get_permutation_list` method unless absolutely required.

pgmpy/metrics/permutation_test.py
JATAYU000: Updated, There is a little reduction in time taken per distribution, with higher permutations this saves a descent amount of time

pgmpy/metrics/permutation_test.py
JATAYU000: That is exactly what `_get_permutation_list` does it takes numpy random permutaions as`tuple(np.random.permutation(nodes))`
when the n_permutation is -1 or greater than max permuations it switches to be deterministic and takes all as `itertools.permutations(nodes)`

pgmpy/metrics/permutation_test.py
JATAYU000: But Falsified depends on falsifiable, if we seperate it out `ClassFalsified` would have to call `ClassFalsifiable.evaluate()` , then the next better method i would think of is having `ClassFalsified` to inherit `ClassFalsifiable` and that would reduce a lot of duplicate code and computes ? what do you think @ankurankan ?

pgmpy/metrics/permutation_test.py
JATAYU000: Would make it once the structure changes is completed

pgmpy/metrics/permutation_test.py
JATAYU000: Also Falsiable would not need X in `_evalute`, But X is required in `evaluate`

JATAYU000:
Hi @ankurankan I have left some followup questions on some reviews please do have a look at them, I have updated others

Nimish-4 (dismissed):

pgmpy/tests/test_metrics/test_permutation_test.py
ankurankan, line 1: We should write some tests that recreate the dowhy examples and see if our values match.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,306 @@
...
+        data_columns = set(X.columns)
+
+        if not set(nodes).issubset(data_columns):
+            missing_vars = set(nodes) - data_columns
+            raise ValueError(f"Data missing variables present in model: {missing_vars}")
+
+        if self.n_permutations is None:
+            n_permutations = max(20, int(1 / self.significance_level))
+        elif self.n_permutations == -1:
+            n_permutations = math.factorial(len(nodes))
```
ankurankan: Can remove this.

pgmpy/metrics/permutation_test.py
```diff
@@ -0,0 +1,306 @@
...
+        else:
+            n_permutations = self.n_permutations
+
+        ci_test = get_ci_test(test=self.ci_test, data=X)
+        permutation_violations = []
+        tpa_violations = []
+        n_within_mec = 0
+
+        # Step 1: Compute LMC violations for the given DAG.
+        n_lmc_violations, _, triples = self._get_violations(ci_test, causal_graph)
```
ankurankan: Let's rename `triples` to `ci_statements`

ankurankan (changes requested):

PR #2971, opened 2026-03-15
Chore: add extension template for new methods(unify format)

daehyun99:

- [ ] Does the PR fully address the linked issue and is within its defined scope? If you are still working on the PR, mark it as draft.

- `None`

- Create a template to provide users with a consistent method documentation experience

### DIFF
--- a/devtools/extension_templates/_method.py
+++ b/devtools/extension_templates/_method.py
@@ -13,23 +13,57 @@ class TemplateClass:
         self,
     ):  # TODO: Provide type hints for the input parameters and the return value.
         """
-        # TODO: Ensure the docstring follows the format provided below.
+        [One line description of the method.]
+
+        [Detailed description of the method.]
 
         Parameters
         ----------
+        hyperparam1: type, optional
+            Description of hyperparam1.
+
+        hyperparam2: type, optional
+            Description of hyperparam2.
+
+        Attributes
+        ----------
+        attribute1: []
+            Description of attribute1.
+
+        attribute2: []
+            Description of attribute2.
 
         Returns
         -------
+        return's type: []
+            Description of return value.
 
         See Also
         --------
+        [List of related methods for user reference.]
 
         Notes
         -----
+        User-facing comments (optional)
 
         Examples
         --------
+        >>> from pgmpy.base import DAG
+        >>> ...
 
+        References
+        ----------
+        .. [1] Citation1
+        .. [2] Citation2
         """
+        # pgmpy-engineer-facing comments (optional)
+        # NOTE: Summarize of implementation background for pgmpy-engineer.
+        # TODO: Summarize of need to work for pgmpy-engineer.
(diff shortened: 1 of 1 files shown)


daehyun99:
If this is necessary, we might need to update the contributing guide as well.

devtools/extension_templates/_method.py
```diff
@@ -0,0 +1,36 @@
+# This extension template provides instructions to add new methods to pgmpy.
+#
+# Please follow the instructions below:
+# 1. Copy the `template_method` below and add it to the file you are working on.
+# 2. Implement the method's core logic and write the corresponding unit tests.
```
DARHWOLF: Would it be a good idea to mention here that the tests must be `pytest` compatible, given that we are moving away from `unittets`?

daehyun99: Oh, that could be confusing. I've fixed it.

daehyun99:
I have added 'Attributes' and 'References' sections, following the structure of `_causal_discovery.py`.

Additionally, I have separated user-facing comments from engineer-facing comments.
This approach ensures that internal development notes are hidden from the official documentation and IDEs while improving overall maintainability.

ankurankan:
@daehyun99 @daehyun99 I am a bit unsure about this. The template would depend heavily on the method that we are implementing (for example, some might require `__init__` method specification; others might not). I think the main value add here is the format of the docstring, but that is also a standard numpydoc style. What do you guys think?

daehyun99:
@ankurankan 
That makes sense.

My suggestion was based on the idea that having a template and deleting what’s unnecessary might be a better way to prevent missing docstrings, rather than simply forgetting to write them during development.(I also considered that this process would automatically maintain the order of comments and ensure consistency.)
I thought it could be a safer approach to minimize such oversight.

If you don't think this is the right approach, feel free to close it.

Thanks.

ankurankan:
@daehyun99, I think a better approach to achieve that would be to write a GitHub workflow that verifies that the docstring has the required sections. Most new contributors don't really look into the templates, and I think people who have been contributing for a while do not really need the template. I am, however, not sure if it is possible to test this using a github workflow and we will need to come up with a precise rule of what is acceptable.

daehyun99:
@ankurankan 
I hadn’t considered using a GitHub workflow for documentation verifies before.
After a quick search, it seems possible to integrate `numpydoc-validation` with both `pre-commit` hooks and `GitHub workflows`.

https://numpydoc.readthedocs.io/en/latest/validation.html

Since `numpydoc` is currently only used for docstring generation, I believe extending its role to include validation would be a good approach.

https://github.com/pgmpy/pgmpy/blob/a885acf6554c5ecbf130b4e1278b0551b0fcbaf0/pyproject.toml#L66-L75

ankurankan:
@daehyun99 That's great. Could you please create an issue for it? I will close this.

Issue #2936, opened 2026-03-13
Add LiNGAM Causal Discovery Algorithm


Jatinbhardwaj-093:
### Description
LiNGAM (Linear Non-Gaussian Acyclic Model) is a causal discovery algorithm that leverages non-Gaussianity in the data to identify the unique causal Directed Acyclic Graph (DAG), even in cases where multiple graphs belong to the same Markov Equivalence Class.

Current causal discovery algorithms in pgmpy (such as PC and GES) rely on conditional independence tests or score-based optimization. These approaches typically return a Partially Directed Acyclic Graph (PDAG) because they cannot distinguish between graphs that are Markov equivalent.

LiNGAM (specifically ICA-LiNGAM) provides a powerful alternative for continuous data when the disturbance variables are non-Gaussian.

Key advantages:
- **Unique Identification:** Under the LiNGAM assumptions, the full causal DAG can be identified without undirected edges.
- **ICA-based Estimation:** The method uses Independent Component Analysis (ICA) to recover the causal structure.

### API Design
```python
from pgmpy.causal_discovery import LiNGAM

# Initialize the model
model = LiNGAM(random_state=42)

# Fit the model to continuous, non-gaussian data
model.fit(X)

# Access the learned causal graph (unique DAG)
dag = model.causal_graph_ 

# Access the estimated adjacency matrix (causal effects/weights)
weights = model.adjacency_matrix_
```

Key Class Core Methods:
- `_fit()`: Uses **sklearn.decomposition.FastICA** to estimate the unmixing matrix W.
- `_estimate_structure()`: Implements the permutation logic to find the causal order and B matrix.

### Example
```python
import pandas as pd
import numpy as np
from pgmpy.causal_discovery import LiNGAM

np.random.seed(42)
n_samples = 2000

# non-Gaussian noise terms
e0 = np.random.uniform(-1, 1, n_samples)
e1 = np.random.exponential(1, n_samples)
e2 = np.random.laplace(0, 1, n_samples)

# Create causal structure: X0 -> X1 -> X2 and X0 -> X2
x0 = e0
x1 = 0.6 * x0 + e1
x2 = -0.4 * x0 + 0.8 * x1 + e2

data = pd.DataFrame({'X0': x0, 'X1': x1, 'X2': x2})

# Discover causal structure using ICA-LiNGAM
model = LiNGAM(random_state=42)
model.fit(data)

print(model.causal_graph_.edges())
[...]

PR #2998, opened 2026-03-16
[ENH] ICA LiNGAM algorithm in casual_discovery

Jatinbhardwaj-093:

	Yes. I used Gemini 3.1 primarily for assistance in understanding and summarizing parts of the LiNGAM research paper. In particular, it helped clarify
	The algorithm for permutation of the weight matrix.
	The overall steps of the LiNGAM algorithm.
	The statistical pruning methods described in Section 6 of the paper (Wald test and A Chi-Square Test for Evaluating the Overall Fit of the Estimated Model), which are not yet implemented in the current code.
	The implementation itself was written by me based on the original paper and my understanding of the algorithm.
	The implementation was written based on the LiNGAM paper and verified by running the algorithm on synthetic datasets with known causal structures.
	A small number of tests were generated with the help of an AI tool and then manually reviewed. Currently, two simple test cases are included to verify that the discovered causal graph matches the expected structure.
	No.
	Some initial test scaffolding was generated with the help of an AI tool and then reviewed manually. However, the test suite is not finalized in this PR. A more complete and compact set of tests will be added in a subsequent update.

- Issue #2936

- Added the algorithm and few test.


Jatinbhardwaj-093:
### Remaining Tasks
- [x] Section 6: Statistical Tests for Pruning Edges.
   - Wald Test for Examining Significance of Edges
- [ ] Tests
  - Add tests on real datasets
  - Add few more synthetic/random tests with known causal structures.
- [ ] Add proper documentation

ankurankan (commented):
@Jatinbhardwaj-093 I see that you are also trying to include expert knowledge in the implementation. Did you find a reference paper for it? If not, I would suggest keeping the initial implementation without it.

Jatinbhardwaj-093:
> @Jatinbhardwaj-093 I see that you are also trying to include expert knowledge in the implementation. Did you find a reference paper for it? If not, I would suggest keeping the initial implementation without it.

I initially assumed that expert knowledge was a native feature in pgmpy, which is why I didn’t explore it in depth during the first draft. I will look for relevant reference papers to support its inclusion. If I’m unable to find sufficient backing, I’ll remove or revise this part accordingly.

Jatinbhardwaj-093:
@ankurankan 

I was thinking about the variant parameter you suggested. Please correct me if I’m wrong, but my understanding is that this would correspond to different types of LiNGAM algorithms, such as ICA, Direct, VAR, Kernel, and others.

If supporting multiple variants is a primary goal, would it make sense to create a single master wrapper class that selects the appropriate LiNGAM implementation based on the variant parameter? Which will have this parameter.

ankurankan:
@Jatinbhardwaj-093 In the case of LiNGAM, I think the two main algorithms are Direct and ICA. I don't think there is a significant overlap between the two, so might be better to implement them in separate classes and not have a variant argument. Sorry, if I suggested to add that earlier.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,333 @@
...
+
+        X_vals = X.values
+        n_samples, n_features = X_vals.shape
+        self.n_features_in_ = n_features
+        self.feature_names_in_ = list(X.columns)
+
+        # Step 1: Apply an ICA algorithm to obtain a decomposition X = AS where S has
+        # the same size as X and contains in its rows the independent components.
+        # From here on, we will exclusively work with W = A^-1.
+        ica = FastICA(random_state=self.random_state, max_iter=1000)
```
ankurankan: I would suggest taking an instance of the `FastICA` as the argument itself. This makes the method composable, and users can fine-tune the FastICA hyperparameters without us having to design our method in a way that allows users to specify that.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,290 @@
...
+    ):
+        self.alpha = alpha
+        self.return_type = return_type
+        self.fast_ica = fast_ica
+
+    def _fit(self, X: pd.DataFrame):
+
+        # Step 0: Validate inputs
+        if self.return_type != "dag":
+            raise NotImplementedError(f"Return type {self.return_type} is not yet implemented. Use 'dag'.")
```
ankurankan: We can convert a DAG to a PDAG using the `DAG.to_pdag` method.

pgmpy/tests/test_causal_discovery/test_LiNGAM.py
ankurankan, line 1: Can we compare the output of these tests to some other implementations, for example, https://github.com/cdt15/lingam

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,290 @@
...
+        self.causal_graph_.add_nodes_from(self.feature_names_in_)
+
+        # Step 6: Add edges to the graph
+        for target_idx in range(n_features):
+            for source_idx in range(n_features):
+                if B_tilde[target_idx, source_idx] != 0:
+                    source_name = self.feature_names_in_[source_idx]
+                    target_name = self.feature_names_in_[target_idx]
+
+                    self.causal_graph_.add_edge(source_name, target_name)
```
ankurankan: Should be able to use `nx.from_numpy_matrix` to directly create a graph from the adjacency matrix.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,290 @@
...
+
+class LiNGAM(_BaseCausalDiscovery):
+    """
+    LiNGAM (Linear Non-Gaussian Acyclic Model) finds the causal order under three assumptions:
+    1. The causal graph is acyclic.
+    2. The causal relationships are linear.
+    3. The noise terms are non-Gaussian.
+    A model with these three properties we call a Linear, Non-Gaussian, Acyclic Model,
+    abbreviated LiNGAM.
+
```
ankurankan: Please try to write the minimal mathematical equations of what the algorithm is computing. For example, https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/ci_tests/pearsonr_equivalence.py#L9

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,290 @@
...
+            potential_parents = causal_order[:i]
+            if len(potential_parents) == 0:
+                continue
+
+            # Extract the target variable and the feature matrix of its potential parents
+            y_target = X[:, target_node]
+            X_parents = X[:, potential_parents]
+
+            # Fit OLS regression: y = X_parents * beta + error
+            reg = LinearRegression().fit(X_parents, y_target)
```
ankurankan: I think the estimator here should also be an argument.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,290 @@
...
+    feature_names_in_: list
+        The feature names in the dataset used to learn the causal graph.
+
+    Examples
+    --------
+    >>> import pandas as pd
+    >>> from pgmpy.causal_discovery import LiNGAM
+    >>> X = pd.DataFrame({"x1": [1, 2, 3], "x2": [2, 4, 6], "x3": [3, 6, 9]})
+    >>> algo = LiNGAM()
+    >>> algo.fit(X)
```
ankurankan: Show some of the learned attributes

pgmpy/tests/test_causal_discovery/test_LiNGAM.py
ankurankan: Can we also try simulating data from a real network (`pgmpy.example_models.load_model`) and trying to learn it back?

ankurankan (changes requested):
I haven't reviewed the PR's algorithmic implementation yet, but I've added some comments on the general implementation. I also think your editor is wrapping line lengths at 80 char, we use a 120 char limit

pgmpy/tests/test_causal_discovery/test_LiNGAM.py
Jatinbhardwaj-093: > Can we compare the output of these tests to some other implementations, for example, [cdt15/lingam](https://github.com/cdt15/lingam)

I’ve already compared the current adjacency matrix with the implementation from `cdt15/lingam` locally and verified that the results are consistent.
If you intend for me to add those to the test file as well, just let me know.

> Can we also try simulating data from a real network (pgmpy.example_models.load_model) and trying to learn it back?

 This is already on my list, and I plan to work on it next as part of this PR.

Jatinbhardwaj-093:
> I haven't reviewed the PR's algorithmic implementation yet, but I've added some comments on the general implementation. I also think your editor is wrapping line lengths at 80 char, we use a 120 char limit

Thanks for the initial review. I’ll make the suggested changes.

Regarding the line wrapping, I had been manually wrapping lines at 80 characters. I wasn’t aware that `pgmpy` uses a 120 character limit. Thanks for pointing that out; since the docs aren’t finalized yet, this will save me a lot of time.

pgmpy/tests/test_causal_discovery/test_LiNGAM.py
ankurankan: > I’ve already compared the current adjacency matrix with the implementation from cdt15/lingam locally and verified that the results are consistent.

Great!! What we have done for other methods is that we create test cases comparing our values with values from other packages. And we add a comment with the code used for obtaining the values from the other package. This helps in reproducing the values later on in case something starts to fail without adding the other package as a dependency.

ankurankan:
> Regarding the line wrapping, I had been manually wrapping lines at 80 characters. I wasn’t aware that pgmpy uses a 120 character limit. Thanks for pointing that out; since the docs aren’t finalized yet, this will save me a lot of time.

I don't know if you have our pre-commit installed? It should have automatically handled this.

Jatinbhardwaj-093:
> I don't know if you have our pre-commit installed? It should have automatically handled this.

I do have pre-commit installed, but it seems like it didn’t enforce checks on the docs, or those files were already within the character limit, that’s my guess.

From here, I’ll first add the comparison with `cdt15/lingam`, since I’ve already validated that locally. Then I’ll work on adding tests with real data, while incorporating the suggested changes along the way.

I’ll open this up for review after adding the tests, so the algorithm and validation can be reviewed together.

My plan is to finalize the documentation at the very end, since some parameters/attributes may still change. This way, I can avoid going back and forth on the docs.

Jatinbhardwaj-093:
@ankurankan 
I’ve added an example validation comment for the computed graph, taking reference from `test_StructureScore.py`. Could you confirm if this convention is appropriate to follow going forward?

```python
# model = lingam.ICALiNGAM(random_state=42)
# Adj_matrix = model.adjacency_matrix_
```
```python
	# print(num_letter(networkx.convert_matrix.from_numpy_array(Adj_matrix).edges()))
    # [('A', 'B'), ('A', 'C'), ('B', 'D'), ('C', 'E')]
    assert graph.has_edge("A", "B")
    assert graph.has_edge("B", "D")
    assert graph.has_edge("C", "E")
    assert graph.has_edge("A", "C")
```

ankurankan:
@Jatinbhardwaj-093 Yes that is good.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,364 @@
...
+        B_tilde = self._prune_edges(X_vals, causal_order)
+
+        self.adjacency_matrix_ = pd.DataFrame(B_tilde, index=self.feature_names_in_, columns=self.feature_names_in_)
+
+        # Step 6: Construct graph
+        self.causal_graph_ = nx.from_numpy_array(B_tilde.T, create_using=DAG())
+        nx.relabel_nodes(
+            self.causal_graph_, mapping={i: name for i, name in enumerate(self.feature_names_in_)}, copy=False
+        )
+
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: Thanks for catching this. This kind of discrepancy might have happened due to comparison and refrence with the LiNGAM library.
I fixed it in a more subtle way without rebuilding the matrix.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,364 @@
...
+        fast_ica: FastICA | None = None,
+        estimator=None,
+        gamma: float = 1.0,
+        return_type: str = "dag",
+    ):
+        if fast_ica is None:
+            self.fast_ica = FastICA(max_iter=1000)
+        else:
+            self.fast_ica = fast_ica
+
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: > it when constructing the default FastICA (or overriding fast_ica.random_state when provided).

I’m not entirely sure about the design choice of overriding the provided fast_ica instance when `random_state or max_iter` are also passed. However, I do support adding these parameters, as they allow minimal usage without requiring users to explicitly import and configure FastICA.

I’m not marking this as resolved yet.

pgmpy/tests/test_causal_discovery/test_LiNGAM.py
```diff
@@ -0,0 +1,206 @@
...
+            [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, -0.70158999, 0.89914671, 0.0],
+        ]
+    )
+
+    np_test.assert_almost_equal(Adj_matrix.to_numpy(), test_matrix)
+
+
+def test_mpg_data(mpg_data):
+    algo = LiNGAM(fast_ica=FastICA(random_state=42))
+    algo.fit(mpg_data)
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: Going to discuss on this

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,364 @@
...
+        # Step 0: Validate inputs
+        if self.return_type != "dag":
+            raise NotImplementedError(f"Return type {self.return_type} is not yet implemented. Use 'dag'.")
+
+        X_vals = X.values
+        n_samples, n_features = X_vals.shape
+        self.n_features_in_ = n_features
+        self.feature_names_in_ = list(X.columns)
+
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: This is was on my notice.

Jatinbhardwaj-093:
@ankurankan 

A few things have been updated, while some are still pending; I’ll summarize them below:

1. Change in testing algorithm
I replaced the `Wald test` with `Adaptive Lasso`. The paper also mention's that more robust methods (e.g., bootstrapping or Lasso) can be used instead of the Wald test. Additionally, `cdt15/lingam` implementations use Adaptive Lasso, so I updated the pruning step to rely on statistical significance via Adaptive Lasso. 
The Wald test also fails to eliminate a confounding edge in the current larger graph used in `test_LiNGAM`. Although the relationship is weak, the coefficient does not shrink to zero. In contrast, Lasso is able to correctly prune this edge. You can see this in the this [commit](https://github.com/pgmpy/pgmpy/pull/2998/changes/634466f4c9ec756b46a31237bef04d1d2a039ca0).

2. Testing on real data
I attempted to evaluate the method on real datasets, but the results were not as expected. I think this is because true causal relationships are generally unknown. Other current approaches have expert knowledge (e.g., forbidden edges, temporal order constraints).

I’m unsure whether I should include my simulation tests (e.g., sachs_continuous, mpg_auto) as part of the test suite.

3. Overall status
I’ve tested most parts, and the implementation should now be ready for review.

There is one Copilot review comment that I haven’t resolved yet; please take a look. Also, regarding your earlier feedback on PDAGs, I haven’t fully explored the graph system yet. I’ll revisit and update that part accordingly.

Please review this at your convenience.

pgmpy/tests/test_causal_discovery/test_LiNGAM.py
```diff
@@ -0,0 +1,245 @@
+# The tests in this file are validated against the lingam package v1.12.2.
+# Helper function to convert node indices to letters (used for validation only):
+# def num_letter(edges):
+#     def num_to_letter(n):
+#         return chr(ord('A') + n)
+#     return [(num_to_letter(u), num_to_letter(v)) for u, v in edges]
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: This is a refrence comment for retesting

pgmpy/tests/test_causal_discovery/test_LiNGAM.py
```diff
@@ -0,0 +1,245 @@
...
+
+    # The test_matrix (adapted from the reference lingam package) stores edge weights in [target, source] orientation.
+    # Transpose it to match pgmpy's standard graph adjacency convention of [source, target].
+    np_test.assert_almost_equal(Adj_matrix.to_numpy(), test_matrix.T, decimal=2)
+
+
+def test_mpg_data(mpg_data):
+    lingam = LiNGAM(random_state=42)
+    lingam.fit(mpg_data)
+
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: This is currently not discussed

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,399 @@
...
+        try:
+            X_vals = np.asarray(X.values, dtype=float)
+        except ValueError as e:
+            raise ValueError("All features must be numeric.") from e
+
+        n_samples, n_features = X_vals.shape
+        self.n_features_in_ = n_features
+        self.feature_names_in_ = list(X.columns)
+
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: These are set for the attribute uniformity

pgmpy/causal_discovery/__init__.py
```diff
@@ -1,13 +1,15 @@
...
+from .PC import PC
+
+__all__ = [
+    "ExpertInLoop",
+    "ExpertKnowledge",
+    "GES",
+    "HillClimbSearch",
+    "LiNGAM",
+    "PC",
+]
```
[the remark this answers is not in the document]
Jatinbhardwaj-093: @copilot apply changes based on [this feedback](https://github.com/pgmpy/pgmpy/pull/2998#discussion_r3033352855)

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+       $\sum_{i} \frac{1}{|(W_{perm})_{ii}|}$.
+    3. **Diagonal Scaling**: Normalize the rows of $W_{perm}$ by dividing each row
+       by its corresponding diagonal element, resulting in a matrix $W_{scaled}$ with
+       ones on the diagonal. Compute the connection strength matrix estimate as
+       $\hat{B} = I - W_{scaled}$.
+    4. **Causal Ordering**: Discover a valid causal ordering of the variables by
+       recursively identifying and removing nodes with no parents from $\hat{B}$.
+    5. **Edge Pruning**: Construct the causal adjacency matrix $\tilde{B}$
+       by applying sparse regression (Adaptive Lasso) to prune statistically
+       insignificant edges based on the discovered causal ordering.
```
ankurankan, line 49: I would suggest not adding a new section here. It will mess up the formatting of the documentation website. I would also suggest to keep this algorithm description much shorter. We already do the step wise description in the code. This description is supposed to be shorter than that and include any important equations. So, this can a paragraph in the following order: The model assumption (the equation) -> uses ICA to compute; add the equation -> does permutation to recover causal order -> and lastly pruning si done to get a sparser structure.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+    Parameters
+    ----------
+    fast_ica : sklearn.decomposition.FastICA
```
ankurankan, line 53: This should be called `ica`, FastICA is a variant.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+        An instance of FastICA to use for independent component analysis. If None,
+        a default FastICA instance with `max_iter=1000` is used.
+
+    estimator : sklearn.linear_model._base.LinearModel
```
ankurankan, line 57: This is the estimator used for pruning? If that is the case, it should be clear from its name.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+        An instance of a linear model to use for estimating the causal relationships.
+        If None, a default LinearRegression instance is used.
+
+    gamma : float, default=1.0
+        The exponent used to calculate the adaptive weights in the Adaptive Lasso.
```
ankurankan, line 62: This description doesn't tell me anything about how it would affect the output.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+    return_type : str, default="dag"
+        The type of graph to return. Currently only "dag" is supported.
```
ankurankan, line 65: Should support both dag and pdag. You can simply do `dag.to_pdag()` to convert a dag to pdag.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+    random_state : int, default=None
+        Seed for the random number generator used by the ICA algorithm. Ensures
+        reproducibility across repeated algorithm runs. If a `fast_ica` instance
+        is passed, this will override its existing random state.
+
+    max_iter : int, default=None
+        Maximum number of iterations for the FastICA algorithm. Defaults to 1000.
+        If a `fast_ica` instance is passed, this will override its existing max_iter.
```
ankurankan, line 74: Why are we asking for this separately? Shouldn't this already be in the `FastICA` object passed? This is not a composable design.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+    Examples
+    --------
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> from pgmpy.causal_discovery import LiNGAM
+    >>> X = pd.DataFrame(np.random.uniform(size=(100, 3)), columns=list("ABC"))
+    >>> X["B"] = 2.0 * X["A"] + X["B"]
+    >>> X["C"] = -1.5 * X["B"] + X["C"]
+    >>> lingam = LiNGAM()
```
ankurankan, line 99: Use an existing dataset from the example datasets module.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+        Returns
+        -------
+        self : pgmpy.causal_discovery.LiNGAM
+            Returns the instance with the fitted attributes.
+        """
+
+        # Step 0: Validate inputs
+        if self.return_type != "dag":
+            raise NotImplementedError("Only return_type='dag' is supported.")
+
```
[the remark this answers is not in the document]
ankurankan: This comment is valid, will need to be addressed.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+        try:
+            X_vals = np.asarray(X.values, dtype=float)
+        except ValueError as e:
+            raise ValueError("All features must be numeric.") from e
```
ankurankan, line 150: Please use `preprocess_data` module from `pgmpy.utils.utils.py`.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+        _, n_features = X_vals.shape
```
ankurankan, line 152: `self.n_features_in_` is already defined in the parent class.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+        # Step 1: Resolve internal estimators and apply ICA
+        if self.estimator is None:
+            self._estimator = LinearRegression()
+        else:
+            self._estimator = clone(self.estimator)
```
ankurankan, line 158: Isn't this supposed to be Lasso by default?

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+        if self.fast_ica is None:
+            _max_iter = self.max_iter if self.max_iter is not None else 1000
+            ica = FastICA(max_iter=_max_iter, random_state=self.random_state)
+        else:
+            ica = clone(self.fast_ica)
+            if self.random_state is not None:
+                ica.set_params(random_state=self.random_state)
+            if self.max_iter is not None:
+                ica.set_params(max_iter=self.max_iter)
```
ankurankan, line 168: I don't think we should set them. Should use what the user provided. This is against composability design pattern.

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+
+        ica.fit(X_vals)
+        W = ica.components_
+
+        if W.shape[0] != n_features:
+            raise ValueError(f"FastICA n_components must equal n_features (got {W.shape[0]} != {n_features}).")
```
ankurankan, line 174: Under what conditoin can this happen?

pgmpy/causal_discovery/LiNGAM.py
```diff
@@ -0,0 +1,397 @@
...
+            raise ValueError("Near-zero diagonal elements in ICA permutation.")
+        W_scaled = W_perm / diag_W_perm[:, np.newaxis]
+        B_hat = np.eye(n_features) - W_scaled
+
+        # Step 4: Find a causal order
+        causal_order = self._causal_order(B_hat)
+
+        # Step 5: Construct the causal adjacency matrix.
+        if causal_order is None:
+            raise ValueError("Graph contains unresolvable cycles.")
```
ankurankan, line 196: When does this happen? The current error message doesn't help to debug.

ankurankan (changes requested):

pgmpy/causal_discovery/LiNGAM.py
Jatinbhardwaj-093: This occurs if the user sets the `n_components` parameter lower than the `number of features` for a FastICA instance. I have already written a test case for this scenario:
```
def test_lingam_error_dimensionality_mismatch(rand_data):
    lingam = LiNGAM(ica=FastICA(n_components=1))
    with pytest.raises(ValueError, match="n_components must equal n_features"):
        lingam.fit(rand_data)
```

I believe I should validate this condition before beginning the computation. I will update the code and the test accordingly.

pgmpy/causal_discovery/LiNGAM.py
Jatinbhardwaj-093: I updated this.

pgmpy/causal_discovery/LiNGAM.py
Jatinbhardwaj-093: This error typically signifies that there are cycles in the data (which can occur due to a lack of non-Gaussianity) or that the dataset is insufficient for the algorithm to converge.

pgmpy/causal_discovery/LiNGAM.py
Jatinbhardwaj-093: The algorithm utilizes `Lasso` primarily for structural pruning (identifying edges to set to zero). Since Lasso is inherently biased due to its penalty term, for correctness algorithm  rely on the `_estimator (OLS)` to provide the true causal effects in their original units. Essentially, the OLS refit is performed on the remaining edges after Lasso has removed the coefficients it deemed `zero`.
 
Additionally, this estimator is used to calculate the weights for the penalty matrix in the initial Lasso stage.

Jatinbhardwaj-093:
@ankurankan 
Sorry for the delay; I had a very busy week. I have now addressed all the review comments. Could you please take another look?

PR #3013, opened 2026-03-17
Add the separator parameter to _BaseDataset.load_dataframe method

### LINKED ISSUE #3003: [ENH] Add a property `sep` to `_BaseDataset.load_dataframe`
The current data loader in `_BaseDataset.load_dataframe` only accepts tab separator files. So, CSV files can't be parsed with the default data loader.

The property `sep` should be added in the class `_BaseDataset`, and `sep` shoul be added as a parameter when initializing `_BaseDataset`.

**Additional context**
I encountered this problem when trying to add a new dataset, which is in this PR #2695 

cc @ankurankan

stephanielees:

_I didn't use LLM._

_I have tested that if the data class needs a tab separator but the `sep` argument in the method is set to `,`, an error message would come out._

_Not applicable (didn't use LLM)._

_No._

- Fixes #3003 

- Add `sep` as the new parameter to the data loader in `_BaseDataset`.
- Revise the docstring to explain the parameter and output of this data loader.


pgmpy/datasets/_base.py
```diff
@@ -145,10 +145,20 @@ def _get_raw_data(cls, data_type, url) -> bytes:
         return raw_data
 
     @classmethod
-    def load_dataframe(cls) -> pd.DataFrame:
-        """Fetches/reads from cache the data associated with the dataset."""
+    def load_dataframe(cls, sep="\t") -> pd.DataFrame:
```
ankurankan: `sep` shouldn't be an argument. You should directly use `cls.sep`, and `sep` should be defined as an attribute in dataset classes. For example, have a look at how `missing_values_marker` is used (line 162).

ankurankan (changes requested):

pgmpy/datasets/_base.py
stephanielees: @ankurankan oh, right. Sorry. I'll change it asap.

Btw, I just thought of this... I think we should also add the `sep` attribute to the dataset template in `pgmpy/devtools`. Is that right?

ankurankan: @stephanielees Yes, we should.

stephanielees: @ankurankan Ok, I'm going to work on it.

ankurankan (changes requested):
@stephanielees I think you still need to update the extension template.

stephanielees:
@ankurankan 
> I think you still need to update the extension template.

I had a problem when running `git commit` for the extension template. It failed the flake8 test because there are some comments that are too long. One example is line 4 which has 110 characters. Apparently flake8 wants to format the code so that each line has 80 characters max.

This is the 4th line of `_dataset.py`:
```# 1. Copy this file to `pgmpy/datasets` and rename the file as `your_dataset_name.py` (e.g., `my_dataset.py`).```
How do you think we should change this? All the lines that are too long are the comments, such as the TODO stuff.
While some TODO instructions are easy to split because there is more than one sentence, the line 4 is only one sentence.
I think we can split it to something like this:
```
# 1. Copy this file to `pgmpy/datasets` and rename the file as
#     `your_dataset_name.py` (e.g., `my_dataset.py`)
```
Or maybe you can do something about the flake8 test?

[Update]
Apparently the `_base.py` also fails the same test. But the difference is that there is one line in this file that is part of a function, not just comments. For example, in line 33 and 34 are the returned value of `__str__(self)`:
```
def __str__(self) -> str:
        return (
            f"Dataset(name={self.name}, \n data=DataFrame of size: {self.data.shape}, \n "
            f"expert_knowledge={self.expert_knowledge}, \n ground_truth={self.ground_truth}, \n tags={self.tags})"
        )
```

ankurankan:
@stephanielees This seems to due to some of my recent commits. I am working on a fix for it.

ankurankan (dismissed):

ankurankan:
@stephanielees Thanks for the PR. I have made some small changes to simplify further.

ankurankan (approved):

Issue #2931, opened 2026-03-12
[ENH] List of causal discovery algorithms


ankurankan:
This is an umbrella issue for maintaining and prioritizing causal discovery algorithms that we would like to implement. Add any suggestions in the comments.

#### Constraint-Based
- [ ] RFCI
- [ ] BCCD 
- [ ] Greedy SP
- [ ] MIIC

#### Score Base
- [ ] Nonparametric GES #2531 
- [ ] GPS: Greedy PAG Search (https://proceedings.mlr.press/v180/claassen22a/claassen22a.pdf)
- [x] FGES 
- [ ] BOSS #2912 
- [x] Sparset Permutation (SP) 
- [ ] GRaSP
- [x] Sortnregress with variants variance and R2. #3019 

#### Optimization Based
- [ ] NOTEARS #1342
- [ ] NOTEARS-NLP  
- [ ] NOTEARS-SOB
- [ ] DAGMA #2513 
- [ ] GOLEM #2906
- [ ] CALM 
- [ ] D-Struct

#### Functional 
- [ ] LiNGAM
- [ ] DirectLiNGAM #3256 
- [ ] ICA-LiNGAM #2936 
- [ ] RESIT
- [ ] SLOPE

#### Deep Learning
- [ ] DAG-GNN
- [ ] GraN-DAG
- [ ] CAREFL
- [ ] DiBS #3233 
- [ ] DiffAN #3147 
- [ ] CASTLE #3140 

#### Interventional data
- [ ] GIES
- [ ] DCDI
- [ ] AVICI

anusa-saha:
@ankurankan I have opened an issue on BOSS algorithm #2912 
working on the structure proposal

anusa-saha:
@ankurankan about GOLEM, should I base my structure on NOTEARS and DAGMA implementation??

ankurankan:
@anusa-saha They will all be similar but we currently do not have any implementation of any of those.

anusa-saha:
@ankurankan so should I propose an API structure from scratch like i am doing for BOSS??

ankurankan:
@anusa-saha Yes.

anusa-saha:
@ankurankan I have made a proposal for the structure in #2912 
Please have a look

sabasiddique1:
Hi @ankurankan,

I’m interested in contributing to causal discovery algorithms.

[...]

PR #3042, opened 2026-03-18
Add GraN-DAG implementation and integration

Aarnav5037:
Your checklist for this pull request

Did you use an LLM for any assistance with this PR? Please describe in detail (around a paragraph) how and what you used it for?

Yes, LLM was used to create a basic skeleton of what the workflow would look like, also it was used to understand some of the more complex functions from the original algorithm's implementation.

Are you able to fully explain your changes? We expect you to fully understand the algorithm and take full responsibility for any changes in this PR.
Yes, I have spent considerable time understanding the actual algorithm. I have gone through the code base and covered most of the important functions. Only then have i tried to implement and integrate them into pgmpy.

What steps have you taken to verify that the changes correctly address the issue? And what edge cases have you considered? Other than running tests, what else have you verified?

I have tried to implement the logic of the algorithm.

Has the LLM added try-except blocks? They will need to be removed; any error handling must be explicit.
No

Have you used LLM for generating tests? They need to be compressed into a smaller number of tests without reducing coverage.
No

Issue number(s) that this pull request fixes
Issue #2931 
Fixes #

Attempting the Implementation of the algorithm

The implementation may not be perfect, but i do believe that i have a good understanding of the algorithm.


Aarnav5037:
@ankurankan 
Most of the main functions of the algorithm like the compute_constraint are taken as is from the official implementation of the gran dag, only after understanding the function. I have only tried to modify the data loading process to suit pgmpy and removed certain logging and plotting that were being done for experiments in the original implementation. Also, just the final conversion of the adjacency matrix into the actual DAG. The main file is the train.py, rest are functions taken from the official implementation.
I've spent most of the time trying to understand the algorithm. it may look like ai overuse since it is 700 lines, but ive only used llms to understand the concept. i also acknowledge that the implementation is far from perfect, but it is only an attempt to demonstrate that i do understand the algorithm on some level and would love the opportunity to work on it with proper guidance. I understand that the implementation of such an algorithm is neither a small nor easy task and will surely take significant time.

ankurankan (changes requested):
- What license does the original code have? We can not simply copy their code. And if you copied something, you must credit the original authors and link their implementation.
- This will need to be rewritten to pgmpy standards. Documentation is missing everywhere. One algorithm should live in a single file; common components should be separated out to mixin classes, etc.
- Why are there changes to the `pgmpy/estimators` module?
- This should only work on pgmpy's pytorch backend.
- Your implementation doesn't match our causal discovery interface. We have an extension template for causal discovery methods.

PR #2638, opened 2026-02-18
Refactor conditional independence tests (CITests.py)

kajal-jotwani:
### Your checklist for this pull request
**Do not** remove this checklist. Your PR **won't** get reviewed if this checklist is not present in the PR.

Please complete the following checklist after creating the PR.

- [x] Make sure you are requesting to **pull a topic/feature/bugfix branch** (right side) against our **dev branch** (left side). Please do not request your **dev branch**.
- [x] Have you followed all the steps from our [Contributing Guide](https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md)?
- [ ] Are all the GitHub Actions checks passing? If not, they will need to be fixed before review. You can reference logs for the failing check to identify the issue.

Did you use a Large language model (LLM) to assist you in making changes? If yes, please go through the following checklist too: 
- [x] Please include a short paragraph (**not bullet points**) describing the changes you have made. This should at least include the problem your PR solves and the implemented solution. This doesn't need to be a polished description, but it **must** be written **manually** by you without any assistance from any AI tools.
- [x] Have you **manually** verified that the changes are doing exactly what is expected? Have you considered edge cases?
- [ ] Has the LLM added try-except blocks? They will need to be removed; any error handling must be explicit.
- [ ] If you used LLM for generating tests, they usually need to be compressed into a smaller set of tests. Please **manually** describe what scenarios the tests are testing for.

### Issue number(s) that this pull request fixes
- Fixes #2547 

### List of changes to the codebase in this pull request
This PR refactors the conditional independence (CI) tests into a class-based structure.
- All CI tests are moved into a new `pgmpy/ci_tests/` folder.
- All CI tests inherit from a new `_BaseCITest` class which has shared logics for input validation and boolean handling.
- The previous `CITestRegistry ` has been removed and replaced with tag-based lookup using `skbase.lookup.all_objects` (reference from metrics)

Previously, CI tests were used as standalone functions:
```
chi_square(X="A", Y="B", Z=["C"], data=data)
```
They are now instantiated as classes:
```
test = ChiSquare(data)
test("A", "B", ["C"])
```

The dataset is now passed during class initialization rather than to the test call. This structure avoids passing the DataFrame repeatedly and allows caching to be introduced later by caching only (X, Y, Z) calls without hashing the DataFrame.

PR #3072, opened 2026-03-20
[ENH] Add estimator argument to Pillai Trace CI test (#3028)

### LINKED ISSUE #3028: [ENH] Add `estimator` argument to pillai trace conditional independence test
Similar to #3027, we should allow the method to accept any sklearn estimator as an argument. In the case of Pillai trace, however, we will need the estimator to have a `predict_proba` method if there are any discrete variables in the data.

### LINKED ISSUE #3200: [ENH] Resual mixin for residualization based CI tests
Many of our current conditional tests depend on a residualization process, for example, GCM, Pillai Trace, pearsonr, and #3198. Currently, each implements its own residualization procedure, leading to duplicated code. A better way to do this would be to define a `_ResidualMixin` class that implements methods for computing the residuals, and each of the test classes can inherit this `_ResidaulMixin` class and use its residualization methods.  We should try to implement `pillai_trace`'s residualization method in the `_ResidualMixin` as it is the most generalized version (supports both continuous and discrete) variables.

hanara2112:

1. Did you use an LLM for any assistance with this PR? Please describe in detail (around a paragraph) how and what you used it for?

- I used a coding assistance to navigate myself to the new CI refactored architecture (PR https://github.com/pgmpy/pgmpy/pull/2638 ) and help me write the logic for one hot encoding for the Z variable (pd.get_dummies) to make them compatible with scikit-learn.

2. Are you able to fully explain your changes? We expect you to fully understand the algorithm and take full responsibility for any changes in this PR.

- Yes
 
  - I added a estimator=None to preserve backwards compatibilitiy with XGBoost.
  - I used the sklearn.base.clone() to prevent the X|Z and Y|Z estimators from sharing trained state.
  - I had to add pd.get_dummies for the Z columns because standard sklearn estimators crash on pandas category strings natively (unlike XGBoost).  
  - I explicitly added a check (hasattr) to raise a ValueError if a discrete target doesn't have predict_proba.

3. What steps have you taken to verify that the changes correctly address the issue? And what edge cases have you considered? Other than running tests, what else have you verified?

- I wrote 3 specific tests, one for pure continuous (GradientBoostingRegressor), one for pure categorical (RandomForestClassifier), and one that safely crashes when predict_proba is missing (LinearRegression). I verified it passes locally using pytest.

4. Has the LLM added try-except blocks? They will need to be removed; any error handling must be explicit.

- No. I did not add any new try-except blocks. I explicitly used raise ValueError if predict_proba is missing.

5. Have you used LLM for generating tests? They need to be compressed into a smaller number of tests without reducing coverage.

- AI helped me draft the tests, but they are minimal, concise, and only specifically target the 3 new custom-estimator edge cases. There is no bloated test coverage.

Fixes: #3028 
Fixes: #3200 

List of changes to the codebase in this pull request:

  1. pgmpy/ci_tests/pillai_trace.py: Added the estimator=None parameter to the PillaiTrace class.
  2. pgmpy/ci_tests/pillai_trace.py: Implemented _get_predictions_custom to handle custom sklearn estimators. This uses sklearn.base.clone to isolate X|Z and Y|Z state, applies pd.get_dummies to categorical Z variables for sklearn compatibility, and validates predict_proba for discrete targets.
  3. pgmpy/tests/test_ci_tests/test_pillai_trace.py: Added 3 tests (test_pillai_custom_estimator_continuous, test_pillai_custom_estimator_discrete, and test_pillai_custom_estimator_no_predict_proba) to verify correct continuous/discrete detection and edge cases.


hanara2112:
@ankurankan I’ve tested the merge request as instructed, and it’s working as expected. Please let me know if any further changes are needed and if i can add the same estimator logic to the GCM CI test if thats not already been taken care of.

Thanks for your support!

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -46,16 +57,23 @@ class PillaiTrace(_BaseCITest):
         "requires_data": True,
     }
 
-    def __init__(self, data: pd.DataFrame, seed=None):
+    def __init__(self, data: pd.DataFrame, seed=None, estimator=None):
```
ankurankan: Should be before `seed`

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -92,6 +110,44 @@ def fit_predict(target_col):
 
         return pred_x, pred_y, x_cat_index, y_cat_index
 
+    def _get_predictions_custom(self, X, Y, Z, data):
```
ankurankan: Why do we need this method? Doesn't XGBoost also have the same fit-predict API? Can this not be combined with it?

ankurankan (changes requested):

pgmpy/ci_tests/pillai_trace.py
hanara2112: The main reason they were separated is because XGBoost (enable_categorical=True) can handle pandas categories natively, whereas standard sklearn estimators crash unless we one-hot encode Z first (with pd.get_dummies). However, you're totally right that they should be combined!
I'll update  _get_predictions to handle both cases smoothly and push an update.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -46,44 +57,64 @@ class PillaiTrace(_BaseCITest):
...
-            ) from None
 
-        enable_categorical = any(data.loc[:, Z].dtypes == "category")
+        if self.estimator is None:
+            try:
+                from xgboost import XGBClassifier, XGBRegressor
+            except ImportError as e:
+                raise ImportError(
+                    f"{e}. xgboost is required for using pillai_trace test. Please install using: pip install xgboost"
+                ) from None
```
ankurankan: Please use scikit-base._safe_import. Have a look at other places it has been used for examples

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -10,16 +11,26 @@ class PillaiTrace(_BaseCITest):
...
         The dataset in which to test the independence condition.
 
     seed : int, optional
-        Random seed used for the underlying XGBoost models.
+        Random seed used for the underlying XGBoost models. Only used when
+        ``estimator`` is ``None``.
+
+    estimator : sklearn estimator instance, optional
+        An sklearn-compatible estimator to use for computing residuals. If
+        ``None`` (default), uses XGBoost (``XGBClassifier`` for categorical
```
ankurankan: As I had mentioned before, it doesn't need to be sklearn estimator as long as it has `fit`, `predict`, or `predict_proba` methods

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -46,44 +57,64 @@ class PillaiTrace(_BaseCITest):
...
+            try:
+                from xgboost import XGBClassifier, XGBRegressor
+            except ImportError as e:
+                raise ImportError(
+                    f"{e}. xgboost is required for using pillai_trace test. Please install using: pip install xgboost"
+                ) from None
+
+        # Sklearn estimators do not natively handle pandas categories, so one-hot encode Z
+        if self.estimator is not None:
+            Z_data = pd.get_dummies(data.loc[:, Z])
```
ankurankan: What if some variables are continuous?

ankurankan (changes requested):

pgmpy/ci_tests/pillai_trace.py
hanara2112: I replaced it, but earlier we had a custom error message — now it’s gone
Are you okay with that tradeoff?

pgmpy/ci_tests/pillai_trace.py
hanara2112: ah yes - continuous variables are left unchanged since pd.get_dummies only encodes categorical columns. I’ve also made this explicit by selecting only categorical columns before encoding, so there’s no ambiguity for future readers.

Alternatively, we could split categorical and numeric columns and concatenate after encoding for even clearer separation — happy to update if you prefer that.

pgmpy/tests/test_ci_tests/test_pillai_trace.py
```diff
@@ -180,6 +180,53 @@ def test_pillai_dependent(self):
...
+    def test_pillai_custom_estimator_no_predict_proba(self):
+        """Test that ValueError is raised when estimator lacks predict_proba for discrete variable."""
+        from sklearn.linear_model import LinearRegression
+
+        estimator = LinearRegression()
+
+        test = PillaiTrace(data=self.df_indep_cat_cat, estimator=estimator)
+
+        with self.assertRaisesRegex(ValueError, "predict_proba"):
+            test("X", "Y", ["Z1", "Z2", "Z3"])
```
ankurankan: Please combine these into two tests. One for testing errors and one for testing custom estimators.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -46,7 +48,17 @@ class PillaiTrace(_BaseCITest):
...
+        ``estimator`` is ``None``.
+
+    estimator : estimator instance, optional
+        Any estimator with `fit`, `predict`, and `predict_proba` (if testing
+        discrete variables) methods. If
+        ``None`` (default), uses XGBoost (``XGBClassifier`` for categorical
+        targets, ``XGBRegressor`` for continuous). When provided, the
+        estimator must have a ``predict_proba`` method if any of the test
+        variables (X or Y) are categorical. A fresh clone of the estimator
+        is used for each variable to avoid state leakage.
```
ankurankan: This language here needs to be a bit tighter. There's some repetition. The last sentence shouldn't be included here as it says something about what the algorithm does and is not relevant for describing the argument.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,44 +85,63 @@ class PillaiTrace(_BaseCITest):
...
+        if self.estimator is not None:
+            # Explicitly select only category/object columns to dummy encode, leaving continuous variables untouched.
+            cat_cols = data.loc[:, Z].select_dtypes(include=["category", "object"]).columns
+            Z_data = pd.get_dummies(data.loc[:, Z], columns=cat_cols)
+            enable_categorical = False
+        else:
+            Z_data = data.loc[:, Z]
+            enable_categorical = any(Z_data.dtypes == "category")
 
         def fit_predict(target_col):
```
ankurankan: Would be better to move this function out of `_get_predictions` method since `PillaiTrace` is a separate class now.

ankurankan (changes requested):

hanara2112:
Hi @ankurankan, i have updated the requested changes and pushed.

Summary of updates:

1. Tests: Merged the discrete and continuous custom estimator tests into a single test_pillai_custom_estimators() function. The ValueError test has been kept separate for clarity.
2. Refactoring: Extracted the nested fit_predict function from _get_predictions into a dedicated private class method _fit_predict, keeping the PillaiTrace namespace cleaner.
3. Docstring: Simplified and tightened the estimator docstring, removed internal implementation details (like object cloning).
4. Consistency Fix: Updated the legacy pillai_trace function in estimators/CITests.py to replace the manual XGBoost import with _safe_import("xgboost"), ensuring consistency across the codebase.

Please let me know if any further adjustments are needed!

hanara2112:
hi @ankurankan , there is a conflict for this pr. I don't want to break anything, can you suggest a fix ?

.github/PULL_REQUEST_TEMPLATE.md
ankurankan, line 1: Why was this file changed?

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -46,7 +48,14 @@ class PillaiTrace(_BaseCITest):
         The dataset in which to test the independence condition.
 
     seed : int, optional
-        Random seed used for the underlying XGBoost models.
+        Random seed used for the underlying XGBoost models. Only used when
+        ``estimator`` is ``None``.
```
ankurankan: This needs to be updated.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -46,7 +48,14 @@ class PillaiTrace(_BaseCITest):
...
+
+    estimator : estimator instance, optional
+        Any estimator with `fit`, `predict`, and `predict_proba` (if testing
+        discrete variables) methods. If ``None`` (default), uses XGBoost
+        (``XGBClassifier`` for categorical targets, ``XGBRegressor`` for
+        continuous).
```
ankurankan: This needs to go above seed. The order should be same as the `__init__` signature.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,49 +82,63 @@ class PillaiTrace(_BaseCITest):
...
-                f"{e}. xgboost is required for using pillai_trace test. Please install using: pip install xgboost"
-            ) from None
-
-        enable_categorical = any(data.loc[:, Z].dtypes == "category")
-
-        def fit_predict(target_col):
-            is_cat = data.loc[:, target_col].dtype == "category"
-            model_cls = XGBClassifier if is_cat else XGBRegressor
+    def _fit_predict(self, data, target_col, Z_data, enable_categorical):
+        is_cat = data.loc[:, target_col].dtype == "category"
```
ankurankan: Why do you compare with only `category` here but in line 133 you have used `category` and `object`?

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,49 +82,63 @@ class PillaiTrace(_BaseCITest):
...
-            raise ImportError(
-                f"{e}. xgboost is required for using pillai_trace test. Please install using: pip install xgboost"
-            ) from None
-
-        enable_categorical = any(data.loc[:, Z].dtypes == "category")
-
-        def fit_predict(target_col):
-            is_cat = data.loc[:, target_col].dtype == "category"
-            model_cls = XGBClassifier if is_cat else XGBRegressor
+    def _fit_predict(self, data, target_col, Z_data, enable_categorical):
```
ankurankan: Why do we need so many arguments here? Why not just send the target column instead of both `data` and `target_col`

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,49 +82,63 @@ class PillaiTrace(_BaseCITest):
...
+        is_cat = data.loc[:, target_col].dtype == "category"
+        target_data = data.loc[:, target_col]
+        cat_index = None
+
+        if self.estimator is not None:
+            model = clone(self.estimator)
+        else:
+            xgboost = _safe_import("xgboost")
```
ankurankan: Needs to go outside this method.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,49 +82,63 @@ class PillaiTrace(_BaseCITest):
...
-        pred_y, y_cat_index = fit_predict(Y)
+        Uses the user-provided ``self.estimator`` if set, otherwise falls
+        back to XGBoost. Uses ``self.seed`` for reproducibility when using
+        XGBoost.
+        """
+        # Sklearn estimators do not natively handle pandas categories, so one-hot encode Z
+        if self.estimator is not None:
+            # Explicitly select only category/object columns to dummy encode, leaving continuous variables untouched.
+            cat_cols = data.loc[:, Z].select_dtypes(include=["category", "object"]).columns
+            Z_data = pd.get_dummies(data.loc[:, Z], columns=cat_cols)
```
ankurankan: Do we need to compute `cat_cols`? Doesn't `columns=None` here already do that?

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,49 +82,63 @@ class PillaiTrace(_BaseCITest):
...
+            enable_categorical = False
```
ankurankan: Doesn't `enable_categorical` depend on whether `X` or `Y` is categorical?

pgmpy/estimators/CITests.py
ankurankan, line 1: No changes in this file is required. This will be deprecated soon.

ankurankan (changes requested):
I don't think you understand how the method works or the changes you are making, and you are using an LLM for all the changes. Please spend some time to understand what the method is supposed to do.

.github/PULL_REQUEST_TEMPLATE.md
hanara2112: i will revert the changes.  sorry for the inconvenience

pgmpy/ci_tests/pillai_trace.py
hanara2112: Makes sense, data is redundant since the method already has self.data. I've simplified the signature to _fit_predict(self, target_col, Z_data). It fits an estimator on Z to predict the target column and returns the raw predictions (probabilities for categorical targets, point predictions for continuous).

pgmpy/ci_tests/pillai_trace.py
hanara2112: i was initially thinking if Python caches imports in sys.modules, so repeated imports would effectively be just dictionary lookups and not reloading whole module, it’s still better for code clarity and style to move the import outside the method. I’ll update that

pgmpy/ci_tests/pillai_trace.py
hanara2112: Right, pd.get_dummies() with columns=None already encodes all category/object columns by default. Removed the explicit cat_cols selection.

pgmpy/ci_tests/pillai_trace.py
hanara2112: Thank you for mentioning this. I feel i should clarify this.
enable_categorical is specific to XGBoost and controls whether the input features (Z) contain categorical columns.

The choice of treating X or Y as categorical is handled separately (via is_cat, deciding between classifier/regressor), so it’s not directly related to this flag.

That said, I agree the current placement was confusing, I’ve now limited enable_categorical to the XGBoost path inside _fit_predict, where it’s derived from Z’s dtypes, and removed it from the generic estimator path.

hanara2112:
> I don't think you understand how the method works or the changes you are making, and you are using an LLM for all the changes. Please spend some time to understand what the method is supposed to do.

Hi @ankurankan , thanks for the feedback. I get why it might look like I didn’t fully understand it, so just wanted to explain how I was thinking about the method and the changes.

At a high level, my understanding is that Pillai trace here is basically: remove the effect of Z from X and Y using a model, then check how much correlation is left between the residuals. That’s done using CCA, and then the statistic is the sum of squared canonical correlations, with an F-approximation for the p-value.

About the implementation:
1. For _fit_predict, the xgboost arg isn’t meant as a flag or anything. I’m just passing the module (or None). The idea was to keep it optional and only import it when we actually need it (i.e., no custom estimator). If a user passes their own estimator, we don’t import XGBoost at all.
2. There are basically two paths:
    1. custom estimator → one-hot encode Z (pd.get_dummies) since most sklearn models won’t handle category dtype
    2. XGBoost → pass Z as-is with enable_categorical=True so we don’t do unnecessary encoding
3. For categorical targets, I used pd.factorize and kept the index. This is mainly so that later in residual computation, the dummy columns line up with the order of predict_proba. Otherwise the subtraction can silently go wrong if columns are in a different order.
4. The predict_proba check is only for custom estimators when the target is categorical, since we need probabilities there. XGBoost classifiers already support that so I didn’t add anything extra there.
5. I used clone(self.estimator) because _fit_predict runs separately for X and Y, so I didn’t want to reuse a fitted model instance across both.

I might be missing something subtle, so happy to go deeper into any part, especially the residual computation or anything in _fit_predict if that’s where the concern is.

ankurankan:
> About the implementation:
> For _fit_predict, the xgboost arg isn’t meant as a flag or anything. I’m just passing the module (or None). The idea was to keep it optional and only import it when we actually need it (i.e., no custom estimator). If a user passes their own estimator, we don’t import XGBoost at all.
There are basically two paths:
custom estimator → one-hot encode Z (pd.get_dummies) since most sklearn models won’t handle category dtype
XGBoost → pass Z as-is with enable_categorical=True so we don’t do unnecessary encoding

Did you use an AI to write this part? I see UTF-8 characters in the text which no one writes manually.

hanara2112:
> > About the implementation:
> > For _fit_predict, the xgboost arg isn’t meant as a flag or anything. I’m just passing the module (or None). The idea was to keep it optional and only import it when we actually need it (i.e., no custom estimator). If a user passes their own estimator, we don’t import XGBoost at all.
> > There are basically two paths:
> > custom estimator → one-hot encode Z (pd.get_dummies) since most sklearn models won’t handle category dtype
> > XGBoost → pass Z as-is with enable_categorical=True so we don’t do unnecessary encoding
> 
> Did you use an AI to write this part? I see UTF-8 characters in the text which no one writes manually.



I agree that the particular part with unicode character was generated by llm to state my thoughts clearly. 
Although now i understand that sharing my own thoughts and getting feedback is far more valuable then using an llm and is also more benefitial for the community.
I apologise for the same and confirm that now on i will use ai purely as an assistant and nothing else.

hanara2112:
A few comments on the current standing of pillai estimator and the _fit_predict() method: 
1. Right now i am thinking that passing xgboost module as a parameter is kind of  awkward. The safe_import call happens in run_test, then the module object is threaded into _fit_predict().
-- A more cleaner approach would be to move safe_import call into fit_predict itself and call only when needed. This would remove the xgboost=None param entirely and safe_import caches the import so calling it twice( Once for X and once for Y) would be negligible. 

2. Secondly i am not fully sure weather keeping both seed and random_state is necessary. as XGboost's seed param is deprecated in favopr of random_state, and passing both might be redundant. 
-- So i propose to just use random_state=self.seed. Thats it. 

3. Also i was trying to test pillai_estimator with some custom estimator and compare it against existing CI tests like Pearsonr or GCM to validate the behaviour. 

Please let me know if this sounds good to you and if any other changes are recommended for this pr. 

Thank you!

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -129,43 +131,70 @@ def run_test(
...
+
+        # Step 1: Add an intercept column so the estimator always has at least one feature.
         Z = Z + ["_intercept_Z"]
         data = data.assign(_intercept_Z=np.ones(data.shape[0]))
 
-        # Step 2: Get the predictions
-        pred_x, pred_y, x_cat_index, y_cat_index = self._get_predictions(X, Y, Z, data)
+        # Step 2: Prepare Z data. Custom sklearn estimators need one-hot encoded Z;
+        # XGBoost handles categorical columns natively.
+        xgboost = None
```
ankurankan: Remove `xgboost`.

pgmpy/estimators/CITests.py
```diff
@@ -633,12 +634,9 @@ def _get_predictions(X, Y, Z, data, **kwargs):
...
     """
-    try:
-        from xgboost import XGBClassifier, XGBRegressor
-    except ImportError as e:
-        raise ImportError(
-            f"{e}. xgboost is required for using pillai_trace test. Please install using: pip install xgboost"
-        ) from None
+    xgboost = _safe_import("xgboost")
+    XGBClassifier = xgboost.XGBClassifier
+    XGBRegressor = xgboost.XGBRegressor
```
ankurankan, line 639: Make the default as `RandomForest`.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,51 +83,43 @@ class PillaiTrace(_BaseCITest):
...
-            ) from None
-
-        enable_categorical = any(data.loc[:, Z].dtypes == "category")
-
-        def fit_predict(target_col):
-            is_cat = data.loc[:, target_col].dtype == "category"
-            model_cls = XGBClassifier if is_cat else XGBRegressor
+    def _fit_predict(self, target_col, Z_data, xgboost=None):
+        """Fit an estimator on Z_data to predict target_col, return predictions and category index."""
+        is_cat = self.data.loc[:, target_col].dtype == "category"
```
ankurankan: Use `pgmpy/utils.py:preprocess_data` method here. It returns a dict of detected variable type for each column in the dataframe.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -129,43 +131,70 @@ def run_test(
...
+            Conditioning variables.
+
+        Returns
+        -------
+        statistic : float
+            The Pillai's trace statistic.
+        p_value : float
+            The p-value.
         """
         data = self.data
```
ankurankan: Remove this line.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -73,51 +83,43 @@ class PillaiTrace(_BaseCITest):
         "requires_data": True,
     }
 
-    def __init__(self, data: pd.DataFrame, seed=None):
-        self.seed = seed
+    def __init__(self, data: pd.DataFrame, estimator=None, seed=None):
         self.data = data
+        self.estimator = estimator
+        self.seed = seed
```
ankurankan, line 91: Check if this is passed to estimators.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -45,8 +47,16 @@ class PillaiTrace(_BaseCITest):
...
+    estimator : estimator instance, optional
+        Any sklearn-compatible estimator with ``fit``, ``predict``, and ``predict_proba``
+        (if testing discrete variables) methods. If ``None`` (default), uses XGBoost
+        (``XGBClassifier`` for categorical targets, ``XGBRegressor`` for continuous).
+        Categorical Z columns are one-hot encoded when a custom estimator is used, since
+        most sklearn estimators do not handle them natively.
+
     seed : int, optional
-        Random seed used for the underlying XGBoost models.
+        Random seed for the default XGBoost estimator. Ignored when a custom
```
ankurankan: XGBoost -> estimator

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -45,8 +47,16 @@ class PillaiTrace(_BaseCITest):
     data : pandas.DataFrame
         The dataset in which to test the independence condition.
 
+    estimator : estimator instance, optional
+        Any sklearn-compatible estimator with ``fit``, ``predict``, and ``predict_proba``
+        (if testing discrete variables) methods. If ``None`` (default), uses XGBoost
```
ankurankan: Update docs. Not using XGBoost anymore.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -11,7 +13,7 @@ class PillaiTrace(_BaseCITest):
     Pillai's trace test for conditional independence with mixed data [1].
 
     This test first residualizes :math:`X` and :math:`Y` with respect to :math:`[1, Z]` using an estimator (XGBoost by
```
ankurankan: Update here as well.

ankurankan (changes requested):

pgmpy/tests/test_ci_tests/test_pillai_trace.py
```diff
@@ -180,6 +182,43 @@ def test_pillai_dependent(self):
             test("X", "Y", ["Z1", "Z2", "Z3"])
             computed_coefs.append(test.statistic_)
             computed_pvalues.append(test.p_value_)
-
         self.assertTrue(np.all(np.array(computed_coefs) >= 0.1))
         self.assertTrue(np.all(np.array(computed_pvalues) <= 0.05))
+
+    def test_pillai_custom_estimators(self):
+        """Test PillaiTrace with custom sklearn estimators on continuous and categorical data."""
+        from sklearn.ensemble import GradientBoostingRegressor, RandomForestClassifier
```
ankurankan: Please put imports on top

pgmpy/tests/test_ci_tests/test_pillai_trace.py
```diff
@@ -180,6 +182,43 @@ def test_pillai_dependent(self):
...
+        clf_estimator = RandomForestClassifier(n_estimators=50, random_state=42)
+
+        test_cat = PillaiTrace(data=self.df_indep_cat_cat, estimator=clf_estimator)
+        result_cat = test_cat("X", "Y", ["Z1", "Z2", "Z3"])
+        self.assertIsInstance(result_cat, (bool, np.bool_))
+        self.assertTrue(test_cat.p_value_ > 0.05, msg=f"Expected independence (high p-value), got {test_cat.p_value_}")
+
+    def test_pillai_custom_estimator_no_predict_proba(self):
+        """Test that ValueError is raised when estimator lacks predict_proba for discrete variable."""
+        from sklearn.linear_model import LinearRegression
```
ankurankan: Same here.

.test_durations
```diff
@@ -3134,4 +3134,4 @@
     "pgmpy/tests/test_utils/test_utils.py::TestGetExampleModel::test_model_categorization": 0.47183115999996517,
     "pgmpy/tests/test_utils/test_utils.py::TestPairwiseOrientation::test_llm": 0.000252161999924283,
     "pgmpy/tests/test_utils/test_utils.py::TestPreprocessData::test_preprocess_data": 0.006515682999861383
-}
\ No newline at end of file
+}
```
ankurankan, line 3137: Check changes before commiting.

ankurankan (changes requested):

pgmpy/ci_tests/pillai_trace.py
hanara2112: currently when a custom estimator is provided, the code does something like this : model= clone(self.estimator) and doesnt pass seed. for deafult estimator path seed is used. 

this is a usual fix for this : 
model = clone(self.estimator)
if self.seed is not None and hasattr(model, "get_params") and "random_state" in model.get_params():
    model.set_params(random_state=self.seed)

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -45,8 +50,15 @@ class PillaiTrace(_BaseCITest):
     data : pandas.DataFrame
         The dataset in which to test the independence condition.
 
+    estimator : estimator instance, optional
+        Any sklearn-compatible estimator with ``fit``, ``predict``, and ``predict_proba``
+        (if testing discrete variables) methods. If ``None`` (default), uses
+        ``RandomForestClassifier`` for categorical targets and ``RandomForestRegressor``
+        for continuous targets. Conditioning variables are one-hot encoded before fitting.
```
ankurankan, line 57: Conditioning variables are one-hot encoded only for categorical variables.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -45,8 +50,15 @@ class PillaiTrace(_BaseCITest):
...
+
     seed : int, optional
-        Random seed used for the underlying XGBoost models.
+        Random seed for the default estimator. Ignored when a custom ``estimator`` is
+        provided.
```
ankurankan, line 61: On second thought, we should remove this. I don't think there is any randomness in the test itself. And if there is randomness in the estimator, it should be provided in the estimator constructor.

pgmpy/ci_tests/pillai_trace.py
```diff
@@ -129,43 +128,62 @@ def run_test(
...
-        # Step 3: Compute the residuals
+        # Step 4: Compute the residuals.
         def get_residuals(col_name, pred, cat_index):
-            if data.loc[:, col_name].dtype == "category":
+            if self.dtypes[col_name] in ("C", "O"):
                 dummies = pd.get_dummies(data.loc[:, col_name]).loc[:, cat_index.categories[cat_index.codes]]
-                # Drop last column to avoid multicollinearity
                 return (dummies - pred).iloc[:, :-1]
             else:
                 return data.loc[:, col_name] - pred
```
ankurankan, line 165: This method should be outside the `run_test` method.

pgmpy/tests/test_ci_tests/test_pillai_trace.py
```diff
@@ -136,10 +133,7 @@ def test_pillai_no_cond(self):
         self.assertTrue(np.all(np.array(computed_coefs) >= 0.1))
```
ankurankan: If values are stable, change the GitHub actions skipping tests to match exact values.

ankurankan (changes requested):

ankurankan:
@hanara2112 Sorry, but I will need to take over this PR. I need to get some results using this method urgently.

hanara2112:
hi @ankurankan , i am not sure why there are changes against the refactored CITests file. Should i remove the file from git history?

ankurankan:
@hanara2112 Thanks, it will be easier for me to take care of that. I also think we can leave that as we are anyway transitioning to using safe imports instead of try except for imports.

ankurankan (approved):

hanara2112:
> @hanara2112 Thanks, it will be easier for me to take care of that. I also think we can leave that as we are anyway transitioning to using safe imports instead of try except for imports.

Got it. Thanks!

PR #3073, opened 2026-03-20
Fix copy method in base graph classes to use inplace=False

### LINKED ISSUE #2875: [BUG] Fix copy method in base graph classes

qianchongyang:
## Problem

After fixing the `with_role` and `without_role` methods to correctly return `None` when `inplace=True`, the `copy()` methods in base graph classes still use the old behavior with `inplace=True`, causing incorrect return values.

## Solution

Updated all `copy()` methods in base graph classes to use `inplace=False` when calling `with_role()`, following the fix pattern in PR #2862.

## Validation

```python
# copy() now returns the correct modified graph
graph = BaseGraph()
copied = graph.copy()  # Returns graph, not None
assert copied is not None
```

Fixes #2875

### DIFF
--- a/pgmpy/base/AncestralBase.py
+++ b/pgmpy/base/AncestralBase.py
@@ -730,6 +730,6 @@ class AncestralBase(nx.Graph, _GraphRolesMixin):
         )
 
         for role, vars in self.get_role_dict().items():
-            ancestral_base.with_role(role=role, variables=vars, inplace=True)
+            ancestral_base.with_role(role=role, variables=vars, inplace=False)
 
         return ancestral_base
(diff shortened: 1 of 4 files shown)


ankurankan:
@qianchongyang Please do not remove the checklist from the PR. Can you add the checklist back? And can you comment on the issue for assignment?

PR #3086, opened 2026-03-20
[ENH] Allow DAG.get_random to generate graphs with a fixed number of …

### LINKED ISSUE #3004: [ENH] Allow DAG.get_random() to generate graphs with a fixed number of edges
For implementing negative control metrics in pgmpy, we need to generate random DAGs with an exact number of edges per simulation run. The required edge count varies each time, but `DAG.get_random() `currently uses `edge_prob`, so the number of edges is unpredictable. This makes it unsuitable for this use case.

Add an optional `n_edges` parameter to `DAG.get_random()`. 

- List all valid forward edges, all pairs `(i→j)` where` i < j`, same upper-triangle logic the current method already uses. This gives `n×(n-1)/2` total possible edges.
- Randomly shuffle the node ordering first before listing edges. Without this, lower-indexed nodes are always the ones that can never be pointed to, which biases the output.
- Sample exactly `n_edges` from that pool using `numpy.random.choice` without replacement.

When `n_edges=None` (default), everything works exactly as before. 

**Additional context**
This is needed for implementing NegativeControlMetric ([issue #2414)](https://github.com/pgmpy/pgmpy/issues/2414). The negative control pipeline requires generating DAGs with edge counts sampled from previous runs, so DAG.get_random(n_edges=k) needs to be called repeatedly with different values of k.

kajal-jotwani:

Yes, I used it for the helping me with writing tests and validating my implementation and code changes.

Yes. I just added an optional `n_edges` parameter to `DAG.get_random()` so it can generate DAGs with an exact number of edges. 

```
(venv) kajal@kajal-HP-Pavilion-Laptop-15-eh2xxx:~/pgmpy$ python
Python 3.12.3 (main, Mar  3 2026, 12:15:18) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
Ctrl click to launch VS Code Native REPL
>>> import networkx as nx
>>> 
>>> for seed in [1, 2, 3, 4, 5]:
...     d = DAG.get_random(n_nodes=5, edge_prob=0.5, seed=seed)
...     print(f'seed={seed}: {len(d.edges())} edges')
... 
seed=1: 4 edges
seed=2: 4 edges
seed=3: 5 edges
seed=4: 8 edges
seed=5: 5 edges
>>> print()

>>> for seed in [1, 2, 3, 4, 5]:
...     d = DAG.get_random(n_nodes=5, n_edges=4, seed=seed)
...     print(f'seed={seed}: {len(d.edges())} edges -> {list(d.edges())}')
... 
seed=1: 4 edges -> [('X_4', 'X_2'), ('X_4', 'X_3'), ('X_0', 'X_2'), ('X_1', 'X_2')]
seed=2: 4 edges -> [('X_2', 'X_0'), ('X_2', 'X_4'), ('X_4', 'X_3'), ('X_4', 'X_1')]
seed=3: 4 edges -> [('X_2', 'X_0'), ('X_2', 'X_3'), ('X_1', 'X_3'), ('X_4', 'X_1')]
seed=4: 4 edges -> [('X_0', 'X_1'), ('X_4', 'X_0'), ('X_4', 'X_3'), ('X_2', 'X_4')]
seed=5: 4 edges -> [('X_3', 'X_1'), ('X_3', 'X_2'), ('X_4', 'X_0'), ('X_4', 'X_2')]
>>> print()
```
 
No, there are no try-except blocks in my code.

yes, but I have reviewed them and the number of tests are small as well

- Fixes #3004

- Added `n_edges` parameter to `DAG.get_random()`
- Implemented fixed-edge DAG generation by sampling from all valid forward edges (i < j)
- Shuffled node ordering to avoid bias in edge direction
- Updated tests for testing the behaviour


Spinachboul:
@kajal-jotwani 
Could you resolve the reviews made by copilot?

kajal-jotwani:
> @kajal-jotwani Could you resolve the reviews made by copilot?

@Spinachboul I’ve addressed the relevant comments. Thanks!

pgmpy/base/DAG.py
```diff
@@ -1319,6 +1320,9 @@ def get_random(
         Returns a randomly generated DAG with `n_nodes` number of nodes with
         edge probability being `edge_prob`.
```
ankurankan: This needs to be modified too.

ankurankan: I would suggest keeping this line as just that it generates random graphs.

In the line below, specify the different behaviors based on the arguments.

pgmpy/base/DAG.py
```diff
@@ -1328,9 +1332,12 @@ def get_random(
             The probability of edge between any two nodes in the topologically
             sorted DAG.
```
ankurankan: Need to specify that this will be ignored if `n_edges` is specified.

ankurankan (changes requested):
I think the implementation can be simplified quite a bit. We don't need such a big if-else condition. The only difference between using `n_edges` vs `edge_prob` is the generation of the adjacency matrix. The rest should be the same. I also think there should be a simpler way to generate the adjacency matrix for the `n_edges` case.

pgmpy/base/DAG.py
```diff
@@ -1311,6 +1311,7 @@ def to_daft(
     def get_random(
         n_nodes=5,
         edge_prob=0.5,
+        n_edges: int | None = None,
```
ankurankan, line 1313: If `n_edges` has priority over `edge_prob` then I think it should be above `edge_prob`.

kajal-jotwani:
> I think the implementation can be simplified quite a bit. We don't need such a big if-else condition. The only difference between using `n_edges` vs `edge_prob` is the generation of the adjacency matrix. The rest should be the same. I also think there should be a simpler way to generate the adjacency matrix for the `n_edges` case.

I have tried to simplify the implementation so instead of building a list of edge tuples, now the only difference as you mentioned is how the adj matrix is generated for `n_edges` we shuffle the upper triangle indices and place 1s at the first `n_edges` positions in a zero matrix, and the rest of the code is same. I have also addressed the other comments. Thanks!!

ankurankan (changes requested):
We will need to modify the signature of random methods in `DiscreteBayesianNetwork` and `LinearGaussianBayesianNetwork` too.

pgmpy/base/DAG.py
```diff
@@ -1353,19 +1360,40 @@ def get_random(
...
 
+        if len(node_names) != n_nodes:
+            raise ValueError(f"Length of node_names ({len(node_names)}) must be equal to n_nodes ({n_nodes}).")
+        if n_edges is not None:
+            max_edges = n_nodes * (n_nodes - 1) // 2
+            if n_edges < 0 or n_edges > max_edges:
+                raise ValueError(f"Invalid n_edges={n_edges}. For n_nodes={n_nodes}, valid range is [0, {max_edges}].")
+
+            # Shuffle upper triangle indices to avoid structural bias
+            upper_idx = np.array([(i, j) for i in range(n_nodes) for j in range(i + 1, n_nodes)])
```
ankurankan: I think `np.tril_indices` would generate something like this and should be faster. We should try to avoid writing `for` loops as much as possible; they are slower than vectorized operations.

pgmpy/base/DAG.py
```diff
@@ -1353,19 +1360,40 @@ def get_random(
...
+            gen.shuffle(upper_idx)
+
+            adj_mat = np.zeros((n_nodes, n_nodes), dtype=int)
+            for i, j in upper_idx[:n_edges]:
+                adj_mat[i, j] = 1
```
ankurankan: Same here.

ankurankan: I think a faster method to implement this could be something like this:

```python
    lower_i, lower_j = np.tril_indices(n_nodes, k=-1)

    # Randomly choose exactly m positions
    chosen = gen.choice(n_nodes, size=n_edges, replace=False)

    A = np.zeros((n_nodes, n_nodes), dtype=int)
    A[lower_i[chosen], lower_j[chosen]] = 1
```

pgmpy/base/DAG.py
```diff
@@ -1310,27 +1310,34 @@ def to_daft(
     @staticmethod
     def get_random(
         n_nodes=5,
+        n_edges: int | None = None,
         edge_prob=0.5,
```
ankurankan: I am a bit unsure of the design here:
1. I find it a bit confusing to have a default value for `edge_prob`, especially given that it gets ignored if `n_edges` is passed.
2. I don't think we can set a default for `n_edges` instead, as I think it would be counterintuitive to many people.
3. The other idea could not have a default for either, but then the user can't run the method without specifying one of the additional arguments.
4. I do think we should throw a warning if both `n_edges` and `edge_prob` are set to make it clear that one of them is going to be ignored.

pgmpy/base/DAG.py
kajal-jotwani: I’ll keep this in mind and try to avoid loops going forward Thanks!! 
I have used `np.triu_indices` instead of `np.tril_indices` because later we apply np.triu(adj_mat, k=1) to extract the upper triangle

pgmpy/base/DAG.py
kajal-jotwani: I thought about adding the warning but I'm unsure about the condition. If we keep `edge_prob=0.5` in the signature and use:
```
if n_edges is not None and edge_prob is not None:
    warnings.warn(...)
```
the warning would fire every time `n_edges` is set, even if the user never explicitly passed `edge_prob`, since `edge_prob` always has the default value 0.5 and is never None. Alternatively:
```
 if n_edges is not None:
if edge_prob != 0.5:
                warnings.warn(
```
misses the case where someone explicitly passes `edge_prob=0.5` alongside `n_edges`  the warning wouldn't fire even though both were set.
>The other idea could not have a default for either, but then the user can't run the method without specifying one of the additional arguments.


what if we change the signature to edge_prob=None and only apply the 0.5 default inside the else branch?
```
if n_edges is not None:
    ...
else:
    if edge_prob is None:
        edge_prob = 0.5
    adj_mat = gen.choice(...)

```

kajal-jotwani:
@ankurankan could you please take a look at this whenever you get a chance. Thanks!!

pgmpy/base/DAG.py
ankurankan: Make `edge_prob=None`.  Make the behavior like this:

```python
if (n_edges is None) and (edge_prob is None):
     edge_prob = 0.5 
```

And add a logger info that we are using the default value for edge_prob.

If both `n_edges` and `edge_prob` are defined by the user, let's throw an error.

pgmpy/base/DAG.py
```diff
@@ -1310,27 +1310,34 @@ def to_daft(
...
         seed: int | None = None,
     ) -> DAG:
         """
-        Returns a randomly generated DAG with `n_nodes` number of nodes with
-        edge probability being `edge_prob`.
+        Returns a randomly generated DAG with `n_nodes` number of nodes.
+
+        If `n_edges` is specified, generates a DAG with exactly that many edges
+        and `edge_prob` is ignored.
+        If `n_edges` is None, edges are added randomly with probability `edge_prob`.
```
ankurankan, line 1323: Add a section on what errors are thrown. So, the case when both `edge_prob` and `n_edges` are defined.

ankurankan (changes requested):

pgmpy/base/DAG.py
```diff
@@ -1353,19 +1367,46 @@ def get_random(
...
+            max_edges = n_nodes * (n_nodes - 1) // 2
+            if n_edges < 0 or n_edges > max_edges:
+                raise ValueError(f"Invalid n_edges={n_edges}. For n_nodes={n_nodes}, valid range is [0, {max_edges}].")
+
+            upper_i, upper_j = np.triu_indices(n_nodes, k=1)
+            chosen = gen.choice(len(upper_i), size=n_edges, replace=False)
+
+            adj_mat = np.zeros((n_nodes, n_nodes), dtype=int)
+            adj_mat[upper_i[chosen], upper_j[chosen]] = 1
+
```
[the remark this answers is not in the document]
ankurankan: @kajal-jotwani I think this is a valid issue.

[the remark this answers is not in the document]
kajal-jotwani: Done! Added a random node permutation before placing edges to fix the label-order bias.

ankurankan (dismissed):
Thanks for the PR. I have added node name shuffling as well.

kajal-jotwani:
@ankurankan the CI was failing because `gen.shuffle(node_names)` was mutating the list in-place fixed it by shuffling a copy instead. Could you please re-approve it.  Thanks!!

ankurankan (approved):
Thanks @kajal-jotwani

PR #3105, opened 2026-03-21
[ENH] Fix seed handling in DiscreteBayesianNetwork.get_random_cpds

### LINKED ISSUE #2829: [ENH] Add a `get_random_cpds` method to `LinearGaussianBayesianNetwork`
`DiscreteBayesianNetwork` has two methods for creating random models:

1. `get_random`: Creates a random model graph and adds random parameters (TabularCPDs) to it.
2. `get_random_cpds`: Given a graphical structure, adds random parameters (TabularCPDs) to it.

However, `LinearGaussianBayesianNetwork` currently only has the `get_random` method. We should add a `get_random_cpds` method as well with a similar behavior as `DiscreteBN`.

codervinitjangir:
…work (#2829)

I use AI tool like coding partner for this PR. First I check the DiscreteBayesianNetwork class myself and find problem: get_random_cpds was using same seed for all nodes, and get_random was repeating CPD logic. I tell AI with clear steps: change the for loop with enumerate, make current_seed = seed + i, and replace duplicate loop in get_random with call to self.get_random_cpds().
Later one notebook test fail because latent nodes not found in dag.edges(). I read the traceback , find the fix, and implement it.
At the end, I also use AI to quickly translate some Hindi comments into clean English so codebase look professional.

Yes, I fully understand the changes. The issue was that DiscreteBayesianNetwork.get_random_cpds was passing the exact same seed to every node. Because of this, if two nodes had the exact same structure (same number of states and parents), they generated identical CPD probabilities. I fixed this by using enumerate on the nodes and passing seed + i so each node gets a unique seed. I also refactored get_random so it simply calls self.get_random_cpds() instead of duplicating the CPD generation loop.

I verified the fix by a specific unit test (test_get_random_cpds_seed) where created a network with two independent nodes having the same structure. Before the fix, passing a fixed seed made their CPDs identical. After the fix, I assert that their generated probabilities are completely different. I also ran the full test_DiscreteBayesianNetwork.py suite locally to ensure no backward compatibility was broken and all edge cases (like seed=None) still work perfectly.

No, there are no try-except blocks added anywhere.

Yes, I instructed the LLM to draft the test_get_random_cpds_seed test based on my exact test design. I reviewed the generated test to ensure it is concise, strictly targets the seed duplication bug, and does not add unnecessary bloat to the test suite.

- Fixes #2829

- Modified get_random_cpds in DiscreteBayesianNetwork to increment the seed (seed + i) iteratively for each node to prevent identical CPD generation.

- Refactored get_random in DiscreteBayesianNetwork to reuse get_random_cpds instead of duplicating the probability generation loop (DRY principle).

- Added a new unit test test_get_random_cpds_seed to explicitly verify the seed fix.


codervinitjangir:
@ankurankan Please review.

ankurankan:
@codervinitjangir Update your checklist here as well to exactly describe how you have used LLMs.

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1280,9 +1272,17 @@ def get_random_cpds(
             n_states = {var: gen.integers(low=1, high=5, size=1)[0] for var in self.nodes()}
 
         cpds = []
-        for node in self.nodes():
+        for i, node in enumerate(self.nodes()):
             parents = list(self.predecessors(node))
-            cpds.append(TabularCPD.get_random(variable=node, evidence=parents, cardinality=n_states, seed=seed))
+            current_seed = (seed + i) if seed is not None else None
```
ankurankan: I don't think we should take this approach for generating the seed values for random cpds. Have a look at how this is done in LinearGaussianBayesianNetwork. We use a random number selector for the seed value as well.

ankurankan (changes requested):

pgmpy/models/DiscreteBayesianNetwork.py
codervinitjangir: Thanks for the feedback, @ankurankan! I updated the code to use np.random.default_rng(seed) and now each node gets its own seed with rng.integers(). I also made the cardinality and state_names logic cleaner by using dictionary unpacking. Let me know if you’d like me to change anything else.

PR #3145, opened 2026-03-22
Fix recursion in copy() caused by with_role(inplace=False)

### LINKED ISSUE #3142: [BUG] Fix recursion in copy() caused by with_role(inplace=False)
**Describe the bug**
There is a recursion problem between copy() and with_role(inplace=False).

copy() uses with_role(inplace=False) to assign roles
with_role(inplace=False) internally calls copy()

This creates a loop:
```copy() → with_role() → copy() → ...```

which leads to:
```RecursionError: maximum recursion depth exceeded```

**To Reproduce**

```python
from pgmpy.base._base import _CoreGraph

# Create a graph with roles
edges = [("A", "B", "->"), ("B", "C", "->")]

graph = _CoreGraph(
    ebunch=edges,
    exposures=["A"],   # this creates roles internally
)

# This triggers the bug
graph_copy = graph.copy()
```
**Expected behavior**
copy() should return a valid copy of the graph
Roles (exposures, outcomes, latents, custom roles) should be preserved
No recursion should occur

**Fix**
Instead of calling copy() inside with_role(inplace=False),
we manually create a new graph.

**Problematic Part of code is in**
```python
if not inplace:
            new_graph = self.copy()
else:
        new_graph = self
```
line 88-91 in pgmpy/base/_mixin_roles.py

Issue related to #2853 #2862 #2875

anusa-saha:

Yes, to find the replacement of `.copy()` 

Yes

Ran all the tests under `tests/test_base/...`

N/A

N/A

- Fixes #3142 

- Replaced shallow copying with `.deepcopy()` to ensure node and edge attributes (especially mutable ones like sets) are not shared, preventing unintended side effects when modifying roles.
- Standardized role assignment operations to use `inplace=False`, ensuring methods return a new graph instance without mutating the original graph, improving safety and consistency.

### DIFF
--- a/pgmpy/base/AncestralBase.py
+++ b/pgmpy/base/AncestralBase.py
@@ -730,6 +730,6 @@ class AncestralBase(nx.Graph, _GraphRolesMixin):
         )
 
         for role, vars in self.get_role_dict().items():
-            ancestral_base.with_role(role=role, variables=vars, inplace=True)
+            ancestral_base = ancestral_base.with_role(role=role, variables=vars, inplace=False)
 
         return ancestral_base
--- a/pgmpy/base/DAG.py
+++ b/pgmpy/base/DAG.py
@@ -1558,7 +1558,7 @@ class DAG(_GraphRolesMixin, nx.DiGraph):
         dag.add_nodes_from(self.nodes())
 
         for role, vars in self.get_role_dict().items():
-            dag.with_role(role=role, variables=vars, inplace=True)
+            dag = dag.with_role(role=role, variables=vars, inplace=False)
 
         return dag
 
--- a/pgmpy/base/PDAG.py
+++ b/pgmpy/base/PDAG.py
@@ -195,7 +195,7 @@ class PDAG(_GraphRolesMixin, nx.DiGraph):
         pdag.add_nodes_from(self.nodes())
 
         for role, vars in self.get_role_dict().items():
-            pdag.with_role(role=role, variables=vars, inplace=True)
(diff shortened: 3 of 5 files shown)


anusa-saha:
@ankurankan @DARHWOLF fixed the infinite loop

anusa-saha:
@ankurankan the notebook tests are failing

codervinitjangir:
Hi @anusa-saha, thanks for putting up the PR!

I looked at the changes, and I am really worried about using copy.deepcopy(). Deepcopy is notoriously slow and uses a massive amount of memory, especially for large NetworkX graphs (like DAGs or BayesianNetworks). If a user has a large model, this change will make their copy() operations incredibly slow and might even crash their memory.

More importantly, we don't actually need this fix because the bug doesn't exist in the main code. The recursion loop only happens if we force the code to use inplace=False.

The current dev branch uses inplace=True inside the copy() methods. This perfectly and safely updates the newly cloned graph without touching the original one, and it returns it instantly without any recursion loops.

cc: @ankurankan @DARHWOLF — The current code in dev is already fast, safe, and works exactly as intended. I strongly suggest we do not merge this deepcopy approach, as it will introduce a severe performance drop for users.

anusa-saha:
@codervinitjangir but keeping self.copy() also isnt making sense 
I'll think about something else but this is what i could think about right now
maybe i'll come up with something else in the next week

DARHWOLF:
I do not think this is the correct fix. I'm not sure you understand the issue here.

Issue #3106, opened 2026-03-21
[ENH] Implement KCI and KCIT kernel-based conditional independence tests


MAVRICK-1:
Part of umbrella issue #3099.

  ### Description
  Implement the Kernel-based Conditional Independence Test (KCI) and Kernel Conditional
  Independence Test (KCIT) as described in:

  > Zhang, Peters, Janzing, Schölkopf (2011). Kernel-based Conditional Independence Test
  > and Application in Causal Discovery.
  > https://arxiv.org/abs/1202.3775

  KCI/KCIT use kernel embeddings of distributions to test X ⊥ Y | Z for non-linear,
  non-Gaussian relationships — scenarios where pearsonr and chi_square fail.

  ### Reference implementation
  - causal-learn: https://causal-learn.readthedocs.io/en/latest/independence_tests_index/kci.html

hanara2112:
hi @MAVRICK-1 , i am interested in implementing KCI and KCIT. I am currently reading the paper and will update you once i have something concrete. Can you meanwhile assign this issue to me?

Issue #3099, opened 2026-03-21
[ENH] List of Conditional Independence Tests


ankurankan:
This is an umbrella issue to keep a track of Conditional Independence (CI) tests that we would like to implement. To work on any of them please create a separate issue for it.

| CI Test | Reference | Other Implementations | Issue | PR |
|-----------|---------------|---------------------------------|--------|------|
| Fisher's Z | | | | #3092 |
| KCI and KCIT | 1 | a | #3106 | #3166 |
| RCIT and RCoT | 2 | | #3107  | |
| Missing value Fisher's Z | 3, 4 | | #3108  | |
| Empirical null wrapper | | b | | |
| Regression based Likelihood-ratio | 5 | c | #3132 | |
| DFCIT | 6 | | | |
| Projected Distance Covariance | 7 | | #3115 | | 
|Local Permutation Test | 8 | | |

[1] Zhang, Peters, Janzing, Schölkopf (2011), Kernel-based Conditional Independence Test and Application in Causal Discovery
[2] Strobl, Zhang, Visweswaran (2019), Approximate Kernel-Based Conditional Independence Tests for Fast Non-Parametric Causal Discovery
[3] Tu et al. (2019), Causal Discovery in the Presence of Missing Data
[4] Witte et al. (2022), Multiple imputation and test-wise deletion for causal discovery with incomplete cohort data
[5] Tsagris, Michail et al. “Constraint-based causal discovery with mixed data.” International journal of data science and analytics vol. 6,1 (2018): 19-30. doi:10.1007/s41060-018-0097-y
[6] https://jmlr.org/papers/v23/20-682.html
[7] Fan, Jianqing, Yang Feng, and Lucy Xia. "A projection-based conditional dependence measure with applications to high-dimensional undirected graphical models." Journal of Econometrics 218.1 (2020): 119-139.
[8] Kim, Ilmun, et al. "Local permutation tests for conditional independence." The Annals of Statistics 50.6 (2022): 3388-3414.

Implementation Links:
[a] https://causal-learn.readthedocs.io/en/latest/independence_tests_index/kci.html
[b] https://www.bnlearn.com/documentation/man/conditional.independence.tests.html
[c] https://search.r-project.org/CRAN/refmans/MXM/html/MXMCondIndTests.html

hanara2112:
Hi @ankurankan,
I would love to take up the implementation for Kernel-based Conditional Independence Test (Zhang et al., 2011).
[...]

PR #3166, opened 2026-03-23
[ENH] Add Kernel-based Conditional Independence (KCI) test (#3106)            

hanara2112:

- Did you use an LLM for any assistance with this PR? Please describe in detail how and what you used it for?                                        
  - Yes, mainly for scaffolding code and tests. I verified the logic myself. The bandwidth part I figured out by reading causal-learn since the LLM got it wrong.
  
  - Yes. It uses RBF kernels + HSIC for independence. For conditional, Z is regressed out before computing the statistic. Main fix was correcting bandwidth logic for HSIC vs KCI.                                                              

- What steps have you taken to verify that the changes correctly address the 
  issue? And what edge cases have you considered? Other than running tests,   
  what else have you verified?        
  - Matched stats and p-values with causal-learn across a few scenarios. Also checked small samples and edge cases.                                      

  - No
- Have you used LLM for generating tests? They need to be compressed into a 
  smaller number of tests without reducing coverage.         
  - Initially yes, but I simplified and reduced them.               

                                                                              
Issue number(s) that this pull request fixes
  - Solves #3106 issue part of the umbrella issue #3099 

- List of changes to the codebase in this pull request                        
   - pgmpy/ci_tests/kci.py : new addition
   - pgmpy/tests/test_ci_tests/test_kci.py: new addition


pgmpy/tests/test_ci_tests/test_kci.py
```diff
@@ -0,0 +1,103 @@
+import unittest
```
Jatinbhardwaj-093: Use pytest

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,338 @@
...
+        The dataset in which to test the independence condition.
+
+    kernel_X : str, default="rbf"
+        Kernel for variable X. Currently only ``"rbf"`` is supported.
+
+    kernel_Y : str, default="rbf"
+        Kernel for variable Y. Currently only ``"rbf"`` is supported.
+
+    kernel_Z : str, default="rbf"
+        Kernel for conditioning variables Z. Currently only ``"rbf"`` is supported.
```
ankurankan: These should be sklearn Kernel instances. This will allow the users to have finer control over the kernels and also be able to choose any kernel they want. We can keep the default as sklearn's RBF kernel.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,338 @@
...
+
+    width_X : float or None, default=None
+        Bandwidth (sigma) for the X kernel. If ``None``, estimated from the data
+        using a sample-size-based heuristic.
+
+    width_Y : float or None, default=None
+        Bandwidth (sigma) for the Y kernel. If ``None``, estimated from the data.
+
+    width_Z : float or None, default=None
+        Bandwidth (sigma) for the Z kernel. If ``None``, estimated from the data.
```
ankurankan: These would get removed after the above change.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,338 @@
...
+
+    regularization : float, default=1e-3
+        Regularization parameter :math:`\epsilon` used in the conditional test to
+        invert the kernel matrix on :math:`Z`.
+
```
ankurankan: Is this a standard parameter of the test? This feels something just for numerical stability?

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,338 @@
...
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import pandas as pd
+    >>> from pgmpy.ci_tests import KCI
+    >>> rng = np.random.default_rng(seed=42)
+    >>> data = pd.DataFrame(rng.standard_normal((300, 3)), columns=["X", "Y", "Z"])
+    >>> test = KCI(data=data)
+    >>> test("X", "Y", [], significance_level=0.05)
+    True
```
ankurankan, line 111: Show the attributes as well.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,338 @@
...
+        self.kernel_Y = kernel_Y
+        self.kernel_Z = kernel_Z
+        self.width_X = width_X
+        self.width_Y = width_Y
+        self.width_Z = width_Z
+        self.regularization = regularization
+        super().__init__()
+
+    @staticmethod
+    def _rbf_kernel(X, width):
```
ankurankan: Can be removed after accepting sklearn kernels

pgmpy/estimators/CITests.py
```diff
@@ -939,3 +939,65 @@ def pearsonr_equivalence(X, Y, Z, data, boolean=True, delta_threshold=0.1, **kwa
         return p_value < kwargs.get("significance_level", 0.05)
     else:
         return coeff, p_value
+
+
+@ci_registry.register(name="kci", data_types=["continuous"])
```
ankurankan: No need to add here. This will be deprecated in the next couple of releases.

pgmpy/tests/test_ci_tests/test_kci.py
ankurankan: And add tests for comparing values with other packages.

ankurankan (changes requested):
Had a design level review and have added some comments.

hanara2112:
Hi @ankurankan, I went through the KCI implementation again to make sure I properly understand it before pushing more changes. Writing down my current understanding + a few questions, please correct me if anything is off.

Currently the implementation splits into two cases depending on whether Z is empty or not.

When Z is empty (HSIC case):

1. Normalize X and Y
2. Compute RBF kernels
3. Center the kernels
4. Compute the statistic using element-wise product of centered kernels
5. Approximate p-value using gamma fit (based on traces / Frobenius norms)

When Z is present (conditional case):

1. Compute kernels for X, Y, Z
2. For X, use augmented input [X, 0.5 * Z]
3. Center all kernels
4. Residualize using something like ε * pinv(Kz + εI)
5. Compute statistic similarly
6. Estimate moments via eigendecomposition

This might be a bit rough for the conditional case, but this is how I’m currently reading the flow.

hanara2112:
Things i am unsure about and would need discussion:

1. Bandwidth handling: when user passes a kernel _estimate_bandwidth only runs when kernel=None.
If a user passes RBF() directly, it uses default length_scale=1.0, which might not match the median heuristic on normalized data.
Would it be better to: just document this clearly, or expose _estimate_bandwidth so users can compute it before passing a kernel?

2. Augmenting X with 0.5 * Z
This follows the paper / causal-learn, but when a custom kernel_X is passed, the input changes from X → [X, Z].
Should this just be documented, or is there a cleaner way to handle this?

3. Complexity concerns
Since this uses pseudoinverse / eigendecomposition, it looks like ~O(n³) time and O(n²) memory.
Do you think it’s worth adding a warning/check for large n, or is that out of scope for now?

pgmpy/tests/test_ci_tests/test_kci.py
hanara2112: other CI tests use unit test. can you please clarify?

pgmpy/ci_tests/kci.py
hanara2112: yes, it's for numerical stability. it's the ε in ε(Kz + εI)⁻¹. 
The causal-learn implementation hardcodes it to 1e-3 as well. I've removed it as a user-facing parameter in the latest refactor. It's now just an internal constant in _conditional_test.

pgmpy/tests/test_ci_tests/test_kci.py
hanara2112: Added TestKCICompareCausalLearn that compares pgmpy's results against causal-learn's KCI on 4 scenarios       
(unconditional independent/dependent, conditional independent/dependent). Both implementations agree on all cases.

hanara2112:
I revisited the implementation and realized my conditional projection step was incorrect. I wasn’t using the proper residual-forming operator R_z = I - K_z (K_z + \epsilon I)^{-1}.
I’m updating this along with correcting HSIC scaling and reviewing the gamma approximation to better match the original paper.

pgmpy/tests/test_ci_tests/test_kci.py
```diff
@@ -0,0 +1,138 @@
...
+    def test_get_ci_test_factory(self):
+        test = get_ci_test(test="kci", data=self.df_ind)
+        self.assertIsInstance(test, KCI)
+
+
+@unittest.skipUnless(
+    __import__("importlib").util.find_spec("causallearn"),
+    "causallearn not installed",
+)
+class TestKCICompareCausalLearn(unittest.TestCase):
```
ankurankan: We should not just look at whether values are greater than or less than 0.05. We need to compare exactly the statistic and p-values computed by us vs what causallearn's implementation returns. That will help identify any subtle differences in the implementation. Also, please don't compute the values from causallearn in the tests, as this will add a dependency for our package:

- You should run the causallearn tests on your local machine and copy the outputted values in the tests here.
- Add the causal learn code used to generate the values as comment in the code, so that we can reproduce it later if required.

pgmpy/tests/test_ci_tests/test_kci.py
```diff
@@ -0,0 +1,138 @@
+import unittest
+
+import numpy as np
+import pandas as pd
+from sklearn.gaussian_process.kernels import RBF
+
+from pgmpy.ci_tests import KCI, get_ci_test
+
+
+class TestKCI(unittest.TestCase):
```
ankurankan: Tests need to be compressed. These look LLM generated.

ankurankan:
> Would it be better to: just document this clearly, or expose _estimate_bandwidth so users can compute it before passing a kernel?

This can be a boolean argument in the constructor. If yes, we can use sklearn's `set_params` method to update the Kernal constructors. This behavior should be documented clearly.

 > Augmenting X with 0.5 * Z
This follows the paper / causal-learn, but when a custom kernel_X is passed, the input changes from X → [X, Z].
Should this just be documented, or is there a cleaner way to handle this?

Can you point me to where exactly is this in the paper? 

> Since this uses pseudoinverse / eigendecomposition, it looks like ~O(n³) time and O(n²) memory.
Do you think it’s worth adding a warning/check for large n, or is that out of scope for now?

This is fine. It's a very well known issue of KCI; RCIT improves on this.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,297 @@
...
+
+
+class KCI(_BaseCITest):
+    r"""
+    Kernel-based Conditional Independence (KCI) test [1].
+
+    When :math:`Z = \emptyset`, this reduces to the HSIC (Hilbert-Schmidt Independence
+    Criterion) test for unconditional independence. When :math:`Z \neq \emptyset`, it
+    performs a kernel-based conditional independence test (KCIT) by regressing out the
+    effect of :math:`Z` from the kernel matrices of :math:`X` and :math:`Y`.
```
ankurankan: Try to add the exact equations of what we are computing in the test. Refer to other tests for example.

ankurankan:
@hanara2112 Did you use an LLM to write the PR checklist?

pgmpy/ci_tests/kci.py
hanara2112: should this be included in the parameter docstring?

ankurankan: It can stay as an argument but its purpose muct be documented properly.

hanara2112:
Hi @ankurankan, while reading the paper and doing some finals check for the implementation i noticed that in Section 3.5 (page 6 from the KCI paper) , the authors suggest to use half the kernel width for K_z compared to what's used for Kx and Ky. 

In our current implementation and in causal-learn, all 3 kernels share the same bandwidth from empirical_width_kci. 
Should we keep it this way to stay consistent with causal-learn, or should we follow the paper and use ls / 2 for the Z kernel?

hanara2112:
> > Would it be better to: just document this clearly, or expose _estimate_bandwidth so users can compute it before passing a kernel?

> This can be a boolean argument in the constructor. If yes, we can use sklearn's `set_params` method to update the Kernal constructors. This behavior should be documented clearly.

Right now, if a custom kernel_X is passed, it’s used as is, ie. no bandwidth estimation is applied. Empirical bandwidth is only used when the kernel is None.

We could introduce an auto_bandwidth flag (default True) with this behavior:

  1. auto_bandwidth=True: estimate bandwidth via _empirical_width_kci(Z) and override length_scale using set_params (for compatible kernels like RBF).
  2. auto_bandwidth=False (or no Z): use the kernel exactly as provided.

Although one important thing to notice here is that this only works for kernels with a length_scale parameter. For custom kernels, we can either:

  - restrict support to such kernels, or
  - use a try/except and document the requirement.


> > Augmenting X with 0.5 * Z
> > This follows the paper / causal-learn, but when a custom kernel_X is passed, the input changes from X → [X, Z].
> > Should this just be documented, or is there a cleaner way to handle this?
> 
> Can you point me to where exactly is this in the paper?

The 0.5Z augmentation isn't actually from the paper but from the casual-learn codebase. The paper defines X" = (X,Z) with a product kernel i.e. two seperate kernels multiplied element-wise. The concatenation of X with 0.5Z usinga single RBF is a casual-learn design choice for easier reference value comparison.

Since the paper uses separate kernels for X and Z, a cleaner approach matching the paper would be to compute kernel_X(X) * kernel_Z(Z) element-wise for Kx" rather than feeding [X, 0.5*Z] into kernel_X. That way a custom kernel_X would operate on X alone, as the user would expect. Should I refactor it this way, or keep it consistent with causal-learn's concatenation approach ? but that would probably change the output values.

hanara2112:
> @hanara2112 Did you use an LLM to write the PR checklist?

that was a intial checklist while starting to code. i will change it and write it by myself now since the kci implementation seems close to completion after a review. 
Apologies for any inconvienience.

pgmpy/tests/test_ci_tests/test_kci.py
hanara2112: Done. TestKCICompareCausalLearn now directly checks stats and p-values against hardcoded causal-learn reference values (generated locally from v0.1.4.5), so there’s no external dependency. The docstring aslo includes the full reproduction code. After fixing the bandwidth issue, pgmpy matches causal-learn exactly across the test cases.

hanara2112:
Hi @ankurankan , this PR is ready for review. Can you please let me know if any changes required ?

hanara2112:
Working on this!

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

hanara2112:
superseded by https://github.com/pgmpy/pgmpy/pull/3297.

PR #3191, opened 2026-03-26
fix: use global X/Y states for contingency tables in power_divergence

### LINKED ISSUE #2886: [BUG] Incorrect contingency table computation in `power_divergence` test

thakoreh:
Fixes #2886

## Problem

When computing conditional independence tests (power_divergence, chi_square, g_sq, log_likelihood), the contingency table for each Z stratum was built using np.unique on only that stratum's data. If a state of X or Y didn't appear in a particular stratum, it was silently dropped from the table.

This produced two issues:
1. Contingency tables had wrong dimensions across strata
2. The zero-row/column check failed to fire because the missing state was invisible

## Example

With data where X=2 only appears when Z=0:
- Old: Z=1 stratum computed a 2x2 table (missing X=2), dof=3
- New: Z=1 stratum computes a 3x2 table with a zero row, which is correctly skipped, dof=2

## Fix

Compute unique_x and unique_y once from the full dataset before the loop, then use np.searchsorted to index into them per stratum.

Applied to both:
- pgmpy/ci_tests/power_divergence.py (new API)
- pgmpy/estimators/CITests.py (deprecated API)

## Tests

- Added regression test test_contingency_table_all_states_conditioned verifying correct dof
- Updated two expected values in test_chisquare_adult_dataset for conditioned tests that now correctly skip strata with zero-count rows

### DIFF
--- a/pgmpy/ci_tests/power_divergence.py
+++ b/pgmpy/ci_tests/power_divergence.py
@@ -136,10 +136,12 @@ class PowerDivergence(_BaseCITest):
         else:
             chi = 0
             dof = 0
+            unique_x = np.unique(data[X])
+            unique_y = np.unique(data[Y])
             for z_state, df in data.groupby(list(Z), observed=True):
-                # Compute the contingency table
-                unique_x, x_inv = np.unique(df[X], return_inverse=True)
-                unique_y, y_inv = np.unique(df[Y], return_inverse=True)
+                # Compute the contingency table using global states
+                x_inv = np.searchsorted(unique_x, df[X])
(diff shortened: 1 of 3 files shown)


ankurankan:
1. Please add the PR checklist back
2. Tests are failing.

ankurankan:
Closing. Issue wasn't assigned to you, and no activity.

PR #3201, opened 2026-03-27
Fix dtype handling in FunctionalCPD.sample for torch backend

### LINKED ISSUE #3154: [DOC] Fixing the doctest errors in pgmpy/factors/hybrid/FunctionalCPD.py
This issue is related to #2832 
```
**********************************************************************
File "D:\pgmpy\pgmpy\pgmpy/factors/hybrid\FunctionalCPD.py", line 95, in FunctionalCPD.FunctionalCPD.sample
Failed example:
    cpd.sample(2, parent_samples)
Exception raised:
      File "C:\Python313\Lib\doctest.py", line 1398, in __run
        exec(compile(example.source, filename, "single",
        ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                     compileflags, True), test.globs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "<doctest FunctionalCPD.FunctionalCPD.sample[6]>", line 1, in <module>
        cpd.sample(2, parent_samples)
        ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
      File "D:\pgmpy\pgmpy\pgmpy\factors\hybrid\FunctionalCPD.py", line 118, in sample
        p: torch.as_tensor(row[p], dtype=config.get_dtype(), device=config.get_device())
           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    TypeError: as_tensor(): argument 'dtype' must be torch.dtype, not str
**********************************************************************
1 item had failures:
   1 of   7 in FunctionalCPD.FunctionalCPD.sample
***Test Failed*** 1 failure.
```

codejasleen:

1 .To explore possible approaches (local fix vs global fix)
2. To reason about backend-dependent dtype handling in the config system
3. To understand why small floating-point differences appeared and how to handle them in doctests using ellipsis

1. I reproduced the issues locally, traced the source of dtype mismatch in code.
2. Implemented the fix for dtype conversion.
3. Validated the solution by running doctests and verifying correct behavior

I added a minimal targeted test to validate the new behavior of Config.set_dtype() for the torch backend (string to torch.dtype conversion). The test is concise and focused on the newly introduced logic, and does not duplicate existing coverage.

- Fixes #3154 

-Fixed dtype handling in Config.set_dtype() by converting string dtype to torch.dtype when using torch backend.
-Updated doctest in FunctionalCPD.sample to explicitly set backend to torch
-Added a new test in test_global_vars.py to validate dtype conversion behavior for torch backend


pgmpy/factors/hybrid/FunctionalCPD.py
```diff
@@ -114,9 +114,12 @@ def sample(self, n_samples=100, parent_sample=None):
             else:
                 for i in range(n_samples):
                     row = parent_sample.iloc[i]
+
+                    dtype = config.get_dtype()
+                    dtype = getattr(torch, dtype) if isinstance(dtype, str) else dtype
+
```
ankurankan: Why is this required?

codejasleen: `config.get_dtype `return a string but `torch.as_tensor` expects a torch dtype object and passing the string directly leads to a TypeError. 
so, the `isinstance` check is to handle cases where dtype is already a torch type object, so we only convert when it's a string.

ankurankan: This change should then go to the `Config.get_dtype`, not here.

codejasleen: One concern I had was that NumPy usages of `config.get_dtype()` expect a string, so changing it globally to always return a `torch.dtype` could break those cases.

Based on the usages across the codebase, it seems `get_dtype()` is expected to be backend dependent:

* NumPy paths expect a string dtype
* Torch paths expect a `torch.dtype`

So I plan to update `Config.get_dtype()` to return a `torch.dtype` only when the backend is set to torch, and keep returning the string dtype for numpy. This way we fix the root issue without affecting existing numpy behavior.

NSKY0808: I noticed that `Config.set_backend("torch")` ensures that `get_dtype() `returns a torch `dtype` instead of a string.

This suggests that the issue might arise when `FunctionalCPD` is used without explicitly setting the backend to torch, leading to `get_dtype()` returning a numpy-style `dtype (string)`, which is incompatible with `torch.as_tensor`.

Since FunctionalCPD relies on torch and pyro, would it make sense to either:

- Enforce that the backend is set to torch when using this class, or
- Handle dtype conversion locally as a safeguard in case the backend is not set correctly?

codejasleen: That’s a good observation. Initially, I also considered handling this either by enforcing the torch backend or converting dtype locally in `FunctionalCPD`.

However, after looking into the implementation, the issue actually stems from `Config.set_dtype()` allowing string dtypes to be stored even when the backend is torch. This leads to inconsistencies, since some torch APIs (like `torch.as_tensor`) require a `torch.dtype`.

So instead of enforcing backend changes or adding local safeguards, I fixed the issue at the source by ensuring that `set_dtype()` converts string dtypes to `torch.dtype` when the backend is torch.

With this change, `config.get_dtype()` always returns the correct type for the active backend, and we avoid duplicating logic or introducing hidden behavior in `FunctionalCPD`.

ankurankan: @codejasleen Can you create a separate issue for this showing reproducible code?

NSKY0808: @codejasleen that makes sense.. converting string dtypes in `set_dtype()` ensures consistency when using the torch backend.

I was also thinking that if an invalid dtype string is passed (for example somethingg like "`float128`"), the conversion using getattr(torch, dtype) would Still fail, but later as an AttributeError.

Would it make sense to validate the dtype at this stage as well, so that such cases raise a clearer `ValueError `early instead of failing at runtime?

codejasleen: @ankurankan 
To address this, I moved the fix to the source by updating `Config.set_dtype()` so that when the backend is torch, any string dtype is converted to the corresponding `torch.dtype`. This ensures that `config.get_dtype()` always returns a backend compatible type and avoids having to handle dtype conversion at individual usage sites like FunctionalCPD.

After fixing this, the doctest in `FunctionalCPD.sample` was still failing because it was using torch without explicitly setting the backend. Since pgmpy defaults to `numpy, get_dtype()` was returning a string in that context. To resolve this, I updated the doctest to explicitly set the backend to torch before using torch based functionality.

Finally, I added a test in `test_global_vars.py` to verify that passing a string dtype with the torch backend correctly results in a `torch.dtype`. The test is conditionally skipped when torch is not available to ensure compatibility with CI environments without optional dependencies.

Please once review this.

codejasleen: > @codejasleen Can you create a separate issue for this showing reproducible code?

I’m a bit confused here,I thought this was already being handled as a standalone issue through this PR/discussion.

codejasleen:
Hii @ankurankan 
Should i create new issue for this PR?

PR #3212, opened 2026-03-28
Refractors TreeSearch Algorithm to be sklearn compatible 

### LINKED ISSUE #2575: [ENH] Refactor `TreeSearch` algorithm
Refactor the TreeSearch algorithm to be sklearn compatible. Refer parent issue for details.

### LINKED ISSUE #3298: [BUG] Fix TAN root selection to exclude class node during automatic selection
In the TAN causal discovery algorithm where, during automatic root selection, the algorithm could incorrectly choose the class_node as the root. This happens because the current implementation computes weights over all variables, including the class node, and selects the one with the highest score. In some datasets, the class node naturally has the highest mutual information, which leads to a conflict (root_node == class_node) and raises a ValueError, even though the user did not provide a root. 

Finding the fix to this 
Will need to read the paper in depth to understand why??

cc: @ankurankan

anusa-saha:

I used both Claude and ChatGPT as learning and support tools while working on this task. I first went through the existing code of TreeSearch myself and tried to understand how to refractor using the examples of the `ExpertInLoop`, `ExpertKnowledge` and `GES` refractoring and what is needed to be changed to make it sklearn compatible. I refactored the TreeSearch class to make it follow a sklearn-style design instead of the old pgmpy estimator pattern. Earlier, the class stored the data inside `__init__` and used an estimate() method that directly returned the graph. I changed this by removing data from `__init__` and keeping only hyperparameters there, and moved the learning logic into a `_fit(X)` method that takes the data as input. Instead of returning the graph, the model now stores results as attributes like causal_graph_ and adjacency_matrix_. I also made sure not to modify input parameters like root_node during fitting by using a local variable, so the behavior stays consistent across multiple runs. Along with this, I updated the tests and examples to use .fit() and access results through attributes, and adjusted static method usage to be called directly from the class. I did not change implementation of the algorithm rather just copy pasted it into the new file. Also added the warning message in `pgmpy/estimators/TreeSearch.py`

Kept consistency in the structure by comparing it with other refractored estimators. Also copy pasted the logic from `pgmpy/estimators/TreeSearch.py` so there is minimum errors from my side which can be introduced by an AI. Also added the doctests needed and validated them.  

Didnt use AI to generate any tests, the tests were already present in `tests/test_estimators/test_TreeSearch.py`. Earlier, the tests were creating the model with data in the constructor and calling estimate() to get the result. I changed this to first initialize the model with parameters, then call .fit(data), and access the learned graph using est.causal_graph_.

- Fixes #2575
- Fixes #3298 

- Refractored TreeSearch Algorithm into sklearn compatible 
- Also refractored the tests and added them
- Added the warning in `pgmpy/estimators/TreeSearch.py`
- Added in doctest_modules.txt


ankurankan (changes requested):
@anusa-saha Thanks for the PR. I would suggest splitting Chow Liu and TAN into two separate algorithms. While both are related methods, they are separate algorithms. I think splitting them would also make the class constructors a bit cleaner (relevant to each of the algorithms). If you see any shared functionality between these two, you can create a `_TreeSearchMixin` class that both can inherit and use the shared functionality.

anusa-saha:
@ankurankan okay I'll look into this

anusa-saha:
@ankurankan I think `from pgmpy.inference import CausalInference` is resolving to the `pgmpy.inference.CausalInference` module instead of the `CausalInference class`, causing TypeError: 'module' object is not callable at runtime.

ankurankan (approved):

PR #3224, opened 2026-03-28
fix(CPD): raise ValueError when normalize encounters zero-sum columns

### LINKED ISSUE #3213: [BUG] TabularCPD.normalize produces NaN values for zero-sum columns
**Describe the bug**
While working with TabularCPD , I noticed that calling normalize() on a CPD with a zero-sum column results in NaN values instead of being handled explicitly. Internally normalization divides by the column sum. If a column sums to zero, this leads to division by zero and produces Nans. A RuntimeWarning is raised, but might be easy to miss or suppress, and the resulting CPD becomes invalid.

Since TabularCPD is meant to represent probability distributions,  NaN values can (silently) break downstream inference or sampling.

**To Reproduce**
```
import numpy as np
cpd = TabularCPD(variable="A", variable_card=2, values=np.array([[0],[0]]))
cpd.normalize()
print(cpd.values)
```
gives us [nan nan]
also happens in cases where a valid CPD becomes zero during intermediate computations, like:
```
cpd = TabularCPD(variable="A", variable_card=2, values=np.array([[1],[0]]))
cpd.values[0] = 0
cpd.values[1] = 0
cpd.normalize()
print(cpd.values)
```

**Expected behavior**
Zero-sum columns should be handled explicitly instead of producing NaNs, it should either-
- use a fallback
- raise the error

**Additional context**
Currently, the method seems to rely on numpy division without checking for zero column sums. Since warnings can be suppressed, this can lead to silent corruption of CPDs.

BlocksecPHD:
## Summary

Fixes #3213

**Problem:** When calling `normalize()` on a `TabularCPD` with a column that sums to zero, the method would silently divide by zero and produce NaN values, making the CPD invalid for downstream inference and sampling.

**Solution:** Added a check in `normalize()` to detect zero-sum columns before division. Now raises `ValueError` with an informative message listing the affected column indices.

## Changes

- `pgmpy/factors/discrete/CPD.py`: Added zero-sum column check in `normalize()` method
- `pgmpy/tests/test_factors/test_discrete/test_Factor.py`: Added tests for the new behavior

## Before

```python
>>> import numpy as np
>>> from pgmpy.factors.discrete import TabularCPD
>>> cpd = TabularCPD(variable='A', variable_card=2, values=np.array([[0],[0]]))
>>> cpd.normalize()
>>> print(cpd.values)
[nan nan]  # Invalid CPD!
```

## After

```python
>>> cpd = TabularCPD(variable='A', variable_card=2, values=np.array([[0],[0]]))
>>> cpd.normalize()
ValueError: Cannot normalize CPD: column(s) [0] sum to zero
```

## Testing

- Added `test_normalize_zero_sum_column_raises_error`: Verifies ValueError is raised
- Added `test_normalize_zero_sum_column_informative_error_message`: Verifies error message includes column indices

### DIFF
--- a/pgmpy/factors/discrete/CPD.py
+++ b/pgmpy/factors/discrete/CPD.py
@@ -421,6 +421,11 @@ class TabularCPD(DiscreteFactor):
             If inplace=True it will modify the CPD itself, else would return
             a new CPD
 
+        Raises
+        ------
+        ValueError
+            If any column sums to zero (normalization would produce NaN values)
+
         Examples
         --------
         >>> from pgmpy.factors.discrete import TabularCPD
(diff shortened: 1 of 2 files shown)


ankurankan:
No checklist; incorrect implementation. Closing.

PR #3240, opened 2026-03-30
Fix: Resolved incorrect state handling in BayesianModel (Fixes #3239)

### LINKED ISSUE #3239: Fix Pandas DataFrame concatenation FutureWarning in Dynamic_BN.ipynb
In one of the cells, the code iteratively appends data to an empty dataframe using pd.concat in a loop. 
In modern versions of Pandas, concatenating on empty is deprecated. As a result, this loop  triggers the  following warning again and again

Image

this is making the output unnecessarily jumbled up and the code not future proof

i can simply refactored the code to collect all the separated samples_t dataframes into a standard Python list first, and then run pd.concat()

now the output is much better 

Image

Zaki-Haider1:
This pull request addresses a warning emitted by Pandas in the Dynamic_BN.ipynbnotebook regarding the concatenation of empty DataFrames.

This pull request fixes the issue #3239 

the changes made are very simple, i used AI to identify why the issue was occuring.

- changed the data processing loop in Dynamic_BN.ipynb to collect samples_t into a standard Python list samples_list first.

- Replaced the iterative concatination calls single pd.concat at the end.


Nimish-4:
Ref: https://github.com/pgmpy/pgmpy/issues/3239

PR #3297, opened 2026-04-04
HSIC and KCI CI tests

### LINKED ISSUE #3254: [ENH] Add HSIC as a standalone CI test
1. **Is your feature request related to a problem? Please describe.**
   - there is not a dedicated implementation of HSIC available as a standalone conditional independence test in pgmpy yet.

2. **Describe the solution you'd like**

  - I want to implement HSIC as a standalone function in CI Tests suite. It should correctly measure the marginal dependence between 2 continous variables. HSIC should be exposed as a test object compatible with _BaseCITest, similar to the other CI tests in the module. Given variables X and Y, the test should first standardize both variables, construct kernel matrices for each variable using Gaussian RBF kernels, and then center those kernel matrices. Use a V-statistic test to sum over all the pairs of data points; including self-pairs. the final trace of the kernel metrices (similarity maps of X and Y) are centerd and used to estimate dependence between variables. 
  - We use a V-statistic as it follow natural flow of hsic implementation rather than using an unbiased estimator for hsic. It measures the dependence in RKHS and can capture both linear and non-linear relationships.
  - bandwidth selection
  - null distribution and the p-value
  - The test should store the observed statistic and p-value as attributes and return whether independence is accepted or rejected.
  - Since HSIC is only a marginal independence test, it should reject any non-empty conditioning set Z and direct users to use KCI for conditional independence testing.
  - custom kernel .. default to RBF

Concretely, this would involve:

  - adding a new module - pgmpy/ci_tests/hsic.py,
  - implementing an HSIC class that inherits from _BaseCITest,
  - adding a new module - tests/test_ci_tests/test_hsic.py .

3. **Describe alternatives you've considered**
  I have studied the causal-learn's implementation and observed how they emphasize certain constraints on the actual methodology from the work of Gretton et al. Also i took help from a matlab implementation of the dependence measure. Then i have finally decided the methodology in good faith of pgmpy's code convention. 

4. **Additional context**
   - I have refactored the HSIC and KCI class and now they exist as standalone classes to be used by users. I implemented KCI (conditional + unconditional test) - from where i seperate out the uncond. test as the HSIC class which is essentially measuring dependence between x and y without any effect of the confounding variable z and other wise for KCI algorithm.

References:
  - A Kernel Statistical Test of Independence
  - Gretton et al., “Measuring Statistical Dependence with Hilbert-Schmidt Norms”, ALT 2005
  - https://github.com/py-why/causal-learn

hanara2112:

  - I used AI to add permutation based p-value method to the hsic class. I asked it to read the internal source code of casual-learn(specifically the KCI_UInd and GaussianKernel.py) to figure out exactly why the medium bandwidths were diverging. The AI helped identify that causal-learn uses sqrt(2) x median(euclidean_dist) and for understanding git workflow and commit messages.

  - Verified that the centering matrix formulation (O(n²) trick) and the Gamma approximation moments (V-statistic vs U-statistic) matched both causal-learn and the Gretton 2007 / Zhang 2011 papers.
  - Confirmed that the empirical and median bandwidth heuristics exactly match causal-learn's local implementation mathematically.
  - Ensured HSIC formally rejects conditional inputs (raising a ValueError if Z is not empty), preventing users from accidentally using it for KCI. Handled degenerate cases where the calculated variance drops below zero.

  - Compact tests cover all HSIC configs and validate correctness against causal-learn benchmarks without extra dependencies.

- Fixes #3254

  pgmpy/ci_tests/hsic.py: Added standalone HSIC class inheriting from _BaseCITest.
  pgmpy/ci_tests/kci.py: Refactored KCI to inherit from HSIC, removing duplicated core logic while keeping KCI-specific conditional residualization.
  pgmpy/ci_tests/__init__.py: Exposed HSIC in the public API.
  pgmpy/tests/test_ci_tests/test_hsic.py: Added compressed test suite including causal-learn parity checks.


hanara2112:
This PR depends on #3166 and is based on feature/issue-3099-kci to keep the diff minimal.

Changes:
- Added HSIC CI test implementation
- Integrated with existing CI framework

hanara2112:
The tests appear to have failed due to a joblib KeyError in TestBayesianNetworkFitPredict.test_predict. As this PR only modifies ci_tests and this error does not reproduce locally, I believe this is a flaky GitHub Actions runner issue with the ExactInference multiprocessor. @ankurankan could you please re-trigger the CI checks?

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,263 @@
...
+        Measuring Statistical Dependence with Hilbert-Schmidt Norms.
+        ALT 2005. https://doi.org/10.1007/11564089_7
+    .. [2] Gretton, A., Fukumizu, K., Teo, C. H., Song, L., Schölkopf, B., & Smola, A. J. (2007).
+        A Kernel Statistical Test of Independence.
+        NeurIPS 2007. https://papers.nips.cc/paper/3201
+    .. [3] Zhang, K., Peters, J., Janzing, D., & Schölkopf, B. (2011).
+        Kernel-based Conditional Independence Test and Application in Causal Discovery.
+        UAI 2011. https://arxiv.org/abs/1202.3775
+    .. [4] Gretton, A., Borgwardt, K. M., Rasch, M. J., Schölkopf, B., & Smola, A. (2012).
+        A Kernel Two-Sample Test. JMLR, 13, 723–773.
```
ankurankan: References need to be precise, 1 or 2 papers max. This should be the paper that our implementation matches.

hanara2112:
Hi @ankurankan , I have a small clarification regarding this PR - would you consider merging the kernels for X and Y into a single K applied to both? In most cases , the same kernel is used anyway , and this would simplify the API while avoiding ambiguity around missing kernel types. since bandwidth is still comptued per varible internally, we would'nty loose flexibilty , just wanted to check your preference before finalizing.

I was thinking something like kernel: `Kernel | str | None = None ` (defaulting to RBF), applied to both. if someone really needs different kernels, they could subclass. Also, minor thought: n_permutations could be moved to run_test since it’s more of a runtime parameter, but I’m okay keeping it in the constructor for consistency.

hanara2112:
adding the final report and example notebook for reference...
[hsic_final_report.pdf](https://github.com/user-attachments/files/26639460/hsic_final_report.pdf)
[HSIC_Genomics.ipynb](https://github.com/user-attachments/files/26639461/HSIC_Genomics.ipynb)

hanara2112:
The remaining uncovered lines are defensive guards in the gamma p-value computation. 
With valid kernel matrices these conditions shouldn’t be triggered, so they’re effectively unreachable in normal usage.

Would you prefer adding `# pragma: no cover` to these lines, or leaving them uncovered as-is?

hanara2112:
hi @ankurankan as suggested , i have verfied the implementation locally beyond unit tests to confirm the statistics actually are meaningful.
1. I ran the test n times on independent gaussian data at alpha = 0.05. The false-positive rate was 4.5%, consistent with the nominal 5%. This confimed me that the gamma null approximation is correctly caliberated.
2. I compared the gamma and permutation p-values on 30 independent datasets. | delta Mean | = 0.021 and max = 0.052. this approximation closely tracks the model-free permutation null. 
3. Also the test correctly rejects for liner dependence and non linear dependence and pearson misses the quadratic case entirely (r = -0.11), which is the key motivation for a kernel based test. in a signal to noise sweep , the statistic and p-value transition smoothly from rejection to non-rejection as noise increases.
4. lastly to cross validate the reuslts , both the statistic and the p-value mathc the causal-learn's KCI_IInd to within abs=0.01.

pgmpy/ci_tests/hsic.py
ankurankan, line 1: We follow 120 char line length. Can change that to improve readability.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+from scipy import stats
+from scipy.spatial.distance import pdist
+from sklearn.gaussian_process.kernels import RBF, Kernel
+
+from ._base import _BaseCITest
+
+
+class HSIC(_BaseCITest):
+    r"""
+    Hilbert-Schmidt Independence Criterion (HSIC) test for marginal independence [1][2].
```
ankurankan, line 12: References need to be precise.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+    >>> from pgmpy.ci_tests import KCI
+    >>> rng = np.random.default_rng(seed=42)
+    >>> data = pd.DataFrame(rng.standard_normal((300, 3)), columns=["X", "Y", "Z"])
+    >>> test = KCI(data=data)
+    >>> test("X", "Y", ["Z"], significance_level=0.05)
+    True
+    >>> test.statistic_  # doctest: +SKIP
+    2.66...
+    >>> test.p_value_  # doctest: +SKIP
+    0.55...
```
DARHWOLF, line 73: would be better to not skip doctests here

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+
+    HSIC detects nonlinear dependencies by measuring the distance between the
+    joint distribution and the product of marginals in a reproducing kernel
+    Hilbert space (RKHS). The V-statistic is:
+
+    .. math::
+        T = \operatorname{Tr}(\tilde{K}_X \tilde{K}_Y),
+        \quad \tilde{K} = HKH, \quad H = I - \tfrac{1}{n}\mathbf{1}\mathbf{1}^T.
+
```
ankurankan, line 21: Please add what the null distribution is or how the p-value is computed from the statistic

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+        kernel_Z: Kernel | None = None,
+        bandwidth: str = "empirical",
+        epsilon: float = 1e-3,
+    ):
+        super().__init__(data=data, kernel_X=kernel_X, kernel_Y=kernel_Y, bandwidth=bandwidth)
+        self.kernel_Z = kernel_Z
+        self.epsilon = epsilon
+
+    def _empirical_width_kci(self, Z: np.ndarray) -> float:
+        """Piecewise RBF length-scale for the conditional path [2]."""
```
DARHWOLF: would be nice to explicitly mention the heuristic you're implementing from this paper

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+        "name": "hsic",
+        "data_types": ("continuous",),
+        "default_for": None,
+        "requires_data": True,
+    }
+
+    def __init__(
+        self,
+        data: pd.DataFrame,
+        kernel_X: Kernel | None = None,
```
ankurankan, line 75: Can make this a tuple of (kernel_X, kernel_Y). If a single object is passed, it gets used for both the kernels.

ankurankan:
```python
test = HSIC(data=data, kernel=(RFB(...), RBF(...))

test = HSIC(data=data, kernel=RFB(...))

test = HSIC(data=data, kernel=None)
```

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+        T = \operatorname{Tr}(\tilde{K}_X \tilde{K}_Y),
+        \quad \tilde{K} = HKH, \quad H = I - \tfrac{1}{n}\mathbf{1}\mathbf{1}^T.
+
+    Parameters
+    ----------
+    data : pandas.DataFrame
+        Dataset containing the variables to test.
+    kernel_X, kernel_Y : sklearn.gaussian_process.kernels.Kernel or None
+        Kernels for X and Y. Default: RBF with bandwidth heuristic.
+    bandwidth : {"empirical", "median"}, default="empirical"
```
ankurankan, line 28: empirical -> heuristic

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+        Independence. NeurIPS 2007.
+    .. [3] Zhang et al. (2011). Kernel-based conditional independence
+        test and application in causal discovery. UAI 2011.
+
+    Examples
+    --------
+    >>> import numpy as np, pandas as pd
+    >>> from pgmpy.ci_tests import HSIC
+    >>> rng = np.random.default_rng(seed=42)
+    >>> data = pd.DataFrame(rng.standard_normal((300, 3)), columns=["X", "Y", "Z"])
```
ankurankan, line 59: Use `load_dataset('tubingen/<dataset_no>")`.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+    >>> test = HSIC(data=data)
+    >>> test("X", "Y", [], significance_level=0.05)
```
ankurankan, line 61: Remove `[]`.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+    True
+    """
+
+    _tags = {
+        "name": "hsic",
+        "data_types": ("continuous",),
+        "default_for": None,
```
ankurankan, line 68: Should add a new tag for non-coniditional test. Please create an issue.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+        return test_stat, _gamma_pvalue_from_moments(test_stat, mean, var)
+
+    def run_test(self, X: str, Y: str, Z: list):
+        """
+        Compute KCI statistic and p-value.
+
+        Sets ``self.statistic_`` and ``self.p_value_``.
+        """
+        if len(Z) == 0:
+            return super().run_test(X, Y, Z)
```
ankurankan, line 170: Doesn't super require the arguments to be passed as well?

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+    Extends :class:`HSIC` to the conditional case :math:`X \perp\!\!\!\perp Y \mid Z`.
+    When ``Z`` is empty the test falls back to HSIC. The conditional statistic is
+
+    .. math::
+        T_{CI} = \frac{1}{n}\operatorname{Tr}
+            \bigl(\tilde{K}_{\ddot{X}|Z}\tilde{K}_{Y|Z}\bigr),
+
+    where :math:`\tilde{K}_{\cdot|Z} = R_Z \tilde{K}_\cdot R_Z` and
+    :math:`R_Z = \varepsilon (K_Z + \varepsilon I)^{-1}`. A Gamma approximation
+    is used for the null distribution.
```
ankurankan, line 32: Specify the form of the gamma approximation used for p-value.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+
+    Parameters
+    ----------
+    data : pandas.DataFrame
+        Dataset containing variables X, Y, and optionally Z.
+    kernel_X, kernel_Y, kernel_Z : sklearn.gaussian_process.kernels.Kernel or None
```
ankurankan, line 38: Can use the tuple form as before.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+        Kernels for each variable. Default: RBF with bandwidth heuristic.
+    bandwidth : {"empirical", "median"}, default="empirical"
```
ankurankan, line 40: Same.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+        vx = vx @ np.diag(np.sqrt(wx))
+        vy = vy @ np.diag(np.sqrt(wy))
+
+        n = KxR.shape[0]
+        num_eigx, num_eigy = vx.shape[1], vy.shape[1]
+        size_u = num_eigx * num_eigy
+        uu = np.zeros((n, size_u))
+        for i in range(num_eigx):
+            for j in range(num_eigy):
+                uu[:, i * num_eigy + j] = vx[:, i] * vy[:, j]
```
ankurankan, line 134: This needs to be vectorized.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+        wx, vx = np.linalg.eigh(0.5 * (KxR + KxR.T))
+        wy, vy = np.linalg.eigh(0.5 * (KyR + KyR.T))
+
+        wx, vx = wx[np.argsort(-wx)], vx[:, np.argsort(-wx)]
+        wy, vy = wy[np.argsort(-wy)], vy[:, np.argsort(-wy)]
+
+        vx = vx[:, wx > np.max(wx) * thresh]
+        wx = wx[wx > np.max(wx) * thresh]
+        vy = vy[:, wy > np.max(wy) * thresh]
+        wy = wy[wy > np.max(wy) * thresh]
```
ankurankan, line 123: Verify what this thing is doing?

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+
+        n = KxR.shape[0]
+        num_eigx, num_eigy = vx.shape[1], vy.shape[1]
+        size_u = num_eigx * num_eigy
+        uu = np.zeros((n, size_u))
+        for i in range(num_eigx):
+            for j in range(num_eigy):
+                uu[:, i * num_eigy + j] = vx[:, i] * vy[:, j]
+
+        return uu @ uu.T if size_u > n else uu.T @ uu
```
ankurankan, line 136: Check if `size_u > n` is required.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+    def _make_kernel(self, user_kernel, X):
+        if user_kernel is not None:
+            return user_kernel
+        return RBF(length_scale=self._length_scale(X))
+
+    def _hsic_gamma_pvalue(self, test_stat, K, L, Kc, Lc):
+        """Gamma p-value using exact null moments (Proposition 6(i) of [2])."""
+        n = K.shape[0]
+        if n < 6:
+            return 1.0
```
ankurankan, line 129: Always document heurstics like these.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # Null variance for T/n.
+        M_sq = (Kc * Lc / 6.0) ** 2
+        var_hsic = (M_sq.sum() - np.trace(M_sq)) / (n * (n - 1))
+        var_hsic *= 72.0 * (n - 4) * (n - 5) / (n * (n - 1) * (n - 2) * (n - 3))
```
ankurankan, line 134: Where is `6` and `72` coming from?

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # Null mean for T/n from off-diagonal kernel means.
+        mu_x = (K.sum() - np.trace(K)) / (n * (n - 1))
+        mu_y = (L.sum() - np.trace(L)) / (n * (n - 1))
+        mean_hsic = (1.0 + mu_x * mu_y - mu_x - mu_y) / n
+
+        if var_hsic <= 0 or mean_hsic <= 0:
+            return 1.0
```
ankurankan, line 142: Check if this ever gets executed.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,186 @@
...
+        The p-value for the test. Set after calling the test.
+
+    References
+    ----------
+    .. [1] Gretton et al. (2005). Measuring Statistical Dependence
+        with Hilbert-Schmidt Norms. ALT 2005.
+    .. [2] Gretton et al. (2007). A Kernel Statistical Test of
+        Independence. NeurIPS 2007.
+    .. [3] Zhang et al. (2011). Kernel-based conditional independence
+        test and application in causal discovery. UAI 2011.
```
ankurankan, line 52: Look into the performance of these methods and select the one that works best if there is any variation

pgmpy/tests/test_ci_tests/test_hsic.py
```diff
@@ -0,0 +1,119 @@
...
+
+    return df_ind, df_dep, df_nonlinear
+
+
+class TestHSIC:
+    def test_independent(self, hsic_data):
+        df_ind, _, _ = hsic_data
+        test = HSIC(data=df_ind)
+        assert test("X", "Y", [], significance_level=0.05)
+        assert test.p_value_ > 0.05
```
ankurankan, line 31: Test exact values.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,181 @@
...
+    def _conditional_test(self, x: np.ndarray, y: np.ndarray, z: np.ndarray):
+        n = x.shape[0]
+        ls = self._get_length_scale_kci(z)
+
+        kernel_x = self.kernel_X if self.kernel_X is not None else RBF(length_scale=ls)
+        kernel_y = self.kernel_Y if self.kernel_Y is not None else RBF(length_scale=ls)
+        kernel_z = self.kernel_Z if self.kernel_Z is not None else RBF(length_scale=ls)
+
+        # Augment X with half-weighted Z, following the causal-learn reference [2].
+        Kx = self._center_kernel(kernel_x(np.hstack([x, 0.5 * z])))
```
ankurankan, line 147: Check why this is being done.

ankurankan (changes requested):

pgmpy/ci_tests/hsic.py
hanara2112: I used the 3 references for the below reasons:

1. HSIC V-statistic and the premutation testwas taken from [1].
2. Gamma null approximation from [2].
3. [3] ref was used for citing the bandwidth heuristics.

Can you please clarify which should i remove?

pgmpy/ci_tests/kci.py
hanara2112: if you are talking about super(), i think its valid and auto resolves to super(KCI, self) via __class__ cell.
so i don't think we need args there. let me know what you think!

pgmpy/ci_tests/hsic.py
hanara2112: It does get executed for degenerate inputs (e.g., constant columns). In such cases, both mean_hsic and var_hsic collapse to 0, triggering this branch. Added a test case and documented the behavior.

pgmpy/ci_tests/hsic.py
hanara2112: Benchmarked gamma vs permutation (n=100/300/1000, seed=0): p-values agree within ~0.02; permutation is 10–172× slower. Documented this in the param so users know why gamma is the default and when to switch to permutation (small n, non-RBF kernel).

pgmpy/ci_tests/kci.py
hanara2112: Concatenating X and Z lets the kernel evaluate similarity using both sources together, so any dependence between them can influence the result. The 0.5 scaling simply keeps Z from dominating this comparison, especially when it has higher dimensionality. So it contributes as context while X remains the primary signal.

pgmpy/ci_tests/kci.py
hanara2112: KxR and KyR should be PSD, but due to numerical noise we get tiny/negative eigenvalues; we drop values below thresh × max_eigenvalue to keep only meaningful signal (same as causal-learn).

pgmpy/ci_tests/kci.py
hanara2112: This is for efficiency. both matrix forms give the same trace, so we compute using the smaller one to avoid expensive large matrix multiplications.

pgmpy/ci_tests/hsic.py
hanara2112: They come from the HSIC variance formula (Gretton et al.) Appendix; 6 is from normalizing the U-statistic, and 72 = 2 × 6² is the resulting constant in the final expression.

pgmpy/ci_tests/hsic.py
ankurankan: https://github.com/pgmpy/pgmpy/issues/3347

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+           Hilbert-Schmidt Norms. ALT 2005.
+    .. [2] Gretton et al. (2007). A Kernel Statistical Test of Independence.
+           NeurIPS 2007.
+    """
+
+    _tags = {
+        "name": "hsic",
+        "data_types": ("continuous",),
+        "default_for": None,
+        "requires_data": True,
```
ankurankan, line 101: We have added a new tag: `is_symmetric`. Should add it for both these tests too.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+        if null_dist not in ("gamma", "permutation"):
+            raise ValueError(f"null_dist must be 'gamma' or 'permutation', got {null_dist!r}")
+        if n_permutations < 1:
+            raise ValueError(f"n_permutations must be >= 1, got {n_permutations!r}")
+
+        self.data = data
+        if kernel is None or isinstance(kernel, Kernel):
+            self.kernel_X = self.kernel_Y = kernel
+        else:
+            self.kernel_X, self.kernel_Y = kernel
```
ankurankan: Have all the input validation steps here.

ankurankan: This should also be renamed to: `self.kernel_X_` and `self.kernel_Y_`

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+        "requires_data": True,
+    }
+
+    def __init__(
+        self,
+        data: pd.DataFrame,
+        kernel: Kernel | tuple | None = None,
+        bandwidth: str = "heuristic",
+        null_dist: str = "gamma",
+        n_permutations: int = 500,
```
ankurankan: Let's default it to 100.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+        """Empirical p-value by permuting Y (Section 5.1 of [1])."""
+        rng = np.random.default_rng(self.random_state)
+        n = y.shape[0]
+        null_stats = np.empty(self.n_permutations)
+        for i in range(self.n_permutations):
+            Lyc = self._center_kernel(kernel_y(y[rng.permutation(n)]))
+            null_stats[i] = np.sum(Kxc * Lyc)
+        return (1 + np.sum(null_stats >= test_stat)) / (1 + self.n_permutations)
+
+    def run_test(self, X: str, Y: str, Z: list):
```
ankurankan: This needs to be `_compute_result`

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+
+        Returns (statistic, p_value) and sets self.statistic_, self.p_value_.
+        """
+        if Z:
+            raise ValueError("HSIC does not support conditioning variables. Use KCI instead.")
+
+        x = self.data[X].to_numpy(dtype=float).reshape(-1, 1)
+        y = self.data[Y].to_numpy(dtype=float).reshape(-1, 1)
+        x = np.nan_to_num(stats.zscore(x, ddof=1, axis=0))
+        y = np.nan_to_num(stats.zscore(y, ddof=1, axis=0))
```
ankurankan: Simplify this:

if X.std() != 0:
       X - X.mean() / X.std()
else:
      X = 0

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+
+        kernel_x = self.kernel_X if self.kernel_X is not None else RBF(length_scale=self._length_scale(x))
+        kernel_y = self.kernel_Y if self.kernel_Y is not None else RBF(length_scale=self._length_scale(y))
```
ankurankan: Move the preprocessing to the init method.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+        Kxc, Kyc = self._center_kernel(Kx), self._center_kernel(Ky)
+        test_stat = np.sum(Kxc * Kyc)
+
+        if self.null_dist == "permutation":
+            p_value = self._permutation_pvalue(test_stat, Kxc, y, kernel_y)
+        else:
+            p_value = self._hsic_gamma_pvalue(test_stat, Kx, Ky, Kxc, Kyc)
+
+        self.statistic_ = float(test_stat)
+        self.p_value_ = float(p_value)
```
ankurankan: Need to set an `effect_size_` attribute as well. Keep it as None for HSIC.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+
+    def _permutation_pvalue(self, test_stat, Kxc, y, kernel_y):
+        """Empirical p-value by permuting Y (Section 5.1 of [1])."""
+        rng = np.random.default_rng(self.random_state)
+        n = y.shape[0]
+        null_stats = np.empty(self.n_permutations)
+        for i in range(self.n_permutations):
+            Lyc = self._center_kernel(kernel_y(y[rng.permutation(n)]))
+            null_stats[i] = np.sum(Kxc * Lyc)
+        return (1 + np.sum(null_stats >= test_stat)) / (1 + self.n_permutations)
```
ankurankan: Move this in the main `run_test` method itself.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+        beta = var_hsic * n / mean_hsic
+        return 1.0 - stats.gamma.cdf(test_stat / n, a=alpha, scale=beta)
+
+    def _permutation_pvalue(self, test_stat, Kxc, y, kernel_y):
+        """Empirical p-value by permuting Y (Section 5.1 of [1])."""
+        rng = np.random.default_rng(self.random_state)
+        n = y.shape[0]
+        null_stats = np.empty(self.n_permutations)
+        for i in range(self.n_permutations):
+            Lyc = self._center_kernel(kernel_y(y[rng.permutation(n)]))
```
ankurankan: Do we need to center it in every loop?

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,193 @@
...
+        """Median heuristic: length-scale = sqrt(2) * median(euclidean_dists)."""
+        med = np.median(pdist(X, metric="euclidean"))
+        return np.sqrt(2.0) * med if med > 0 else 1.0
+
+    def _length_scale(self, X: np.ndarray) -> float:
+        """RBF length-scale from ``bandwidth`` (piecewise heuristic or median)."""
+        if self.bandwidth == "median":
+            return self._median_width(X)
+        n = X.shape[0]
+        width = 0.8 if n < 200 else (0.5 if n < 1200 else 0.3)
```
ankurankan, line 163: Please document it in the docstring.

pgmpy/ci_tests/hsic.py
ankurankan: Should also be in the docstring.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,173 @@
...
+    >>> round(test.p_value_, 2)
+    0.33
+
+    Attributes
+    ----------
+    statistic_ : float
+        The KCI test statistic. Set after calling the test.
+    p_value_ : float
+        The p-value for the test. Set after calling the test.
+
```
ankurankan, line 76: Need to set an `effect_size_` attribute. Set it to None for KCI.

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,173 @@
...
+           Causal Discovery. UAI 2011.
+    .. [2] causal-learn: Causal Discovery in Python.
+           https://causal-learn.readthedocs.io/en/latest/
+    """
+
+    _tags = {
+        "name": "kci",
+        "data_types": ("continuous",),
+        "default_for": None,
+        "requires_data": True,
```
ankurankan, line 109: `is_symmetric`

pgmpy/ci_tests/kci.py
```diff
@@ -0,0 +1,173 @@
...
+        uu = (vx[:, :, None] * vy[:, None, :]).reshape(n, -1)
+        return uu @ uu.T if uu.shape[1] > n else uu.T @ uu
+
+    def _conditional_test(self, x: np.ndarray, y: np.ndarray, z: np.ndarray):
+        n = x.shape[0]
+        ls = self._length_scale_kci(z)
+
+        kernel_x = self.kernel_X if self.kernel_X is not None else RBF(length_scale=ls)
+        kernel_y = self.kernel_Y if self.kernel_Y is not None else RBF(length_scale=ls)
+        kernel_z = self.kernel_Z if self.kernel_Z is not None else RBF(length_scale=ls)
```
ankurankan: Move this into preprocessing step.

pgmpy/tests/test_ci_tests/test_kci.py
```diff
@@ -0,0 +1,118 @@
...
+    """Cross-validate against causal-learn v0.1.4.5 (numpy 2.4.3).
+
+    Reproduction code for reference values::
+
+        import numpy as np
+        from causallearn.utils.KCI.KCI import KCI_CInd
+
+        rng = np.random.default_rng(seed=7)
+        Z = rng.normal(size=500)
+        X = 2*Z + rng.normal(scale=0.5, size=500)
```
ankurankan: Use values under 1 for beta.

pgmpy/tests/test_ci_tests/test_kci.py
```diff
@@ -0,0 +1,118 @@
...
+        Z = 2*X + 2*Y + rng.normal(scale=0.3, size=300)
+        KCI_CInd().compute_pvalue(X[:,None], Y[:,None], Z[:,None])  # stat=1084.59, p=0.0
+    """
+
+    def test_matches_causal_learn_kci_cind(self):
+        rng = np.random.default_rng(seed=7)
+        Z = rng.normal(size=500)
+        X = 2 * Z + rng.normal(scale=0.5, size=500)
+        Y = 3 * Z + rng.normal(scale=0.5, size=500)
+        df = pd.DataFrame({"X": X, "Y": Y, "Z": Z})
```
ankurankan, line 115: Simpler way to write this in pgmpy:
```
lgbn = DAG.from_dagitty("dag{ Z -> X [beta=2] ; Z -> Y [beta=3]}")
df = lgbn.simulate(500)
```

ankurankan (changes requested):

pgmpy/ci_tests/hsic.py
hanara2112: added kernel input validation which rejects non-kernel/non-tuple values and tuples of wrong lenght.

pgmpy/ci_tests/hsic.py
hanara2112: done

pgmpy/ci_tests/hsic.py
hanara2112: simplified to explicit z-score calculation and fallback to 0 when x_std==0. 
for kci , the same change applies whereas z uses the per-column form since z can be multi-column.

pgmpy/ci_tests/hsic.py
hanara2112: effect size set to None for both hsic and kci.

pgmpy/ci_tests/hsic.py
hanara2112: No, I fixed it by centering the kernel matrix only once before the permutation loop instead of re-centering it every iteration. Inside the loop, I now just permute the already-centered matrix using Kyc[perm][:, perm], which gives the same result much faster.

pgmpy/tests/test_ci_tests/test_kci.py
hanara2112: should i test with pther beta values like 0.9 or keep the beta 0.5/0.8 and rather bump the number of samples ?

pgmpy/tests/test_ci_tests/test_kci.py
hanara2112: Tried simulate() and smaller betas, but very low beta made KCI tests noisy/unstable and expected vals started drifting.

pgmpy/ci_tests/kci.py
hanara2112: noticed that i can't move this cleanly into _init_ because kernel depends on column choice , which isn't known until run_test and same for hsic. 
@ankurankan Do you wnat me to move it anyway or extract it into a helper function?

pgmpy/tests/test_ci_tests/test_kci.py
hanara2112: Also same changes to be applied for hsic tests ? (after discussion)

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,198 @@
...
+    or from a permutation test on Y [1]. Returns ``p_value_ = 1.0`` if
+    ``n < 6`` (variance estimator is undefined).
+
+    Parameters
+    ----------
+    data : pandas.DataFrame
+        The dataset in which to test the independence condition.
+    kernel : sklearn Kernel, tuple of two Kernels, or None
+        Kernel(s) for X and Y; a single object is shared by both. Default: RBF
+        with bandwidth heuristic.
```
ankurankan: Look up numpydoc convention for specifying object type.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,198 @@
...
+    bandwidth : {"heuristic", "median"}, default="heuristic"
```
ankurankan, line 30: this should be just `str`.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    def __init__(
+        self,
+        data: pd.DataFrame,
+        kernel: Kernel | tuple | None = None,
+        bandwidth: str = "heuristic",
+        null_dist: str = "gamma",
+        n_permutations: int = 100,
+        random_state: int | np.random.Generator | None = None,
```
ankurankan: This needs to be called seed to be consistent with the rest of the package.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,198 @@
...
+        ``"median"`` uses ``sqrt(2) * median`` of pairwise Euclidean
+        distances (falling back to ``1.0`` if the median is zero).
+    null_dist : {"gamma", "permutation"}, default="gamma"
+        Null approximation. ``"gamma"`` [2] is faster but assumes RBF-like
+        kernels; use ``"permutation"`` [1] for non-RBF kernels.
+    n_permutations : int, default=100
+        Number of permutations (used only when ``null_dist="permutation"``).
+    random_state : int, np.random.Generator, or None, default=None
+        Seed for permutation reproducibility.
+
```
ankurankan, line 44: Add a `Notes` section here and add the heuristics in the notes.

ankurankan: Make a bullet point for each heuristic used.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,198 @@
...
+        r"""
+        Compute HSIC statistic and p-value for :math:`X \perp Y`.
+
+        Z must be empty; use :class:`KCI` for conditional tests.
+        """
+        if Z:
+            raise ValueError("HSIC does not support conditioning variables; use KCI.")
+
+        x = self.data[X].to_numpy(dtype=float).reshape(-1, 1)
+        y = self.data[Y].to_numpy(dtype=float).reshape(-1, 1)
```
ankurankan: We shouldn't specify the `dtype` here. Use the one that is already used in `self.data`.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,198 @@
...
+
+        x_std, y_std = x.std(ddof=1), y.std(ddof=1)
+        x = (x - x.mean()) / x_std if x_std > 0 else np.zeros_like(x)
+        y = (y - y.mean()) / y_std if y_std > 0 else np.zeros_like(y)
```
ankurankan: Shouldn't do it like this. Check separately whether x or y are constants and throw a error.

pgmpy/ci_tests/hsic.py
```diff
@@ -0,0 +1,198 @@
...
+
+        kernel_x = self.kernel_X_ if self.kernel_X_ is not None else RBF(length_scale=self._length_scale(x))
+        kernel_y = self.kernel_Y_ if self.kernel_Y_ is not None else RBF(length_scale=self._length_scale(y))
```
ankurankan: Add this preprocessing step in the `__init__` method.

ankurankan (changes requested):

hanara2112:
@ankurankan , i have updated hsic and kci code for clarity and added correct citation support and ensured that all tests are passing. I have attached a md file for PR context and development.

[HSIC_KCI_PR_NOTES.md](https://github.com/user-attachments/files/31346972/HSIC_KCI_PR_NOTES.md)

PR #3338, opened 2026-04-19
Add tags to Casual Discovery methods

### LINKED ISSUE #3264: [WIP][ENH] List of tags for causal discovery methods
1. Supported data types (continuous, discrete, or ordinal, depends on hyperparameter)
2. Assumed relationship (linear, non-linear, depends on hyperparameter)
3. Supports expert knowledge (which attributes of ExpertKnowledge)
4. Assumed noise term. (non-gaussian, additive, etc.)

Nimish-4:
- Fixes #3264

### DIFF
--- a/pgmpy/causal_discovery/ExpertInLoop.py
+++ b/pgmpy/causal_discovery/ExpertInLoop.py
@@ -149,6 +149,17 @@ class ExpertInLoop(_BaseCausalDiscovery):
     and the GES algorithm.
     """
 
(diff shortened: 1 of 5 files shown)


ankurankan (dismissed):
Let's implement the `summary` method only in the `_BaseCausalDiscovery`. It should check what type is `self.causal_graph_` is, and depending on that compute  a set of summary statistics for that causal graph type.

PR #3381, opened 2026-05-07
Dibs causal discovery

### LINKED ISSUE #3233: [ENH] DiBS Algorithm
DiBS is a deep learning based technique for causal discovery, currently not yet implemented in pgmpy. 

**Describe the solution you'd like** A clear and concise description of what you want to happen, ideally taking into consideration the existing toolbox design, classes and methods. I will build a class for this algorithm, using code from the codebase of the authors or the paperhttps://arxiv.org/pdf/2105.11839: https://github.com/larslorch/dibs/tree/master/dibs. The authors of this paper used the JAX framework for their implementation.

This is what the API could look like:

```python
from pgmpy.causal_discovery.base import _BaseCausalDiscovery

class DiBS(_BaseCausalDiscovery):
    def __init__(
        self,
        n_particles=20,
        n_steps=3000,
        learning_rate=1e-2,
    ):
        ...

    def _fit(self, X: pd.DataFrame):
        values = X.values
        nodes = list(X.columns)

        self._initialize_particles(n_nodes=len(nodes))
        self._run_inference(values)
        self.graph_samples_ = self._sample_graphs(nodes)
        self.causal_graph_ = self._summarize_graphs(self.graph_samples_)
        return self

    def _initialize_particles(self, n_nodes):
    def _run_inference(self, X):
    def _sample_graphs(self, nodes):
    def _summarize_graphs(self, graph_samples):
```

Here, I decided to summarize the graphs in some way, but as DiBS is finding a posterior over graphs, we could also return a number of samples that represent the posterior. What would be preferred?

**Additional context** This issue is opened as discussed with @ankurankan as part of the MSc course "Open Source Software Development" at Radboud University 

~ Yves van Haaren

winkelwagentje:

This page does not seem to exist.

- [ ] Does the PR fully address the linked issue and is within its defined scope? If you are still working on the PR, mark it as draft.
Yes, it fully addresses the linked issue and is within its defined scope. However, we may choose to extend the implementation with some more log likelihood models for the graphs and such.

How do I know in advance? Not familiar with this.

- [ ] Have you reviewed our [AI usage policy](https://github.com/pgmpy/pgmpy/blob/dev/Contributing.md#ai-usage-policy)?
- [ ] Are you able to fully explain your changes? We expect you to fully understand the algorithm and take full responsibility for any changes in this PR.  

I used AI to write some tests. I fed the LLM examples of other test files for other implementations as well as other algorithm implementations and my own algorithm's implementation, and asked it to write some tests similar in style to the other files. 

I mostly checked if the objects returned are of the correct types and shapes as explicit testing of the contents of the returned objects is very difficult.

Yes, I did compress them into a smaller amount after revising my own code for a bit and then removed some redundant tests that did not add to the quality of the implementation.

- Fixes #3233 

- Implementation of the DiBS framework/algorithm, including multiple implementations for computing gradients: score and reparameterization. 
- Tests for the DiBS algorithm


ankurankan:
@winkelwagentje 
1. I think you have accidentally committed your `.idea` directory. It will need to be removed, and that should automatically fix the linting checks.
2. `torch` is an optional dependency for us, and `pytest` is a test dependency. They are already listed in the optional-dependencies section of `pyproject.toml`. So, your changes in the `pyproject.toml` will need to be reverted.
3. The other test is failing because it requires pyro to be installed: https://github.com/pgmpy/pgmpy/actions/runs/25549005372/job/74992070324?pr=3381. This "Core deps" test suite is only for testing the core dependencies of the package. We use safe import and dependency checks to handle tests that depend on optional dependencies. You can have a look here for an example of how to use them: https://github.com/pgmpy/pgmpy/blob/dev/pgmpy/tests/test_models/test_FunctionalBayesianNetwork.py#L21

These changes should get the tests to pass.

ankurankan:
@winkelwagentje In case you don't have the `pre-commit` hook installed, we have the setup instructions here: https://github.com/pgmpy/pgmpy/blob/dev/CONTRIBUTING.md#pre-commit-hooks . It should automatically take care of all the linting fixes when you commit.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+            Observational data where columns are variables and rows are samples.
+
+        Returns
+        -------
+        self : DiBS
+            Fitted estimator.
+        """
+
+        self.n_features_in_ = X.shape[1]
+        self.feature_names_in_ = X.columns.tolist()
```
ankurankan: These attributes do not need to be set. They automatically get defined in the `_BaseCausalDiscovery.fit` method.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+        self.beta = lambda t: beta_linear * (t + 1)
+        self.latent_dim = latent_dim  # dimension of each U_i and V_i.
+        self.n_grad_mc_samples = n_grad_mc_samples
+        self.n_acyclicity_mc_samples = n_acyclicity_mc_samples
+        self.latent_prior_std = latent_prior_std
+        self.tau = tau
+        if torch.cuda.is_available():
+            self.device = torch.device("cuda")
+        else:
+            self.device = torch.device("cpu")
```
ankurankan: The device and dtypes should be read from `pgmpy.global_vars.config` object.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+        Returns
+        -------
+        torch.Tensor
+            Boolean adjacency matrices of shape
+            ``(n_particles, n_nodes, n_nodes)``.
+        """
+
+        X_t = torch.tensor(X.to_numpy(), device=self.device, dtype=torch.float32)
+        n_nodes = X.shape[1]
+        particles = torch.nn.Parameter(self._initialize_particles(n_nodes).to(self.device))
```
ankurankan: I would suggest not defining methods such as `_initialize_particles`. This method is a single-line operation and is only called from this line of code. To help readability in pgmpy, we try to keep the algorithm logic linear/local and only creating methods if:
1. We reuse the same logic at multiple places.
2. The method by itself is obvious without any context around it.
I would recommend having a look at: https://htmx.org/essays/locality-of-behaviour/ . We try to have this locality in terms of methods as well.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+                        dim=1,
+                    )
+                else:
+                    design = torch.ones((n_samples, 1), device=device, dtype=dtype)
+
+                beta = torch.linalg.lstsq(design, y.unsqueeze(1)).solution
+                resid = y.unsqueeze(1) - design @ beta
+                rss = (resid.squeeze(1) ** 2).sum()
+
+                sigma2 = (rss / n_samples).clamp_min(1e-8)
```
ankurankan, line 249: This seems to be the standard log-likelihood. Can we reuse it from the `structure_score` module?

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+        ----------
+        graph_samples : list[nx.DiGraph]
+            Posterior graph samples.
+
+        Returns
+        -------
+        tuple
+            ``(summary_graph, edge_probs, adjacency_matrix)``.
+        """
+        nodes = list(graph_samples[0].nodes())
```
ankurankan: Would suggest using `feature_names_in_` instead everywhere. It would be uniform way to access the node names and will make the code more readable.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+            Observational data with shape ``(n_samples, n_nodes)``.
+
+        Returns
+        -------
+        torch.Tensor
+            Boolean adjacency matrices of shape
+            ``(n_particles, n_nodes, n_nodes)``.
+        """
+
+        X_t = torch.tensor(X.to_numpy(), device=self.device, dtype=torch.float32)
```
ankurankan: Same here. `device` and `dtype` should come from the global config object.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+        n_nodes = X.shape[1]
```
ankurankan: We should use `n_features_in_` everywhere.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+        -------
+        self : DiBS
+            Fitted estimator.
+        """
+
+        self.n_features_in_ = X.shape[1]
+        self.feature_names_in_ = X.columns.tolist()
+
+        # Run inference and store particle graph samples
+        self._run_inference(X)
```
ankurankan: I would recommend writing such methods in a way that they return values rather than setting attributes. The core logic/flow of the method should be obvious in the `_fit` method. Having methods like `_run_inference` set attributes makes it harder to understand what it does without looking at the method itself.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+
+        # Convert particle samples to graphs
+        self.graph_samples = self._sample_graphs(X.columns.tolist())
```
ankurankan: `X.columns.tolist()` should be `self.feature_names_in`. We should uniformly use that wherever the variable names is needed.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+        for adj in graph_samples:
+            graph = nx.DiGraph()
+            graph.add_nodes_from(nodes)
+            src_idx, dst_idx = np.where(adj)
+            graph.add_edges_from((nodes[i], nodes[j]) for i, j in zip(src_idx, dst_idx) if i != j)
+            sampled_graphs.append(graph)
+
+        return sampled_graphs
+
+    def _summarize_graphs(self, graph_samples):
```
ankurankan: I don't think graph summarization methods should be part of this class as they aren't part of the algorithm and can be applied to use cases outside this algorithm. Would suggest moving this to `pgmpy.utils`.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+        ----------
+        X : pd.DataFrame
+            Observational data where columns are variables and rows are samples.
+
+        Returns
+        -------
+        self : DiBS
+            Fitted estimator.
+        """
+
```
ankurankan: I see that there are various checks in methods that throw error, e.g., check for the kernel argument. All of these checks should ideally be done before the actual algorithm runs. This also keeps the rest of our methods clean, as in the current approach, we are mixing validation and algorithm logic in the methods. Ideally, they should be separate.

ankurankan: You can also have a look at some of our other causal discovery algorithm implementations for examples.

pgmpy/causal_discovery/DiBS.py
```diff
@@ -0,0 +1,843 @@
...
+    def __init__(
+        self,
+        n_particles: int = 30,
+        n_steps: int = 2000,
+        log_likelihood: Callable | None = None,
+        learning_rate: float = 5e-3,
+        edge_prob_threshold: float = 0.0,
+        kernel: str = "frobenius",
+        kernel_bandwidth: float | str = "median",
```
ankurankan: If we only support a single value for these arguments, we shouldn't have them as arguments.

pgmpy/tests/test_causal_discovery/test_DiBS.py
```diff
@@ -0,0 +1,130 @@
...
+
+import unittest
+
+import networkx as nx
+import numpy as np
+import pandas as pd
+from skbase.utils.dependencies import _check_soft_dependencies, _safe_import
+
+torch = _safe_import("torch")
+
```
ankurankan, line 15: The sklearn compatibility checks will need to be added. Have a look at the `parameterize_with_checks` tests in the other causal discovery test files such as `test_PC.py`, `test_HillClimbSearch.py`. This would also most like fail for the current implementation as sklearn doesn't allow for any code logic in the `__init__` method (I see some if-else assignment in the init currently)

ankurankan (changes requested):
@winkelwagentje Thanks for the PR. I had a quick look, and I think the code needs some work in terms of organization and readability. I have added a few inline comments, but broadly:

1. We would like to have the core logic of the algorithm in the `_fit` method. Ideally, the main steps of the algorithm should be obvious from just looking at this method. We also try to write comments for each main step of the algorithm (have a look at other causal discovery algorithm implementations). We would like these steps to be written in a way such that it gives us the pseudocode of the algorithm.
2. I think we should reduce the number of methods in this class. Feel free to make the `_fit` method larger as long as they are the steps of the algorithm is clear.
3. I see the number and names of nodes have been accessed in many different ways in the code. These should be uniform using `n_features_in_` and `feature_names_in_`.
4. I see some conversions between torch and numpy. I would recommend converting everything to torch as part of the preprocessing step (Step 0 in `_fit`). This would help us in having a clear separation between algorithm logic and preprocessing steps.

pgmpy/causal_discovery/DiBS.py
winkelwagentje: The issue I am having for that is that I can't set self.causal_graph_ if we don't summarize it in some way. Do you suggest to just move this functionality or also remove it altogether from DiBS as it is not part of the algorithm?

pgmpy/causal_discovery/DiBS.py
winkelwagentje: I think in theory we could but it would degrade performance severely I suspect; the _lgbn_log_likelihood can take batches of sampled graphs. If we want to reuse the LogLikelihoodGauss API which is pandas backed, we would have to do conversions of tensors to numpy arrays and/or pandas dataframes at each iteration of the algorithm for many graphs/samples. This would make everything really slow probably.

pgmpy/causal_discovery/DiBS.py
winkelwagentje: Changed that! Also, I was wondering what you think about how we should initialize the particles as the paper doesn't explicitly discusses this. We use a latent dimension for the embeddings u_i and v_j, but I think we should maybe divide by the sqrt of the latent dimension to keep the vectors at unit variance. What do you think?

pgmpy/causal_discovery/DiBS.py
ankurankan: You can leave the summarization for now. For now, you can just assign one of the sampled DAGs as `self.causal_graph_`. We are already implementing some summarization techniques in https://github.com/pgmpy/pgmpy/pull/3464. We can later on hook those methods here as well.

winkelwagentje: Alright, will do!

winkelwagentje:
Hi, I didn't see any comments after you merged dev into my branch. Are there still things I should change before we can merge?

Issue #3245, opened 2026-03-30
[BUG] `DiscreteBayesianNetwork.copy` doesn't copy variable roles.


ankurankan:
```python
In [1]: from pgmpy.example_models import load_model 

In [9]: model = load_model('bnlearn/asia')

In [13]: model.with_role(role='exposure', variables='asia', inplace=True)

In [14]: model.get_role_dict()
Out[14]: {'exposure': ['asia']}

In [15]: model_copy = model.copy()

In [16]: model_copy.get_role_dict()
Out[16]: {}

```

Theavinash02:
Hi @ankurankan ,
I would like to work on this issue. Could you please assign this to me?
Thanks
Avinash

ankurankan:
Fixed in #3410

PR #3410, opened 2026-05-31
DiscreteBayesianNetwork.copy doesn't copy variable roles.

Theavinash02:
Issue no: #3245 
This PR contains code changes that will resolve the bug that we noticed in `DiscreteBayesianNetwork.copy()` was not preserving all model state when creating a copy during our community call.
- This PR contain code changes that replaced the manual `add_nodes_from` + `add_edges_fro`m  with `DiscreteBayesianNetwork(self)`, which lets networkx handle copying nodes, edges, and node attributes.
- Passed  `latents` , `exposures`, and `outcomes ` to the constructor so they are preserved.
- Added tests to verify roles and latents are correctly reserved and that the copy is independent from the original.


Theavinash02:
Hi @ankurankan ,
Could you please review this code and kindly share your feedback?
Thanks
Avinash

Theavinash02:
> Hi @ankurankan , Could you please review this code and kindly share your feedback? Thanks Avinash

Hi @ankurankan ,
I noticed that this fix causes some CI failures in `test_EM.py`. These are not a bug introduced by this PR but rather a pre-existing issue that was being masked by the old broken `copy()` behavior.

The old `copy()` was silently dropping `latents`, so any internal copy of the model had an empty `latents` set. The EM estimator calls `copy()` internally and was accidentally relying on this with `latents` dropped, latent variables like `Smoker` were being treated as observed nodes and the dataset validation was passing incorrectly.

Now that `copy()` correctly preserves `latents`, the EM estimator hits the validation check and raises:
```
ValueError: Nodes detected in the model that are not present in the dataset: {'Smoker'}
```

Could you please let me know how can we proceed this further?

Thanks,
Avinash

pgmpy/models/DiscreteBayesianNetwork.py
```diff
@@ -1089,12 +1089,14 @@ def copy(self) -> DiscreteBayesianNetwork:
         >>> len(copy_model.get_cpds())
         3
         """
-        model_copy = DiscreteBayesianNetwork()
-        model_copy.add_nodes_from(self.nodes())
-        model_copy.add_edges_from(self.edges())
```
ankurankan, line 1094: model_copy.latents = self.latents.copy()
...exposure, outcome.

Theavinash02:
Hi @ankurankan ,
As discussed on our debugging session, i have created an issue #3514.
As a temp fix, i have update the code as suggested. Could you please review and kindly approve this PR?
Thanks

ankurankan (approved):

PR #3459, opened 2026-06-09
01-01. Implement `BaseParameter`

### LINKED ISSUE #3453: 01-01. Implement `BaseParameter`
It is child issue of #3452 for tracking.

### Proposal Solutions

- scikit-learn and skpro each define tag lists tailored to their own libraries.
- This follow `skpro`'s tagging system.
- ~~rename exist path `pgmpy/base/` -> `pgmpy/graph/`~~
- ~~create path `pgmpy/base/`. This path has a responsibility of `BaseEstimator` and `Mixin` class.~~

> In scikit-learn, tags are used to dynamically run tests based on estimator properties, such as input data validation and automated common unit tests. <br>
> In skpro, tags provide users with a way to search for models that can perform a specific workflow, for example through `all_objects()`, `all_tags()`. <br>
> In pgmpy, a model’s tags are used to determine whether a specific algorithm can be run.

```py
from skbase.base import BaseEstimator as _BaseEstimator

class _ParameterTags:
    _config = {}

    _tags = {
        "variable_type" :"discrete"
        "produces_factor" : False
        "is_linear_gaussian": False
        "supports_fit_joint": False
        "python_dependencies" = []
    }

class BaseParameter(_ParameterTags, _BaseEstimator):
    ...
```

- [skpro/base/_base.py](https://github.com/sktime/skpro/blob/main/skpro/base/_base.py)

### Alternative Solutions

- rename exist path `pgmpy/base/` -> `pgmpy/graph/`
- create path `pgmpy/base/`. This path has a resopnsibity of `BaseEstimator` and Tagging system.

```py
# pgmpy/utils/_tags.py
@dataclass(slots=True)
class Tags:
    accessor_type: str | None
    parameter_tags: ParameterTags | None = None

@dataclass(slots=True)
class ParameterTags:
    variable_type: str | None # "discrete" or "continous"
    produces_factor: bool = False
    is_linear_gaussian: bool = False
    supports_fit_joint: bool = False
    distribution: bool = False # Is this CPD can be root node's distribution? (P(X))
    python_dependencies = []

```

```py
# pgmpy/base/_base.py
from sklearn.base import BaseEstimator as _BaseEstimator

class BaseParameter(_BaseEstimator):
    def __pgmpy_tags__(self):
        return Tags(
            accessor_type="parameter",
            parameter_tags=ParameterTags(),
        )

```

- [sklearn/base.py](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/base.py)
- [sklearn/utils/_tags.py](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_tags.py)

- ENH proposal link: https://github.com/pgmpy/enhancement_proposals/pull/12

daehyun99:

- Fixes #3453
- Part of #3527

- Feat: Implement `BaseParameter`
- Feat: Implement test code


daehyun99:
@ankurankan I think this PR ready for merge.
Please review when you have some time.

ankurankan (changes requested):
- I think it might be better to rename `parameterization` -> `parameters`. It would be a bit easier to type.
- In the base class, we shouldn't define the tag values as they are not valid for the base class. We can use them to store the expected type, for example, `variable_type: str`, "supports_fit_joint: bool", etc.
- Can you also look into what tags skpro set for its regressor/classifiers? Should we also use some of them?
- I wouldn't expose the `name` property. I think directly using `self.__class__.__name__` is a fairly standard way to access the name.
- I would suggest implementing methods such as `fit`, `log_pdf`, etc., and throw a `NotImplementedError`. These methods can later also be modified to contain various common checks, for example, checking if the data passed is supported (data types, size, etc.) by the chosen estimator.

daehyun99:
@ankurankan 

> * I think it might be better to rename `parameterization` -> `parameters`. It would be a bit easier to type.

Apply

> * In the base class, we shouldn't define the tag values as they are not valid for the base class. We can use them to store the expected type, for example, `variable_type: str`, "supports_fit_joint: bool", etc.

Apply

> * Can you also look into what tags skpro set for its regressor/classifiers? Should we also use some of them?

skpro currently supports only regression models. The reason there is no categorical distribution is also that classifier models are not supported yet.

Among the tags for regression models, `capability:missing` looks like a good tag for indicating whether missing data can be handled.

However, since the tags in pgmpy are separate from the tags in skpro and sklearn, I implemented it as a `missing` tag.


```py
    _tags = {
        "object_type": "regressor_proba",  # type of object, e.g., "distribution"
        "estimator_type": "regressor_proba",  # type of estimator, e.g., "regressor"
        "capability:survival": False,
        "capability:multioutput": False,
        "capability:missing": True,
        "capability:update": False,
        "X_inner_mtype": "pd_DataFrame_Table",
        "y_inner_mtype": "pd_DataFrame_Table",
        "C_inner_mtype": "pd_DataFrame_Table",
    }
```

> * I wouldn't expose the `name` property. I think directly using `self.__class__.__name__` is a fairly standard way to access the name.

Apply

> * I would suggest implementing methods such as `fit`, `log_pdf`, etc., and throw a `NotImplementedError`. These methods can later also be modified to contain various common checks, for example, checking if the data passed is supported (data types, size, etc.) by the chosen estimator.

Apply.
`log_pdf()` already exists on the distribution class, so it seems better not to duplicate its implementation in the CPD parameterization class.

ankurankan:
@daehyun99 Looks good. I will wait for one of the refactors (class inheriting this) to be done before merging this.

daehyun99:
Please review when you have a chances.
- I enhance this PR's `_check_data` logic.(Checking `X`, `y`, `sample_weight`)
- This logic will cover every parameter's input data.
- This is the initial version.

fkiraly (changes requested):
Nice!

To me it looks like the `BaseParameter` class is identical to `BaseRegressor` in `skpro`, with a few exceptions or required extensions:

* support for categorical return distributions - should be added to `skpro`. Potentially naming the respective base class `BaseClassifier`.
* support for fixed conditional distributions, where `fit` does nothing. As in your PR, I would indeed expose them under the *same* API and not as a separate base class, and distinguish by a tag. This has to be introduced.
* the other tags - we should review how much this is duplicative or extensible, only introducing as much as needed.
* the `sample` method. As far as I understand, `sample(X, n_samples)` is equivalent to `predict_proba(X).sample(n_samples)`, so it is disputable whether one needs this additional syntactic sugar, or whether one wants to keep the `predict_proba` API clean. Design-wise, one could invoke the "law of demeter" in which a class should avoid reaching into an adjacent class, but I also see the conceptual appeal of having `sample` right here, especially for fixed conditional distributions.

Hence, my key suggestion is to work on this as follows:

1. try another version of this PR, but where you inherit from `BaseRegressor`. Override only what you need! That should result in a much leaner class. For instance, I believe you do not need to override `fit` or `_fit`, or even `predict_proba`, or `_predict_proba`, and most of the conversion/checking also becomes redundant.
2. work on the downstream PR until you are 90% sure that you have everything what you need there in this PR.
3. contribute the changes to `skpro`.

daehyun99:
@fkiraly 

Thank you for the review!!

> To me it looks like the BaseParameter class is identical to `BaseRegressor` in `skpro`, with a few exceptions or required extensions:
> * support for categorical return distributions - should be added to `skpro`. Potentially naming the respective base class `BaseClassifier`.

We can revisit this discussion after #3462 is added to skpro.

- Related:
    - https://github.com/sktime/skpro/issues/1003
    - https://github.com/sktime/skpro/issues/1071

> * support for fixed conditional distributions, where `fit` does nothing. As in your PR, I would indeed expose them under the _same_ API and not as a separate base class, and distinguish by a tag. This has to be introduced.

I think the `DistributionAdapter` currently under development is a good example.

Does `skpro` also need adding `DistributionAdapter`?

- Issues:
    - [04-02. Implement `DistributionAdapter`](https://github.com/pgmpy/pgmpy/issues/3525)
    - [[ENH] Make a `DistributionAdapter` class to fittable class when the `self.distributions` is `scipy.stats`'s distribution](https://github.com/pgmpy/pgmpy/issues/3531)
- PR:
    - https://github.com/daehyun99/pgmpy/blob/f63bf83abaad160f7508707f7910995b10e6772e/pgmpy/parameter/adapter/DistributionAdapter.py

I also think that the extension direction of `DistributionAdapter`(make to fittable) is related to skpro’s `NaiveProbaRegressor` (https://github.com/sktime/skpro/pull/1108/changes).

> * the other tags - we should review how much this is duplicative or extensible, only introducing as much as needed.

This could be discussed in #3527.

> * the `sample` method. As far as I understand, `sample(X, n_samples)` is equivalent to `predict_proba(X).sample(n_samples)`, so it is disputable whether one needs this additional syntactic sugar, or whether one wants to keep the `predict_proba` API clean. Design-wise, one could invoke the "law of demeter" in which a class should avoid reaching into an adjacent class, but I also see the conceptual appeal of having `sample` right here, especially for fixed conditional distributions.

I missed this in yesterday’s meeting, but there are two reasons for introducing the `sample()` method.

1. The contract between `BayesianFunctionalRegression` (which uses `pyro`) and the output of `predict_proba()`

We have an API rule that `predict_proba(X)` must return an skpro distribution.

On the other hand, in `BayesianFunctionalRegression` using `pyro`, after training with `mcmc` or `svi`, we can sample from the posterior. Those samples are in `pytorch.tensor` format.

- https://github.com/pgmpy/pgmpy/pull/3533

To satisfy the `predict_proba(X)` contract, we would need something like `PyroToSkpro` as in the PR. `PyroToSkpro` wraps posterior samples into an skpro distribution class. However, the resulting distribution may no longer have the same meaning as the original posterior.

Therefore, for `BayesianFunctionalRegression`, I think it is better not to support `predict_proba(X)` and instead only support a `sample(X)` method that directly returns posterior samples.

2. The meaning of `predict_proba()` at a root node

We can use `TabularCPD`, `LinearGaussianCPD`, and `DistributionAdapter` on root nodes.
However, a root node has no parents, so calling `predict_proba(X)` would return a distribution that does not depend on `X`.

(I think the same case also occurs when calling `predict_proba(X)` on skpro’s `NaiveProbaRegressor`. - https://github.com/sktime/skpro/pull/1108/changes)

That clashes with the name `predict`, which suggests prediction from `X`. In that case, `parameter.sample(n_samples)` seems more appropriate.

```py
from skpro.distributions.normal import Normal
parameter = DistributionAdapter(Normal(mu=0, sigma=1))
```

The cases I have in mind are as follows, and I believe a `sample()` method is needed.

- Root node case:
    - `predict_proba(X)` and `sample(X)` cannot be used.
    - Sampling is possible via `sample(n_samples)`.
- Non-root node case (has parent nodes):
    - `predict_proba(X)` can return a probability distribution.
    - `sample(n_samples)` cannot be used because the conditioning values `X` are not given.
    - `sample(X)` can sample from the conditional distribution.


> Hence, my key suggestion is to work on this as follows:

> 1. try another version of this PR, but where you inherit from `BaseRegressor`. Override only what you need! That should result in a much leaner class. For instance, I believe you do not need to override `fit` or `_fit`, or even `predict_proba`, or `_predict_proba`, and most of the conversion/checking also becomes redundant.
> 2. work on the downstream PR until you are 90% sure that you have everything what you need there in this PR.
> 3. contribute the changes to `skpro`.

I got it, I’ll proceed with that.

Thank you.

PR #3460, opened 2026-06-10
[ENH]Add ProbabilityExpression Class and tests for ID algorithms

### LINKED ISSUE #3435: Add ProbabilityExpression Class and tests for formula returning Identification algorithms
The non-role based identification algorithms (ID, IDC, ID*, σ-ID) need to return symbolic closed-form formulas over observed distributions. Currently pgmpy has no way to represent or render such expressions. This module provides the data structure that all four algorithms will output. Tests are essential here because bugs in the expression tree or incorrect LaTeX rendering could silently corrupt algorithm output
 
Add a new module `pgmpy/identification/probability_expressions.py` containing:

`_TreeNode` :  private base class for all expression tree nodes, providing `node_type`, `children`, `to_latex()`, and `__repr__()`
`Prob` : leaf node for atomic probability terms, with fields variables, do, cond, sumset
`Marginal`:  internal node for `∑sumset​(⋅)`, one child
`Product` : internal node for products of two or more expressions
`Division` : internal node for a ratio, exactly two children (numerator, denominator)
`ProbabilityExpressionTree` : public container returned by identify(); holds a root node and exposes `to_latex()` and `simplify()`

Tests (in `pgmpy/tests/test_identification/test_probability_expressions.py`):

- Construction: correct children length and `node_type` for each node class, default arguments
- Validation: Product with fewer than two factors, ProbabilityExpression with a non-_TreeNode root
- Tree traversal: a nested tree can be fully walked via .children; Prob always terminates
- LaTeX rendering: individual node types render correctly, and two end-to-end tests: the frontdoor formula and the bow-arc formula (both have known closed-form outputs from the literature)

**Additional context**
Reference Proposal :  [Expanding Causal Identification module](https://github.com/pgmpy/enhancement_proposals/tree/main/3_causal_identification)

kajal-jotwani:

I've used Claude Sonnet 4.6.

The design for the probabiltyExpression class was already decided so I just had to majorly implement the `to_latex` methods in all the classes. I used ai assitance for validation my logic and helping me with the examples. It recommened to add the `_vars_to_latex` as we need sorted set of variables otherwise the tests may fail. I also used it to help me with the test and edge cases which I was missing while implementing the tests. 

The implementation follows the design discussed in the proposal. I incrementally built the bow-arc tree in the REPL, checked each node's to_latex output at each level, and verified the final output against the formula in Shpitser & Pearl (2006).

Yes I used AI assistance to identify the test and edge cases I was missing The tests follow one test per method in a class so the tests are minimal with maximum code coverage

- Fixes #3435 

- Added `pgmpy/identification/probability_expression.py` with `_TreeNode`, `Prob`, `Marginal`, `Product`, `Division`, `ProbabilityExpressionTree`
- Added `pgmpy/tests/test_identification/test_probability_expression.py` with 18 tests
- Updated `pgmpy/identification/__init__.py` to export `ProbabilityExpressionTree`


pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+    cond : frozenset of str, optional
+        Variables being passively conditioned on.
+        Example: ``frozenset({"Z"})`` renders after the conditioning bar.
+        Default: ``frozenset()``.
+
+    sumset : frozenset of str, optional
+        Variables summed out directly at this node (inline marginalisation).
+        Example: ``frozenset({"Z"})`` prepends ``\\sum_{Z}`` to the term.
+        Default: ``frozenset()``.
+
```
ankurankan: Why are we doing sumset here when we have a `Marginal` class? Wouldn't it be cleaner to keep them separate?

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
+from abc import ABC, abstractmethod
+
+
+class _TreeNode(ABC):
```
ankurankan: I would suggest not using ABC. It usually becomes too strict, and we have followed duck typing package-wide. We can keep a very lean base class here.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+        """Return an unambiguous string representation of this node.
+
+        Returns
+        -------
+        str
+        """
+        raise NotImplementedError
+
+
+class Prob(_TreeNode):
```
ankurankan: Maybe we can rename this to `ProbabilityNode` ?

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+        if self.do:
+            parts.append(f"do={set(self.do)!r}")
+        if self.cond:
+            parts.append(f"cond={set(self.cond)!r}")
+        if self.sumset:
+            parts.append(f"sumset={set(self.sumset)!r}")
+        return "Prob(" + ", ".join(parts) + ")"
+
+
+class Marginal(_TreeNode):
```
ankurankan: Similarly `MarginalNode`. Same for the rest.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+        self.variables = frozenset(variables)
+        self.do = frozenset(do)
+        self.cond = frozenset(cond)
+        self.sumset = frozenset(sumset)
+        self.children = []
+
+    @staticmethod
+    def _vars_to_latex(var_set):
+        """Sort and join variable names for deterministic LaTeX output."""
+        return ", ".join(sorted(var_set))
```
ankurankan: I think this method is used in other classes as well. Might be better to keep it in the base class so that it is available in all child classes.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+    children : list
+        Always ``[]``. ``Prob`` is always a leaf node.
+
+    Examples
+    --------
+    >>> Prob(frozenset({"Y"})).to_latex()
+    'P(Y)'
+
+    >>> Prob(frozenset({"Y"}), cond=frozenset({"X"})).to_latex()
+    'P(Y \\\\mid X)'
```
ankurankan: Why are the latex strings showing with so many `\`s ?

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+    --------
+    >>> m = Marginal(
+    ...     Prob(frozenset({"Y"}), cond=frozenset({"X"})),
+    ...     sumset=frozenset({"X"}),
+    ... )
+    >>> m.to_latex()
+    '\\\\sum_{X} P(Y \\\\mid X)'
+    """
+
+    node_type = "sum"
```
ankurankan: Do we need the `node_type` anywhere? Can we not simply compare to the type of object given that each class has a unique name.

ankurankan (changes requested):

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+    -----
+    The `to_latex()` method delegates to `self.root.to_latex()` and
+    returns a complete LaTeX representation suitable for math
+    environments. Rendering proceeds bottom-up: leaf `Prob` nodes render
+    themselves and internal nodes combine their children's strings using
+    the appropriate notation (e.g. `\\sum`, `\frac`, brackets).
+
+    Examples
+    --------
+    Build the bow-arc formula
```
DARHWOLF, line 392: Could you add the import statements here

pgmpy/tests/test_identification/test_probability_expression.py
```diff
@@ -0,0 +1,298 @@
+import pytest
+
+from pgmpy.identification.probability_expression import (
+    Division,
+    Marginal,
+    Prob,
+    ProbabilityExpressionTree,
+    Product,
+    _TreeNode,
+)
```
DARHWOLF, line 10: Is it a good idea to have so many imports for users? I'm seeing in the example above that we have to import every internal node implementation that is present in the equation. 

If there are any users who want to use this as a standalone implementation, wouldn't it be better and cleaner to have one public import and not have to import the rest of the internal classes?

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+
+class Division(_TreeNode):
+    """
+    Internal node representing a ratio of two sub-expressions::
+
+        numerator / denominator
+
+    Parameters
+    ----------
+    numerator : Prob or Marginal or Product or Division
```
DARHWOLF: wwould it be better to specify that these are "node-types" here?

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+
+class Product(_TreeNode):
+    """
+    Internal node representing a product of two or more sub-expressions::
+
+        factors[0] * factors[1] * ... * factors[n]
+
+    Parameters
+    ----------
+    factors : list of (Prob or Marginal or Product or Division)
```
DARHWOLF: I would prefer `list of node-types`

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+        """
+        Return LaTeX for this atomic probability term.
+
+        The conditioning bar is omitted when both ``do`` and ``cond`` are
+        empty.  ``do(·)`` is always rendered before passive conditioning.
+        The ``sumset`` prefix is prepended when non-empty.
+
+        Returns
+        -------
+        str
```
DARHWOLF, line 105: missing desciption

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,433 @@
...
+    returned by causal identification routines and is not itself a tree
+    node.
+
+    The expression tree is composed of `Prob`, `Marginal`, `Product`, and
+    `Division` nodes (all subclasses of `_TreeNode`). Inspect or traverse the
+    tree via the `root` attribute and the `children` lists on nodes.
+
+    Parameters
+    ----------
+    root : Prob | Marginal | Product | Division
```
DARHWOLF: I would suggest using `or` rather than `|` and following a uniform style here, as you are using elsewhere.

pgmpy/tests/test_identification/test_probability_expression.py
```diff
@@ -0,0 +1,298 @@
...
+    """
+    Bow-arc formula: sum_Z [ P(Z|X) * sum_{X'} [ P(Y|Z,X') P(X') ] ]
+
+    P(Y|do(X)) for the bow-arc graph (X->Z->Y, X<->Z). Role-based methods
+    cannot derive this; it requires the ID algorithm.
+    Reference: Shpitser & Pearl (2006), AAAI-06.
+    """
+
+    def test_to_latex(self, bowarc_expr):
+        latex = bowarc_expr.to_latex()
```
DARHWOLF: The bow-arc formula is defined as `sum_Z [ P(Z|X) * sum_{X'} [ P(Y|Z,X') P(X') ] ]` in the docstring on line 279.
However, the output that I am seeing for the variable latex is:-

`'\\sum_{Z} P(Z \\mid X) \\left[ \\sum_{X} P(Y \\mid X, Z) P(X) \\right]'`

In the docstring you're using the x' convention, but the output here does not render the x-"dash". From what I can guess, I think it is supposed to be a smaller X. If this is the case, this would be confusing. I suggest changing the docstring to match the convention

Also:-
I am not sure why there are `\\` between the terms? It does not render correct on my latex editor. It is putting the terms in new lines.

The output that renders correct for me is 
`\[\sum_{Z} P(Z \mid X) \left[ \sum_{X} P(Y \mid X, Z) P(X) \right]\]`

Moreover, I do not see any difference between `X` and `mid X` in the rendered formula. Both look exactly the same to me. This needs correction.

DARHWOLF (changes requested):
Please see my comment on line 287. 

- I think there is a mismatch between the conventional representation of the frontdoor formula and the bow-arc with the output that I am seeing in the tests.

- The latex formula does not seem to be correct with the use of `mid M` and the double `//`. Please have a look here. 

- I would suggest ensuing that the outputs follow the convention that you have described in the docstrings for tests.  

- I would also suggest checking the outputs simultaneously on a latex editor to see whether they render correct.

kajal-jotwani:
```
(pgmpy) kajal@kajal-HP-Pavilion-Laptop-15-eh2xxx:~/pgmpy$ uv run python3
Python 3.14.5 (main, May 10 2026, 19:28:16) [Clang 22.1.3 ] on linux
Type "help", "copyright", "credits" or "license" for more information.
Ctrl click to launch VS Code Native REPL
>>> from pgmpy.identification.probability_expression import (
...     ProbabilityNode,
...     MarginalNode,
...     ProductNode,
...     ProbabilityExpressionTree,
... )
>>> 
>>> inner_product = ProductNode([
...     ProbabilityNode(
...         frozenset({"Y"}),
...         cond=frozenset({"Z", "X"}),
...     ),
...     ProbabilityNode(
...         frozenset({"X"}),
...     ),
... ])
>>> 
>>> inner_sum = MarginalNode(
...     inner_product,
...     sumset=frozenset({"X"}),
... )
>>> 
>>> outer_product = ProductNode([
...     ProbabilityNode(
...         frozenset({"Z"}),
...         cond=frozenset({"X"}),
...     ),
...     inner_sum,
... ])
>>> 
>>> expr = ProbabilityExpressionTree(
...     root=MarginalNode(
...         outer_product,
...         sumset=frozenset({"Z"}),
...     )
... )
>>> 
>>> latex = expr.to_latex()
>>> 
>>> print(repr(latex))
'\\sum_{Z} P(Z \\mid X) \\left[ \\sum_{X} P(Y \\mid X, Z) P(X) \\right]'
>>> print()

>>> print(latex)
\sum_{Z} P(Z \mid X) \left[ \sum_{X} P(Y \mid X, Z) P(X) \right]
>>> 
```
The four backslashes appeared because the docstrings were non-raw strings  `\\mid` in a non-raw docstring collapses to `\mid` at the string-object level, so doctest sees the expected output as 'P(Y \mid X)' (one backslash), but repr() of the actual returned string shows 'P(Y \\mid X)' (two backslashes), causing a mismatch. they're now raw strings to avoid the issue

pgmpy/tests/test_identification/test_probability_expression.py
kajal-jotwani: I think the typical end-user flow doesn't involve importing any of these classes they're returned inside a `ProbabilityExpressionTree` from the identification functions, and users just call `.to_latex()` on the result. The import list only shows up when constructing trees manually (as the tests do, and as internal algorithm implementations will do)

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+    Rendering proceeds bottom-up: leaf ``ProbabilityNode`` nodes render
+    themselves and internal nodes combine their children's strings using
+    the appropriate notation (e.g. ``\sum``, ``\frac``, brackets).
+
+    Examples
+    --------
+    Build the bow-arc formula
+    ``\sum_Z P(Z|X) [ \sum_X P(Y|Z,X) P(X) ]``:
+
+    >>> from pgmpy.identification import ProbabilityExpressionTree
```
DARHWOLF: This example isn't working without the remaining imports. You'll have to import MarginalNode, ProbabilityNode and ProductNode too.


> I think the typical end-user flow doesn't involve importing any of these classes 

This makes me question whether the users would be required to have all of these imports (refering to my earlier comment -- and your response to it.)

You can also check from doctests which are failing.

pgmpy/tests/test_identification/test_probability_expression.py
```diff
@@ -0,0 +1,311 @@
...
+
+        d = DivisionNode(prob_y, prob_x)
+        assert len(d.children) == 2
+        assert d.children[0] is prob_y
+        assert d.children[1] is prob_x
+
+
+class TestFrontdoorLatex:
+    """
+    Frontdoor formula: sum_M [ P(M|X) * sum_X [ P(Y|M,X) P(X) ] ]
```
DARHWOLF: please cross check this. This does not seem correct. It was correct earlier.
From my previous comments, I was referring to updating the incorrect latex output.

pgmpy/tests/test_identification/test_probability_expression.py
```diff
@@ -0,0 +1,311 @@
...
+        assert r"\right]" in latex
+
+    def test_tree_structure(self, frontdoor_expr):
+        assert isinstance(frontdoor_expr.root, MarginalNode)
+        assert frontdoor_expr.root.sumset == frozenset({"M"})
+
+
+class TestBowArcLatex:
+    """
+    Bow-arc formula: sum_Z [ P(Z|X) * sum_X [ P(Y|Z,X) P(X) ] ]
```
DARHWOLF: is this correct?

pgmpy/tests/test_identification/test_probability_expression.py
DARHWOLF: I don't think this has been addressed yet. I still see double slashes and an incorrect formula being generated. @ankurankan could you please cross check.

DARHWOLF (commented):
Please refer comments

pgmpy/identification/probability_expression.py
ankurankan, line 1: Please set your editor to use 120 character line wrap

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+
+    Examples
+    --------
+    >>> ProbabilityNode(frozenset({"Y"})).to_latex()
+    'P(Y)'
+
+    >>> ProbabilityNode(frozenset({"Y"}), cond=frozenset({"X"})).to_latex()
+    'P(Y \\mid X)'
+
+    >>> ProbabilityNode(frozenset({"Y"}), do=frozenset({"X"})).to_latex()
```
ankurankan, line 84: The import statements are missing. The code in `Examples` block needs to run independently; doctests test for that.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+
+        Returns
+        -------
+        str
+            LaTeX string of the form ``P(variables \mid do(...), cond)``.
+
+        Examples
+        --------
+        >>> ProbabilityNode(frozenset({"Y"}), do=frozenset({"X"}), cond=frozenset({"Z"})).to_latex()
+        'P(Y \\mid do(X), Z)'
```
ankurankan, line 116: I would recommend returning raw strings here. It would be easier to both copy-paste and to pass it to any latex rendering tool.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+            conditioning_parts.append("do(" + self._vars_to_latex(self.do) + ")")
+        if self.cond:
+            conditioning_parts.append(self._vars_to_latex(self.cond))
+
+        if conditioning_parts:
+            inner = variables_str + r" \mid " + ", ".join(conditioning_parts)
+        else:
+            inner = variables_str
+
+        return "P(" + inner + ")"
```
ankurankan: Would recommend working only with raw strings in this method and then return that directly without converting them to normal string.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+        if self.cond:
+            parts.append(f"cond={set(self.cond)!r}")
+        return "ProbabilityNode(" + ", ".join(parts) + ")"
+
+
+class MarginalNode(_TreeNode):
+    r"""
+    Internal node representing a marginalisation over a sub-expression::
+
+        sum_{sumset} child
```
ankurankan: I would suggest writing this in latex. You can use the `math` directive. Refer numpydoc documentation.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+        ``children[0]`` is the child expression.
+
+    Examples
+    --------
+    >>> m = MarginalNode(
+    ...     ProbabilityNode(frozenset({"Y"}), cond=frozenset({"X"})),
+    ...     sumset=frozenset({"X"}),
+    ... )
+    >>> m.to_latex()
+    '\\sum_{X} P(Y \\mid X)'
```
ankurankan, line 184: Import statement is missing.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+    """
+
+    def __init__(self, child, sumset):
+        self.sumset = frozenset(sumset)
+        self.children = [child]
+
+    def to_latex(self):
+        """
+        Return LaTeX for ``\\sum_{sumset} child``.
```
ankurankan: Use `math` directive here as well for the equation part.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def __repr__(self):
+        return f"MarginalNode(child={self.children[0]!r}, sumset={set(self.sumset)!r})"
+
+
+class ProductNode(_TreeNode):
+    r"""
+    Internal node representing a product of two or more sub-expressions::
+
+        factors[0] * factors[1] * ... * factors[n]
```
ankurankan: This needs to be improved. We don't know what `factors` mean here.

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+
+    def __repr__(self):
+        return "ProductNode([" + ", ".join(repr(c) for c in self.children) + "])"
+
+
+class DivisionNode(_TreeNode):
+    r"""
+    Internal node representing a ratio of two sub-expressions::
+
+        numerator / denominator
```
ankurankan: I would suggest writing this in equation form as well and showing the actual representation for example: `P(X_1 | do(Y_1), Z_1) / P(X_2 | do (Y_2), Z_2)`

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,393 @@
...
+        ``root.children``.
+
+    Notes
+    -----
+    The ``to_latex()`` method delegates to ``self.root.to_latex()`` and
+    returns a complete LaTeX representation suitable for math environments.
+    Rendering proceeds bottom-up: leaf ``ProbabilityNode`` nodes render
+    themselves and internal nodes combine their children's strings using
+    the appropriate notation (e.g. ``\sum``, ``\frac``, brackets).
+
```
ankurankan: I don't think we need to mention this.

pgmpy/tests/test_identification/test_probability_expression.py
ankurankan: Yes, I think these would be used mainly internally, and users shouldn't need to import them manually.

pgmpy/tests/test_identification/test_probability_expression.py
```diff
@@ -0,0 +1,311 @@
...
+        ]
+    )
+    inner_sum = MarginalNode(inner_product, sumset=frozenset({"X"}))
+    outer_product = ProductNode(
+        [
+            ProbabilityNode(frozenset({"Z"}), cond=frozenset({"X"})),
+            inner_sum,
+        ]
+    )
+    return ProbabilityExpressionTree(root=MarginalNode(outer_product, sumset=frozenset({"Z"})))
```
ankurankan, line 72: This test is very similar to the previous one. Can we replace this with something that also has a do statement? Or just modify this to add some do variables.

pgmpy/tests/test_identification/test_probability_expression.py
```diff
@@ -0,0 +1,311 @@
...
+    return types
+
+
+def find_leaves(node):
+    if isinstance(node, ProbabilityNode):
+        return [node]
+    leaves = []
+    for child in node.children:
+        leaves.extend(find_leaves(child))
+    return leaves
```
ankurankan: These helper methods should be methods in the `ProbabilityExpressionTreee` class itself.

pgmpy/tests/test_identification/test_probability_expression.py
```diff
@@ -0,0 +1,311 @@
...
+        assert len(find_leaves(frontdoor_expr.root)) == 3
+        assert len(find_leaves(bowarc_expr.root)) == 3
+
+        d = DivisionNode(prob_y, prob_x)
+        assert len(d.children) == 2
+        assert d.children[0] is prob_y
+        assert d.children[1] is prob_x
+
+
+class TestFrontdoorLatex:
```
ankurankan: I would suggest keeping this and the next test together with the probability expression tests for these two equations.

pgmpy/tests/test_identification/test_probability_expression.py
ankurankan, line 1: I am seeing that you have created class for most of the tests even if they only have a single test method. The idea of test classes is to group things together. For example, all tests for the frondoor equation can be in a single class and then you can have methods to test different things such as whether the expression matches, or `to_latex` method is working correctly on it, etc. Essentially, the test class is a way to group tests together so that it is easier for us to read it later.

ankurankan (changes requested):

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,448 @@
...
+        Always ``[]``. ``ProbabilityNode`` is always a leaf node.
+
+    Examples
+    --------
+    >>> from pgmpy.identification.probability_expression import ProbabilityNode
+    >>> ProbabilityNode(frozenset({"Y"})).to_latex()
+    'P(Y)'
+    >>> ProbabilityNode(frozenset({"Y"}), cond=frozenset({"X"})).to_latex()
+    'P(Y \\mid X)'
+    >>> ProbabilityNode(frozenset({"Y"}), do=frozenset({"X"})).to_latex()
```
ankurankan, line 84: If we are returning raw strings now, does this need to be updated?

pgmpy/identification/probability_expression.py
```diff
@@ -0,0 +1,448 @@
...
+
+    do : frozenset of str, optional
+        Variables being intervened on via the do-operator.
+        Example: ``frozenset({"X"})`` renders as ``do(X)`` inside the conditioning bar.
+        Default: ``frozenset()``.
+
+    cond : frozenset of str, optional
+        Variables being passively conditioned on.
+        Example: ``frozenset({"Z"})`` renders after the conditioning bar.
+        Default: ``frozenset()``.
```
ankurankan, line 65: I think in our case, variables can be any hashable object not just str

ankurankan (dismissed):

pgmpy/identification/probability_expression.py
kajal-jotwani: No, doesn't need updating. The `repr()` shown here always doubles the backslash that's just Python's display convention, not the actual latex output I just checked
```
Ctrl click to launch VS Code Native REPL
>>> from pgmpy.identification.probability_expression import ProbabilityNode
>>> result = ProbabilityNode(frozenset({"Y"}), cond=frozenset({"X"})).to_latex()
>>> result
'P(Y \\mid X)'
>>> print(result)
P(Y \mid X)
>>> 
```

pgmpy/identification/probability_expression.py
kajal-jotwani: Done! I updated the docstrings to accept any hashable. I also updated the `_vars_to_latex` helper to cast these hashables to strings before sorting

DARHWOLF (approved):

ankurankan (approved):

PR #3462, opened 2026-06-11
02. Implement `NominalDistribution`

### LINKED ISSUE #3454: 02. Implement `NominalDistribution`
It is child issue of #3452 for tracking.

### Proposal Solutions

- Implement `CategoricalDistribution` class.
- This class follow skpro's distribution templates and tag system.

```py
# pgmpy/distribution/categorical.py
from skpro.distributions.base import BaseDistribution

class CategoricalDistribution(BaseDistribution):
    _tags = {
        # packaging info
        # --------------
        "python_version": None,
        # estimator tags
        # --------------
        "python_dependencies": [],
        "distr:measuretype": "discrete",
        "distr:paramtype": "parametric",
        "capabilities:approx": [],
        "capabilities:exact": ["mean", "var", "log_pmf", "pmf", "cdf"],
        "broadcast_init": "on",
    }

    def _log_pmf(self, y):
        ...
        return np.ndarray # 2D

```

- [extension_templates/distributions.py](https://github.com/sktime/skpro/blob/main/extension_templates/distributions.py)

### Alternative Solutions
- The `distribution` class has sklearn-style tags:
    - we designed the `distribution` object to follow the skpro structure.
- Implement this class directly in skpro lib. (https://github.com/sktime/skpro/issues/1003)

- ENH proposal link: https://github.com/pgmpy/enhancement_proposals/pull/12

daehyun99:

- Fixes #3454

- Implement `CategoricalDistribution`
- Implement test code
- Add `skpro` dependency


daehyun99:
1. Because `mean()` and `var()` cannot be computed when `state_name` is a string, these methods were not implemented.

2. In `skpro`, the output shape of a method must match `dist.shape`.
However, when this setting is `"on"`, `dist.shape` is fixed to `(index, state_names)`.
To ensure that `dist.shape` is defined as `(index, columns)`, the setting should be changed to `"off"`.

3. Adding validation in `__init__()` to check whether each row of `values` sums to 1 causes an error during `skpro`’s `check_estimator()` process.
Therefore, the validation was omitted to maintain compatibility with the `skpro` specification.
Instead, this responsibility could be handled by adding the corresponding validation logic to `TabularCPD.predict_proba()`.

pyproject.toml
```diff
@@ -37,6 +37,7 @@ dependencies = [
   "joblib>=1.2",
   "opt_einsum>=3.3",
   "scikit-base>=0.12.4",
+  "skpro>=2.13.0",
```
daehyun99: 1. Each parameter return skpro's distribution.
So, I think `skpro`'s dependencies is not optional.

2. I just set the `skpro`'s version working in my env.
Are the role exist setting version?

pgmpy/distributions/categorical.py
```diff
@@ -0,0 +1,339 @@
...
+    def __init__(self, values, state_names, index=None, columns=None):
+
+        self.values = values
+        self.state_names = state_names
+
+        if len(state_names) != len(set(state_names)):
+            raise ValueError("state_names must not contain duplicate values")
+
+        if len(values[0]) != len(state_names):
+            raise ValueError("mismatch values and state_names's shape")
```
DARHWOLF: can you improve the error message here? perhaps something like:-
"mismatch between the shape of state_names and values"

pgmpy/distributions/categorical.py
```diff
@@ -0,0 +1,339 @@
...
+        if index is None:
+            index = pd.RangeIndex(len(values))
+        elif len(index) != len(values):
+            raise ValueError("wrong index's len")
+
+        if columns is None:
+            columns = ["variable"]
+        elif len(columns) != 1:
+            raise ValueError("wrong columns's len")
+
```
DARHWOLF: Could you improve the error messages here as well

pgmpy/distributions/categorical.py
```diff
@@ -0,0 +1,339 @@
...
+        super().__init__(index=index, columns=columns)
+
+    def _pmf(self, x):
+        """Probability mass function.
+
+        Parameters
+        ----------
+        x : 2D np.ndarray, same shape as ``self``
+            values to evaluate the pmf at
```
DARHWOLF: can you cross check if this has been cut?
"values to evaluate the pmf at the given points"

pgmpy/distributions/categorical.py
```diff
@@ -0,0 +1,339 @@
...
+
+    def __init__(self, values, state_names, index=None, columns=None):
+
+        self.values = values
+        self.state_names = state_names
+
+        if len(state_names) != len(set(state_names)):
+            raise ValueError("state_names must not contain duplicate values")
+
+        if len(values[0]) != len(state_names):
```
DARHWOLF: would it be a good idea to validate the values to ensure they are non-negative?

pgmpy/tests/test_distributions/test_categorical.py
```diff
@@ -0,0 +1,475 @@
...
+
+        # Case 10: wrong columns
+        values = [[0.1, 0.2, 0.8], [0.5, 0.3, 0.2]]
+        state_names = [1, 2, 3]
+        columns = ["A", "B", "C"]
+
+        with pytest.raises(ValueError):
+            dist = CategoricalDistribution(values=values, state_names=state_names, columns=columns)
+
+        # Case 13: worng shape(values, state_names)
```
DARHWOLF: looks like a typo here

pgmpy/tests/test_distributions/test_categorical.py
```diff
@@ -0,0 +1,475 @@
...
+        # # NOTE: An error occurs when `state_names` contains strings.
+        # values = [[0.1, 0.9], [0.7, 0.3]]
+        # dist = CategoricalDistribution(values=values, state_names=["A", "B"])
+        # check_estimator(dist, raise_exceptions=True, verbose=False)
+
+    def test_init(self):
+        """test"""
+        # Case 0: default
+        values = [[0.1, 0.2, 0.7], [0.5, 0.3, 0.2]]
+        state_names = [1, 2, "C"]
```
DARHWOLF: Having these mixed state_names isn't working in the case of pmf. I suspect it is because of the `np.asarray(state_names)` conversion.

DARHWOLF (changes requested):
Nice Work! 

Can you implement tests for `pmf`, `cdf`, `ppf` and `sample` and feed them with mixed-type `state_names`. 

I suspect there is a bug here because of the np.asarray conversion.

daehyun99:
@DARHWOLF 
Thank you for your review.

I apply your review.
please review when you have a chance.

pgmpy/distributions/categorical.py
```diff
@@ -0,0 +1,399 @@
...
+
+        res = np.empty(probs.shape[0], dtype=float)
+
+        for i, xi in enumerate(x):
+            target = xi[0]
+
+            idx_arr = np.where(categories == target)[0]
+
+            if len(idx_arr) == 0:
+                res[i] = 0.0
```
DARHWOLF: are you sure this is correct?

daehyun99: Yes. Something i missed?

when `x` value is not exist in `categories`; then, i believe the `cdf` values should be `0.0`.

```py
probs = [[0.1, 0.2, 0.7], [0.5, 0.3, 0.2]]
categories = [1, 2, 3]
x = [[1], [4]]

dist = CategoricalDistribution(probs=probs, categories=categories)

expected = pd.DataFrame({"variable": [0.1, 0.0]})
try:
    pd.testing.assert_frame_equal(dist.cdf(x), expected)
    print("Success")
except:
    print("Fail")

```

in `target==4` case, `categories == target` would be `[False, False, False]` and `idx_arr` would be `[]`

pgmpy/distributions/nominal.py
ankurankan, line 1: We use 120 character line limit. You can change your editor setting to wrap lines on 120 char.

pgmpy/distributions/categorical.py
```diff
@@ -0,0 +1,399 @@
...
+        Returns
+        -------
+        2D np.ndarray
+
+        """
+        probs = self.probs
+        categories = self.categories
+
+        probs = np.asarray(probs, dtype=float)
+        categories = np.asarray(categories)
```
ankurankan: We can remove 114 and 115 and use `self.probs` directly in 117.

ankurankan:
@daehyun99 I have pushed some changes in this PR for it to represent only nominal / categorical distributions. Your original implementation was a mix of nominal and ordinal variables. I think we should keep those two separate and maybe even rename this to `NominalDistribution`. Some of the test cases that I have pushed can be removed.

daehyun99:
@ankurankan 
> Your original implementation was a mix of nominal and ordinal variables.

Oh, yes. I missed that.

I will rename of dist to `NominalDistribution`.

---

The below one is intentional. right?
And I have a one suggestion

if we set up the seed(`rng = np.random.default_rng(self.random_state)`) in `_sample()`.
then, I think it is always determestic.
I think if user repeat the `cpd.sample(n)` code. then, the result always return equal value.

So, What do you think about that set up the seed(`self.rng = np.random.default_rng(self.random_state)`) once in `__init__()` method?

image

fkiraly (changes requested):
I think this (and categorical distributions in general) would be a great addition to `skpro`!

My main comments are as follows:
* it might be better to split `BaseDistribution` into `BaseNominalDistribution` and `BaseRealDistribution` eventually.
(taking into account that eventually ordinal or other exotics might also come into play)
* unlike other distributions in `skpro`, this does not allow multiple variables. I think representing 2D array-like distributions has been the main trouble in `skpro` when attacking this issue, so it might be worth thinking about that too. E.g., would `probs` be a list? Would `categories` be a list of 1D arrays? (and in the univariate case, the list can be omitted?)

daehyun99:
@fkiraly 

> it might be better to split BaseDistribution into BaseNominalDistribution and BaseRealDistribution eventually.

Yes, I agree.

> I think representing 2D array-like distributions has been the main trouble in skpro when attacking this issue

Quick question:

If we simply consider the case where all variables are discrete, couldn’t this be represented as a joint probability distribution?

Referring to `pgmpy`’s `JointProbabilityDistribution`, it looks like the following:

```py
"""
Examples
--------
>>> import numpy as np
>>> from pgmpy.factors.discrete import JointProbabilityDistribution
>>> prob = JointProbabilityDistribution(
...     variables=["x1", "x2", "x3"],
...     cardinality=[2, 2, 2],
...     values=np.ones(8) / 8,
... )
>>> print(prob)
+-------+-------+-------+---------------+
| x1    | x2    | x3    |   P(x1,x2,x3) |
+=======+=======+=======+===============+
| x1(0) | x2(0) | x3(0) |        0.1250 |
+-------+-------+-------+---------------+
| x1(0) | x2(0) | x3(1) |        0.1250 |
+-------+-------+-------+---------------+
| x1(0) | x2(1) | x3(0) |        0.1250 |
+-------+-------+-------+---------------+
| x1(0) | x2(1) | x3(1) |        0.1250 |
+-------+-------+-------+---------------+
| x1(1) | x2(0) | x3(0) |        0.1250 |
+-------+-------+-------+---------------+
| x1(1) | x2(0) | x3(1) |        0.1250 |
+-------+-------+-------+---------------+
| x1(1) | x2(1) | x3(0) |        0.1250 |
+-------+-------+-------+---------------+
| x1(1) | x2(1) | x3(1) |        0.1250 |
+-------+-------+-------+---------------+
"""
```

If a `NominalDistribution` is represented by a variable `A` with 2 states and a variable `B` with 3 states, it could take a form like the following.
In that case, I think it could have `probs` and `categories` as 1D arrays.

```py
NominalDistribution(
    proba = [
        0.1, 0.1, 0.1,
        0.1, 0.1, 0.5,
    ],
    categories = [
        "a1_b1" ,"a1_b2", "a1_b3",
        "a2_b1" ,"a2_b2", "a2_b3",
    ],
    columns =[
        "A_B"
    ]
)
```

Issue #3450, opened 2026-06-09
[ENH] Bootstrap Estimator for Causal Discovery


Jatinbhardwaj-093:
### Is your feature request related to a problem? Please describe.
Causal discovery estimators in `pgmpy` (e.g., `HillClimbSearch`, `PC`) output a single graph that is highly sensitive to noise, small sample size, or sampling variability. There is currently no way to evaluate the stability or confidence of the learned edges and orientations.

### Describe the solution you'd like
Adding a new `BootstrapEstimator` wrapper class under `pgmpy.causal_discovery` to support non-parametric bootstrap estimation.

**Implementation Scope:**
- Subclass `_BaseCausalDiscovery` and wrap a base causal discovery estimator.
- Run the estimator over `n_resamples` with `sample_frac` data fraction (parallelized via `n_jobs`).
- Store resampled row indices (via `bootstrap_samples_`) for memory efficiency.
- Generate stability statistics (`edge_prob_` and `direction_prob_`).
- Implement `get_consensus_graph()` with a greedy cycle-prevention algorithm (sorting edges by probability and iteratively adding them only if they do not introduce cycles) to ensure the consensus graph is a DAG.

Example API usage:
```python
from pgmpy.causal_discovery import BootstrapEstimator, HillClimbSearch

hc = HillClimbSearch()
est = BootstrapEstimator(estimator=hc, n_resamples=100, threshold=0.5, n_jobs=-1)
est.fit(data)

# Retrieve valid DAG consensus graph & probabilities
consensus_dag = est.get_consensus_graph(threshold=0.8)
print(est.edge_prob_)
```

### Describe alternatives you've considered
**Describe alternatives you've considered**

- **Parametric Bootstrap**: Requires a correctly specified parametric model, making it less robust than the non-parametric approach.
- **Jackknife Resampling**: Mentions omitting one sample at a time, but lacks the parameter flexibility of bootstrap resampling.

### Additional context
- **Reference Paper**: [Causal Discovery with Resampling-Based Confidence Measures](https://arxiv.org/pdf/1301.6695)
- **Original Issue**: #3326
- **Proposal for details**: [draft.md](https://github.com/pgmpy/enhancement_proposals/pull/9)

PR #3464, opened 2026-06-11
Non-Parametric Bootstrap Estimator for Causal Discovery.

Jatinbhardwaj-093:
Adds BootstrapEstimator, a non-parametric bootstrap meta-estimator under pgmpy.causal_discovery to quantify edge and direction confidence in aggrregated bootstrap causal graphs.

`Why is this needed??`
Current structure learning algorithms (like PC or HillClimbSearch) are highly sensitive to the specific dataset they run on; even slight changes or noise in the data can result in an entirely incorrect causal structure. pgmpy currently lacks a native way to evaluate the stability or confidence of learned edges and orientations.

- [ ] Are you able to fully explain your changes? We expect you to fully understand the algorithm and take full responsibility for any changes in this PR.

- Issue: https://github.com/pgmpy/pgmpy/issues/3450
- Full Draft: https://github.com/pgmpy/enhancement_proposals/blob/main/6_bootstrap_causal_discovery/draft.md

* **`BootstrapEstimator` Wrapper:** Subclasses `_BaseCausalDiscovery` and wraps any base causal discovery estimator.
* **Efficient Parallelization:** Runs across `n_resamples` using `n_jobs`, storing resampled indices (`bootstrap_samples_`) instead of raw data to keep memory usage low.
* **Stability Metrics:** Exposes `edge_prob_` and `direction_prob_` to track the empirical frequency of learned features.
* **Optimal Consensus DAG:** Employs a greedy approach to construct a valid PDAG and DAG by leveraging `threshold`, `edge_prob_`, and `direction_prob_`.


pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,303 @@
...
+    show_progress : bool, default=True
+        If True, shows a progress bar while fitting the bootstrap estimators.
+
+    seed : int, default=None
+        Seed for the random number generator to ensure reproducibility.
+
+    Attributes
+    ----------
+    causal_graph_ : DAG or PDAG
+        The learned robust consensus causal graph. The reuturn type will be of
```
ankurankan: Typo: `reuturn` -> `return`

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,303 @@
...
+        base_estimator: BaseCausalDiscovery,
+        bootstrap_sample_size: int,
+        seed: int | np.random.SeedSequence,
+    ) -> tuple[list[int], BaseCausalDiscovery]:
+        """Helper function to run a single bootstrap iteration."""
+
+        # Create new sample by resampling
+        rng = np.random.default_rng(seed)
+        row_indices = list(rng.choice(X.index, size=bootstrap_sample_size, replace=True))
+        sample = X.iloc[row_indices]
```
ankurankan: This will fail if the index is not the same as element position. `index` can be anything but `iloc` will start counting from 0.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Fit the resample data on base estimator.
+        base_estimator = cast(BaseCausalDiscovery, clone(base_estimator))
```
ankurankan, line 212: Why do we need `cast` here?

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,303 @@
...
+        X: pd.DataFrame,
+        base_estimator: BaseCausalDiscovery,
+        bootstrap_sample_size: int,
+        seed: int | np.random.SeedSequence,
+    ) -> tuple[list[int], BaseCausalDiscovery]:
+        """Helper function to run a single bootstrap iteration."""
+
+        # Create new sample by resampling
+        rng = np.random.default_rng(seed)
+        row_indices = list(rng.choice(X.index, size=bootstrap_sample_size, replace=True))
```
ankurankan: I would suggest computing all the row_indices in the main method in a single `choice` call. You can specify the size as: `(n_bootstrap, bootstrap_sample_size)` and then pass one row of that array to each process. This should be faster as it reduces the number of function calls.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,303 @@
...
+            raise ValueError("estimator must be an instance of BaseCausalDiscovery Class.")
+
+        N = self.n_features_in_
+        variables = self.feature_names_in_
+
+        self.bootstrap_samples_ = list()
+        self.bootstrap_graphs_ = list()
+
+        seed_seq = np.random.SeedSequence(self.seed)
+        child_seeds = seed_seq.spawn(self.n_bootstraps)
```
ankurankan: This will also get simplified if we generate all the indexes in a single call in this method. Refer comment in the `_bootstrap_iteration` method.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,303 @@
...
+        )
+
+        undirected_counts = pd.DataFrame(
+            np.zeros((N, N)),
+            index=variables,
+            columns=variables,
+        )
+
+        # Step 1: Run bootstrap iterations
+        results = cast(
```
ankurankan: Why do we need to cast here? When would the returned values be of a different type?

ankurankan (changes requested):

pgmpy/causal_discovery/BootstrapEstimator.py
Jatinbhardwaj-093: The `clone` function from sklearn returns `BaseEstimator` as its generic type, and we are adopting `base_estimator` as the type for the pgmpy `BaseCausalDiscovery` class.

pgmpy/causal_discovery/BootstrapEstimator.py
Jatinbhardwaj-093: `Parallel` simplifies the return type to just `list[Any]`. Casting here helps with type checking later in the loop, making it more specific.

> When would the returned values be of a different type?

I actually don't get what you mean by this.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,303 @@
...
+
+                # Add the edge and check for cyclicity
+                if target == "undirected":
+                    pdag.add_edge(u, v)
+                    pdag.add_edge(v, u)
+                    pdag.calibrate_directed_undirected_edges()
+                    if not pdag.has_acyclic_extension():
+                        pdag.remove_edge(u, v)
+                        pdag.remove_edge(v, u)
+                        pdag.calibrate_directed_undirected_edges()
```
Jatinbhardwaj-093: For undirected edges in a PDAG, should I remove the acyclicity test? We actually discuss this in our 1-on-1.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,354 @@
...
+
+        undirected_counts = pd.DataFrame(
+            np.zeros((N, N)),
+            index=variables,
+            columns=variables,
+        )
+
+        rng = np.random.default_rng(self.seed)
+        bootstrap_sample_size = int(len(X) * self.sample_size)
+
```
ankurankan: Step 1 is missing.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,354 @@
...
+        for _ in range(self.n_bootstraps):
```
ankurankan: `rng.choice(size=(bootstrap_sample_size, len(X)),  replace=True)`

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,354 @@
...
+            bootstrap_sample = list(rng.choice(len(X), size=bootstrap_sample_size, replace=True))
+            self.bootstrap_samples_.append(bootstrap_sample)
+
+        results = cast(
+            list[BaseCausalDiscovery],
+            Parallel(n_jobs=self.n_jobs)(
+                delayed(self._bootstrap_iteration)(X, self.estimator, self.bootstrap_samples_[i])
```
ankurankan: `self.bootstrap_samples_[i]` -> `self.bootstra_samples_[i, :]`

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,354 @@
...
+                    desc="Bootstrapping",
+                    disable=not (self.show_progress and config.SHOW_PROGRESS),
+                )
+            ),
+        )
+
+        # Step 2: Aggregating the bootstrap results.
+        for est in results:
+            causal_graph = est.causal_graph_
+            adjacency_matrix = est.adjacency_matrix_
```
ankurankan: An idea here could be to represent the set of adjacency matrices using a 3D array (each 2D adjacency matrix is stacked, the size would be `(boostrap_sample_size, N, N)`. Then `edge_presence = np.sum(adjacency_matrices, axis=0)`

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def __init__(
+        self,
+        estimator: BaseCausalDiscovery,
+        n_bootstraps: int = 10,
+        sample_size: float = 1,
+        threshold: float = 0.5,
+        show_progress: bool = True,
+        seed: int | None = None,
+        n_jobs: int = -1,
```
ankurankan: Should add another argument called `warm_start`. If `warm_start=True`, use the previously fitted samples and add new bootstrapped graphs to it.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,354 @@
...
+            dag = DAG()
+            dag.add_nodes_from(variables)
+
+            for u, v, _ in candidate_edges:
+                if not nx.has_path(dag, v, u):
+                    dag.add_edge(u, v)
+
+            return dag
+
+    def get_causal_graph(self, threshold: float) -> DAG | PDAG:
```
ankurankan: `get_causal_graph` -> `get_consensus_graph`.

pgmpy/tests/test_causal_discovery/test_BootstrapEstimator.py
```diff
@@ -0,0 +1,226 @@
...
+    return data.astype("category")
+
+
+def test_bootstrap_rand_data(rand_data):
+    """
+    Tests edge discovery on 5-node causal graph:
+    A -> C <- B, C -> D -> E, B -> E
+    """
+    hc = BootstrapEstimator(estimator=HillClimbSearch(return_type="dag"), show_progress=False, seed=42)
+    pc = BootstrapEstimator(estimator=PC(), show_progress=False, seed=42)
```
ankurankan: Just use one estimator in tests.

pgmpy/tests/test_causal_discovery/test_BootstrapEstimator.py
```diff
@@ -0,0 +1,226 @@
...
+        ("lung", "smoke"),
+    }
+
+    assert {tuple(sorted(edge)) for edge in est_pc.causal_graph_.undirected_edges} == {
+        ("bronc", "dysp"),
+        ("bronc", "smoke"),
+    }
+
+    assert ((est_pc.edge_prob_ > 0.0) & (est_pc.edge_prob_ < 1.0)).any().any()
+
```
ankurankan: Just test with one estimator.

pgmpy/tests/test_causal_discovery/test_BootstrapEstimator.py
```diff
@@ -0,0 +1,226 @@
...
+    # Test with HillClimbSearch
+    base_est = HillClimbSearch(return_type="dag")
+    est_hc = BootstrapEstimator(
+        estimator=base_est,
+        n_bootstraps=6,
+        sample_size=0.75,
+        threshold=0.35,
+        seed=42,
+    )
+    est_hc.fit(df)
```
ankurankan: Don't use the alarm model in tests. It will take too much time.

pgmpy/tests/test_causal_discovery/test_BootstrapEstimator.py
```diff
@@ -0,0 +1,226 @@
...
+    hc.fit(rand_data)
+    pc.fit(rand_data)
+
+    hc_graph = hc.causal_graph_
+    pc_graph = pc.causal_graph_
+
+    expected_edges_hc = {("C", "B"), ("C", "D"), ("D", "E"), ("C", "E"), ("B", "E"), ("A", "C")}
+    expected_edges_pc = {("B", "C"), ("C", "D"), ("D", "E"), ("B", "E"), ("A", "C")}
+    assert set(hc_graph.edges()) == expected_edges_hc
+    assert set(pc_graph.edges()) == expected_edges_pc
```
ankurankan: Test for all of the attributes that are set in the fit method.

pgmpy/tests/test_causal_discovery/test_BootstrapEstimator.py
```diff
@@ -0,0 +1,226 @@
...
+    est = BootstrapEstimator(estimator=HillClimbSearch(return_type="dag"), show_progress=False)
+    est.fit(rand_data)
+    for invalid_threshold in [-0.5, 1.5]:
+        with pytest.raises(ValueError):
+            est.get_causal_graph(threshold=invalid_threshold)
+        with pytest.raises(ValueError):
+            est.get_adjacency_matrix(threshold=invalid_threshold)
+
+
+def test_bootstrap_threshold_variation():
```
ankurankan: Tests for `get_causal_graph` and `get_adjacency_matrix` should test that the returned graphs/values are correct for the specified threshold.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,501 @@
...
+        direction u -> v, given that an edge exists between u and v.
+
+        Calculated as n1 / (n1 + n2 + n3), where:
+        - n1: Number of bootstraps where u -> v is directed (only u -> v).
+        - n2: Number of bootstraps where v -> u is directed (only v -> u).
+        - n3: Number of bootstraps where u -- v is undirected (both exist).
+
+        Both keys (u, v) and (v, u) are present for any connected pair. The
+        probability of the undirected edge u - v is
+        1 - direction_prob_[(u, v)] - direction_prob_[(v, u)].
```
ankurankan: Shouldn't we have this in an adjacency matrix format as well?

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,501 @@
...
+
+    bootstrap_samples_ : np.ndarray
+        2D array containing row indices used in each bootstrap sample.
+
+    bootstrap_graphs_ : np.ndarray
+        3D array containing adjacency matrices learned from each bootstrap.
```
ankurankan: Please define which axis represents what. Is the first axis representing each graph?

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,501 @@
...
+        - n1: Number of bootstraps where u -> v is directed (only u -> v).
+        - n2: Number of bootstraps where v -> u is directed (only v -> u).
+        - n3: Number of bootstraps where u -- v is undirected (both exist).
+
+        Both keys (u, v) and (v, u) are present for any connected pair. The
+        probability of the undirected edge u - v is
+        1 - direction_prob_[(u, v)] - direction_prob_[(v, u)].
+
+    bootstrap_samples_ : np.ndarray
+        2D array containing row indices used in each bootstrap sample.
```
ankurankan: Here as well. Please define whether the row is the indexes or the columns.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,501 @@
...
+    --------
+    # Simulate dataset and fit BootstrapEstimator with HillClimbSearch:
+    >>> from pgmpy.causal_discovery import BootstrapEstimator, HillClimbSearch
+    >>> from pgmpy.example_models import load_model
+    >>> data = load_model("bnlearn/asia").simulate(n_samples=500, seed=42)
+    >>> hc = HillClimbSearch(return_type="dag")
+    >>> est = BootstrapEstimator(hc, seed=42, show_progress=False)
+    >>> est = est.fit(data)
+    >>> isinstance(est.causal_graph_, DAG)
+    True
```
ankurankan: Show other attributes here as well.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,501 @@
...
+        if not isinstance(self.estimator, BaseCausalDiscovery):
+            raise ValueError("estimator must be an instance of BaseCausalDiscovery Class.")
+
+        variables = self.feature_names_in_
+
+        # Step 1: Resample dataset and fit base estimators
+        rng = np.random.default_rng(self.seed)
+        bootstrap_sample_size = int(len(X) * self.sample_size)
+
+        if self.warm_start and hasattr(self, "bootstrap_samples_"):
```
ankurankan, line 248: I think we can simplify this quite a bit. The only thing `warm_start` needs to modify is the data structure initialization (where we store our results). We do not really need repeated code here.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,501 @@
...
+                    new_graph_matrices.append(adj_df.values)
+                new_graphs = np.array(new_graph_matrices)
+
+                # Append new samples and graphs onto existing warm start arrays
+                self.bootstrap_samples_ = np.concatenate([self.bootstrap_samples_, new_samples], axis=0)
+                self.bootstrap_graphs_ = np.concatenate([self.bootstrap_graphs_, new_graphs], axis=0)
+            else:
+                # Trim arrays if requested n_bootstraps is less than existing count
+                self.bootstrap_samples_ = self.bootstrap_samples_[: self.n_bootstraps]
+                self.bootstrap_graphs_ = self.bootstrap_graphs_[: self.n_bootstraps]
```
ankurankan: I think this behavior diverges from the sklearn behavior. This should throw a `ValueError`. Have a look at how sklearn handles this condition.

ankurankan (changes requested):
@Jatinbhardwaj-093 Pr looks mostly good. You will also need to add this class to the `pgmpy/docs` to make the docs render on the website.

pgmpy/causal_discovery/BootstrapEstimator.py
Jatinbhardwaj-093: Do I also need to show the results for the attributes, or just the usage?

pgmpy/causal_discovery/BootstrapEstimator.py
Jatinbhardwaj-093: We can definitely use an adjacency matrix here, but the main decision is how we want to handle the missing edges, as we can't mark them as zero. This will break the undirected edge frequency formula (1 - n1 - n2). Should I fill the missing edges with NaN, or would -1 be better?

pgmpy/causal_discovery/BootstrapEstimator.py
ankurankan: This should be results as well.

Jatinbhardwaj-093:
@ankurankan I have changed the code for adding edges to the PDAG according to the new implementation. Also, I made a few cleanups for the docs and code. 

This is ready for review.

pgmpy/causal_discovery/BootstrapEstimator.py
```diff
@@ -0,0 +1,545 @@
...
+        rng = np.random.default_rng(self.seed)
+        bootstrap_sample_size = int(len(X) * self.sample_size)
+
+        # Generate all required bootstrap sample index arrays
+        sample_indices = []
+        for _ in range(self.n_bootstraps):
+            sample_idx = rng.choice(len(X), size=bootstrap_sample_size, replace=True)
+            sample_indices.append(sample_idx)
+        all_samples = np.array(sample_indices)
+
```
Nimish-4, line 247: @ankurankan Not very sure, but could this part result in repeated calculations if `sample_size`=1? Will it need to be handled?

Jatinbhardwaj-093: Repeated calculations would only happen if replace=False with sample_size=1.0 (which would just permute full dataset). Since replacement is enabled, all bootstrap runs operate on different sample distributions.

Nimish-4 (dismissed):
Looks good to me.

Jatinbhardwaj-093:
@ankurankan Does this pull request require anything further?

Issue #3455, opened 2026-06-09
03. Implement `TabularCPD`, `LinearGaussianCPD`, `FunctionalCPD`


daehyun99:
It is child issue of #3452 for tracking.

- [ ] #3495
- [ ] #3496 
- [ ] #3497

### Proposal Solutions

```py
# pgmpy/parameterization/TabularCPD.py
class TabularCPD(BaseParameter):
    def __init__(
        self,
        variable_card,
        evidence_card=None,
        state_names=None,
        prior_type=None,
        equivalent_sample_size=10,
        pseudo_counts=None
    ):
    ...
    # Fitted attributes (sklearn convention)
    values_: np.ndarray             # (variable_card, prod(evidence_card))
    classes_: np.ndarray            # child state labels
    feature_names_in_: np.ndarray   # parent labels in column order (if known)
    is_fitted_: bool

    @classmethod
    def from_values(
        cls,
        variable_card,
        values,
        evidence_card=None,
        state_names=None,
        parent_order=None,
    ):
        return TabularCPD(...)

    def fit(self, X, y, sample_weight=None):
        return self

    def predict_proba(self, X):
        ...
        return CategoricalDistribution() # (len(X), variable_card)

    def sample(self, X, n_samples=1):
        dist = self.predict_proba(X)
        samples = dist.sample()
        return pd.Series(samples, ...)
```

```py
# pgmpy/parameterization/LinearGaussianCPD.py
class LinearGaussianCPD(BaseParameter):
    def __init__(self):
        ...

    beta_: np.ndarray               # length n_parents + 1; beta_[0] = intercept
    std_: float
    feature_names_in_: np.ndarray   # parent labels in beta_[1:] order (if known)

    @classmethod
    def from_values(cls, beta, std, parent_order=None):
        ...
        return LinearGaussianCPD()

    def fit(self, X, y, sample_weight=None):
        ...
        return self

    def predict_proba(self, X):
        ...
        return skpro.distributions.Normal(mu, sigma)

    def sample(self, X, n_samples=1):
        dist = self.predict_proba(X)
        samples = dist.sample()
        return pd.Series(samples, ...)

```

```py
# pgmpy/parameterization/FunctionalCPD.py
class FunctionalCPD(BaseParameter):
    def __init__(self):
        ...

    @classmethod
    def from_values(cls):
        ...
        return FunctionalCPD()

    def fit(self, X, y):
        ...
        return mu, sigma

```

### Alternative Solutions

[...]

PR #3472, opened 2026-06-16
03-01. Implement `TabularCPD`

### LINKED ISSUE #3495: 03-01. Implement `TabularCPD`
It is child issue of #3455 for tracking.

#### 1. Parameter Learning way
- Frequentist
    - MLE
- Bayesian
    - posterior
    - posterior mean
    - posterior median
    - MAP
- EM

#### 2. Conjugate prior distribution for bayesian estimate
- Beta
- Dirichlet

#### 3. Return distribution
- Frequentist
    - NominalDistribution
-  Bayesian
    - Beta
    - Dirichlet

daehyun99:

- Part of #3455
- Closes #3495

- This PR is based on #3462 , #3459  PRs.
- Feat: Implement `TabularCPD`'s code logic
- Test: Implement `TabularCPD`'s test code


pgmpy/parameter/TabularCPD.py
```diff
@@ -0,0 +1,128 @@
...
+    def _fit(self, X, y=None, sample_weight=None):
+        if not hasattr(self, "state_names_"):
+            self._label_binarizer = LabelBinarizer()
+            if y is None:  # if root node
+                self._label_binarizer.fit(X)
+            else:
+                self._label_binarizer.fit(y)
+            self.categories_ = self._label_binarizer.classes_
+
+        estimator_cls = _ESTIMATOR_REGISTRY[self.estimator.lower()]
```
daehyun99: Should we implement this so that CPD classes use a `parameter_estimator` through `get_parameter_estimator`, similar to how Causal Discovery loads CI test classes through `get_ci_test`?

https://github.com/pgmpy/pgmpy/blob/6f834cdb60be0392df16c645baa388886d7f6a25/pgmpy/ci_tests/_base.py#L271

This would be analogous to how estimators are configured for models in scikit-learn.

It would also improve maintainability by keeping the structure consistent with Causal Discovery.

pgmpy/parameter/TabularCPD.py
```diff
@@ -0,0 +1,125 @@
...
+        self.estimator_ = estimator_cls()
+        self.estimator_.fit(
+            X,
+            y,
+            sample_weight,
+            self.categories_,
+            prior_type=self.prior_type,
+            equivalent_sample_size=self.equivalent_sample_size,
+            pseudo_counts=self.pseudo_counts,
+        )
```
ankurankan: This is something that I would like to avoid. TabularCPD has 3 parameter learning methods currently:
1. DiscreteMLE: This fits the tabularCPD objects one by one.
2. BayesianEstimator: This also fits the CPD objects one by one.
3. Expectation Maximization (EM): This internally calls DiscreteMLE and runs a few iterations of fitting.

Because both DiscreteMLE and BayesianEstimator fit the CPD objects one by one, I think the logic for fitting one CPD should be in the `_fit` method here instead of being in the Estimator classes. The estimator class can then initialize all the TabularCPD objects for a given model and then call the fit methods on them, so act as an orchestrator to just initialize CPD objects and call the fit methods on them. Similarly, EM would then become an orchestrator which call the fit methods on the CPD objects multiple times while also doing some computation.

This would mean that we will have to:
1. Remove the `estimator` argument from the `__init__` method. Change all the default argument values to None. If none of these arguments are specified by the user, we do an MLE fit. If any of the arguments are specified, we do a Bayesian fit. 
2. We move the fitting logic from the Estimator classes to the `_fit` method here.

fkiraly (changes requested):
Nice!

I like the design idea of having conditional and unconditional in the same class.

However, there are currently *discrepancies* between:

* the pre-PR specification of `TabularCPD` which allows specification without data; and the post-PR version which does not
* the conditional and unconditional case, in the API proposed by this PR.

I think these must be resolved cleanly, or this will introduce unnecessary friction between inconsistent APIs.

### specification with / without data

At the root of this discrepancy is what the pre-PR vs post-PR `TabularCPD` represent. For simplicity, I will consider the post-PR object only in the unfitted case.

* the pre-PR `TabularCPD` represents symbolically the entire conditional distribution, that is, `P(B|A)` for conditioning rv `A` and conditioned rv `B`.
* the post-PR `TabularCPD` only allows to compute, for an concrete value conditioned upon, a symbolic representation of the unconditional distribution after conditioning. That is, a computational rule that takes `a` and produces a symbolic representation of the (unconditional!) distribution `P(B|A=a)`. That is, a computational (and not symbolic!) representation of the function `f: a\mapsto P(B|A=a)`.

### conditional vs unconditional use

If `y` is provided, the distribution estimated is that of `y`, conditioned on `X` (row-wise).

If `y` is not provided, the distribution estimated is that of `X`.

This is inconsistent between cases!

To be consistent here, in both cases, the distribution of `y` would have to be estimated - otherwise the mathematical semantics change, depending on which variables are present.

Inconsistencies like this have impact on downstream code, consider the case of estimation based on data.

Under the current API:

```python
for n in nodes:
    if is_root_node(n):
        X = Xes[n]
        y = None
    else:
        X = Xes[parent(n)]
        y = ys[n]
    dists[n] = tcpd.fit(X, y)
```

If in both cases th distribution of `y` is estimated:

```python
for n in nodes:
    if is_root_node(n):
        X = None
    else:
        X = Xes[parent(n)]
    y = ys[n]
    dists[n] = tcpd.fit(X, y)
```

or even sth like
```python
for n in nodes:
    X = Xes.get(parent(n), None)
    y = ys[n]
    dists[n] = tcpd.fit(X, y)
```

fkiraly:
Here is how I would, personally, resolve the discrepancies:

#### specification with/without data

* allow the same args as `TabularCPD` currently has. If they are provided, potential data in `_fit` gets ignored, and always the fixed distribution is produced.
* alternatively, separate the following two types of object types (with base classes): fixed conditional distributions; and conditional distribution fitters (aka `skpro`-like `BaseRegressor`/`BaseClassifier`). The unconditional distribution fitter case remains to be treated too then.

#### conditional vs unconditional use

Depending on the ultimate use case you are aiming for, I can see two options:

* move consistency to always estimate distribution of `y` (conditional or not). Change the defaulting such that `X` can be `None`.
* alternatively, introduce a parameter to `__init__` which determines whether `X` gets ignored or not, in fitting. Instead of `n_samples` in `predict_proba`, you simply use the length of `X`, without conditioning on it.

daehyun99:
@fkiraly 

> specification with/without data

Yes, I also think using `Classifier` / `Regression` would be better than representations like `TabularCPD` or `LinearGaussianCPD`.
We also provide a separate API that lets users set parameters directly, without learning from data.

```py
class TabularCPD(BaseParameter):

    ...

    def set_fitted_params(self, CPT, columns, categories, evidences, is_fitted):
        self.CPT_ = CPT
        self.categories_ = categories
        self.evidences_ = evidences
        self._is_fitted = is_fitted
        return self

```

> conditional vs unconditional use

I designed the API less from a “conditional vs unconditional” perspective, and more from a “supervised vs unsupervised” perspective.

As far as I know, `skpro` has so far mainly covered supervised regression models.
So for the unsupervised case, I referred to the APIs of scikit-learn’s `GaussianMixture` and `BayesianGaussianMixture`.
In scikit-learn mixture models (unsupervised), `y` is ignored.

On the other hand, the recently added `NaiveProbaRegressor` in `skpro` ignores `X`.

I see this as a difference in perspective, and I think it would be better to unify on one convention, whichever criterion we choose.
What do you think would be the best way to do this?

https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/mixture/_base.py

```py
class BaseMixture(DensityMixin, BaseEstimator, metaclass=ABCMeta):

    ...

    def fit(self, X, y=None):
        """Estimate model parameters with the EM algorithm.

        The method fits the model ``n_init`` times and sets the parameters with
        which the model has the largest likelihood or lower bound. Within each
        trial, the method iterates between E-step and M-step for ``max_iter``
        times until the change of likelihood or lower bound is less than
        ``tol``, otherwise, a ``ConvergenceWarning`` is raised.
        If ``warm_start`` is ``True``, then ``n_init`` is ignored and a single
        initialization is performed upon the first call. Upon consecutive
        calls, training starts where it left off.

        Parameters
        ----------
        X : array-like of shape (n_samples, n_features)
            List of n_features-dimensional data points. Each row
            corresponds to a single data point.

        y : Ignored
            Not used, present for API consistency by convention.

        Returns
        -------
        self : object
            The fitted mixture.
        """
        # parameters are validated in fit_predict
        self.fit_predict(X, y)
        return self
```

```py
class NaiveProbaRegressor(BaseProbaRegressor):

    ...

    def _fit(self, X, y):
        """Fit regressor by fitting a distribution to the target values.

        Parameters
        ----------
        X : pandas DataFrame
            Feature instances (ignored).
        y : pandas DataFrame
            Target values to fit the distribution to.

        Returns
        -------
        self : reference to self
        """
        from skpro.distfitter import NormalFitter

        self._y_columns = y.columns

        if self.distfitter is None:
            distfitter = NormalFitter(method="unbiased")
        else:
            distfitter = self.distfitter.clone()

        distfitter.fit(y)
        self.distfitter_ = distfitter
        self.distribution_ = distfitter.proba()

        return self
```

PR #3504, opened 2026-07-03
fix: use np.random.default_rng() instead of deprecated np.random.seed() in sampling

wali-reheman:
replaces the deprecated global-state np.random.seed() + np.random.choice() pattern with the modern isolated default_rng() + rng.choice() in sample_discrete and sample_discrete_maps, making both functions thread-safe since concurrent calls no longer mutate shared global RNG state


ankurankan:
This seems to have already been fixed upstream. Thanks for the PR anyway. Closing.

PR #3530, opened 2026-07-27
fix: prevent double-removal of bidirected edges in do()

### LINKED ISSUE #3528: [BUG] do() raises ValueError when intervening on both endpoints of a bidirected edge
**Describe the bug**

Calling `do()` on multiple nodes connected by a bidirected (<>) edge raises a `ValueError`.

For example, for an ADMG containing `A <> B`, calling `do(["A", "B"])` attempts to remove the same bidirected edge twice and fails with:

`ValueError: Edge (B, A, <>) not in graph.`

Calling `do()` on only one endpoint works as expected.

**To Reproduce**

```python
from pgmpy.base import ADMG

g = ADMG(
    edge_list=[
        ("A", "B", "<>"),
    ]
)

print(g.get_edges(data=True))
# [('A', 'B', '<>')]

g.do(["A", "B"])
```

This raises:
`ValueError: Edge (B, A, <>) not in graph.`

**Expected behavior**

`do(["A", "B"])` should successfully return an ADMG containing both nodes with the bidirected edge removed:

```
result = g.do(["A", "B"])
list(result.nodes())
# ['A', 'B']
result.get_edges()
# []
```

**Additional context**

**Versions**
<details>

</details>

abhi-0203:
## Summary

Fix `ValueError` when `do()` is called with both endpoints of a bidirected edge.

**Bug:** Calling `do(["A", "B"])` on an ADMG with `A <> B` raises `ValueError: Edge (B, A, <>) not in graph.` because the bidirected edge is removed when processing node A, then the same edge is attempted again when processing node B.

**Fix:** Collect all edges to remove first (with `has_edge()` guard), then remove them in a single pass. This prevents the double-removal problem when both endpoints of a bidirected edge are in the intervention set.

**Reproducer:**
```python
from pgmpy.base import ADMG
g = ADMG(edge_list=[("A", "B", "<>")])
g.do(["A", "B"])  # ValueError before fix
```

**Expected after fix:** Returns an ADMG with both nodes and no edges.

Closes #3528

### DIFF
--- a/pgmpy/base/_base.py
+++ b/pgmpy/base/_base.py
@@ -890,10 +890,16 @@ class _CoreGraph(nx.MultiGraph, _GraphAlgorithms, _GraphRolesMixin, _GraphPlotti
                 )
 
         # Step 2: Remove every incoming (arrowhead-at-`node`) edge.
+        # Collect all edges first to avoid double-removal when both endpoints
+        # of a bidirected edge (A <> B) are in `nodes`.
+        edges_to_remove = []
         for node in nodes:
             for edge_type in arrowhead_types:
                 for neighbor in self.get_neighbors(node, edge_type):
-                    graph.remove_edge(node, neighbor, edge_type)
+                    if graph.has_edge(node, neighbor, edge_type):
+                        edges_to_remove.append((node, neighbor, edge_type))
+        for u, v, et in edges_to_remove:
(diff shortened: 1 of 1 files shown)


DARHWOLF:
@abhi-0203 Thanks for the PR. This issue might be part of an ongoing mentorship project. May I request you to also take up some other issue for the time being?

Issue #1766, opened 2024-05-23
Add functions to compute Structural Hamming Distance and Structural Intervention Distance


ankurankan:
Structural Intervention Distance: https://arxiv.org/pdf/1306.1043

jihyeseo:
I see that @Nimish-4 already implemented SHD. I am happy to try SID unless you are on it.

Nimish-4:
@jihyeseo In case you were asking me, I am not working on SID. @ankurankan may clarify if someone else is working on it.

ankurankan:
@Nimish-4 @jihyeseo Nobody is currently working on SID, please feel free to implement it.

jihyeseo:
Thanks for the info. I will start on it. Feel free to assign me to this.

aniket2405:
I'd like to work on this.

ankurankan:
@aniket2405 Sure. @jihyeseo had already gotten quite far in implementing this (#1927). You can branch off her branch, and 1. Update the code to some of the changes in the package (for example, BayesianNetwork is now called DiscreteBayesianNetwork), 2. See if you can simplify the implementation as we were discussing in the PR.

kchirag2023:
@ankurankan Do anyone implemented SID? if not please ,let me know i will work on it

ankurankan:
@kchirag2023 No, the PR has a partial implementation and is still open. Would suggest continuing from that.

rpp5501:
Hi , I'm Rishabh a student in the CodePath AI301 program. I'm interested in picking up the Structural Intervention Distance (SID) implementation from PR #1927 to get it across the finish line. I plan to branch off the existing work, update the deprecated BayesianNetwork classes to DiscreteBayesianNetwork, and finalize the adjustment set logic. Let me know if anyone is currently working on this, otherwise I will open a fresh PR linking back to this one soon!

ankurankan:
@rpp5501 Sorry for the delay. The PR for this has stalled. Assigning it to you.

rpp5501:
@ankurankan could you please take a look at my pull request and let me know if you know if you have any feedback.

rpp5501:
@ankurankan

PR #3535, opened 2026-08-04
Feature/sid metric

### LINKED ISSUE #2263: [ENH] Add Structural Intervention Distance (SID) for comparing DAGs.
Compared to Structure Hamming Distance (SHD), SID is more useful when comparing DAGs with a causal interpretation. Instead of just comparing structural differences (as with SHD), SID compares the implied interventional distributions in the two DAGs. 

Reference Paper: Structural Intervention Distance: https://arxiv.org/pdf/1306.1043

There is an already open PR (#1927) that implements parts of the functionality, but it needs to be updated to utilize some of the existing functionality in pgmpy.

rpp5501:

Yes
Yes
Yes

Yes
Yes. Also, added comments while reviewing opus code for easy understanding and review. 

Opus 5 , gpt sol and gemini/notebooklm for comments. 
Opus 5 and sol for implementing the algorithm code from the paper and gemini and opus for detailed comments .
Cross verified with the paper for consistency. 
Used opus 5 for writing tests. 
- Fixes #
https://github.com/pgmpy/pgmpy/issues/1766
- Fixes #2263

-sid.py:
-SID class extends the BaseSupervisedMetric; returns the count of ordered pairs whose interventional distribution the estimated DAG gets wrong by summing the incorrect interventions matrix returned by _sid_matrix .
- _compute_path_matrix : Computes transitive closure of all paths by repeatedly squaring.
-  _reachable_on_non_directed_path: Implements algorithm 2 to find the nodes reachable from the source on a non directed path (looks for colliders and back paths) given the adjustment set. 
- _sid_matrix : Implements algorithm 1, Also, calls and uses _reachable_on_non_directed_path function to find the incorrect interventions . Algorithm 1 finds if any estimated parent that is in the adjustment set is a descendant of any node on the causal path(excluding source).  If algorithm 1 doesn't find any incorrect interventions it uses the outputs from _reachable_on_non_directed_path that implements algorithm 2. If the target is in the adjustment set then the (source, target) pair is marked as having no effect in the estimated graph and this is verified with the true graph. 
-is_default = False, so CausalDiscovery.score still selects SHD.
-test_sid.py:
-Tests the paper's worked example; asymmetry; node-order independence; input validation.
- Reference matrices from PR#1927 across 6 graph pairings.
- Brute-force oracle deriving expected values from the adjustment criterion. Checked against both the implementation and reference matrices. 
- Tests edge cases: edgeless/isolated graphs, mediator adjustment and default-metric registry. 
- __init__.py: Importing SID, making SID discoverable via get_metrics.
- metrics.rst: autosummary entry under supervised metrics
- references.bib : Cited the paper

### DIFF
--- a/pgmpy/metrics/sid.py
+++ b/pgmpy/metrics/sid.py
@@ -168,6 +168,11 @@ class SID(BaseSupervisedMetric):
     Graphs are converted to aligned adjacency matrices once, and the path and
     state-doubled reachability computations use NumPy arrays.
 
+    As in the reference implementation, the transitive closure is recomputed for
+    each source node, so the cost grows roughly with the fourth power of the
+    number of nodes: a 100-node pair takes a few seconds and a 200-node pair
+    around a minute.
+
     Examples
     --------
     >>> from pgmpy.base import DAG
@@ -179,9 +184,8 @@ class SID(BaseSupervisedMetric):
 
     References
     ----------
-    Peters, J. and Buehlmann, P. (2015). Structural Intervention Distance (SID)
-    for Evaluating Causal Graphs. Neural Computation, 27(3), 771-799.
-    https://doi.org/10.1162/NECO_a_00708
+    - :footcite:t:`peters_buhlmann_2015`
+
     """
 
     _tags = {
@@ -191,12 +195,11 @@ class SID(BaseSupervisedMetric):
         "lower_is_better": True,
         "is_symmetric": False,
         "supported_graph_types": (DAG,),
-        "is_default": True,
+        "is_default": False,
     }
 
     def _evaluate(self, true_causal_graph: DAG, est_causal_graph: DAG) -> int:
-        if set(true_causal_graph.nodes()) != set(est_causal_graph.nodes()):
-            raise ValueError("true_causal_graph and est_causal_graph must have the exact same set of nodes.")
+        # `BaseSupervisedMetric.evaluate` has already checked the types and the shared node set.
         nodes = list(true_causal_graph.nodes())
         true_adjacency = true_causal_graph.to_adjacency(encoding="binary", nodelist=nodes).to_numpy(dtype=bool)
         est_adjacency = est_causal_graph.to_adjacency(encoding="binary", nodelist=nodes).to_numpy(dtype=bool)
--- a/pgmpy/metrics/sid.py
+++ b/pgmpy/metrics/sid.py
@@ -80,19 +80,40 @@ def _reachable_on_non_directed_path(
             continue
         checked[current_node] = True
 
+        # The arrival state and the departure direction together fix which d-separation role
+        # ``current_node`` (written c below) plays, so each of the four combinations gets exactly one
+        # transition. Writing a for a parent and b, d for children:
+        #
+        #   from     to              walk           role at c    guarded by
+        #   c        child           a -> c -> d    chain        not conditioned[c]
+        #   n+c      child           b <- c -> d    FORK         not conditioned[c]
+        #   n+c      n+parent        b <- c <- a    chain        not conditioned[c]
+        #   c        n+parent        a -> c <- a'   collider     ancestors_of_conditioned[c]
+        #
+        # Chains and forks block under the same condition, which is why all three non-collider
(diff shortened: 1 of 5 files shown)


pgmpy/metrics/sid.py
```diff
@@ -0,0 +1,308 @@
...
+
+    # Each squaring doubles the path length covered, so ceil(log2(p)) rounds reach every path: the
+    # longest simple path in a p-node DAG has p - 1 edges. On a bool array ``@`` is logical matrix
+    # multiplication (+ is OR, * is AND), so "a reaches b and b reaches c" composes into "a reaches c".
+    for _ in range(math.ceil(math.log2(n_nodes))):
+        path_matrix = path_matrix @ path_matrix
+    return path_matrix
+
+
+def _reachable_on_non_directed_path(
```
ankurankan, line 29: This looks like something that we already have implemented in the `DAG` class.

pgmpy/metrics/sid.py
```diff
@@ -0,0 +1,308 @@
...
+
+    def _evaluate(self, true_causal_graph: DAG, est_causal_graph: DAG) -> int:
+        # `BaseSupervisedMetric.evaluate` has already checked the types and the shared node set.
+        # Take the node order once and reuse it as the ``nodelist`` for both conversions, so the two
+        # matrices stay row- and column-aligned even when the graphs enumerate their nodes differently.
+        nodes = list(true_causal_graph.nodes())
+        true_adjacency = true_causal_graph.to_adjacency(encoding="binary", nodelist=nodes).to_numpy(dtype=bool)
+        est_adjacency = est_causal_graph.to_adjacency(encoding="binary", nodelist=nodes).to_numpy(dtype=bool)
+        # Each True cell is one ordered pair with a falsely inferred intervention distribution.
+        return int(_sid_matrix(true_adjacency, est_adjacency).sum())
```
ankurankan, line 308: I don't think `_sid_matrix` should be separate function. Since, it is being called only from here, it would be better to have that logic within this method. It makes it easier to read without jumping between methods.

pgmpy/metrics/sid.py
ankurankan, line 1: Use 120 char line length.

pgmpy/metrics/sid.py
```diff
@@ -0,0 +1,308 @@
...
+        #
+        # Chains and forks block under the same condition, which is why all three non-collider
+        # transitions share the single ``not conditioned`` guard rather than being cased apart.
+        #
+        # Every one of these transitions is recorded twice, once from each end of the edge: the
+        # parents block below writes them while visiting the child, the children block while
+        # visiting the parent. The duplication is harmless but total -- deleting either block
+        # wholesale leaves the results unchanged, because parents are always queued (a conditioned
+        # node is its own ancestor, so one of the two guards below always fires) and therefore a
+        # visited node's parents are visited too. Kept as-is to match Algorithm 2 as published.
```
ankurankan, line 101: I would not recommend writing comment like this in the code. It needs to be 1-2 lines max. I would suggest looking at how we have written comments as Steps (have a look at any file in pgmpy/causal_discovery). The comments then act as the pseudo code for the method.

pgmpy/metrics/sid.py
```diff
@@ -0,0 +1,308 @@
...
+                ],
+            ] = False
+
+        reachable_on_non_directed_path |= reachability_matrix[reachable_on_non_directed_path, :].any(axis=0)
+
+    # Collapse the doubled state space: a node counts as reached if either arrival state was reached.
+    return reachable_on_non_directed_path[:n_nodes] | reachable_on_non_directed_path[n_nodes:]
+
+
+def _sid_matrix(true_graph: np.ndarray, est_graph: np.ndarray) -> np.ndarray:
```
ankurankan, line 183: This looks like this is implementing an identification algorithm. This should be reusing code from `pgmpy/identification`.

ankurankan (changes requested):
I believe this is a mostly AI generated PR and you might not have understood the method well enough to figure which parts of the codebase can be reused. I would recommend spending some time understanding the method before implementing it.

PR #3537, opened 2026-08-05
Implementing Greedy Sparsest Permutation Algorithm

### LINKED ISSUE #3540: [ENH] Add Greedy Sparsest Permutation (GSP) as a `variant` of `SP`
### Summary
 
Add a greedy variant to `pgmpy.causal_discovery.SP` implementing the Greedy Sparsest Permutation algorithm. Exhaustive `SP` scales as O(p!) and is already documented as impractical past ~10 variables. GSP is a bounded approximation (search depth `d`, restarts `r`) that scales to hundreds of variables while remaining consistent under a weaker-than-faithfulness assumption.
 
Proposal: extend the existing `SP` class with `variant="greedy"` rather than adding a separate `GSP` class, the two algorithms share their core primitive (`_build_imap_edges`) and public output attributes, so a second class would just duplicate plumbing.
 
### Motivation
 
- Exhaustive `SP` is currently unusable above ~10 nodes.
- GSP is shown in the paper to outperform PC and GES on both simulated and
  real (gene regulatory network) data, while staying within the same
  `SP`-style consistency guarantees.
- Users should be able to opt in without changing existing code.

### Proposed API
 
```python
SP()                                          # unchanged — exhaustive SP
SP(variant="greedy")                          # GSP, default depth/runs
SP(variant="greedy", search_depth=4, n_runs=10)
```
 
```python
def __init__(
    self,
    ci_test: str | Callable | None = None,
    significance_level: float = 0.01,
    max_iter: int | None = None,
    return_type: str = "dag",
    show_progress: bool = True,
    seed: int | None = None,
    variant: str = "exhaustive",
    search_depth: int = 4,
    n_runs: int = 10,
):
```
 
| Parameter | Default | Used by | Meaning |
|---|---:|---|---|
| `ci_test` | `None` | Both | CI test |
| `significance_level` | `0.01` | Both | CI-test significance level |
| `max_iter` | `None` | Exhaustive only | Max permutations evaluated |
| `return_type` | `"dag"` | Both | `"dag"` or `"pdag"` |
| `show_progress` | `True` | Both | Progress bar |
| `seed` | `None` | Both | Random seed |
| `variant` | `"exhaustive"` | Both | `"exhaustive"` or `"greedy"` |
| `search_depth` | `4` | Greedy only | Max DFS depth per run (paper's `d`) |
| `n_runs` | `10` | Greedy only | Independent restarts (paper's `r`) |
 
`variant="exhaustive"` is the default, so **all existing code keeps working
unchanged** (`SP(ci_test="chi_square", max_iter=100)` is unaffected).
 
`max_iter` and `n_runs` are mutually exclusive by variant — validated only
for the variant that uses them, so `SP(search_depth=0)` (exhaustive) does
not raise.
 
### Internal structure — one new helper only
 
```
SP
├── __init__()
├── _build_imap_edges()        existing, shared by both variants
├── _reverse_if_covered()      new helper
└── _fit()
    ├── common init + validation
    ├── variant == "exhaustive"   → current code, unchanged
    └── variant == "greedy"       → GSP loop (Algorithm 4)
```
 
Both algorithms remain inline in `_fit()` (not split into a separate
`_gsp_search()` method) so the branch logic is easy to diff and review.
 
### Implementation note: the sparsifying step needs two parts
 
A single covered-edge reversal only produces a **Markov-equivalent** graph —
same skeleton, same edge count (this is the `Gπ → G` move in Figure 1 of the
paper). It is *not* by itself a sparser candidate. To reach `Gτ`, a genuinely
sparser minimal I-MAP, each DFS step needs to:
 
1. Reverse a covered arrow `u → v` (`Pa(u) == Pa(v) \ {u}`) → Markov-equivalent `G'`.
2. Take a linear extension (topological sort) `τ` of `G'`.
3. Re-run `_build_imap_edges(τ, n_edge_limit=current_edges)` — this re-tests
   CI relations against the new ordering and is what can actually drop
   edges.
4. Accept the move if edge count is weakly decreasing (`≤` current); accept
   as an *improvement* if strictly less. Bound total steps without strict
   improvement by `search_depth`; restart from a new random permutation
   after `n_runs` exhausted restarts with no improvement, per Algorithm 4.
Skipping step 3 (i.e., driving the DFS off the reversed graph directly)
would make the search a no-op that never reduces edge count, so this should
be called out explicitly in the PR description and covered by a test.
 
### Shared output attributes (unchanged for both variants)
 
```python
self.causal_graph_
self.adjacency_matrix_
self.optimal_permutations_
self.n_features_in_
self.feature_names_in_
```

anusa-saha:

- Fixes #3540 

-


pgmpy/causal_discovery/SP.py
```diff
@@ -151,6 +193,39 @@ def _build_imap_edges(
...
+        Returns
+        -------
+        list[DAG]
+            One DAG per covered arrow in G, each with that arrow reversed.
+        """
+        neighbors = []
+
+        for u, v in G.edges():
+            parents_u = set(G.predecessors(u))
+            parents_v = set(G.predecessors(v))
```
ankurankan: Use `get_parents`  instead of `predecessors`

pgmpy/causal_discovery/SP.py
```diff
@@ -166,52 +241,117 @@ def _fit(self, X: pd.DataFrame):
...
+                    model = DAG()
+                    model.add_nodes_from(permutation)
+                    model.add_edges_from(edges)
+
+                    depth = 0
+                    while depth < self.search_depth:
+                        moved = False
+
+                        for neighbor in self._reverse_covered_edges(model):
+                            candidate_permutation = tuple(nx.topological_sort(neighbor))
```
ankurankan: Check if there is a simple way to compute the topological order of each neighbour using the topological order of the `model`.

ankurankan: Could possibly be a simple list manipulation which can give the neighbour's topological order.

Issue #1459, opened 2021-09-12
CausalInference: Same Variables in Variables/Evidence Error when not the same?


rokapre:
### Subject of the issue

Hi, I seem to be running into a bug where the CausalInference.query method complains that the variables in the "variables" and evidence arg are the same when they actually aren't. Below is the code: 

### Your environment
* pgmpy version:  0.1.15
* Python version: 3.8
* Operating System:MacOS Catalina

### Steps to reproduce

Here is some simulated data

```
import numpy as np
import pandas as pd
import pgmpy
from pgmpy.models import BayesianModel
from pgmpy.estimators import MaximumLikelihoodEstimator,BayesianEstimator
from pgmpy.inference import VariableElimination,CausalInference


n = 10000

#P(W)

pW = 0.3
W = np.random.binomial(1,pW,n)

#P(X=1|W=0) = 0.3, P(X=1,|W=1) = 0.6
nW1 = np.sum(W==1)
nW0 = np.sum(W==0)

pX_W0 = 0.3
pX_W1 = 0.6
X = np.zeros(n)
X[:] = np.nan

X[W==0] = np.random.binomial(1,pX_W0,nW0)
X[W==1] = np.random.binomial(1,pX_W1,nW1)

#P(Y=1|X=1,W=0) = 0.3, P(Y=1|X=0,W=0) = 0.3, P(Y=1|X=1,W=1) = 0.9, P(Y=1|X=0,W=1) = 0.5

nX1W0 = np.sum((X==1) & (W==0))
nX0W0 = np.sum((X==0) & (W==0))
nX1W1 = np.sum((X==1) & (W==1))
nX0W1 = np.sum((X==0) & (W==1))

pY_X1W0 = 0.3
pY_X0W0 = 0.3
pY_X1W1 = 0.9
pY_X0W1 = 0.5

Y = np.zeros(n)
Y[:] = np.nan 
Y[(X==1) & (W==0)] = np.random.binomial(1,pY_X1W0,nX1W0)
Y[(X==0) & (W==0)] = np.random.binomial(1,pY_X0W0,nX0W0)
Y[(X==1) & (W==1)] = np.random.binomial(1,pY_X1W1,nX1W1)
Y[(X==0) & (W==1)] = np.random.binomial(1,pY_X0W1,nX0W1)

simdata = pd.DataFrame({"Y":Y,"X":X,"W":W})

bn = BayesianModel([("X","Y"),("W","X"),("W","Y")])
[...]

PR #3538, opened 2026-08-06
inference: CausalInference: Fix query() dropping evidence with do

### LINKED ISSUE #1701: do operator on evidence
When using the `.query `method to obtain the effect of Y on X using the **do** operator, and some evidence Z, is it possible that depending on which nodes are in Z it may not be possible to isolate the causal path from association paths? Would the API return an error?

NAME-ASHWANIYADAV:

If you have used AI/LLMs for any assistance, please answer the following questions. Please refer #2622 for an example of the level of detail we expect:

Anthropic Claude Opus 5 via Claude Code.

I found this bug while reading through the causal inference module and used the tool throughout the workflow, with me directing each step and reviewing the output. Concretely: I used it to trace the root cause (the adjustment loop at `query()` overwrites the user's evidence dict, and the weights get conditioned on the evidence when they should not be), to check my derivation of the correct formula, and to write the first draft of the implementation and the tests. I reviewed every line of the draft, and the key design decisions were mine: fixing the computation to return the correct value instead of raising an error (see my comment on #1701), using inference on the explicitly mutilated network as the ground truth for tests instead of hand-computed numbers, and splitting the evidence by graph position (see below) after my verification runs caught a mismatch in the first version. I can walk through every line of the diff and I take full responsibility for it.

The regression tests were written first and failed on dev with the exact wrong numbers before the fix was applied (test-driven). All correctness tests compare against inference on the explicitly mutilated network (cut incoming edges of the do-variables, uniform prior on them, condition on do + evidence), which is the definition of do(), so the tests are self-verifying.

Beyond the unit tests, I ran a 111-check verification battery: a 90-case fuzz over random networks (4-6 nodes, 2-3 states, varying edge probability) comparing `query()` against the mutilated network for every evidence shape (evidence inside the adjustment set, outside it, mixed, multiple evidence variables, joint multi-variable queries), plus hand-built cases for multi-do
interventions, string state names, 3-state variables, custom adjustment sets, VE/BP parity, repeated-call determinism, latent-parent error behaviour, and impossible-evidence behaviour (NaN, same as plain conditional inference).

This battery is also what caught a subtle issue in my first version: with a user-specified (non-parent) adjustment set and evidence on a non-descendant of the do-variable, the weights must be p(z | that evidence), not the marginal
prior. The final version therefore splits evidence by graph position: evidence on non-descendants conditions the weights and the per-state queries directly, evidence on descendants is queried jointly and conditioned at the end. For the
default (parents) adjustment set both formulations are exact; for custom sets only the split is correct.

Also verified: the evidence-inside-adjustment-set values from #1459/#1461 check out by hand and stay green; the full inference test suite passes (the one ApproxInference failure is a pre-existing flake - it fails identically on clean dev, which I confirmed by stashing the changes); CausalInference doctests pass; pre-commit hooks pass.

Yes, the first drafts. They are already compact: one method covers the four evidence configurations against a shared ground-truth helper (on both VE and BP), one covers custom adjustment sets, and one is a property check over 8
random networks. Collapsing further would hide which configuration regressed.

- Fixes #1701

- `CausalInference.query()`: the per-state queries inside the adjustment sum now keep the user's evidence instead of overwriting     the evidence dict with the adjustment-set states.
- Evidence is split by graph position: evidence on non-descendants of the do-variables conditions the adjustment weights (p(z |    evidence)) and the per-state queries; evidence on descendants is queried jointly with the target variables and conditioned on at the end (it must not enter the weights).
- Added an early `ValueError` when a variable appears in both `variables` and `evidence`, matching the behaviour of the non-do path.
- Tests: `test_query_do_with_evidence` (four evidence configurations x VE/BP  against a mutilated-network ground-truth helper),`test_query_do_with_evidence_custom_adjustment` (user-specified adjustment sets with evidence), and `test_query_do_with_evidence_random_models`
  (property check on 8 random networks). The #1459/#1461 case stays green.


NAME-ASHWANIYADAV:
@ankurankan may you please review this PR !!

pgmpy/inference/CausalInference.py
```diff
@@ -1020,17 +1023,34 @@ def query(
...
-        # Step 4: For other cases, compute \sum_{z} p(variables | do, z) p(z)
+        # Step 4: For other cases, compute the interventional joint over
+        # `variables` and the post-intervention evidence as
+        # \sum_{z} p(variables, evidence_post | do, z, evidence_pre) p(z | evidence_pre),
+        # with z clamped to any evidence on the adjustment set, and then
+        # condition the sum on evidence_post. Evidence on non-descendants of
+        # the do variables (evidence_pre) is unaffected by the intervention, so
+        # it conditions the weights and the per-state queries directly;
+        # evidence on descendants (evidence_post) must not be conditioned into
+        # the weights and is applied to the summed joint at the end.
```
DARHWOLF: this looks superfluous.

PR #3554, opened 2026-08-28
identification: Adjustment: Fix validate() missing forbidden-set check

### LINKED ISSUE #3553: [BUG] identification.Adjustment.validate() accepts forbidden adjustment sets (mediators, colliders)
**Describe the bug**

`Adjustment.validate()` only checks the d-separation part of the backdoor criterion and never checks the forbidden-set condition - "no node in Z may be a descendant of the exposure on a causal path" (condition (i) of the backdoor criterion / the forb(X, Y, G) condition in Perkovic et al. 2018, which the class docstring cites). On top of that, the check works through a parents-of-exposure shortcut (`_validate` in `pgmpy/identification/adjustment.py`): when the exposure has no parents outside the conditioning set, the `all()` is vacuously true and literally any adjustment set validates.

Concretely, conditioning on a mediator or on a collider both pass validation, even though adjusting for either biases the effect estimate. The collider case is also a regression against the deprecated `CausalInference.is_valid_adjustment_set`, which correctly rejects it via the proper-backdoor-graph d-separation test - and the deprecation warnings now point users to this module as the replacement.

**To Reproduce**

```python
from pgmpy.base import DAG
from pgmpy.identification import Adjustment

# 1. Mediator: X -> M -> Y. M sits on the causal path.
d1 = DAG([("X", "M"), ("M", "Y")],
         roles={"exposures": "X", "outcomes": "Y", "adjustment": ["M"]})
print(Adjustment().validate(d1))   # True, should be False

# 2. Collider: X -> Y, X -> C <- Y. Conditioning on C opens a spurious path.
d2 = DAG([("X", "Y"), ("X", "C"), ("Y", "C")],
         roles={"exposures": "X", "outcomes": "Y", "adjustment": ["C"]})
print(Adjustment().validate(d2))   # True, should be False

# 3. A genuine confounder A plus the mediator M.
d3 = DAG([("A", "X"), ("A", "Y"), ("X", "M"), ("M", "Y")],
         roles={"exposures": "X", "outcomes": "Y", "adjustment": ["A", "M"]})
print(Adjustment().validate(d3))   # True, should be False (M is forbidden)

# Sanity: the valid set {A} alone is correctly accepted.
d4 = DAG([("A", "X"), ("A", "Y"), ("X", "M"), ("M", "Y")],
         roles={"exposures": "X", "outcomes": "Y", "adjustment": ["A"]})
print(Adjustment().validate(d4))   # True, correct
```

**Expected behavior**

Cases 1-3 should return False. Per the adjustment criterion (Perkovic et al. 2018, already referenced in the docstring), Z is a valid adjustment set iff (a) Z contains no node from the forbidden set forb(X, Y, G) - descendants of nodes on proper causal paths from X to Y - and (b) Z blocks all proper non-causal paths from X to Y, which is equivalent to X being m-separated from Y given Z in the proper backdoor graph that `_get_proper_backdoor_graph` already builds.

**Additional context**

The root cause is in `_validate` (`pgmpy/identification/adjustment.py`): it tests that the unconditioned parents of the exposure are separated from the outcome, which covers neither condition (a) nor, in the no-parents case, anything at all. The deprecated `CausalInference.is_valid_adjustment_set` has the same missing forbidden-set condition (it also accepts the mediator in case 1), though its proper-backdoor-graph test does reject the collider.

I would like to work on this - happy to implement validate() as the two conditions above with regression tests for the cases here. Could you assign this to me?

**Versions**
<details>

```
black==26.5.1
huggingface_hub==1.19.0
joblib==1.5.3
networkx==3.6.1
numpy==2.4.6
opt_einsum==3.4.0
pandas==3.0.3
pgmpy==1.1.2 (dev branch, source install)
pre_commit==4.6.0
pyparsing==3.3.2
pytest==9.1.0
scikit-learn==1.9.0
scipy==1.17.1
statsmodels==0.14.6
tqdm==4.68.2
```

</details>

NAME-ASHWANIYADAV:

If you have used AI/LLMs for any assistance, please answer the following questions. Please refer #2622 for an example of the level of detail we expect:

Anthropic Claude (opus 5) via Claude Code.

I found this while reading the identification module (it is where the deprecation warnings on the old CausalInference backdoor methods now point users). I used the tool throughout, directing each step and reviewing the output: tracing why validate() accepts mediators and colliders (the parents-of-exposure shortcut covers only part of the criterion and is vacuously true when the exposure has no unconditioned parents), checking my reading of the adjustment criterion from Perkovic et al. 2018 (already cited in the docstring), and drafting the implementation and tests, which I then reviewed line by line and re-ran. The key design decisions were mine: validating both conditions of the criterion (forbidden set + separation in the proper backdoor graph), computing paths on the directed projection so the same code works for ADMGs (which also resolves the "TODO: Make this work for all graph types" in _get_proper_backdoor_graph), and correcting one existing test expectation after proving numerically that it was wrong (details below). I can explain every line and take full responsibility for the changes.

The regression tests were written first and failed on dev exactly as the issue describes (mediator/collider/confounder+mediator accepted, ADMG bidirected confounding vacuously accepted), then passed after the fix.

Beyond the unit tests, I ran a numeric ground-truth battery: for a fully specified network, Z is a usable adjustment set iff sum_z P(Y|X,z)P(z) equals P(Y|do(X)) computed on the explicitly mutilated network. Across 68 checks (random DAGs with random CPDs, every small candidate Z, plus the hand-built cases), every set the new validate() accepts matches the mutilated network to
machine precision (~1e-16) and every set it rejects is measurably wrong (1e-3 to 2.4e-1). The battery was run twice with different RNG seeds, plus 13 adversarial cases: a descendant of a mediator (rejected), an off-path child of the exposure (correctly accepted - the forbidden set is the one from the paper, not the stricter "no descendant of X"), a descendant of the outcome (rejected), exposure chains under joint intervention (the proper-path filter does not over-remove edges), ADMG bidirected confounding, and a PDAG smoke test (PDAG still goes through the m-separation branch, as before).

One existing test expectation was wrong and is corrected in this PR: test_is_valid_adjustment_set asserted that {z2} is a valid adjustment set for X={x1,x2}, Y={y1,y2} in the UAI-2014 example model. Conditioning on z2 opens the proper non-causal path x1 -> z1 -> z2 <- y2 (z2 is a collider on it), so only {z1,z2} is valid - this is exactly the point that example is used for in
the paper. The old code accepted {z2} because the cross pair (x1,y2) was never tested. Numerically: with random CPDs the adjustment formula for {z2} is off from the mutilated-network ground truth by ~2e-2, while {z1,z2} matches to ~1e-16.

Also verified: the Frontdoor identification class (which calls Adjustment().validate internally) passes its tests, identify(variant="all") on the Book-of-Why game5 model still returns exactly the five textbook adjustment sets, all identification doctests pass, and pre-commit hooks pass.

Yes, the first drafts. They are compact: one test covers the forbidden-set cases from the issue against a shared model set, one covers the ADMG bidirected case, and the corrected {z2} assertion lives in the existing test_is_valid_adjustment_set. Collapsing further would hide which criterion condition regressed.

- Fixes #3553

- `Adjustment._validate()` now checks both conditions of the adjustment  criterion (Perkovic et al. 2018, already cited in the docstring): the adjustment set must not contain a forbidden node (an exposure, a node on a proper causal path, or a descendant of such a node), and the exposures must be separated from the outcomes given the adjustment set in the proper backdoor graph. The old parents-of-exposure shortcut checked neither in  general and was vacuously true when the exposure had no unconditioned parents.
- Added `Adjustment._proper_causal_paths()` helper; proper causal paths are computed on the directed projection (`get_directed_graph()`) so ADMG/MAG  inputs are handled correctly, and paths through other exposures are excluded.

- `_get_proper_backdoor_graph()` now uses the same helper (resolves the "TODO: Make this work for all graph types"): on ADMGs it previously enumerated paths on the undirected MultiGraph structure and removed untyped edges; it now removes exactly the first `->` edge of every proper causal path.
- Corrected one wrong expectation in `test_is_valid_adjustment_set`: {z2} alone is not a valid adjustment set in the UAI-2014 example model (see verification section above).
- New tests: `test_validate_rejects_forbidden_sets` (mediator, collider, confounder+mediator, valid-set sanity) and
  `test_validate_admg_bidirected_confounding`.

### DIFF
--- a/pgmpy/identification/adjustment.py
+++ b/pgmpy/identification/adjustment.py
@@ -61,6 +61,25 @@ class Adjustment(BaseIdentification):
         elif self.variant == "minimal_variance":
             self.supported_graph_types = (DAG, PDAG)
 
+    @staticmethod
+    def _proper_causal_paths(causal_graph):
+        """
+        Yield the node paths of all proper causal paths from `exposures` to `outcomes`.
+
+        A proper causal path is a directed path from an exposure to an outcome
+        that does not pass through another exposure. For graph types that also
+        carry non-directed edges (ADMG, MAG), the paths are computed on the
+        directed projection of the graph.
+        """
+        exposures = causal_graph.get_role("exposures")
+        outcomes = causal_graph.get_role("outcomes")
+        directed = causal_graph if isinstance(causal_graph, DAG) else causal_graph.get_directed_graph()
+        for source in exposures:
+            for path in nx.all_simple_paths(directed, source, outcomes):
+                if set(path[1:]).intersection(exposures):
+                    continue
+                yield path
+
     def _get_proper_backdoor_graph(self, causal_graph, inplace=False):
         """
         Returns a proper backdoor graph of the `causal_graph`.
@@ -101,14 +120,12 @@ class Adjustment(BaseIdentification):
         ----------
         - :footcite:t:`perkovic_2018`
         """
-        # TODO: Make this work for all graph types.
         model = causal_graph if inplace else causal_graph.copy()
-        edges_to_remove = []
-        for source in causal_graph.get_role("exposures"):
-            paths = nx.all_simple_edge_paths(causal_graph, source, causal_graph.get_role("outcomes"))
-            for path in paths:
-                edges_to_remove.append(path[0])
-        model.remove_edges_from(edges_to_remove)
+        first_edges = {(path[0], path[1]) for path in self._proper_causal_paths(causal_graph)}
+        if isinstance(causal_graph, DAG):
+            model.remove_edges_from(first_edges)
+        else:
+            model.remove_edges_from([(u, v, "->") for u, v in first_edges])
         return model
 
     def _identify(self, causal_graph):
@@ -181,7 +198,11 @@ class Adjustment(BaseIdentification):
 
         Given a `causal_graph` with variable roles `exposures`, `outcomes`, and
         `adjustment` defined, this method checks if the given `adjustment` set
-        is valid.
+        satisfies the adjustment criterion [1]: it must not contain a forbidden
+        node (an exposure, a node on a proper causal path from the exposures to
+        the outcomes, or a descendant of such a node), and it must block every
+        proper non-causal path, i.e. the exposures must be separated from the
+        outcomes given the adjustment set in the proper backdoor graph.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


NAME-ASHWANIYADAV:
@ankurankan PTAL !!

PR #3560, opened 2026-09-03
Fix PC search_space handling: seed skeleton from search_space and pre…

### LINKED ISSUE #2155: [ENH] Improve search space handling in PC
**Description**
Currently, search space is not handled correctly in PC. When a search space is specified and the expert knowledge is enforced with PC, the edges may be completely missing from the final graph, going against the idea of a search space.

**Solution**
The search space needs to be handled by using it as the skeleton, or some better approach.

Shweta-nasc:
The following checklist was written by @ankurankan 

- The skeleton was always initialized as a complete graph, ignoring expert_knowledge.search_space entirely. This meant CI tests could condition on irrelevant variables, causing search_space edges to be incorrectly dropped.
- required_edges outside search_space were never added as candidates, so they could never appear in the final graph.
- Fixed by seeding the initial skeleton from search_space_ (the fitted, normalized attribute) when non-empty, and explicitly adding required_edges as candidates.
- Added a regression test using a deterministic fake CI test to verify both behaviors.

###  Please list the AI tool(s) used, along with the model and its version used.
 Used Claude mainly for codebase navigation and debugging support, since this is my first PR in this repo and I wasn't familiar with the structure or tooling.

Had my own initial fix already — seeding the skeleton from search_space instead of a complete graph. Used Claude to find _build_skeleton in _base.py and go through the skeleton-pruning loop and required_edges/forbidden_edges handling.

Claude flagged that my fix didn't cover required_edges outside search_space. Verified with a repro script, then fixed it. That edit broke 36 tests. Used git stash to confirm the regression was from my change and not pre-existing on dev. Traced it to two bugs, verified by reading ExpertKnowledge source directly: search_space defaults to set() not None, so is not None always passed; and the fix needed to read search_space_ (fitted attribute), not the raw one.

For the regression test, first version used real simulated data and gave a misleading failure — diagnosed it as expected PC behavior (search_space restricts candidates, doesn't guarantee they survive), rewrote with a deterministic fake CI function instead.

Also ran black on Claude's suggestion, wrong formatter for this repo (uses ruff), corrupted a file's formatting. Caught it, git checkout -- to discard, reapplied only the intended change.

### Please describe in **detail** how and what you used AI assistance for? Please outline your whole workflow with the AI tool.

Used Claude for codebase navigation and debugging support on my first PR in this repo.

- Had initial fix already: seed skeleton from `search_space` instead of complete graph.
- Located `_build_skeleton` in `pgmpy/causal_discovery/_base.py` with Claude. Read through skeleton-pruning loop, `required_edges`/`forbidden_edges` handling.
- Claude flagged `required_edges` outside `search_space` not handled. Wrote repro script, confirmed. Added fix.
- Ran test suite: 36 failures. `git stash`, reran on clean `dev`: 86 passed. Confirmed regression from my edit.
- Read `ExpertKnowledge` source directly. Found `search_space` defaults to `set()`, not `None` — `is not None` check always true. Found `search_space_` (fitted, normalized) should be used instead of raw `search_space`.
- Checked `expert_knowledge.fit(X)` runs before `_build_skeleton` in `PC.py` — `search_space_` guaranteed populated.
- Applied both fixes. Reran suite: 86 passed, baseline restored.
- Wrote regression test. First version used real Asia data, got a failure — edge from `search_space` correctly dropped by CI test (search_space restricts candidates, doesn't force survival). Rewrote with deterministic fake CI function.
- Ran `black --check` — wrong formatter for repo (uses `ruff`, `.pre-commit-config.yaml`). Reformatted entire test file. Caught via diff size. `git checkout --` to discard, reapplied fix via exact-string-match script insert.
- Pre-commit verification: `ast.parse` both files, `ruff check` + `ruff format --diff` clean, full suite 87 passed (86 + 1 new), manual `git diff` read.

### What steps have you taken to verify that the changes correctly address the issue? What edge cases have you considered? Other than running tests, what else have you verified?

- Full test suite before/after change (git stash isolation): 86 passed baseline on clean dev, 87 passed with fix + new test.
- Manual repro script: ExpertKnowledge(search_space=..., required_edges=[edge not in search_space]) on Asia network — confirmed edge missing before fix, present after.
- Edge case: expert_knowledge=None. Existing test calls _build_skeleton this way directly — confirmed is not None guard still handles it.
- Edge case: empty search_space (default when unspecified). Confirmed falls back to complete graph, not an empty restricted skeleton — this was itself a live bug (is not None vs truthiness) found and fixed during verification.
- Edge case: search_space edge that's genuinely conditionally independent in data. Confirmed it's correctly dropped by CI test — search_space restricts candidates, doesn't force survival. Used to rewrite regression test with deterministic fake CI instead of real data.
- Confirmed expert_knowledge.fit(X) always runs before _build_skeleton in PC.py, so search_space_ is populated when read.
- ruff check + ruff format --diff on both changed files: clean, no findings.
- Manual read of full git diff before staging — confirmed only intended lines changed in both files.
- Used Codecov's patch-coverage report to catch that two lines in an early version of the regression test were unreachable dead code (a `fake_ci` branch for an edge pair that could never be queried, since it was excluded from `search_space`). Rewrote the test to add that pair to `search_space` instead and assert it's correctly dropped by real CI-based pruning — this also closed a real coverage gap: the original test only proved edges *outside* `search_space` were absent, not that genuine independence testing still works correctly on edges *inside* `search_space`.
-  Running the example notebook test suite in CI (Python 3.14 job) surfaced a pre-existing typo in `examples/Expert_Knowledge.ipynb` (`"Gruntin"` instead of `"Grunting"` in a `required_edges` list). This typo previously had no effect because `required_edges` outside `search_space` wasn't actually being applied to the skeleton before this fix — it only surfaced as a `KeyError` once `required_edges` started being correctly added as a candidate. Fixed the typo as part of this PR since it directly demonstrates the fix changing real behavior.
### Have you used AI for generating tests? Can you compress them into a smaller number of tests without losing coverage?

Yes, one test: test_pc_search_space_and_required_edges.

Covers both fixed behaviors in a single test — search_space restricting candidate edges, and required_edges surviving outside search_space — using shared setup (one dataset, one deterministic fake CI function) rather than two separate tests. No further compression possible without losing the distinct assertions per behavior (edge excluded by search_space absent; edges included and dependent present; required edge outside search_space present).

- Fixes #2155 

- pgmpy/causal_discovery/_base.py: _build_skeleton seeds initial candidate graph from expert_knowledge.search_space_ when non-empty, instead of always using a complete graph. required_edges explicitly added as candidates so they aren't dropped for being outside search_space.
- pgmpy/tests/test_causal_discovery/test_PC.py: added test_pc_search_space_and_required_edges — regression test for both fixed behaviors.
-  examples/Expert_Knowledge.ipynb: fixed a pre-existing typo (`"Gruntin"` → `"Grunting"`) in a `required_edges` list, uncovered by this fix (previously silently ignored, now surfaces as a `KeyError` since `required_edges` is correctly applied).

### DIFF
--- a/pgmpy/tests/test_causal_discovery/test_PC.py
+++ b/pgmpy/tests/test_causal_discovery/test_PC.py
@@ -463,13 +463,11 @@ def test_pc_search_space_and_required_edges():
         pair = frozenset((X, Y))
         if pair == frozenset(("A", "C")) and set(Z) == {"B"}:
             return True
-        if pair == frozenset(("A", "D")) and set(Z) == set():
-            return True
         return False
 
     data = pd.DataFrame(np.random.randint(0, 2, size=(100, 4)), columns=["A", "B", "C", "D"])
 
-    search_space = [("A", "B"), ("B", "C"), ("B", "D")]
+    search_space = [("A", "B"), ("B", "C"), ("B", "D"), ("A", "C")]
     background = ExpertKnowledge(
         search_space=search_space,
         required_edges=[("B", "D")],
@@ -484,6 +482,7 @@ def test_pc_search_space_and_required_edges():
 
     edges = set(est.skeleton_.edges())
 
+    assert ("A", "C") not in edges and ("C", "A") not in edges
     assert ("A", "D") not in edges and ("D", "A") not in edges
     assert ("A", "B") in edges or ("B", "A") in edges
     assert ("B", "C") in edges or ("C", "B") in edges
--- a/pgmpy/tests/test_causal_discovery/test_PC.py
+++ b/pgmpy/tests/test_causal_discovery/test_PC.py
@@ -448,6 +448,48 @@ def test_required_edge_enforced_against_independent_data():
     assert ("A", "B") in pdag.directed_edges
 
 
+def test_pc_search_space_and_required_edges():
+    """
+    Regression test for two related bugs in `_build_skeleton`:
+    1. The initial skeleton must be seeded from `search_space` (not a complete
+       graph), otherwise CI tests can condition on irrelevant variables and
+       incorrectly drop edges that are genuinely dependent.
+    2. An edge in `required_edges` must be added as a candidate even if it
+       wasn't included in `search_space`, otherwise it can never appear in
+       the final graph.
+    """
+
+    def fake_ci(X, Y, Z=tuple(), **kwargs):
+        pair = frozenset((X, Y))
+        if pair == frozenset(("A", "C")) and set(Z) == {"B"}:
+            return True
+        if pair == frozenset(("A", "D")) and set(Z) == set():
+            return True
+        return False
+
+    data = pd.DataFrame(np.random.randint(0, 2, size=(100, 4)), columns=["A", "B", "C", "D"])
+
+    search_space = [("A", "B"), ("B", "C"), ("B", "D")]
+    background = ExpertKnowledge(
+        search_space=search_space,
+        required_edges=[("B", "D")],
+    )
+    est = PC(
+        variant="stable",
+        ci_test=fake_ci,
+        expert_knowledge=background,
+        show_progress=False,
+    )
+    est.fit(X=data)
+
+    edges = set(est.skeleton_.edges())
+
+    assert ("A", "D") not in edges and ("D", "A") not in edges
+    assert ("A", "B") in edges or ("B", "A") in edges
+    assert ("B", "C") in edges or ("C", "B") in edges
+    assert ("B", "D") in edges or ("D", "B") in edges
+
+
 @pytest.mark.parametrize("ci_test", ["pearsonr", "pillai", "gcm"])
 @pytest.mark.parametrize("variant", ["orig", "stable", "parallel"])
 def test_build_skeleton_continuous(ci_test, variant):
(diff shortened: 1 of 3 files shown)


Shweta-nasc:
As a side effect, Codecov's patch-coverage report also now shows some pre-existing, unrelated lines in the old `fake_ci_t` mock (lines 48-49, 53, 56, 62) as "uncovered" — this is because the LF normalization makes the whole file appear in the diff. These lines predate this PR and aren't part of any change I made.

pgmpy/tests/test_causal_discovery/test_PC.py
ankurankan, line 1: @Shweta-nasc Why is the diff in this file so large?

ankurankan (changes requested):

pgmpy/tests/test_causal_discovery/test_PC.py
Shweta-nasc: Hi @ankurankan — that's the CRLF→LF line-ending normalization mentioned in my previous comment about the PR. It seems that in one file (and that's the only one in the repo) line endings were CRLF (Windows-style), whereas the whole repo (including the ruff formatted files) uses LF (Unix-style). This commit fixes that (done in the previous commit on this branch), and also makes a small adjustment to the new regression test.

The actual code changes are only 5 lines — you can see them with

git diff --ignore-all-space

which shows that one unreachable branch was removed from the fake_ci test, one edge added to the search space and one new assert line added.

If it's easier to review, I can split this into two commits — one being the line-ending normalization, another the actual code change. Let me know if that's preferable.

Shweta-nasc:
Hi @ankurankan, just following up on this PR. Whenever you get a chance, could you please take another look and let me know if any changes are needed from my side? I’d be happy to make the updates. Thanks!

ankurankan:
@Shweta-nasc Sorry for the delay, the tests seem to be failing. Can you have a look and see if you can fix them?

PR #3562, opened 2026-09-04
Move zero-variance (constant column) check to BaseCausalDiscovery

### LINKED ISSUE #3516: Add shared zero-variance validation to `BaseCausalDiscovery`
The zero-variance (constant column) check currently implemented in `SortnRegress._fit(self,X)` is not specific to that algorithm - most causal discovery algorithms that rely on variance-based computations (correlation, regression, etc.) will need the same check. Rather than each algorithm reimplementing it independently, it should be moved to `BaseCausalDiscovery` and reused across algorithms.

Idea:
- Add a shared method (e.g. `_check_zero_variance(X)`) to `BaseCausalDiscovery`
- Update `SortnRegress` and other applicable estimators to call this shared method instead of their own inline checks
- Future estimators inherit the check automatically

Reference: #3393 

@ankurankan

HasselNot7:

opencode CLI (model: qwen3.8-flash)

I cloned the repo and first confirmed on current dev that the `sortnregress` module mentioned in the issue is gone, and that the same constant-column check now lives inline in `ANM._fit` (ANM.py:141-143) and `IGCI._fit` (IGCI.py:126-128). I used the AI agent to help read the codebase and to draft the first version of the helper on `BaseCausalDiscovery` and the two call-site replacements. I went through every changed line against the original loops, kept the error message text identical, and dropped a redundant dtype re-check from the draft since both call sites already validate continuous data right before. The final result is the three-file diff below.

1. Ran `test_ANM.py` and `test_IGCI.py`: 32 passed, including the parametrized constant-column cases that assert the same `is constant; ... requires non-constant variables` message.
2. Ran `test_TAN.py` and `test_ChowLiu.py` as a sanity check on other `BaseCausalDiscovery` subclasses: 15 passed.
3. Checked the two error strings are byte-identical to the previous inline messages.
4. Repo-wide search confirmed `_check_zero_variance` has no naming collision and no other module duplicated the removed loop.
5. Ran `ruff check` and `ruff format --check` on the three changed files (clean). Environment: Python 3.13, editable install from `dev` @ a8e1640.

No. No tests were added: this is a behavior-preserving refactor and the existing parametrized tests already cover both call sites.

- Fixes #3516

- Added `_check_zero_variance(X, algorithm_name)` to `BaseCausalDiscovery`, which raises `ValueError` when a column of `X` has zero variance.
- `ANM._fit` and `IGCI._fit` now call the shared helper instead of their own inline loop.
- No change in behavior or error messages.


HasselNot7:
Since the issue hasn't been assigned to me yet, I've set it to draft.

ankurankan (changes requested):
@HasselNot7 Thanks for the PR.
1. Your current implementation doesn't cover the case when the variable is discrete.
2. This should be an automated check in the `fit` method and should throw a warning; not an error. There shouldn't be a separate method for this.

HasselNot7:
Thanks! @ankurankan — reworked per your feedback:

1. Now uses `nunique() <= 1` instead of `std() == 0`, so constant discrete columns are caught too.
2. It's a `UserWarning` (not an error), raised inline in `BaseCausalDiscovery.fit`; the separate
   `_check_zero_variance` helper and the ANM/IGCI calls are removed.

One thing to flag: `_ConstraintMixin.fit` (PC/FCI) doesn't go through `BaseCausalDiscovery.fit`, so
those won't emit the check yet. Want me to add it there too, or is base `fit` enough for this issue? :D

Tests: added `test_base.py` (fit warns on a constant column, continuous and discrete) and updated the
ANM/IGCI invalid-input cases that asserted the old error.

pgmpy/causal_discovery/_base.py
```diff
@@ -79,6 +80,15 @@ def fit(self, X: pd.DataFrame, y=None):
         discovery algorithm inheriting from `BaseCausalDiscovery`.
         """
         X = self._check_fit_data(X)
+
+        for col in X.columns:
+            if X[col].nunique() <= 1:
```
ankurankan: Shouldn't this be `== 1`? When could it be lower than 1 ?

pgmpy/tests/test_causal_discovery/test_base.py
```diff
@@ -0,0 +1,35 @@
...
+    with pytest.warns(UserWarning, match="constant"):
+        ChowLiu().fit(data)
+
+
+def test_fit_no_warning_without_constant_column():
+    data = pd.DataFrame({"x": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0], "y": [2.0, 1.0, 3.0, 2.0, 1.0, 3.0]})
+    with warnings.catch_warnings(record=True) as record:
+        warnings.simplefilter("always")
+        ChowLiu().fit(data)
+    assert not [w for w in record if issubclass(w.category, UserWarning) and "constant" in str(w.message)]
```
ankurankan: I don't see the purpose of this test? This is already covered by the rest of the tests.

pgmpy/tests/test_causal_discovery/test_base.py
```diff
@@ -0,0 +1,35 @@
...
+        "y": [2.0, 1.0, 3.0, 2.0, 1.0, 3.0],
+    }
+)
+constant_discrete = pd.DataFrame(
+    {
+        "const": ["a", "a", "a", "a", "a", "a"],
+        "x": ["p", "q", "p", "q", "q", "p"],
+        "y": ["m", "n", "n", "m", "n", "m"],
+    }
+)
```
ankurankan: If there is a single test using these variables, they should be defined within that test function.

pgmpy/causal_discovery/_base.py
```diff
@@ -79,6 +80,15 @@ def fit(self, X: pd.DataFrame, y=None):
...
+                warnings.warn(
+                    f"Variable '{col}' is constant (zero variance), which can lead to unreliable "
+                    f"results for {type(self).__name__}. Consider removing it before fitting.",
+                    UserWarning,
+                )
```
ankurankan, line 90: We use 120 char line length. Could you please verify if this is 120 character line?

ankurankan (changes requested):

HasselNot7:
Thanks for the catches @ankurankan — all sorted:

- `nunique() == 1` now. I just typed `<= 1` out of habit, you're right it can't go below 1 here :D
- Dropped the no-warning test, it was redundant anyway.
- Moved the DataFrames inside the test.
- Checked line length — all lines are within the 120 limit.

pgmpy/causal_discovery/_base.py
HasselNot7: `nunique() == 1` now. I just typed `<= 1` out of habit 😂, you're right it can't go below 1 here. :D

pgmpy/tests/test_causal_discovery/test_base.py
HasselNot7: Good catch — dropped it since it was redundant. Thanks!

pgmpy/tests/test_causal_discovery/test_base.py
HasselNot7: moved them inside the test.

pgmpy/causal_discovery/_base.py
HasselNot7: I checked, longest line here is 98 chars, well under 120.

ankurankan (approved):

HasselNot7:
Thanks for the quick review and for syncing dev in! Happy to help — ping me if anything comes up with this or SortnRegress.😄

